PR #4, opened 2026-04-01
feat: add claw mode (SaFE RayJob) support to inference-optimization s…

xiaofei-zheng:
…kill

Merge local-dev runtime/GEAK changes into PRISM architecture:

- Add dual execution mode: Local (Cursor IDE) + Claw (SaFE RayJob multi-node)
- Add Iron Rules IR-1~IR-8 (parallel GEAK, no pre-edit, mandatory integrate, safe kills, exec_on_gpu, RayJob kind, patch_inductor --target-file)
- Add GEAK & Tooling Constants table (step_limit=150, image selection)
- Add [CLAW MODE] blocks to all 9 action files (setup, baseline, profile, backends, params, kernel-opt, integrate, sweep, report)
- Add GEAK prompt rules: 4 mandatory items (kernel path, repo path, heterogeneous mode max_rounds=3, 1.5x speedup target)
- Add modes/LOCAL.md and modes/CLAW.md supplementary docs
- Add scripts: executor.sh, ray_submit.py, patch_inductor.py
- Update common.sh: MODE detection, SERVER_KILL_WAIT_S variable
- Update KNOWLEDGE-BASE.md: Qwen3-30B full loop +27.6%/+27.8%
- Update GEAK-INFERENCE-KERNEL.md: prompt rules, validation checklist
- Clean /shared_nfs/xiaofei paths from README.md

PRISM core (DFS orchestrator, heuristic scoring, LLM parallel race, KB RAG, state schema) unchanged.

Made-with: Cursor


.cursor/skills/inference-optimization/SKILL.md
```diff
@@ -109,6 +253,18 @@ the default starting structure — the agent should actively look for opportunit
...
+
+**This skill runs end-to-end without human confirmation.** Do NOT ask the user before:
+- Creating/stopping RayJob on SaFE (claw mode)
+- Running baseline/profiling scripts via Ray (claw mode) or locally
+- Submitting GEAK tasks
+- Killing/restarting servers (inside RayJob or locally)
+- Patching kernels (Inductor cache or source files)
+- Reverting failed patches
+
+Execute the full DFS pipeline autonomously. Present only the **final optimization report** to the user.
```
nehaprakriya: This leads to the agent skipping steps because it is in a hurry to finish and just provide **only** the report. Instead, we should stress that it needs to focus on reproducibility with each new optimization, have a standalone repo with all the patches and changes that can be separately evaluated by us. We can retain just one sentence about not asking anything, but it skips steps like KB lookup when I tested this on cursor.

nehaprakriya (commented):
We need to separate out the local and claw modes. When tested locally, I noticed that the agent keeps getting confused with the tasks allowed in different modes, and it said it is having "cognitive overload". The intermittent [CLAW] [LOCAL] blocks can be confusing, and I don't think LLMs are good at solving if-else style problems. The overload leads to the agent skipping steps and the output / steps differ when running this version vs the version in main. It would be better if we had separated, clearer instructions. It might also be easier to maintain over time.

PR #9, opened 2026-04-03
README: add Magpie/Apex/Intellikit references for TraceLens and OOB k…

sinarafati-amd:
update readme


nehaprakriya:
Hi, I added magpie to the readme as requested

PR #149, opened 2026-05-09
Create Weekly Issue Report

arsalanfrq-a11y:
Adding an action to create weekly report on statistics for the issues created, resolved and are still actively worked on.


chaojhou (changes requested):
**Markdown rendering will break (lines 74–104)**

The `summary` f-string sits inside a function + python heredoc + YAML `run` block, so every line ends up with ~10 leading spaces. Markdown treats 4+ leading spaces as a code block, which will collapse the headings, tables, and `---` separators once it's written to `$GITHUB_STEP_SUMMARY`.

Suggested fix: wrap the final string with `textwrap.dedent(...)`, or rewrite the f-string flush-left (no indentation).

chaojhou (approved):

PR #176, opened 2026-05-11
fix(inference-optimizer): harden runtime environment setup

lishuoshuo-amd:
Ensure Hyperloom generates pod-local runtime config and fails fast instead of falling back to stale GEAK or kernel-agent defaults.


ZhengGong-amd (approved):
No modification to actionable code and prompt, safe for merge.

PR #198, opened 2026-05-13
Update README setup instructions

lishuoshuo-amd:


ZhengGong-amd (approved):
Limited modification, no effect on actionable files. Approved.

PR #199, opened 2026-05-13
fix(kernel-agent): harden backend auth

lishuoshuo-amd:


ZhengGong-amd (approved):
Validated by Shuoshuo Li

Issue #210, opened 2026-05-14
Profiling Fix: InferenceX in HyperLoom


tsrikris:
### Target release

v0.3

### Optimization domain

Inference

### Problem / use case

Reported by @rodosingh /@mohbasit in https://github.com/AMD-AGI/Hyperloom/issues/203

We see that the collected trace for the SGLang example contained EXTEND and DECODE in the trace file name. This usually occurs when the profile_by_stage is set to True in the SGLang profiling extra arguments body. From a visual check on the code, this shouldnt be occuring as the yaml file included for SGLang has it set as False. 

This is because the InferenceX was earlier a repository in the HyperLoom codebase, but now they have replaced it with a fresh copy from SemiAnalysis repository. As a result the PROFILE_EXTRA_BODY isnt being read from the yaml file. The earlier InferenceX used this environment variable.

### Proposed solution

TBD

devalshahamd:
I am attaching the example output (how should the trace folder should look like, including the trace splitting output) and a unit test (check_torch_trace.py). Please tag me and @mohbasit  if you have any questions.  

```
vLLM:

torch_trace/
├── capture_traces
 │   ├── graph_capture_rank_*.pt.trace.json.gz
 │   ├── graph_capture_rank_*.pt.trace.json.gz
 │   ├── graph_capture_rank_*.pt.trace.json.gz
 │    :
 │    └── graph_capture_rank_*.pt.trace.json.gz
├── dp0_pp0_tp0_dcp0_ep0_rank0.*.pt.trace.json.gz
├── profiler_out_0.txt
 └── trace_split
    ├── decode_only_steady_state_*.json.gz
    ├── execution_details.csv
    ├── execution_details.json
    ├── mixed_steady_state_*.json.gz
     └── prefilldecode_steady_state_*.json.gz



SGLang:

torch_trace/
├── 1776409856.2485812-TP-0.trace.json.gz
 :
├── 1776409856.2485812-TP-7.trace.json.gz
├── capture_traces
│   ├── bs_104_rank0.json.gz
│   ├── bs_112_rank0.json.gz
│   ├── bs_120_rank0.json.gz
│    :
│   ├── bs_12_rank0.json.gz
 └── trace_split
    ├── decode_only_steady_state_*.json.gz
    ├── execution_details.csv
    ├── execution_details.json
    ├── mixed_steady_state_*.json.gz
     └── prefilldecode_steady_state_*.json.gz


```

Unit test:

```
How to run:
[...]

PR #207, opened 2026-05-13
feat(inference): TraceLens server-patcher robustness (#194 / PR #200 follow-up)

rodosingh:
Closes #205, closes #210. Follow-up to #194 / PR #200 — addresses 6 robustness gaps in `_server_patcher.py` and `_inferencex_patcher.py` identified during code review of PR #200. #210 (InferenceX SemiAnalysis fork dropping `PROFILE_EXTRA_BODY`) is fixed by the `benchmark_serving.py` patcher commit.

## Summary

PR #200 closed #194 by shipping the TraceLens server patcher + InferenceX env-var injection. Code review surfaced six follow-ups: SGLang version pin too tight, wheel-install layout unsupported, InferenceX silently dropping `PROFILE_EXTRA_BODY`, `git`/`patch` not pinned, vLLM sentinel false-positive risk, and a fuzzy-fallback `--fuzz=10` that was loose enough to mis-apply patches to wrong call sites. All six fixed here.

## Commits (8, self-contained)

**C — Tolerate point-release drift**
1. `feat(server-patcher): fuzzy patch -p1 --fuzz=10 fallback when git apply --check rejects` — falls back when context has whitespace / single-line drift.
2. `feat(server-patcher): relax SGLang version pin from frozenset to minor allowlist` — `0.5.x` covers all point releases; `HYPERLOOM_SGLANG_PATCH_EXACT_VERSIONS` / `_ALLOWED_MINORS` env overrides.

**D — Hardening + correctness**

3. `feat(server-patcher): wheel-install SGLang patching via -p3 strip` — patches `site-packages/sglang/…` in place; no symlinks (`git apply` symlink-safety check rules them out).
4. `feat(inferencex-patcher): patch benchmark_serving.py to consume PROFILE_EXTRA_BODY` — fixes the silently-ignored env var that was making `shape_discovery` / `roofline_annotations` / steady-state windows unreachable.
5. `chore(install.sh): pin git + patch for TraceLens server patcher` — apt-installs missing binaries.
6. `feat(server-patcher): tuple-of-substrings sentinel for vLLM` — require both `capture_torch_profiler_dir` AND `detailed_trace_annotation`; collapses false-positive risk.
7. `feat(server-patcher): discover SGLang versions from TraceLens manifest` — when `<patches_dir>/SUPPORTED_VERSIONS.txt` ships, it fully replaces the hardcoded default. Forward-compatible.
[...]

PR #206, opened 2026-05-13
feat(inference): surface TraceLens prose + source-function aggregation to GEAK

rodosingh:
Closes #204. Contributes to #125, #144 (last comment).

## Summary

TraceLens emits Reasoning / Resolution / Identification / Impact prose, per-row shape & bound classification, and resolved kernel-launcher paths — but Hyperloom's hot-kernel cascade either bypasses it (#125) or pre-filters GEAK out of the auto-pick (#144 last comment Layer 1) or pins the dispatch to Claude regardless of what auto-pick wanted (#144 last comment Layer 2). On a typical Qwen3-30B-A3B run that means 5/10 top kernels end up with `recommended_backends=[]` or get routed to Claude even when GEAK could rewrite them.

This PR keeps TraceLens's per-kernel data alive end-to-end **and** ensures every rewritable kernel reaches GEAK first.

> Rebased onto `main` after #207, #217 (closing #203 + #209), #201 (closing #189), and #212 (Cursor backend) merged. Merge resolution combined: GEAK-in-every-default-ladder rule (this PR) with Cursor's conditional `$CURSOR_API_KEY` gating (from main); kept PR #217's silent-fallthrough fix (no fabricated Markdown when SDK orchestrator returns empty); aligned my Layer-2 fix with main's restructured `kernel_opt` payload section.

## Commits (15, self-contained)

**A — TraceLens prose + audit + capability gate**

1. `feat(tracelens): extract Reasoning/Resolution/Impact prose from analysis.md`
2. `feat(tracelens): unified classify_patchability gate with skip_reason` — replaces silent `recommended_backends=[]` with `(bool, skip_reason)`.
3. `feat(tracelens): emit summary.json audit sidecar + handler surface`
4. `feat(kernel-agent): TraceLens hypothesis block in build_prompt`

**B — Source-function aggregation**

5. `feat(tracelens): launcher-path parser + AST function resolver` — fixes `fused_moe_kernel` resolving to `bench_moe.py`.
6. `feat(tracelens): aggregate_by_source_function groups TraceLens rows`
7. `feat(tracelens): plumb task_groups[] into write_reports + summary.json`
8. `feat(orchestrator): collapse task_group members in _batch_kernel_candidates`
9. `feat(kernel-agent): render multi-row benchmark cases when task_group attached`
10. `feat(kernel-agent): bound-keyed optimization priority block in build_prompt`

**Backend-routing fixes (#144 last comment Layers 1 & 2)**

[...]

PR #224, opened 2026-05-15
fix(inferencex-patcher): patch every InferenceX root + post-profile trace validator (#210)

rodosingh:
Refs #210 — addresses the **InferenceX patch** half of the issue (the original problem statement: `PROFILE_EXTRA_BODY` not read because Magpie loads a different InferenceX checkout). PR #207's earlier fix was incomplete; @devalshahamd / @mohbasit verified the residual symptom on the issue thread. The thread has since expanded to cover three more sub-problems that are **not** in this PR's scope and need separate follow-ups, so #210 should stay open after this lands:

- `capture_folder` threading gap (tsrikris RCA, 2026-05-15): SharedState ↔ Orchestration prompt ↔ kernel-agent handler ↔ `tracelens_analysis.py` — 3 explicit fixes proposed there.
- TraceLens versioned-subdir patch layout (mohbasit, 2026-05-18 → `release/hyperloom_integration_0.3.1`): `_server_patcher.py` currently does flat `glob("*.patch")` and misses the new `sglang_0_5_9/` / `sglang_0_5_11/` subdirs — observed live during this PR's v0.5.9 smoke (worked around locally).
- Idle% > 20% gate suppressing kernel candidates (xiaofei experiments, 2026-05-16).

> **Stacked on PR #206**. Base ref is `feat/aditysin/tracelens-prose-and-aggregation` (PR #206's branch), not `main` directly — the two PRs touch disjoint files (PR #206: `kernel-agent/tools/`; PR #224: `inference_optimizer/orchestrator/action_executors/`) so there's no functional dependency, but stacking signals the merge order: review/merge #206 first, then this. Once #206 merges, GitHub auto-rebases this PR's base back to `main` and the diff stays exactly the three commits below.

## Why

PR #207 (`d7c81928`) patched the `benchmark_serving.py` at `$INFERENCEX_PATH`, but Magpie loads its own bundled InferenceX from `$MAGPIE_DIR/InferenceX` at runtime — when the two paths differ, Magpie's runtime copy stays on the unpatched legacy line and `profile_by_stage=True` leaks through. mohbasit's smoke run on `lmsysorg/sglang:v0.5.11-rocm720-mi30x` (#210 comment 3) confirmed: trace still split into separate EXTEND / DECODE files, roofline annotations absent.

@devalshahamd diagnosed the root cause (#210 comments 4 + 6), shipped a verification harness (`check_torch_trace.py`, comment 1), and pointed out the canonical fix lives in Magpie's `MAGPIE_INFERENCEX_PATH` env var (#210 comment 8) — this PR addresses all three.

### 1. `fix(inferencex-patcher): patch every InferenceX root, not just $INFERENCEX_PATH (#210)` — defense in depth

`_discover_inferencex_roots` returns deduped list of every existing InferenceX checkout: caller-provided arg → `$INFERENCEX_PATH` → `$MAGPIE_DIR/InferenceX`. `ensure_benchmark_lib_patched` and `ensure_benchmark_serving_patched` patch each. Returns True iff at least one was patched. 7 tests pin the multi-root contract.

### 2. `feat(profile): post-profile trace structure validator (#210, full Deval check_torch_trace.py parity)` — detection

Read-only post-profile sanity check (`_validate_trace_structure`) implementing all 5 checks from `check_torch_trace.py` plus one Hyperloom-specific check for the #210 smoking-gun (`_extend_*`/`_decode_*` files in `trace_split/` instead of `_steady_state_*`). Logs WARNINGs (never raises). 11 tests cover the healthy-layout zero-warning baseline + each individual symptom in isolation.

### 3. `fix(magpie): pin MAGPIE_INFERENCEX_PATH=$INFERENCEX_PATH so Magpie loads the patched checkout (#210)` — canonical fix

@devalshahamd's preferred approach (#210 comment 8): set Magpie's documented env var at both Hyperloom→Magpie subprocess launch sites (`_grid_runner._run_magpie` and `BaselineExecutor.__call__`) so Magpie's `_resolve_default_inferencex_dir` picks the same InferenceX checkout that Hyperloom's `_inferencex_patcher` patches. Single source of truth, no patch-every-discoverable-root dance for the standard install. 3 tests pin both invocation sites + the no-`$INFERENCEX_PATH`-set fallthrough.

## Layered fix story

| Layer | What it does | When it fires |
|---|---|---|
| **Primary (commit 3)**: `MAGPIE_INFERENCEX_PATH` env-pin | Tells Magpie which InferenceX to use | Always (when `$INFERENCEX_PATH` set) |
| **Defense (commit 1)**: multi-root patcher | Patches every discoverable InferenceX | Catches older Magpie versions that don't honour `MAGPIE_INFERENCEX_PATH` yet, or operator overrides that bypass the env |
| **Detection (commit 2)**: post-profile validator | Logs actionable warnings on the smoking-gun symptoms | When something still slipped through and operators need to know |

## Test plan

- [x] 89 new + updated tests pass on `feat/aditysin/inferencex-magpie-multiroot`.
- [x] Branched cleanly off PR #206's HEAD (`6711fe5c`); rebase produced zero conflicts.
- [x] Smoke on `lmsysorg/sglang:v0.5.9-rocm700-mi30x` (per @devalshahamd #210 comment 8 — current in-repo SGLang patches target `v0.5.9`) — see smoke verification comment for per-sub-bullet evidence:
  - [x] `MAGPIE_INFERENCEX_PATH` set in the Magpie subprocess env (verified via `/proc/<pid>/environ` on the live subprocesses)
  - [x] `$INFERENCEX_PATH/utils/bench_serving/benchmark_serving.py` carries the `PROFILE_EXTRA_BODY` substring after the profile run (sentinels present pre AND post)
  - [x] trace folder has `trace_split/_steady_state_*` (NO `_extend_*` / `_decode_*`) — splitter produced `decode_only_steady_state_*`, `mixed_steady_state_*`, `prefilldecode_steady_state_*`; commit-2 check `[6]` correctly silent
  - [x] main rank-0 trace contains `kernel_shape_profiler` — 14,472 occurrences in 822 MB main trace; ~3,592 in each of 45 `capture_traces/bs_*_rank0.json.gz` files
- [x] Smoke on standard layout where `$INFERENCEX_PATH = $MAGPIE_DIR/InferenceX`: validator silent (no warnings), patcher dedup works (one patch applied, not two).
- [ ] (Once @mohbasit ships the v0.5.11 patch refresh) repeat smoke on `lmsysorg/sglang:v0.5.11-rocm720-mi30x`.

## Scope

Strictly the #210 fix. Zero overlap with PR #206 (TraceLens prose extraction + GEAK ladder).

cc: @devalshahamd @mohbasit @tsrikris @BaoYunkai (PR #207 author)

### DIFF
--- a/inference_optimizer/orchestrator/action_executors/_grid_runner.py
+++ b/inference_optimizer/orchestrator/action_executors/_grid_runner.py
@@ -965,6 +965,20 @@ def _run_magpie(
     from ._multi_node_env import magpie_remote_env
     env.update(magpie_remote_env())
 
+    # #210 (Deval, comment 8): pin Magpie's InferenceX-resolution to
+    # ``$INFERENCEX_PATH`` so Magpie loads the SAME InferenceX checkout
+    # that Hyperloom's ``_inferencex_patcher`` has patched. Without
+    # this, Magpie's ``_resolve_default_inferencex_dir`` falls through
+    # to ``./InferenceX`` next to its repo or
+    # ``$XDG_CACHE_HOME/magpie/InferenceX``, either of which may be a
+    # separate, unpatched checkout — the symptom reported in #210
+    # comments 4 + 6. ``MAGPIE_INFERENCEX_PATH`` is the highest-
+    # precedence resolution rung in Magpie itself
+    # (``Magpie/modes/benchmark/inferencex.py:43``), so this is the
+    # documented contract for tying the two checkouts together.
+    inferencex_path = os.environ.get("INFERENCEX_PATH", "").strip()
+    if inferencex_path:
+        env["MAGPIE_INFERENCEX_PATH"] = inferencex_path
     # Always-on RESULT_DIR default: covers Magpie scripts that respect
     # the env var (and would otherwise fall back to a hardcoded path).
     # Scripts that ignore RESULT_DIR (e.g. ``dsr1_fp8_mi300x.sh`` with its
--- a/inference_optimizer/orchestrator/action_executors/_inferencex_patcher.py
+++ b/inference_optimizer/orchestrator/action_executors/_inferencex_patcher.py
@@ -113,27 +113,88 @@ _BENCH_SERVING_SENTINEL = "PROFILE_EXTRA_BODY"
 _BENCH_SERVING_LOCK_PATH = "/tmp/hyperloom_benchmark_serving_patcher.lock"
 
 
+def _discover_inferencex_roots(
+    inferencex_path: Path | str | None,
+) -> list[Path]:
+    """Return every InferenceX checkout root Hyperloom should patch.
+
+    #210 root cause (Deval, 2026-05-15 — see issue #210 comments 4 + 6):
+    Magpie loads its own bundled InferenceX from
+    ``$MAGPIE_DIR/InferenceX`` at runtime, NOT from
+    ``$INFERENCEX_PATH``. If those two paths differ, patching only
+    ``$INFERENCEX_PATH`` leaves Magpie's actual runtime copy untouched
+    — ``profile_by_stage=True`` leaks through, ``PROFILE_EXTRA_BODY``
+    is ignored, and the trace folder ends up with separate
+    ``_extend_*`` / ``_decode_*`` files instead of the expected
+    ``_steady_state_*`` (the smoking-gun symptom mohbasit reported on
+    the same issue).
+
+    Patches ALL discovered roots (deduplicated by resolved absolute
+    path) so the patch reaches whichever InferenceX Magpie actually
+    imports at runtime:
+
+    1. ``inferencex_path`` arg (caller-provided override)
+    2. ``$INFERENCEX_PATH`` env (existing behaviour, PR #207's path)
+    3. ``$MAGPIE_DIR/InferenceX`` (the new path — the #210 fix)
+
+    Returns ``[]`` when no roots resolve to existing directories;
+    callers fail-soft as before. Same paths from multiple sources
+    collapse to one entry (the standard install layout where
+    ``$INFERENCEX_PATH = $MAGPIE_DIR/InferenceX`` produces one root,
+    not two).
+    """
+    roots: list[Path] = []
+    seen: set[Path] = set()
+
+    def _add(candidate: Path | str | None) -> None:
+        if not candidate:
+            return
+        try:
+            resolved = Path(candidate).expanduser().resolve()
+        except OSError:
+            return
+        if not resolved.is_dir():
+            return
+        if resolved in seen:
+            return
+        seen.add(resolved)
+        roots.append(resolved)
+
+    _add(inferencex_path)
+    _add(os.environ.get("INFERENCEX_PATH", "").strip() or None)
+    magpie_dir = os.environ.get("MAGPIE_DIR", "").strip()
+    if magpie_dir:
+        _add(Path(magpie_dir) / "InferenceX")
+    return roots
+
+
+def _resolve_benchmark_lib_paths(
+    inferencex_path: Path | str | None,
+) -> list[Path]:
+    """Return every existing ``<root>/benchmarks/benchmark_lib.sh`` to
+    patch (one per InferenceX root from :func:`_discover_inferencex_roots`).
+
+    Returns ``[]`` when no root has the expected file so callers can
+    treat "no InferenceX checkout" as "skip patching".
+    """
+    out: list[Path] = []
+    for root in _discover_inferencex_roots(inferencex_path):
+        candidate = root / "benchmarks" / "benchmark_lib.sh"
+        if candidate.is_file():
+            out.append(candidate)
+    return out
+
+
+# Back-compat: existing callers / tests that import this single-path
+# helper get the first discovered candidate (or None). New code must
+# use :func:`_resolve_benchmark_lib_paths` to patch every root.
 def _resolve_benchmark_lib_path(
     inferencex_path: Path | str | None,
 ) -> Path | None:
-    """Resolve ``<inferencex_root>/benchmarks/benchmark_lib.sh``.
-
-    Returns ``None`` when the path is unconfigured or missing on disk
-    so callers can treat "no InferenceX checkout" as "skip patching"
-    (this is what unit tests need — they exercise profile YAML
-    rendering without provisioning a real InferenceX tree).
-    """
-    root: Path | None = None
-    if inferencex_path:
-        root = Path(inferencex_path)
-    else:
-        env = os.environ.get("INFERENCEX_PATH", "").strip()
-        if env:
-            root = Path(env)
-    if root is None:
-        return None
-    candidate = root / "benchmarks" / "benchmark_lib.sh"
-    return candidate if candidate.is_file() else None
+    """Single-path back-compat shim — returns first
+    :func:`_resolve_benchmark_lib_paths` result."""
+    paths = _resolve_benchmark_lib_paths(inferencex_path)
+    return paths[0] if paths else None
 
 
 @contextmanager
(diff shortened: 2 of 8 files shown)


rodosingh:
## Smoke verification

Ran end-to-end on a `lmsysorg/sglang:v0.5.9-rocm700-mi30x` container (MI300X, fresh install via `inference_optimizer/scripts/install.sh`). Driving `ProfileExecutor` directly against Qwen3-32B / sglang (TP=1, CONC=8, ISL=OSL=256, NUM_PROMPTS=112) so the three PR-224 codepaths fire in their natural order.

### All four bullet-3 sub-assertions verified live

**(a) `MAGPIE_INFERENCEX_PATH` set in the Magpie subprocess env — PASS.** Captured `/proc/<pid>/environ` mid-run on the Magpie subprocess, the `sglang_mi300x.sh` wrapper, and `sglang.launch_server`. All three carry `MAGPIE_INFERENCEX_PATH=/wekafs/InferenceX`. Matches `baseline.py:407` setting `env["MAGPIE_INFERENCEX_PATH"] = inferencex_path` before `subprocess.run`.

**(b) `$INFERENCEX_PATH/utils/bench_serving/benchmark_serving.py` carries `PROFILE_EXTRA_BODY` after profile — PASS.** Sentinels present pre AND post; patcher's idempotent fast-path fired (no re-mutate). And `ps` showed `python3 /wekafs/InferenceX/utils/bench_serving/benchmark_serving.py …` running — Magpie loaded the *patched* checkout (proves commit 1 + commit 3 work together at runtime). Materialised YAML pinned `inferencex_path: /wekafs/InferenceX` and `PROFILE_EXTRA_BODY={…, "profile_by_stage": false, "start_step": 1280, "num_steps": 512}` — start/num match the documented OSL=256 / CONC=8 / RR=1 formula.

**(c) `trace_split/_steady_state_*` (no `_extend_*`/`_decode_*`) — PASS.** Ran `python -m TraceLens.TraceUtils.split_inference_trace_annotation … --find-steady-state` on the produced trace. Output:

```
trace_split/
├── decode_only_steady_state_prefilldecode_0_decode_32_bs8_conc8_*.trace.json.gz
├── mixed_steady_state_prefilldecode_0_decode_32_bs8_conc8_*.trace.json.gz
└── prefilldecode_steady_state_prefilldecode_0_decode_0_bs1024_conc4_*.trace.json.gz
```

Zero `_extend_*` / `_decode_*` per-step files — exactly the layout this PR's commit-2 check `[6]` confirms (verified separately: re-ran `_validate_trace_structure` with the produced `trace_split/` and check `[6]` correctly stayed silent).

**(d) Main rank-0 trace contains `kernel_shape_profiler` — PASS.** 14,472 occurrences in the 822 MB main trace. Each of the 45 `capture_traces/bs_*_rank0.json.gz` files (~24 MB each) carries ~3,592 occurrences too. Shape-discovery instrumentation fully active.

### What this required (and a finding for the team)

This container ships `sglang 0.5.10.dev4+ge714db81a`. `_server_patcher` initially logged `SGLang patches directory empty; skip` — not because the version was rejected (it was accepted via the minor allowlist) but because **TraceLens has reorganised its patch layout into versioned subdirs** (`sglang_0_5_9/`, `sglang_0_5_11/`) and Hyperloom's `_server_patcher.py:414` still does `patches_dir.glob("*.patch")` non-recursively at the parent. Worked around locally by `cp -r`'ing TraceLens to a writable path and symlinking `sglang_0_5_9/*.patch` up to the parent so the flat-glob resolver finds them. With that workaround in place: all 10 patches applied strict (no fuzzy), `--enable-shape-discovery-for-cuda-graph-profile` was auto-appended to `EXTRA_SGLANG_ARGS`, throughput jumped from 50.6 → 178.97 tok/s/gpu, and `capture_traces/` came alive with 45 per-shape capture files.

The layout-vs-glob mismatch is outside PR 224's scope but worth filing — it's likely the same root cause why @mohbasit's v0.5.11 smoke was stuck without server-side instrumentation. Once `_server_patcher` learns the versioned-subdir layout, bullet 5 of the test plan should unblock on the v0.5.11 image automatically (the `sglang_0_5_11/*.patch` files already exist in TraceLens).

### Two follow-up tickets worth filing (none block PR 224)

1. **`_server_patcher` doesn't handle TraceLens's versioned-subdir patch layout** (separate from PR 224; explains the false-negative "patches directory empty" path on this container).

2. **PR 224 commit 2 validator: `_TRACE_INSPECT_BYTES = 2_000_000` is too small for large profile traces.** On this 822 MB trace, validator check `[5]` fires a false positive (`kernel_shape_profiler` events are present 14,472 times, just not in the first 2 MB), and `[2]` similarly false-positives on the 24 MB capture files. The detector's smoking-gun check `[6]` is unaffected and works correctly. Suggest bumping to ~20 MB and/or strided sampling — small in-PR fix if you'd like to address before merge.

3. **3 of the new patcher tests don't `monkeypatch.delenv("INFERENCEX_PATH"/"MAGPIE_DIR")`** so they fail on hosts that export those (like this pod):
   - `test_returns_false_when_benchmark_lib_missing`
   - `test_returns_false_when_legacy_line_missing`
   - `test_benchmark_serving_patch_returns_false_when_legacy_line_missing`

   Peer `test_returns_false_when_inferencex_path_unset` does monkeypatch and passes — same 2-line fix per test. Could land here or as a follow-up.

### On commit 1's necessity (since the layered fix story raised the question)

Checked main: only single-root `_resolve_benchmark_lib_path() -> Path | None`. The multi-root behaviour is genuinely new and not redundant with commit 3 — they cover different rungs of Magpie's resolver chain. Magpie reads `config.inferencex_path` from the materialised YAML *first* (`Magpie/modes/benchmark/benchmarker.py:105-106`); only if unset does it fall back to `_resolve_default_inferencex_dir()` where `MAGPIE_INFERENCEX_PATH` wins. So commit 3 covers the YAML-unset path; commit 1's multi-root patcher catches the case where YAML pins an explicit `inferencex_path` that differs from `$INFERENCEX_PATH` (the original #210 scenario). Dedup keeps overhead at zero in the standard install layout — verified live (single root discovered) + by the new unit test.

All 89 unit tests pass on a clean env (CI's `Analyze (python)` doesn't run pytest; verified manually).

Ready for review from this side. cc @devalshahamd @mohbasit @tsrikris @BaoYunkai

ZhengGong-amd:
In recent PR, the default INFERENCEX_PATH changed, please review the new code.

rodosingh:
>2. feat(profile): post-profile trace structure validator (#210, full Deval check_torch_trace.py parity) — detection
Read-only post-profile sanity check (_validate_trace_structure) implementing all 5 checks from check_torch_trace.py plus one Hyperloom-specific check for the https://github.com/AMD-AGI/Hyperloom/issues/210 smoking-gun (_extend_*/_decode_* files in trace_split/ instead of _steady_state_*). Logs WARNINGs (never raises). 11 tests cover the healthy-layout zero-warning baseline + each individual symptom in isolation.

>
>3. fix(magpie): pin MAGPIE_INFERENCEX_PATH=$INFERENCEX_PATH so Magpie loads the patched checkout (#210) — canonical fix
@devalshahamd's preferred approach (https://github.com/AMD-AGI/Hyperloom/issues/210 comment 8): set Magpie's documented env var at both Hyperloom→Magpie subprocess launch sites (_grid_runner._run_magpie and BaselineExecutor.__call__) so Magpie's _resolve_default_inferencex_dir picks the same InferenceX checkout that Hyperloom's _inferencex_patcher patches. Single source of truth, no patch-every-discoverable-root dance for the standard install. 3 tests pin both invocation sites + the no-$INFERENCEX_PATH-set fallthrough.

Has this feature been integrated @ZhengGong-amd  ?
Thanks!

ZhengGong-amd (approved):
commit 1, 2, 3 are valuable revision to main. Please solve the conflict before merge.

rodosingh:
Rebased on the new main and reviewed the INFERENCEX_PATH-related changes — PR #224's logic is unaffected. Quick walk-through:

**What changed on main:** `install.sh` now defines `INFERENCEX_DEFAULT_DIR=${HYPERLOOM_RUNTIME_DIR}/InferenceX` and `ensure_inferencex` clones into it when `INFERENCEX_PATH` isn't already set in the env. Caller-set `INFERENCEX_PATH` still wins (lines 401-405 of main's install.sh). All three PR-224 commits key off the env var `INFERENCEX_PATH`, so the default-dir change is transparent to them.

**Rebase notes:**

- One conflict in `_grid_runner.py:_run_magpie` — main added a multi-node env block (`magpie_remote_env()`), PR-224 added the `MAGPIE_INFERENCEX_PATH` pin in the same area. They're orthogonal: kept both, in order. Verified post-rebase that `magpie_remote_env()`'s no-op single-node guard still applies and the `MAGPIE_INFERENCEX_PATH` set still fires when `INFERENCEX_PATH` is configured.
- 3 tests needed their mock target updated from `subprocess.run` → `run_with_session_kill` (main introduced the wrapper at module level for session-kill semantics; the same pattern is in `test_grid_runner_param_overrides.py` and `test_baseline_param_overrides.py:249` already on main). Folded into the same commit that added them so history stays clean.

**Local verification:** 191 targeted tests pass on the rebased head (including the 3 fixed mocks). PR is now `MERGEABLE`.

If by "default INFERENCEX_PATH changed" you meant something other than the `install.sh` default-dir refactor, let me know and I'll re-check.

PR #228, opened 2026-05-15
docs: streamline README §Quickstart

lishuoshuo-amd:
- Step 1: replace docker run snippet with two explicit setup paths
- Step 2: simplify path table; drop operator-only env vars
- Step 4: use env var placeholders in prompt examples
- .env.template: fix SAFE_API_KEY prefix comment (sk- -> ak-)


ZhengGong-amd (approved):
limited revision, approved.

PR #234, opened 2026-05-16
fix(install): /opt/venv preference + torch-wheel gate + missing robustness_monitor template (#232, #233)

### LINKED ISSUE #232: inference_optimizer/scripts/install.sh: stale PYTHON wins over /opt/venv, no ROCm-torch gate, blanket --break-system-packages
Hit this on a fresh MI355X pod (`gfx950 × 8`, ROCm 7.0.0, `/opt/venv/bin/python` with full ROCm stack present, `/sgl-workspace/{aiter,sglang,...}` checkouts present). `bash inference_optimizer/scripts/install.sh` failed in two stages and quietly produced a broken install before I caught it. Three problems chained, all in `install.sh` lines ~90–140.

### 1. `resolve_python()` honours a stale caller-supplied `$PYTHON` over `/opt/venv/bin/python`

```bash
resolve_python() {
  if [ -n "${PYTHON:-}" ] && [ -x "$PYTHON" ]; then
    return 0
  fi
  if [ -x "/opt/venv/bin/python" ]; then
    PYTHON="/opt/venv/bin/python"
    return 0
  fi
  ...
}
```

If the caller has any `PYTHON` exported (e.g. from a previous shell, or my workaround venv), it wins — even when `/opt/venv/bin/python` exists with the canonical hyperloom ROCm stack. On hyperloom containers the safer default is the other way round: prefer `/opt/venv/bin/python` and require an explicit opt-out (e.g. `INFERENCE_OPTIMIZER_FORCE_PYTHON=1`) to override. Otherwise a stale `PYTHON` silently sends `pip install -e .[test]` into the wrong interpreter.

### 2. No ROCm-torch sanity gate after `pip install -e .[test]`

`.[test]` pulls in `torch` as a transitive dep. On a freshly-created venv (which is what I had after working around problem 3 below), stock PyPI `torch` resolves to **NVIDIA cu130** wheels (`torch 2.12.0+cu130, hip=None, is_available=False`). The install completes "successfully", and the failure surfaces only later when `kernel-agent/scripts/install.sh` runs the GEAK RAG index build:

```
[kernel-agent] building RAG index ... on device=cuda ...
RuntimeError: Found no NVIDIA driver on your system. Please check that you have an NVIDIA GPU ...
```

The auto-detect that picked `device=cuda` is correct for ROCm torch (PyTorch ROCm uses the CUDA API surface) — the fault is upstream: torch was the wrong wheel and we never noticed. A post-install assertion that `import torch; torch.version.hip is not None` on ROCm pods (or `is_available()` succeeds) would fail fast with an actionable message instead of leaving operators with a broken install.

### 3. Blanket `--break-system-packages` regardless of pip version

```bash
if "$PYTHON" - <<'PY' 2>/dev/null
import sys
raise SystemExit(0 if sys.prefix == sys.base_prefix else 1)
PY
then
  PIP_EXTRA=(--break-system-packages)
fi
```

`--break-system-packages` was added in **pip 23.0.1**. On stripped pods that ship pip 22.0.2 (mine did), the very first `pip install` call dies before anything is installed. Detect-and-surface would be cleaner: if pip < 23.0.1 AND we'd append the flag, fail with "upgrade pip OR re-run inside a venv" instead of running pip with an unsupported flag.

### Repro

1. Fresh MI355X pod, `/opt/venv/bin/python` has the ROCm stack, system pip is 22.0.2.
2. `bash inference_optimizer/scripts/install.sh` (no env exports).
3. Hit problem 3 → workaround: create `$USER_DATA_PATH/venv` and re-run with `PYTHON=$USER_DATA_PATH/venv/bin/python`.
4. Install "succeeds" silently with NVIDIA cu130 torch (problem 1 + 2). Discovered later when GEAK RAG build crashed.

### Suggested direction

* `resolve_python()`: prefer `/opt/venv/bin/python` when present, with an `INFERENCE_OPTIMIZER_FORCE_PYTHON=1` escape hatch.
* Export `PYTHON` and prepend its `bin/` to `$PATH` before chaining to `kernel-agent/scripts/install.sh` so its bare `python3 -m pip` calls hit the same interpreter (currently it falls back to whatever `python3` resolves to on `$PATH`).
* Post-install gate: on hosts where `rocm-smi` works, assert `torch.version.hip is not None` on the chosen `PYTHON` and fail with a clear message naming the ROCm wheel index (`https://download.pytorch.org/whl/rocm6.0`) and the `/opt/venv` opt-in.
* Old-pip detection block: only append `--break-system-packages` when pip ≥ 23.0.1; otherwise fail clearly.

Happy to take a small PR if helpful.

### LINKED ISSUE #233: SKILL.md references optimizer_runs/robustness_monitor.sh.example, but the file isn't in the repo
`inference_optimizer/SKILL.md:695-700` instructs operators to `cp` a robustness-monitor template from the repo and run it as a `setsid nohup` watchdog:

```bash
cp "$REPO_ROOT/optimizer_runs/robustness_monitor.sh.example" \
   "$RUN_DIR/robustness_monitor.sh"
chmod +x "$RUN_DIR/robustness_monitor.sh"
setsid nohup bash "$RUN_DIR/robustness_monitor.sh" \
  > "$RUN_DIR/robustness_monitor_$(date +%Y%m%d_%H%M%S).log" \
  2>&1 < /dev/null &
```

The source file (`$REPO_ROOT/optimizer_runs/robustness_monitor.sh.example`) doesn't exist on `main`. `optimizer_runs/` directory isn't tracked either. `git log --all --diff-filter=D -- '*robustness_monitor*'` returns empty, so it wasn't deleted — it was never landed. Other `.example` files referenced by the codebase do exist (`deploy/local/OOB/env.example`, `inference_optimizer/scripts/setup_env.sh.example`), so this looks like a doc / asset gap, not a deliberate removal.

Symptom on a clean clone: any operator following SKILL.md literally hits

```
cp: cannot stat '.../optimizer_runs/robustness_monitor.sh.example': No such file or directory
```

and skips the watchdog. For a 24h budget that's the difference between auto-resume on optimizer crash and silent stall.

### Two unrelated docstring references (informational only)

`inference_optimizer/paths.py:225` and `inference_optimizer/session_paths.py:206` both mention `robustness_monitor*.log` as session-tail artefacts. Those are fine — they're describing where the monitor's log lands, not pointing at the missing template. The only operational caller is the SKILL.md `cp` above.

### Suggested fix

Either:

* (a) Add `optimizer_runs/robustness_monitor.sh.example` to the repo. The script is short (~25 lines, see SKILL §Robustness Monitor for the contract: poll `state.json` every 300 s, exit on terminal `stop_reason`, auto-resume via `--resume` on optimizer crash). I can submit a small PR with the template if helpful.
* (b) Update SKILL.md:695 to point at the actual source of the script (e.g. inline it, or reference an external location), and remove the `cp` step.

(a) seems cleaner because operators following the SKILL today end up writing the monitor inline anyway, which makes the script un-versioned and inconsistent across runs.

rodosingh:
Closes #232, closes #233.

Both bugs hit the same fresh MI355X pod and chain into each other — operator runs `bash inference_optimizer/scripts/install.sh`, the install "succeeds" but with the wrong torch wheel, then the GEAK RAG step explodes with `Found no NVIDIA driver`; even after that's worked around, the SKILL.md `cp` for the robustness monitor fails because the template was never in the repo. So bundling them.

### 1. `fix(install.sh): prefer /opt/venv ROCm stack, gate torch wheel, detect old pip (#232)`

Four changes in `inference_optimizer/scripts/install.sh`:

* `resolve_python()` prefers `/opt/venv/bin/python` over a caller-supplied `$PYTHON` when both exist (escape hatch: `INFERENCE_OPTIMIZER_FORCE_PYTHON=1`). Stops a stale ambient `$PYTHON` from silently bypassing the canonical ROCm stack.
* Exports `$PYTHON` and prepends its `bin/` to `$PATH` before chaining to `kernel-agent/scripts/install.sh`. That installer's own `_resolve_magpie_python` is correct, but its bare `python3 -m pip` calls fall through to `$PATH` resolution — without this, the two installers can land in different interpreters.
* New `ensure_torch_compatible_with_gpu` step. On hosts where `rocm-smi` works, asserts `torch.version.hip is not None`. Two failure modes surfaced separately: torch missing entirely (next `pip install -e .[test]` would pull NVIDIA cu130 from PyPI) and torch-built-against-CUDA (RAG step crashes at `torch._C._cuda_init()`). Both fail with three actionable fixes; both honour `INFERENCE_OPTIMIZER_SKIP_TORCH_GATE=1`.
* Old-pip detection. `--break-system-packages` was added in pip 23.0.1 and is rejected as unknown by older pips. Probe `pip install --break-system-packages --help` before adopting; if unsupported, fail fast with three options (use `/opt/venv`, create a venv, upgrade pip) instead of running pip with an unknown flag.

### 2. `fix(SKILL): ship optimizer_runs/robustness_monitor.sh.example template (#233)`

Adds `optimizer_runs/robustness_monitor.sh.example`, the template `inference_optimizer/SKILL.md:695` already tells operators to `cp` from. The script was never in the repo (`git log --diff-filter=D` empty for the path); SKILL.md had been pointing at a file that didn't exist. Behaviour matches SKILL §Robustness Monitor for Long Runs verbatim: poll `state.json` every 300s, exit on terminal `stop_reason`, auto-resume via `inference_optimizer optimize --resume` when the optimizer disappears from `/proc`.

Polish vs the inline scripts operators have been hand-writing during the gap:
- Required (`PID_FILE`, `REPO_ROOT`) and optional (`MAX_HOURS`, `TARGET_GAIN`, etc.) envs documented in the header with defaults shown; SKILL invocation pattern reproduced inline.
- `: \"\${PID_FILE:?...}\"` preflight fails fast with a clear message instead of letting `set -u` cascade into a generic "unbound variable" later.

`.gitignore` adjustment: `optimizer_runs/` is the runtime per-session output and stays ignored, but the bare directory rule prevents git from recursing — switching to `optimizer_runs/*` + `!optimizer_runs/robustness_monitor.sh.example` keeps everything else ignored while letting this one tracked template through.

## Test plan

- [x] `bash -n` passes on both `inference_optimizer/scripts/install.sh` and the new `.example`.
- [x] All 4 install.sh fixes exercised in 4 dry-run scenarios on the live pod (reported in the original repro session): stale-PYTHON wins fail, FORCE_PYTHON opt-out works, torch-gate fails with actionable message on bare interpreter, fresh shell auto-picks `/opt/venv`.
- [x] `optimizer_runs/robustness_monitor.sh.example` is the same script that ran healthily as PID 22949 on the live pod (multiple resume cycles observed) — adapted only for header docs + preflight-env validation.
- [x] Re-run `install.sh` on a fresh MI355X pod with no env exports: confirm it auto-picks `/opt/venv/bin/python`, torch gate passes, RAG index build succeeds, no `--break-system-packages` invocation.
- [x] Re-run on a fresh ROCm pod with bare `/usr/bin/python3` + no `/opt/venv`: confirm torch gate fails with the three-option message instead of letting `pip install -e .[test]` pull the NVIDIA wheel.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,8 +20,13 @@ venv/
 trace_rank*.json
 .rocprofv3/
 
-# Optimization run artifacts
-optimizer_runs/
+# Optimization run artifacts (runtime per-session output; ignored). One
+# tracked exception: the .example template SKILL.md tells operators to
+# `cp` from at session start. Use ``optimizer_runs/*`` (not the bare
+# ``optimizer_runs/`` directory) so git still recurses into the dir
+# and the negation pattern below can re-include the .example file.
+optimizer_runs/*
+!optimizer_runs/robustness_monitor.sh.example
 optimization_logs/
 optimization_reports/
 optimized_versions/
--- a/inference_optimizer/scripts/install.sh
+++ b/inference_optimizer/scripts/install.sh
@@ -89,16 +89,24 @@ run() {
 }
 
 # --- 0. Resolve PYTHON ---
-# Prefer the existing PYTHON env (callers may have already pinned it),
-# then /opt/venv/bin/python (default for hyperloom containers), then
-# whatever python3 is on PATH. We do NOT hardcode /opt/venv/bin into
-# subsequent PATH; we only use this binary to drive `pip install`.
+# On hyperloom / sgl-workspace containers the canonical ROCm stack lives in
+# /opt/venv (preinstalled torch+rocm, sglang, vllm, aiter, sgl_kernel,
+# triton, Magpie, inference_optimizer, claude_agent_sdk, ray). Always
+# prefer that interpreter — bare-image PYTHONs (e.g. /usr/bin/python3) on a
+# ROCm pod silently pull plain `torch` from PyPI on `pip install -e .[test]`,
+# which is the NVIDIA CUDA wheel and crashes downstream RAG / baseline
+# steps with "Found no NVIDIA driver". Operators who really need a custom
+# interpreter can opt out with INFERENCE_OPTIMIZER_FORCE_PYTHON=1.
 resolve_python() {
-  if [ -n "${PYTHON:-}" ] && [ -x "$PYTHON" ]; then
+  if [ -x "/opt/venv/bin/python" ] && [ "${INFERENCE_OPTIMIZER_FORCE_PYTHON:-0}" != "1" ]; then
+    if [ -n "${PYTHON:-}" ] && [ "${PYTHON}" != "/opt/venv/bin/python" ]; then
+      log "preferring /opt/venv/bin/python over PYTHON=${PYTHON} (canonical ROCm stack)"
+      log "  set INFERENCE_OPTIMIZER_FORCE_PYTHON=1 to honor PYTHON verbatim"
+    fi
+    PYTHON="/opt/venv/bin/python"
     return 0
   fi
-  if [ -x "/opt/venv/bin/python" ]; then
-    PYTHON="/opt/venv/bin/python"
+  if [ -n "${PYTHON:-}" ] && [ -x "$PYTHON" ]; then
     return 0
   fi
   if command -v python3 >/dev/null 2>&1; then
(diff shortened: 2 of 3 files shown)


ZhengGong-amd:
Thanks for putting this together — both fixes are valid and the repro write-up is really helpful. Two notes on our side:

1. install.sh changes (commit 1) — already addressed in parallel

No objection to the direction here. We've actually landed an equivalent fix on our end already: the install script no longer relies on hard-coded interpreter paths and instead auto-resolves the Python/venv at runtime, and --break-system-packages is no longer required (or unconditionally passed) — so the old-pip failure mode you hit is gone. The behavior should converge with what this PR is targeting; we can either rebase your commit on top to confirm parity, or drop commit 1 in favor of the version already on main. Either is fine with us.

2. robustness_monitor.sh.example template (commit 2) — fine to land, but heads-up on churn

The fix itself is correct — SKILL.md has been pointing at a file that doesn't exist, and shipping the template unblocks operators today. Happy to take it as-is.

That said, I want to flag that the robustness agent is getting a substantial overhaul in the next release, with a number of new capabilities, and that work will likely rewrite or replace large portions of the script this PR adds (along with the corresponding .gitignore carve-out and the SKILL.md section). So the template you're shipping here will probably be short-lived.

We're still happy to merge it now to fix the immediate broken-link issue for users on the current release — just want to make sure you're OK with the fact that it'll be churned shortly after. Does that work for you, or would you rather hold commit 2 and fold it into the upcoming robustness-agent PR instead?

Thanks again for the careful breakdown and the live-pod validation.

ZhengGong-amd:
The installer works fine:

image

rodosingh:
Thanks — rebased both commits on latest \`origin/main\` (a2c29a7f).

**Re commit 1 (install.sh):** rebased rather than dropped. Reading current main's \`install.sh\`, the changes here are still substantive vs what's already there:

- \`resolve_python()\` flips the preference order (PYTHON env → /opt/venv) to (/opt/venv → PYTHON env), with \`INFERENCE_OPTIMIZER_FORCE_PYTHON=1\` to opt back to the old order. This is what avoids the trap reported in #232 where a stale \`PYTHON=/usr/bin/python3\` from a prior shell silently bypasses the ROCm stack.
- \`ensure_torch_compatible_with_gpu()\` is entirely new — fail-fast guard for torch missing or CUDA-built on a ROCm pod.
- \`PATH\` thread + \`export PYTHON\` so the chained kernel-agent installer's bare \`python3 -m pip\` lands in the same interpreter.
- Pip-too-old detection on the \`--break-system-packages\` branch.

If you'd prefer I drop these in favor of further main-side work, happy to — but I'd value the torch-gate landing somewhere since it's the bit that turns a silent install-success-then-runtime-CUDA-crash into a clear preflight error.

**Re commit 2 (robustness_monitor template):** keeping it — operators on the current release need the broken-link fix today, and I'm fine accepting that the upcoming robustness-agent rewrite will replace large portions of it. Happy to coordinate with that PR's author when it lands so the rewrite supersedes cleanly.

ZhengGong-amd:
Thanks for your rapid response! 
Please do a full run and report the installation behavior.

ZhengGong-amd (commented):
The !optimizer_runs/robustness_monitor.sh.example will be ignored. Is that designed by purpose?

rodosingh:
> The !optimizer_runs/robustness_monitor.sh.example will be ignored. Is that designed by purpose?

Yes, beacuse `optimizer_runs/` folder contains some *.log files which generates during run.

ZhengGong-amd (approved):

PR #244, opened 2026-05-19
fix(workload_envs): derive TP from ROCR_VISIBLE_DEVICES when not set

### LINKED ISSUE #248: validate_stack uses wrong TP when user pins a GPU subset via ROCR_VISIBLE_DEVICES
## Summary

When a user pins GPUs via `ROCR_VISIBLE_DEVICES` (e.g. `4,5,6,7` for 4 GPUs on a shared 8-GPU node) but doesn't explicitly set `TP`, the Magpie YAML template's hardcoded `TP: 8` wins. This causes `validate_stack` to launch benchmarks on all 8 GPUs instead of the user's 4.

Combined with the mandatory `validate_stack` gate (see #245), this creates an infinite retry loop that burns LLM calls without progress and blocks all other actions including kernel optimization.

## Fix

PR #244 — derives TP from the device count in `ROCR_VISIBLE_DEVICES` when TP is not set in the process env.

Priority order: TP env var > ROCR device count > YAML template TP > default 1.

## Relevant code

- `inference_optimizer/orchestrator/action_executors/_workload_envs.py:179`

sdubagun-amd:
## Summary

- When a user pins GPUs via `ROCR_VISIBLE_DEVICES` (e.g. `4,5,6,7` for 4 GPUs on a shared 8-GPU node), but doesn't explicitly set `TP`, the Magpie YAML template's hardcoded `TP: 8` wins
- This causes `validate_stack` (and other executors) to launch benchmarks on all 8 GPUs instead of the user's 4, producing wrong results or failures
- The mandatory `validate_stack` gate then retries every tick indefinitely, blocking all other actions including kernel optimization

## Fix

Derive `TP` from the device count in `ROCR_VISIBLE_DEVICES` when TP is not explicitly set in the process env and the YAML template doesn't specify it.

Priority order: `TP` env var > ROCR device count > YAML template TP > default 1.

## Code

One file: `inference_optimizer/orchestrator/action_executors/_workload_envs.py:179`

## Test plan

- [ ] Existing tests pass (all set TP explicitly, unaffected by this change)
- [ ] With `ROCR_VISIBLE_DEVICES=4,5,6,7` and no `TP` env var, validate_stack config should have `TP: 4`
- [ ] With `TP=8` set explicitly, behavior unchanged (TP env wins)

### DIFF
--- a/inference_optimizer/orchestrator/action_executors/_workload_envs.py
+++ b/inference_optimizer/orchestrator/action_executors/_workload_envs.py
@@ -176,9 +176,20 @@ def materialize_config_with_envs(
     rocr_env = os.environ.get("ROCR_VISIBLE_DEVICES", "").strip()
     if rocr_env:
         envs["ROCR_VISIBLE_DEVICES"] = rocr_env
-    resolved_tp = int(envs.get("TP") or os.environ.get("TP") or 1)
+    tp_from_env = os.environ.get("TP", "").strip()
+    tp_from_yaml = envs.get("TP")
     rocr_yaml = str(envs.get("ROCR_VISIBLE_DEVICES") or "").strip()
     rocr_devices = [d.strip() for d in rocr_yaml.split(",") if d.strip()]
+    if tp_from_env:
+        resolved_tp = int(tp_from_env)
+    elif rocr_yaml and not tp_from_yaml:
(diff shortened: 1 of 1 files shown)


ZhengGong-amd (approved):
Limited revision to actionable codes. Works fine with current main. No Conflict shown. Approved.

PR #252, opened 2026-05-19
feat(orchestrator/scripts): analysis.md auto-trigger, validate_stack nudge, RUN_EVAL despam, event_counts --all

rodosingh:
Five small, independent improvements observed on a long Qwen3-30B-A3B run.

### 1. Auto-emit select_kernels after profile (`coordinator.py`, `cli.py`, `prompt_builder.py`, `validate_stack.md`)

`analysis.md` was only produced when Orchestration remembered to emit `select_kernels`. Now there's a fire-and-forget auto-trigger after every successful profile, plus a TODO 3/5 analyze gate that surfaces guidance when the auto-trigger is disabled or still in flight.

- Default ON. Opt out with `--no-auto-analyze-after-profile` or `HYPERLOOM_AUTO_ANALYZE_AFTER_PROFILE=0`.
- TODO chain renumbers from 0/4..4/4 to 0/5..5/5 to slot the new step in.
- No PolicyGate denial -- explore actions stay unlocked. Guidance only.

### 2. validate_stack nudge when current_best advances without opt_stack KEEPs (`coordinator.py`)

`params` / `backends` can lift `current_best.tput` without populating `optimization_stack` (variant_name + candidate_args not always paired), so the existing unvalidated-KEEP TODO never fires and the final report quotes `cumulative_gain_validated=0.0% ⚠ never validated`. A new TODO-5/5 branch fires when `cumulative_gain >= 0.5%`, the stack is empty, and validation hasn't run. Same guidance-only design -- no policy denial.

### 3. RUN_EVAL=false warning once per process (`_workload_envs.py`)

Was emitting per variant per round, drowning the log. Module-level flag gates it to one emission; tightened the message to spell out the consequence and override paths.

### 4. `event_counts.py --all` / `--limit` (`event_counts.py`)

The hard-coded 500-event window silently undercounted on long sessions (saw `deep_kernel_analysis=32` on a run that actually did 57). Default stays 500 for back-compat; `--all` scans full history, `--limit N` for custom windows.

### 5. Tests

- `test_required_step_gates.py`: covers the new 5-step numbering, the analyze gate transition, the existing PMC + integrate + validate_stack gates still firing.
- `test_phase2_mission_and_validate_stack.py`: 2 new tests pinning the +0.5% threshold for the current_best nudge.

## Out of scope (deliberately not in this PR)

- `inference_optimizer/scripts/install.sh` + `.gitignore` + `optimizer_runs/robustness_monitor.sh.example` -- already in PR #234.
- TraceLens `--strict-analysis-md` flag -- main's `088c63b8` already removed the fallback ladder unconditionally, making a flag redundant.
- Kernel-agent consumer-side strict gate (`KERNEL_AGENT_STRICT_ANALYSIS_MD_CONSUMER`) -- producer-side enforcement on main already covers this; consumer-side gate would also need test-fixture updates.
- Zombie-aware `proc_alive` for `robustness_monitor.sh.example` and AITER callsite normalization (`_normalize_callsite_path`) -- planned follow-ups.

## Test plan

- [x] Targeted suites green: `test_phase2_mission_and_validate_stack.py` (19), `test_required_step_gates.py` (38), `test_baseline_param_overrides.py`, `test_p2_2_profile_and_handlers.py`, `test_p5_decision.py`, `test_p2_4_integrate_report.py`, `test_tracelens_csv.py` -- 307 passes, 1 pre-existing failure on `origin/main` unrelated to this change.
- [x] Branch rebased on latest `origin/main` (550d24fc), three commits, clean apply.

### DIFF
--- a/inference_optimizer/actions/validate_stack.md
+++ b/inference_optimizer/actions/validate_stack.md
@@ -26,7 +26,7 @@ session has actually arrived.
 
 After a successful `backends` / `params` / `integrate` round produces a
 KEEP entry on `optimization_stack`, the Coordinator's
-`_required_next_step()` adds a `TODO 4/4: validate_stack required`
+`_required_next_step()` adds a `TODO 5/5: validate_stack required`
 entry to the per-tick checklist. Until Orchestration runs
 `validate_stack`, every other action is denied with
 `policy_denied{rule='validate_stack_required'}`. The TODO clears as
--- a/inference_optimizer/orchestrator/action_executors/_workload_envs.py
+++ b/inference_optimizer/orchestrator/action_executors/_workload_envs.py
@@ -52,6 +52,12 @@ from ._server_patcher import (
 
 log = logging.getLogger(__name__)
 
+# Module-level flag — the RUN_EVAL=false default is the expected steady
+# state on most checkouts (see docstring at materialize_config_with_envs);
+# the warning is for first-time awareness, not per-variant noise. Emit it
+# once per process to keep the log readable.
+_RUN_EVAL_DEFAULT_WARN_EMITTED = False
+
 
 def _tracelens_patch_enabled() -> bool:
     """Read the ``HYPERLOOM_ENABLE_PATCH`` kill switch (default on).
@@ -348,11 +354,17 @@ def materialize_config_with_envs(
             envs["RUN_EVAL"] = env_run_eval
         else:
             envs["RUN_EVAL"] = "false"
-            log.warning(
-                "RUN_EVAL defaulted to false: Magpie main / InferenceX main "
-                "disagree on `run_eval --concurrent-requests`. Export "
-                "RUN_EVAL=true once your InferenceX checkout accepts that flag."
-            )
+            global _RUN_EVAL_DEFAULT_WARN_EMITTED
+            if not _RUN_EVAL_DEFAULT_WARN_EMITTED:
+                log.warning(
+                    "RUN_EVAL defaulted to false (no per-variant accuracy "
+                    "gate): Magpie main / InferenceX main disagree on "
+                    "`run_eval --concurrent-requests`. Export RUN_EVAL=true "
+                    "(or pass extra_envs={'RUN_EVAL': 'true'}) once your "
+                    "InferenceX checkout accepts that flag. This warning "
+                    "fires once per process."
+                )
+                _RUN_EVAL_DEFAULT_WARN_EMITTED = True
     output_dir.mkdir(parents=True, exist_ok=True)
     materialized = output_dir / out_name
     with materialized.open("w", encoding="utf-8") as f:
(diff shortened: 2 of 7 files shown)


ZhengGong-amd:
Thanks for the thorough write-up — points 2–5 all look good to me. I do want to push back on point 1 (auto-emit select_kernels after every successful profile) before this lands.

Concern: profile and select_kernels shouldn't be hard-coupled

profile is a general-purpose action; select_kernels is conceptually paired with kernel_opt, not with profile. Auto-firing it on every profile completion creates a few problems:

1. Wasted work in no-kernel mode. When kernel optimization is disabled (e.g. --no-kernel flag), select_kernels should never run. A blanket "fire-and-forget after profile" trigger will still kick it off and burn time producing an analysis.md whose downstream consumer (kernel_opt) is gated off. Even if the action is cheap, it's non-zero, and it adds log noise + spurious TODO state on runs that explicitly opted out of the kernel path.
2. Wrong coupling axis. The dependency graph today is roughly profile → {params, backends, kernel_opt} and kernel_opt → select_kernels → analysis.md. Tying select_kernels to profile collapses that into profile → select_kernels for all runs, which leaks kernel-path concerns into the generic profile step.
3. The existing logic already handles the "only when needed" case. The orchestrator already knows when kernel_opt is in the plan; that's the natural gate for emitting select_kernels. We don't need a profile-side auto-trigger to guarantee analysis.md is produced — we just need select_kernels to be reliably emitted on the kernel_opt path.

rodosingh:
Thanks for the careful review — pushed an update.

**Removed the auto-emit on profile completion entirely** (your point 1). Specifically dropped:

- `_maybe_auto_emit_select_kernels_after_profile` + `_auto_emit_select_kernels` in `coordinator.py`
- `auto_analyze_after_profile` ctor param + `HYPERLOOM_AUTO_ANALYZE_AFTER_PROFILE` env handling
- `--auto-analyze-after-profile` / `--no-auto-analyze-after-profile` CLI flags
- The call site in `_promote_to_shared_state`
- All "P1 auto-trigger" descriptions in the prompt + tests

I agree on the coupling argument — `profile` should stay generic, and the natural gate for `select_kernels` is the kernel_opt path that already has `_sequence_denial_for_request` enforcing the cache.

What I kept (and would like to confirm is OK with you):

- **TODO 3/5 (analyze) as pure guidance** in `_required_next_step`. It only fires when `kernel in role_registry` and `last_profile_trace` is set but `last_select_kernels` is stale. No PolicyGate denial — explore actions remain unlocked. It's just a prompt nudge that surfaces the existing dependency you described ("we just need select_kernels to be reliably emitted on the kernel_opt path"); reframed it in the docstring/comments accordingly.
- The TODO 4/4 → 5/5 renumbering, since TODO 3/5 still exists.
- TODO 5/5 (current_best path) validate_stack nudge from point 2 — unchanged.

Re the github-code-quality inline comment about `_RUN_EVAL_DEFAULT_WARN_EMITTED`: false positive — the flag is declared at module scope, mutated via `global ... = True` inside the function, and read in the `if not ...` immediately above. Standard global-once pattern; the bot's static analyzer misses the `global` keyword. Leaving as-is per your "points 2–5 all look good to me".

Rebased on latest `origin/main` (a2c29a7f). 207 targeted tests pass.

ZhengGong-amd (approved):
Code checked. rodosingh tested the codes.

PR #279, opened 2026-05-21
Feature/zhanglei/framework agent

zhanglei-amd:


ZhengGong-amd (approved):
First update of framework agent. Validated by Zhang, Lei

PR #290, opened 2026-05-25
breakdown: mirror session_breakdown.json into claw-synced subtree

luochen-amd:
## Summary

- Mirror `session_breakdown.json` into `<workspace_root>/hyperloom/<session_id>/` so the claw sandbox checkpoint sync (which only watches the `hyperloom/` subtree) picks it up. Without this, the canonical file under `<workspace>/<model>/<ts>/` is lost the moment the sandbox is reaped, breaking `claw-stats-service` / offline forensics for any historical session.
- Mirror is best-effort and never raises: failures are swallowed + logged so the real `stop_reason` propagation stays intact.
- Helper lives in its own module `breakdown/claw_mirror.py` rather than as a private helper on `cli.py`, so the unit tests can import it without dragging in cli's `orchestrator/action_executors` graph (which imports `fcntl` and other POSIX-only modules).

## Why now

PR #281 (v1.1 schema) just landed on `main`, so the canonical sbd write is in place — this PR closes the last gap by making sure the artifact survives sandbox death and lands somewhere the claw S3 sync can see.

## Diff scope

- `inference_optimizer/cli.py` — +7 lines in the existing `finally` block.
- `inference_optimizer/breakdown/claw_mirror.py` — new, ~30 lines.
- `inference_optimizer/tests/test_cli_breakdown_claw_mirror.py` — new, 4 tests.

## Test plan

- [x] `pytest inference_optimizer/tests/test_cli_breakdown_claw_mirror.py` — 4/4 pass.
- [x] `pytest inference_optimizer/tests/test_breakdown_*.py inference_optimizer/tests/test_cli_breakdown_claw_mirror.py` — 57/57 pass (no breakdown-suite regressions).
- [x] Lint clean.
- [ ] Post-merge: smoke-check on a CI-driven hyperloom run that `<workspace>/hyperloom/<sid>/session_breakdown.json` shows up in claw's MinIO bucket and is fetchable through `claw-api`.

## Notes for reviewers

- The mirror copies the canonical file rather than re-running `build()` — `build()` is idempotent but not free (re-walks all collectors), and `shutil.copy2` preserves mtime which keeps S3 sync diffs sane.
- Path uses only the hyperloom `session_id`. `claw_session_id` is not always set (e.g. local repro runs); downstream consumers can recover it from the sbd JSON itself if they need to reverse-map.
- Empty `session_id` is a deliberate no-op (would otherwise collapse the dir into `hyperloom/` itself).

### DIFF
--- a/inference_optimizer/breakdown/claw_mirror.py
+++ b/inference_optimizer/breakdown/claw_mirror.py
@@ -1,0 +1,55 @@
+"""Mirror ``session_breakdown.json`` into the claw-synced subtree.
+
+Hyperloom's canonical write lands in ``<session_dir>/`` which under the
+default ``per_model_ts`` layout is ``<workspace>/<model>/<ts>/``. The
+claw sandbox checkpoint sync only watches the ``hyperloom/`` subtree
+under ``$USER_DATA_PATH``, so without an explicit mirror the canonical
+artifact is lost the moment the sandbox is reaped.
+
+This module exposes a single best-effort helper, kept off ``cli.py`` so
+the unit tests can exercise it without dragging in cli's full import
+graph (``orchestrator/action_executors`` pulls ``fcntl`` and friends
+that aren't available on every dev box).
+
+Contract:
+
+* On success, returns the absolute mirror path.
+* On any failure (empty ``session_id``, missing source, unwritable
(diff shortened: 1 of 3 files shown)


ZhengGong-amd (approved):
Limited  revision  to session breakdown. Approved

PR #303, opened 2026-05-25
PR-K: source attribution + per-source attempts + aiter jit cache invalidation

xiaofei-zheng:
## Summary

Fixes two bugs in the kernel-opt dispatch chain that caused Hyperloom on Qwen3-30B-A3B-Base to silently waste hot-MoE-kernel attempts:

1. **TraceLens attributes ck_moe / topk_softmax / moe_align_block_size etc. to the python `@compile_ops` wrapper** (`aiter/ops/moe_op.py`). Patching the wrapper has zero runtime effect — the `@compile_ops` decorator dispatches to a JIT-compiled `.so` under `<aiter>/jit/build/module_*/` that bypasses the Python wrapper entirely. Pre-PR-K kernel-opt sessions wasted GEAK budget on wrapper rewrites that produced 0 KEEP.
2. **`_batch_kernel_candidates._is_live` consults a cumulative `attempts` counter**, so a single failed wrapper attempt locked every member of the kernel's task_group to `not_live` on the next batch — even though the device source `.cu` had never been tried. Qwen3-30B-A3B-Base session 20260523T162026Z burned 2 h producing `group_exhausted` for every reusable hot kernel.

PR-K is split into 4 independent commits, each self-contained (apply / revert friendly):

| commit | scope | files |
|---|---|---|
| 1/4 (D) | Move `<aiter>/jit/build/` aside before `setup.py develop` so the post-rebuild first import re-codegens + re-compiles cleanly. mv-based, atomic, zero-copy on same fs. Restore on revert; clear regenerated dir first so pre-patch state is restored bit-for-bit. **Aiter-only** (sglang/vllm have no JIT codegen layer; no-op for those targets). | `apply_kernel_patch.py` + 20-test new file |
| 2/4 (A) | `upgrade_aiter_compile_ops_launcher`: kernel-name allowlist (ck_moe_stage1/2, topk_softmax, topk_softmax_group, moe_align_block_size, moe_fused_gate) → device codegen `.cu` under `csrc/`. Falls back to /sgl-workspace/aiter for wheel-install layouts. `_finalize_candidates` stamps `launcher_source_file` + `source_promoted_from_launcher=True` on promotion. | `tracelens_analysis.py` + 13-test new file |
| 3/4 (B) | `build_prompt` renders a `SOURCE ATTRIBUTION NOTE — READ FIRST` block above the kernel_metadata JSON when `source_promoted_from_launcher=True`, with three hard rules telling the LLM not to modify the wrapper, the device source may be a codegen entry, the orchestrator clears matching jit/build/ entries before rebuild. Un-promoted candidates produce a byte-identical prompt to pre-PR-K. | `kernel_optimization.py` + 6-test new file |
| 4/4 (C) | `record_kernel_opt` writes a per-source `attempts_per_source[source_file]` ledger alongside cumulative `attempts`. `_is_live(kid, current_source)` consults `attempts_per_source[current_source]` first; falls back to cumulative for legacy v1 entries (resume from a state.json that predates this PR). Wrapper attempts no longer lock out a fresh device-source attempt. | `shared_state.py` + `kernel_request_handlers.py` + 7-test new file |

## Test plan

- [x] 46 new unit tests across the 4 commits (D=20, A=13, B=6, C=7), all passing
- [x] Full regression on related modules: `apply_kernel_patch`, `tracelens`, `kernel_optimization`, `shared_state`, `kernel_request_handlers`, `integrate`, `p2_2`, `p2_4` — **389 passed, 0 regressions**
- [x] Lint clean
- [x] Schema-compatible: legacy `state.json` from before this PR resumes byte-for-byte (no surprise reattempts on resume)
- [x] **Production dry-run on Qwen3-30B-A3B-Base** (session 20260524T020943Z, 6h41min, 19 ticks):
  - A verified: 4 ck_moe kernels (k001-k004) all promote `aiter/ops/moe_op.py` → `csrc/.../gemm_moe_ck2stages.cu`
  - B verified: SOURCE ATTRIBUTION NOTE renders in GEAK prompt (`prompts/geak-c5d9823b.md`) with all three hard rules
  - C verified: `kernel_opt_attempts[k001-k004].attempts_per_source['..gemm_moe_ck2stages.cu']=1` written to `state.json`
  - C verified end-to-end: after k001/k004 reject, k002/k003 dispatched in next batch (was `group_exhausted` pre-PR-K)
  - D **NOT verified in production**: 0 KEEP candidates → 0 integrate calls → `_invalidate_aiter_jit_build` never reached. Only single-test coverage.

## Known limitations / scope

- **Does NOT solve the GEAK-can't-rewrite-CK-codegen-`.cu` issue**: this PR fixes the upstream attribution + ledger bugs so GEAK at least sees the right rewrite target with a fresh attempt budget. Whether GEAK can actually produce a useful patch on `gemm_moe_ck2stages.cu` is a separate problem (left to GEAK side). The session above produced 0 KEEP because GEAK timed out at 5580s/93min on every device-source attempt without producing a `.cu` rewrite.
- **D is a logical-necessity fix without production trigger data**: aiter's `@compile_ops` persistent `.so` cache means a patched `csrc/.../gemm_moe_ck2stages.cu` would otherwise rebuild the wheel but the next import would still load the pre-patch `module_moe_ck2stages_*.so` from `jit/build/` — so the patch silently no-ops at runtime. B-prompt explicitly promises the LLM that "the orchestrator clears the matching jit/build/ entries before rebuild"; removing D would make that prompt lie. Single-test coverage exercises mv/restore boundaries (target detection, missing/empty/populated jit/build, backup-collision refusal, restore round-trip, restore clears regenerated dir, full apply→revert E2E with mocked rebuild, skip_rebuild noop), but the end-to-end "patched .cu actually takes effect on next import" link can only be verified once GEAK produces a KEEP.
- **Promotion allowlist is intentionally narrow**: only kernels we've validated against real aiter trees (ck_moe / topk_softmax / moe_align_block_size / moe_fused_gate). Anything else falls through with the wrapper unchanged so the LLM still gets a valid signal. Add new entries to `_AITER_COMPILE_OPS_PROMOTIONS` as new high-leverage MoE-style ops surface.
- **Multi-node**: aiter csrc patch + jit_build invalidation operate on the sandbox node only. Existing `_dispatch_multinode_apply` fan-out still mirrors the source patch to every pod; pod-side jit caches are invalidated at server restart in `integrate_handler`, not by this PR.

## Production session as evidence

| metric | pre-PR-K (162026Z) | this PR (020943Z) |
|---|---|---|
| Run time | 2h 11min | 6h 41min |
| Ticks | 8 | 19 |
| ck_moe members tried | k001 + k002 (wrapper, group_exhausted at batch 2) | k001-k004 all (device .cu) |
| GEAK source target | `aiter/ops/moe_op.py` | `csrc/.../gemm_moe_ck2stages.cu` |
| Group fallback | locked at `group_exhausted` | k002/k003 auto-dispatched after k001/k004 reject |
| validated_gain | 0% | 0.26% (from `--attention-backend aiter`, not kernel-opt) |
| stop_reason | report_emitted (LLM gave up early) | report_emitted (ran out of leverage to test) |

PR-K does not deliver a positive throughput gain on this model on this run — that requires GEAK to produce a working CK kernel rewrite, which is out of scope. PR-K's contribution is letting the dispatch chain reach that decision instead of being silently locked out at the wrong target.

## Branch hygiene

- Cherry-picked from `feature/xiaofei/in-flight-pid-liveness-check` (PR-J) to a clean branch on top of `origin/main` (`3e016e1`); resolved one trivial conflict in `kernel_optimization.py::build_prompt` against the PR-301 budget_min change (kept both — backend-specific budget_min + new promotion_block).

### DIFF
--- a/inference_optimizer/orchestrator/kernel_request_handlers.py
+++ b/inference_optimizer/orchestrator/kernel_request_handlers.py
@@ -1015,16 +1015,43 @@ def _batch_kernel_candidates(
                 session_dir,
             )
 
-    def _is_live(kid: str) -> bool:
+    def _is_live(kid: str, current_source: str = "") -> bool:
         """A kernel_id is live (eligible for batch) iff it is NOT
         rejected, NOT in-flight, and has fewer than max_attempts
-        recorded attempts. ``max_attempts = 1`` (default) means: any
-        prior attempt at all -> not live."""
+        recorded attempts AGAINST THE CURRENT CANDIDATE'S source_file.
+
+        ``max_attempts = 1`` (default) means: any prior attempt against
+        the same source_file -> not live, but a prior attempt against a
+        DIFFERENT source_file is ignored. This is what lets PR-K's
+        launcher → device source promotion unlock a fresh attempt:
+        when ``aiter::ck_moe_stage1`` was first dispatched against the
+        python wrapper ``aiter/ops/moe_op.py`` and PARTIAL'd, a
+        subsequent dispatch with ``current_source`` pointing at the
+        promoted device file ``csrc/.../gemm_moe_ck2stages.cu`` is
+        treated as a fresh target with its own quota. Without this,
+        the wrapper's first failed attempt would lock the entire
+        task_group as ``group_exhausted`` even though the device path
+        had never been tried (Qwen3-30B-A3B-Base session
+        20260523T162026Z burned 2 hours on this).
+
+        Falls back to the cumulative ``attempts`` counter when:
+          * ``current_source`` is empty (legacy callers / synthetic
+            test fixtures that don't carry source_file);
+          * the entry was written by a v1 ``record_kernel_opt`` that
+            predated ``attempts_per_source`` (resumed state.json from
+            before this PR).
+        Both fallbacks preserve the pre-PR-K behaviour byte-for-byte.
+        """
         if kid in rejected_kernel_ids:
             return False
         if kid in in_flight:
             return False
         entry = attempts_by_kid.get(kid) or {}
+        if current_source:
+            per_source = entry.get("attempts_per_source")
+            if isinstance(per_source, dict):
+                src_attempts = int(per_source.get(current_source, 0))
+                return src_attempts < max_attempts
         if int(entry.get("attempts", 0)) >= max_attempts:
             return False
         return True
@@ -1072,7 +1099,7 @@ def _batch_kernel_candidates(
             primary_cand is not None
             and primary_cand.get("reusable_native_kernel") is True
             and bool(primary_cand.get("source_file"))
-            and _is_live(primary)
+            and _is_live(primary, str(primary_cand.get("source_file") or ""))
         )
         if not primary_live:
             primary_cand = next(
@@ -1082,7 +1109,7 @@ def _batch_kernel_candidates(
                     if m in kernel_by_id
                     and kernel_by_id[m].get("reusable_native_kernel") is True
                     and kernel_by_id[m].get("source_file")
-                    and _is_live(m)
+                    and _is_live(m, str(kernel_by_id[m].get("source_file") or ""))
                 ),
                 None,
             )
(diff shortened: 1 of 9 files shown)


ZhengGong-amd (approved):
Revision to kernel agents, validated by xiaofei. Approved.

PR #313, opened 2026-05-26
feat(orchestrator): default integrate_handler inputs from SharedState

### LINKED ISSUE #319: integrate_handler early-outs on missing base_tput even when SharedState has it
## Problem

`integrate_handler` in `inference_optimizer/orchestrator/kernel_request_handlers.py` rejects payloads with `integrate_handler requires base_tput > 0` *before* `_resolve_integrate_payload` runs. The existing resolver already defaults `patch_path` and `source_file` from SharedState, but the other three commonly-omitted inputs — `base_tput`, `config_path`, `extra_sglang_args` — are also already on disk:

- `state.baseline_tput`
- `state.baseline_config_path`
- `state.current_best["extra_sglang_args"]`

…yet the hard early-out forces the KEEP-queue replay path to re-plumb `base_tput` through every Orchestration call. The natural payload after a successful `run_optimization` is just `{"kernel_id": ...}`, and that should be enough.

## Scope

- File: `inference_optimizer/orchestrator/kernel_request_handlers.py`
- Severity: forces redundant plumbing on a common call path; "missing input" failure mode must be preserved
- No public-contract change — explicit payload values must continue to win

## Acceptance

- A payload of `{"kernel_id": ...}` succeeds when SharedState has the three fields.
- Explicit payload values always override SharedState defaults.
- When neither payload nor state has `base_tput`, the existing hard-error still fires with the same wording.
- Helper never mutates the caller's payload.

Fixed by #313.

sdubagun-amd:
## Summary

- Add `_fill_integrate_defaults_from_state` helper in [inference_optimizer/orchestrator/kernel_request_handlers.py](inference_optimizer/orchestrator/kernel_request_handlers.py).
- Run it at the *top* of `integrate_handler` — **before** the hard `base_tput > 0` early-out — so a payload of just `{"kernel_id": ...}` can complete by pulling `base_tput`, `config_path`, and `extra_sglang_args` from SharedState.
- 10 new tests in [test_integrate_payload_defaults.py](inference_optimizer/tests/test_integrate_payload_defaults.py).

## Motivation

The existing `_resolve_integrate_payload` already defaults `patch_path` and `source_file` out of SharedState. That's two of the five inputs Orchestration tends to omit when it sends `{"kernel_id": ...}` after a successful `run_optimization`. The other three — `base_tput`, `config_path`, `extra_sglang_args` — are also already on disk:

- `state.baseline_tput`
- `state.baseline_config_path`
- `state.current_best["extra_sglang_args"]`

…but `integrate_handler` short-circuits with `integrate_handler requires base_tput > 0` *before* the existing resolver runs, so the KEEP-queue replay path has to re-plumb `base_tput` through every Orchestration call. This PR closes that gap by running the new helper first.

## Behavior

- Explicit payload values always win. Helper only fills when the field is missing or (for `base_tput`) zero.
- Helper returns a shallow copy of the input dict; never mutates the caller's payload.
- When neither payload nor state has a `base_tput` the existing hard-error still fires with the same wording, so the "missing input" failure mode is preserved.

## Tests

```
pytest inference_optimizer/tests/test_kernel_integrate_and_report.py \
       inference_optimizer/tests/test_integrate_payload_defaults.py \
       inference_optimizer/tests/test_profile_and_kernel_handlers.py -q
# 126 passed in 31.90s
```

New tests:
- 8 helper tests: all-three-fired / each-field-wins (3) / empty-state-no-op / shallow-copy / zero-base_tput triggers fallback / zero-state does not overwrite explicit payload.
- 2 `integrate_handler` wiring tests: missing `base_tput` in payload + present in state advances past the early-out; missing in both still fails with the unchanged error message.

## Test plan

- [x] All 116 existing tests in `test_kernel_integrate_and_report.py` + `test_profile_and_kernel_handlers.py` still pass alongside the 10 new tests.
- [x] No change to public payload contract; only adds defaulting for previously-required fields.
- [ ] Reviewer sanity-check: visually trace `integrate_handler` to confirm the helper runs once at the top and never inside the `_resolve_integrate_payload` early-return path.

Closes #319

### DIFF
--- a/inference_optimizer/orchestrator/kernel_request_handlers.py
+++ b/inference_optimizer/orchestrator/kernel_request_handlers.py
@@ -497,6 +497,46 @@ def _find_selected_kernel_source(state: Any, kernel_id: str) -> str:
     return ""
 
 
+def _fill_integrate_defaults_from_state(
+    payload: dict, *, session_dir: Path,
+) -> dict:
+    """Pull ``base_tput`` / ``config_path`` / ``extra_sglang_args`` defaults from SharedState.
+
+    Sibling of ``_resolve_integrate_payload`` but runs *before* the
+    ``base_tput > 0`` hard-check at the top of ``integrate_handler``.
+    ``_resolve_integrate_payload`` already handles ``patch_path`` /
+    ``source_file`` defaulting; the three fields filled here are the
+    other Magpie re-baseline inputs that Orchestration tends to omit
+    when it sends a bare ``{"kernel_id": ...}`` payload.
+
+    Always returns a (shallow) copy of ``payload`` so the caller can
+    treat it as a fresh dict; never raises on a missing SharedState
+    snapshot (returns the input dict unchanged in that case).
+    """
+    from .shared_state import SharedState
+
+    resolved = dict(payload)
+    state = SharedState.load_or_init(session_dir)
+
+    if float(resolved.get("base_tput", 0.0) or 0.0) <= 0:
+        bt = float(getattr(state, "baseline_tput", 0.0) or 0.0)
+        if bt > 0:
+            resolved["base_tput"] = bt
+
+    if not resolved.get("config_path"):
+        cfg = getattr(state, "baseline_config_path", "") or ""
+        if cfg:
+            resolved["config_path"] = cfg
+
+    current_best = getattr(state, "current_best", None) or {}
+    if not resolved.get("extra_sglang_args") and isinstance(current_best, dict):
+        cb_args = current_best.get("extra_sglang_args") or ""
+        if cb_args:
+            resolved["extra_sglang_args"] = cb_args
+
+    return resolved
+
+
 def _resolve_integrate_payload(payload: dict, *, session_dir: Path) -> tuple[dict, HandlerResult | None]:
     """Fill integrate inputs from SharedState when Orchestration sends only kernel_id.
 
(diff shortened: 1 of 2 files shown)


chaojhou (approved):
## Approve — small, focused, correctly placed, well-tested

Adds `_fill_integrate_defaults_from_state` to backfill `base_tput` / `config_path` / `extra_sglang_args` from SharedState at the very top of `integrate_handler`, before the `base_tput > 0` hard-check. This closes the gap left by `_resolve_integrate_payload` (which only defaults `patch_path` / `source_file`), so a bare `{kernel_id}` replay no longer has to re-plumb `base_tput` through every Orchestration call.

### Correctness (traced line-by-line)
- **Placement is right**: the helper runs once at the top, before the `base_tput > 0` check and before `_resolve_integrate_payload`. The reviewer self-check in the PR description holds.
- **Explicit-wins semantics correct**: `base_tput` is gated on `<= 0` (fills only when missing/0/None); `config_path` / `extra_sglang_args` are gated on falsy. Explicit payload values always win.
- **No input mutation**: shallow `dict(payload)` copy; locked by `test_returns_shallow_copy_not_mutating_input`.
- **Disjoint fields**: the two resolvers fill non-overlapping keys, so there is no override interaction.
- **Error contract preserved**: when neither payload nor state has `base_tput`, the original `requires base_tput > 0` message still fires (locked by `test_no_base_tput_anywhere_still_fails_with_clear_error`).
- **Tests**: 8 helper unit tests + 2 handler wiring tests, covering per-field defaulting, explicit-wins, empty-state no-op, shallow-copy, zero-fallback, and the wiring past/at the early-out.

### CI
All code checks are green (`test`, `Analyze (python)`, `analyze`, `CodeQL`). The only red check — `project-automation` — fails at the "Get project data" step (Project Board Automation / Projects GraphQL); it is infra/permissions and unrelated to this change.

### Non-blocking nit
`SharedState.load_or_init()` is now called twice per `integrate_handler` (once in the new helper, once in `_resolve_integrate_payload`). Functionally fine. A cleaner shape would fold the three-field defaulting into `_resolve_integrate_payload` and move the `base_tput > 0` check after it (single load, single resolver) — but the current split intentionally avoids touching the existing resolver's early-return contract, a reasonable minimal-risk tradeoff. Not a merge blocker.

PR #320, opened 2026-05-26
fix(kernel-opt): stop bloating prompts with full analysis.md; filter non-routable kernels from candidates

### LINKED ISSUE #307: kernel-opt prompt: full tracelens/analysis.md is appended unnecessarily, bloats GEAK prompt
## Summary

The GEAK kernel-opt backend prompt assembled by `build_prompt` in `kernel-agent/tools/kernel_optimization.py` appends the **entire** per-run `tracelens/analysis.md` report as a trailing `## TraceLens Context` section. That report is ~300+ lines and is mostly redundant — the per-kernel TraceLens signal we actually want the agent to act on is **already extracted and emitted near the top of the prompt** by `_format_tracelens_hypothesis` (the `## TraceLens Hypothesis [validate before acting]` block) and the `## Optimization priorities (TraceLens bound: ...)` block. The appended report bloats the prompt with content for unrelated P-items that the same prompt explicitly tells the agent NOT to optimize.

## Where the bloat is added

In `kernel-agent/tools/kernel_optimization.py` (HEAD `c0a17c1`), `build_prompt` constructs `tracelens_context_block` by reading the full report off disk:

- L1298–1324: the block is built as `\"\\n## TraceLens Context\\n\\n\" + focus_line + \"\\n\" + full_report`, where `full_report = report_path.read_text(...)` is the entire `analysis.md`.
- L1357: that block is then joined into the final prompt as the last section.

The very `focus_line` we emit immediately above the dump is itself a red flag:

> `Focus on **P{rank}: {title}** in the report below. Other P-items are context only — do not optimize them.`

If \"other P-items are context only — do not optimize them\", we're shipping ~300 lines of \"do-not-optimize\" content to the model on every kernel-opt run.

## Why it is redundant

The per-kernel TraceLens hypothesis is already on the prompt's top section, populated from the same source:

- `_format_tracelens_hypothesis` (around L819 / L870) emits `## TraceLens Hypothesis [validate before acting]` with **Identification**, **Reasoning for slowdown**, **Recommended direction**, and the numeric **Estimated impact range** — i.e. the prose tuple already lifted out of the report for the specific P-item this kernel belongs to.
- `## Optimization priorities (TraceLens bound: <bound_type>)` (around L737) emits the ordered lever list tied to the kernel's bound type.
- The kernel's own shapes are already on the prompt as `Shapes: [...]` and inside the `kernel_metadata` JSON (`input_shapes`, `kernel_params`, etc.).

So everything the prompt is asking the agent to act on for *this* kernel is already in the body. The appended report mainly carries:

- Findings for **other** categories (e.g. GEMM compute-bound, elementwise `silu_and_mul`) — explicitly out of scope per the focus line.
- Fusion candidates and system-level (host/device sync, GPU graph) recommendations — not per-kernel actionable inside a kernel-opt run.
- The appendix (hardware reference, model arch).

## Concrete example

From a recent local Qwen3-32B run (`rmsnorm` P2 focus):

| File | Lines |
|---|---|
| `tracelens/analysis.md` | 323 |
| `prompts/geak-20a9e3d7.md` (final GEAK prompt) | **823** |

In that prompt:

- L1–L495 — the actual kernel-opt prompt (task, metadata JSON, hardware notes, TraceLens **Hypothesis** for P2, **Optimization priorities** for memory-bound, repo/safety rules, source content).
- **L497–L823 — the appended `## TraceLens Context` section**, i.e. the full `analysis.md` verbatim (~40% of the whole prompt).

Side-by-side, the P2 prose at L94–L102 of the prompt (Identification / Reasoning / Recommended direction / Impact estimate for `aiter::rmsnorm` narrow-feature shape) is identical to the P2 prose at L667–L686 of the appended report. The rest of the appended report (P1 GEMMs, P3 silu_and_mul, kernel fusion P1, system-level P1/P2/P3, appendix) is content the agent is told to ignore.

## Suggested fix

Drop the `tracelens_context_block` from the `return` join of `build_prompt`. The `## TraceLens Hypothesis` and `## Optimization priorities` blocks already carry the per-kernel TraceLens signal in a compact form, and the kernel's own shapes/metadata are already on the prompt body.

If we want to keep a fallback path for runs that produce *no* per-kernel hypothesis (e.g. raw-trace / csv-fallback with no `task_group`), gate the dump on \"hypothesis block is empty\" rather than \"trace report exists on disk\", so the full report only appears when there's truly nothing else to anchor the agent on.

## Impact

- Smaller, cheaper, more focused prompts for every GEAK kernel-opt invocation (current ~40% reduction on the example above).
- Less off-topic context for the LLM to wander into (other P-items are flagged \"do not optimize them\" but still influence sampling).
- No information loss for the on-kernel task — the relevant prose is already on the prompt's top section.

## Repro / references

- Code: `kernel-agent/tools/kernel_optimization.py` — `build_prompt`, L1298–1324 and L1357 (HEAD `c0a17c1`).
- Example prompt: any `runs/<ts>/prompts/geak-*.md` produced by the kernel-opt backend when a tracelens report is present at `runs/<ts>/tracelens/analysis.md` — the trailing `## TraceLens Context` section is the duplication.

### LINKED ISSUE #314: kernel_candidates.json::hot_kernels is unfiltered; should only carry kernels routable to a kernel-opt backend (GEAK/OOB)
## Summary

The `hot_kernels[]` array in `runs/<ts>/kernel_candidates.json` is a byte-identical pass-through of `runs/<ts>/tracelens/tracelens_report.json::hot_kernels[]`. The filename implies \"candidates ready for kernel-opt dispatch\", but in practice it carries the union of:

- routable kernels — `reusable_native_kernel=True`, real `source_file`, routable to GEAK / OOB / Claude / Codex.
- non-routable kernels — `reusable_native_kernel=False`, with a populated `skip_reason` (e.g. `\"source file not resolved\"`, `\"vendor binary (no rewritable source)\"`, `\"PyTorch native op 'aten::mm' backed by Tensile\"`).

The filter that distinguishes the two already exists (`classify_patchability` in `kernel-agent/tools/tracelens_analysis.py:745`), enriches every candidate with `reusable_native_kernel: bool` + `skip_reason: str`, and is correctly used to build the `tracelens/summary.json` audit sidecar (`tasks[]` vs `skipped[]`). It just isn't applied to the `hot_kernels[]` we hand to the dispatch path.

## Where the pass-through happens

In `kernel-agent/tools/tracelens_analysis.py::write_reports` (HEAD `c0a17c1`, L1986–1991):

```python
atomic_write_json(run_dir / \"trace_input_manifest.json\", manifest)
atomic_write_json(tracelens_dir / \"tracelens_report.json\", report)
atomic_write_json(
    run_dir / \"kernel_candidates.json\",
    {\"hot_kernels\": candidates, \"task_groups\": task_groups, **report},
)
```

The same `candidates` list (unfiltered, already enriched with `reusable_native_kernel` / `skip_reason` at L1540) is written into both `tracelens/tracelens_report.json` and `runs/<ts>/kernel_candidates.json`. Only the latter (`kernel_candidates.json`) is consumed by the kernel-opt dispatch path.

## Concrete example

Qwen3-32B / sglang / MI300X run (`runs/20260525T120517Z/`):

| Artifact | `hot_kernels[]` length | Routable | Non-routable |
|---|---:|---:|---:|
| `tracelens/tracelens_report.json` | 10 | 2 | 8 |
| `kernel_candidates.json` | **10** | 2 | 8 |
| `tracelens/summary.json` | — | `task_count`: **2** | `skipped_count`: **8** |

The 8 non-routable entries are `aten::mm` GEMMs (`k001`–`k008`), every one with `source_file=\"\"`, `kernel_name=None`, `kernel_path=None`, `ast_resolved=None`, and `skip_reason=\"source file not resolved\"` in `summary.json`. The 2 routable entries are the AITER rmsnorm kernels (`k009`, `k010`) with `source_file=/sgl-workspace/aiter/aiter/ops/rmsnorm.py` and `recommended_backends=[\"geak\", \"claude\", \"codex\"]`.

`tracelens/summary.json` already gets the split right — it just isn't reflected in `kernel_candidates.json`.

## Cost of leaving it unfiltered

Because `kernel_candidates.json` is the file every downstream consumer treats as the dispatch payload, each consumer has to re-apply the filter independently. Examples in HEAD:

- `inference_optimizer/orchestrator/shared_state.py:1298` skips entries where `reusable_native_kernel is not True` when building the optimization queue.
- `inference_optimizer/orchestrator/kernel_request_handlers.py:363` (`_validate_reusable_native_kernel`) is a runtime guard that fails the dispatch with `error_class=\"non_reusable_kernel\"` when a non-routable kernel sneaks through.
- `kernel-agent/tools/kernel_optimization.py:97` (`load_candidates`) does NO filtering — it returns the full list, and `find_candidate` happily returns a `source_file=\"\"` candidate by `kernel_id`.
- `kernel-agent/tools/parallel_e2e_runner.py:398–407` reads `hot_kernels` unfiltered into `kernel_candidates.json`, then leans on `choose_candidate` for selection.

Net effect: the filter is duplicated across consumers, drift between them is invisible at load time, and any new consumer that reads `kernel_candidates.json::hot_kernels` directly is one accidental missed check away from dispatching a `source_file=\"\"` kernel to GEAK/OOB and getting a generic \"source file not resolved\" / `non_reusable_kernel` failure mid-run.

There is also a smaller cosmetic cost: status / prompt / log lines that report `num_hot_kernels` (e.g. `parallel_e2e_runner.py:422`) overcount by ~5× on this run (10 vs. 2), which is misleading in dashboards and resume logic.

## Suggested fix

Filter `hot_kernels` at the write site so `kernel_candidates.json` carries only routable kernels by default. The unfiltered list stays available in two existing places (`tracelens_report.json` and `summary.json::skipped[]`), so no audit information is lost.

Minimum change in `kernel-agent/tools/tracelens_analysis.py::write_reports`:

```python
routable = [
    c for c in candidates
    if isinstance(c, dict) and c.get(\"reusable_native_kernel\") is True
]
atomic_write_json(tracelens_dir / \"tracelens_report.json\", report)  # unchanged: full audit list
atomic_write_json(
    run_dir / \"kernel_candidates.json\",
    {
        **report,
        \"hot_kernels\": routable,             # filtered: dispatch payload
        \"skipped_kernels\": [                 # explicit audit half, with reasons
            {
                \"kernel_id\": c.get(\"kernel_id\"),
                \"name\": c.get(\"name\"),
                \"skip_reason\": c.get(\"skip_reason\") or \"unknown\",
            }
            for c in candidates
            if isinstance(c, dict) and c.get(\"reusable_native_kernel\") is not True
        ],
        \"task_groups\": task_groups,
    },
)
```

Pair changes worth considering:

- Update `kernel-agent/tools/kernel_optimization.py::load_candidates` to apply the same filter defensively, so older `kernel_candidates.json` files written before the fix still produce a routable dispatch list at read time.
- Update `parallel_e2e_runner.py`'s `num_hot_kernels` accounting to use the filtered length (or report both routable / skipped counts) so dashboards stop conflating \"kernels TraceLens noticed\" with \"kernels we will actually attempt\".

## Why now

This is the same data plumbing called out in #307 (the kernel-opt prompt was carrying the full `analysis.md` even though the per-kernel signal was already extracted at the top). Both bugs share the underlying shape: the audit / source-of-truth list and the dispatch payload are being conflated. Fixing them together gives kernel-opt a clean contract: `kernel_candidates.json` = what we send to a backend; `tracelens_report.json` + `tracelens/summary.json` = the full audit trail with skip reasons.

## Repro / references

- Code (HEAD `c0a17c1`):
  - `kernel-agent/tools/tracelens_analysis.py:745` — `classify_patchability` (the routing gate)
  - `kernel-agent/tools/tracelens_analysis.py:1540` — per-candidate enrichment with `reusable_native_kernel` / `skip_reason`
  - `kernel-agent/tools/tracelens_analysis.py:1988-1991` — pass-through write of `kernel_candidates.json`
  - `kernel-agent/tools/tracelens_analysis.py:1832` — `build_audit_summary` (already does the right split, but only into `summary.json`)
  - `kernel-agent/tools/kernel_optimization.py:97` — `load_candidates` (consumer, no filter)
  - `inference_optimizer/orchestrator/kernel_request_handlers.py:363` — runtime guard that fails on non-routable kernels
- Example artifacts: any `runs/<ts>/` with both `tracelens/tracelens_report.json` and `kernel_candidates.json` will show the identical `hot_kernels[]` lengths; the Qwen3-32B run quoted above is one such case (10 / 10 / 2 routable).

rodosingh:
## Summary

Fixes two kernel-opt data-plumbing bugs uncovered by a Qwen3-32B / sglang / MI300X run. Both share the same shape — the audit / source-of-truth list and the dispatch payload were being conflated — so they're worth addressing together.

- Closes #307 — full \`tracelens/analysis.md\` was unconditionally appended to every kernel-opt prompt (~300 lines, ~40% of the prompt body in the example Qwen3-32B run), even though the per-kernel TraceLens prose was already emitted at the top via \`_build_hypothesis_block\`. The dump also surfaced other P-items that the prompt itself flags as out of scope (\"Other P-items are context only — do not optimize them\").
- Closes #314 — \`kernel_candidates.json::hot_kernels\` was a byte-identical pass-through of the unfiltered TraceLens list. In the Qwen3-32B run 8 of 10 \`hot_kernels\` entries were \`aten::mm\` GEMMs with \`source_file=\"\"\` — already flagged \`reusable_native_kernel=false\` upstream — yet downstream consumers (\`shared_state\`, \`kernel_request_handlers\`, \`parallel_e2e_runner\`) each had to re-apply the filter themselves, and \`num_hot_kernels\` accounting overcounted by ~5x.

### \`fix(kernel-opt): stop appending full analysis.md to GEAK prompt (#307)\`

\`kernel-agent/tools/kernel_optimization.py\`: gate the \`tracelens_context_block\` fallback on \`not hypothesis_block.strip()\`, so the full report is only attached when no per-kernel hypothesis could be rendered (raw-trace / csv-fallback path). Production runs that have per-kernel TraceLens prose now skip the dump entirely.

### \`fix(kernel-opt): filter non-routable kernels from kernel_candidates.json (#314)\`

- \`kernel-agent/tools/tracelens_analysis.py::write_reports\`: filter \`kernel_candidates.json::hot_kernels\` to \`reusable_native_kernel is True\` and write a parallel \`skipped_kernels[]\` sidecar containing the FULL candidate dicts (not just a compact audit projection) for the rejected ones.
- \`kernel-agent/tools/kernel_optimization.py::load_candidates\`: read-side union over \`hot_kernels\` + \`skipped_kernels\` so the CLI's direct ID-lookup path (\`find_candidate(load_candidates(...), kid)\`) still resolves a non-routable kernel by id (operator debugging, and the dispatcher's \`_validate_reusable_native_kernel\` guard fires with the real \`reusable_native_kernel=False\` candidate instead of an empty \"missing native source\" stub). Batch dispatchers continue to read the \`hot_kernels\` field directly and see only routable kernels — the contract #314 was after.
- \`kernel-agent/tools/test_tracelens_csv.py\`: two unit-test fixtures that bypass \`_finalize_candidates\` and pass a raw candidate dict straight into \`write_reports\` now set \`reusable_native_kernel: True\` explicitly (mirroring what production \`_finalize_candidates\` would do) and use a source path under a reusable framework root, so the post-write \`payload[\"hot_kernels\"][0]\` assertions stay valid under the new contract.

## Audit surface (what stays where)

| Artifact | Before | After |
|---|---|---|
| \`tracelens/tracelens_report.json::hot_kernels\` | full unfiltered list | full unfiltered list (unchanged — TraceLens-team contract) |
| \`kernel_candidates.json::hot_kernels\` | full unfiltered list | **routable only** (new dispatch contract) |
| \`kernel_candidates.json::skipped_kernels\` | (did not exist) | non-routable candidates with skip_reason (full dicts) |
| \`tracelens/summary.json::tasks[]\`/\`skipped[]\` | already split correctly | unchanged |
| \`load_candidates(path)\` | unfiltered | union of routable + skipped (back-compat for CLI direct lookup) |

## Test plan

Verified end-to-end on a Qwen3-32B-shaped fixture (10 candidates: 8 GEMM \`aten::mm\` with empty source_file + 2 routable \`aiter::rmsnorm\` at \`/sgl-workspace/aiter/aiter/ops/rmsnorm.py\`).

**#307**
- [x] Production case (TraceLens prose attached): prompt is **175 lines**, no \`## TraceLens Context\` section.
- [x] Fallback case (no prose, e.g. raw-trace / csv-fallback): prompt is **465 lines**, full report attached.
- [x] ~290 lines of bloat eliminated per kernel in the common case.

**#314**
- [x] \`kernel_candidates.json::hot_kernels\` = 2 (k009, k010 — routable rmsnorm)
- [x] \`kernel_candidates.json::skipped_kernels\` = 8 (k001–k008 — full dicts with \`skip_reason: \"source file not resolved\"\`)
- [x] \`tracelens_report.json::hot_kernels\` = 10 (full audit list — unchanged)
- [x] \`load_candidates(...)\` returns 10 entries (union, back-compat for CLI lookup)
- [x] \`find_candidate(..., \"k001\")\` succeeds and returns a candidate with \`reusable_native_kernel=False\` (dispatcher's \`_validate_reusable_native_kernel\` correctly rejects it with \`non_reusable_kernel\` error class)

**Unit tests**
- [x] \`kernel-agent/tools/test_tracelens_csv.py\` — all 132 tests pass (6 \`test_write_reports_*\` cases verified against the new contract).
- [x] \`kernel-agent/tools/test_kernel_optimization_verification.py\` — 54/55 pass; the one failure is the pre-existing \`test_build_prompt_includes_benchmark_cases_when_task_group_present\` env issue unrelated to this PR.
- [x] \`kernel-agent/tools/test_prompt_promotion_block.py\` — all 6 pass.
- [x] \`kernel-agent/tests/test_kernel_agent.py\` — 20/23 pass (the 3 failures — \`test_install_help_and_dry_run_backend_flags\`, \`test_geak_submit_build_cmd_omits_when_none\`, \`test_geak_submit_build_cmd_propagates_zero\` — all fail with the same pre-existing \`GEAK_CONFIG is required\` env error on clean \`main\` too; unrelated to this PR).

No new test regressions are introduced. The two unit-test fixture updates are minimal and explained inline; they mirror what \`_finalize_candidates\` does in the production candidate-enrichment path.

## Backward compatibility

- \`tracelens/tracelens_report.json\` is byte-for-byte unchanged (TraceLens-team interface preserved).
- \`tracelens/summary.json\` is unchanged (the audit \`tasks[]\`/\`skipped[]\` split was already correct upstream).
- Older \`kernel_candidates.json\` files written before this change (with the full unfiltered \`hot_kernels\`) continue to work — \`load_candidates\` and \`_load_candidate_metadata\` already handle the absence of \`skipped_kernels\` gracefully (\`payload.get(\"skipped_kernels\") or []\`).

### DIFF
--- a/kernel-agent/tools/kernel_optimization.py
+++ b/kernel-agent/tools/kernel_optimization.py
@@ -95,11 +95,42 @@ def update_status(
 
 
 def load_candidates(path: Path) -> list[dict[str, Any]]:
-    """Load kernel candidates from JSON, normalizing legacy shapes."""
+    """Load kernel candidates from JSON, normalizing legacy shapes.
+
+    Per AMD-AGI/Hyperloom#314 the canonical ``hot_kernels`` field now only
+    carries kernels that ``classify_patchability`` marked routable, with the
+    rejected ones moved to ``skipped_kernels`` (full candidate dicts, not a
+    compact audit projection). Batch dispatchers (parallel_e2e_runner,
+    ``_batch_kernel_candidates``) read ``hot_kernels`` directly and benefit
+    from the filter. The kernel-opt CLI's direct lookup path
+    (``find_candidate(load_candidates(...), kid)``) still needs to be able
+    to resolve a non-routable kernel by id — both for operator debugging
+    ("what does the backend selector say about k001?") and so the
+    dispatcher's ``_validate_reusable_native_kernel`` guard fires with the
+    real ``reusable_native_kernel=False`` candidate instead of an empty
+    "missing native source" stub. Return the union so both call sites
+    work; the older flat-list / ``kernel_candidates`` legacy shapes are
+    still respected.
+    """
     payload = json.loads(path.read_text(encoding="utf-8"))
     if isinstance(payload, list):
         return payload
     candidates = list(payload.get("hot_kernels") or payload.get("kernel_candidates") or [])
+    skipped = payload.get("skipped_kernels") or []
+    if isinstance(skipped, list):
+        seen_ids = {
+            c.get("kernel_id") for c in candidates
+            if isinstance(c, dict) and c.get("kernel_id")
+        }
+        for entry in skipped:
+            if not isinstance(entry, dict):
+                continue
+            kid = entry.get("kernel_id")
+            if kid and kid in seen_ids:
+                continue
+            candidates.append(entry)
+            if kid:
+                seen_ids.add(kid)
     artifact_paths = payload.get("artifact_paths")
     if not isinstance(artifact_paths, dict):
         artifact_paths = {}
--- a/kernel-agent/tools/kernel_optimization.py
+++ b/kernel-agent/tools/kernel_optimization.py
@@ -1296,32 +1296,45 @@ def build_prompt(
             "still measure compute/IO improvements.\n"
         )
     tracelens_context_block = ""
-    report_path_str = str(candidate.get("trace_report_path") or "")
-    report_path = Path(report_path_str) if report_path_str else None
-    if report_path and report_path.exists():
-        try:
-            full_report = report_path.read_text(encoding="utf-8", errors="replace")
-        except Exception:
-            full_report = ""
-        if full_report:
-            from inference_optimizer.tracelens_md import strip_base64_data_urls
+    # Per AMD-AGI/Hyperloom#307: the per-kernel TraceLens prose
+    # (Identification / Reasoning / Recommended direction / Impact
+    # estimate) is already extracted into ``hypothesis_block`` above
+    # via ``_build_hypothesis_block``, and the bound-specific lever
+    # list is in ``priority_block``. Dumping the full analysis.md on
+    # top of that bloats the prompt by ~300 lines (~40% of the body
+    # in a real Qwen3-32B run) and surfaces other P-items that this
+    # very prompt tells the agent NOT to optimize. Only fall back to
+    # the full-report dump when no per-kernel hypothesis could be
+    # rendered (raw-trace / csv-fallback path with empty prose), so
+    # the agent still has *some* TraceLens grounding in that edge
+    # case.
+    if not hypothesis_block.strip():
+        report_path_str = str(candidate.get("trace_report_path") or "")
+        report_path = Path(report_path_str) if report_path_str else None
+        if report_path and report_path.exists():
+            try:
+                full_report = report_path.read_text(encoding="utf-8", errors="replace")
+            except Exception:
+                full_report = ""
+            if full_report:
+                from inference_optimizer.tracelens_md import strip_base64_data_urls
 
-            full_report = strip_base64_data_urls(full_report)
-            rank = candidate.get("tracelens_pitem_rank")
-            title = candidate.get("tracelens_pitem_title", "")
-            if rank:
-                focus_line = (
-                    f"Focus on **P{rank}: {title}** in the report below. "
-                    "Other P-items are context only — do not optimize them.\n"
+                full_report = strip_base64_data_urls(full_report)
+                rank = candidate.get("tracelens_pitem_rank")
+                title = candidate.get("tracelens_pitem_title", "")
+                if rank:
+                    focus_line = (
+                        f"Focus on **P{rank}: {title}** in the report below. "
+                        "Other P-items are context only — do not optimize them.\n"
+                    )
+                else:
+                    focus_line = "Use the report below as full context for this kernel.\n"
+                tracelens_context_block = (
+                    "\n## TraceLens Context\n\n"
+                    + focus_line
+                    + "\n"
+                    + full_report
                 )
-            else:
-                focus_line = "Use the report below as full context for this kernel.\n"
-            tracelens_context_block = (
-                "\n## TraceLens Context\n\n"
-                + focus_line
-                + "\n"
-                + full_report
-            )
     # Use GEAK task_parser field names (kernel_name/kernel_url/kernel_type/repo)
     # so its LLM-based parser can extract them; OOB agents read the same body
     # as a normal natural-language prompt.
--- a/kernel-agent/tools/test_tracelens_csv.py
+++ b/kernel-agent/tools/test_tracelens_csv.py
@@ -290,10 +290,19 @@ def test_write_reports_enriches_candidates_with_runtime_metadata(tmp_path):
         "duration_us": 100.0,
         "call_count": 2,
         "gpu_pct": 10.0,
-        "source_file": "/tmp/paged_attention.py",
+        "source_file": "/sgl-workspace/aiter/paged_attention.py",
         "shapes": [[1, 32, 128]],
         "is_multigpu": False,
         "num_gpus_recommended": 1,
+        # Per AMD-AGI/Hyperloom#314, ``kernel_candidates.json::hot_kernels``
+        # now only carries candidates that ``classify_patchability`` marked
+        # routable. In production this field is set by
+        # ``_finalize_candidates`` before ``write_reports`` runs; this
+        # unit test bypasses ``_finalize_candidates`` and passes a raw
+        # candidate straight into ``write_reports``, so set the routing
+        # marker explicitly to mirror the production fixture and keep
+        # the downstream assertions on ``hot_kernels[0]`` valid.
+        "reusable_native_kernel": True,
     }
     args = Namespace(
         trace_input=str(trace),
@@ -422,8 +431,10 @@ def test_write_reports_enriches_head_size_from_model_config(tmp_path):
         "duration_us": 100.0,
         "call_count": 2,
         "gpu_pct": 10.0,
-        "source_file": "/tmp/paged_attention.py",
+        "source_file": "/sgl-workspace/aiter/paged_attention.py",
         "shapes": [[1, 32, 128]],
+        # Per AMD-AGI/Hyperloom#314, see twin fixture above.
+        "reusable_native_kernel": True,
     }
     args = Namespace(
         trace_input=str(trace),
--- a/kernel-agent/tools/tracelens_analysis.py
+++ b/kernel-agent/tools/tracelens_analysis.py
@@ -1985,9 +1985,33 @@ def write_reports(
     }
     atomic_write_json(run_dir / "trace_input_manifest.json", manifest)
     atomic_write_json(tracelens_dir / "tracelens_report.json", report)
+    # Per AMD-AGI/Hyperloom#314: ``kernel_candidates.json::hot_kernels``
+    # is the dispatch payload consumed by the kernel-opt path
+    # (``kernel_optimization.load_candidates``, ``parallel_e2e_runner``,
+    # ``kernel_request_handlers``). Filter it down to candidates that
+    # ``classify_patchability`` actually marked routable so downstream
+    # batch dispatchers no longer have to re-apply the same filter and
+    # so ``num_hot_kernels`` accounting reflects what we will really
+    # send to a backend. The full unfiltered list stays available in
+    # ``tracelens/tracelens_report.json`` for audit, and the routable
+    # vs. skipped split is also surfaced in ``tracelens/summary.json``
+    # (``tasks[]`` / ``skipped[]``). ``skipped_kernels[]`` carries the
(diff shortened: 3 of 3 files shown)


ZhengGong-amd (approved):
Valid revision, approved.

PR #325, opened 2026-05-27
fix(orch): unblock slow-workload optimization (auto-derive explore timeout, raise static defaults, pin GEAK fix branch)

### LINKED ISSUE #323: Default action timeouts under-budget Qwen3-32B-class slow workloads
## Symptom

On Qwen3-32B (TP=1, BF16, CONC=64, ISL/OSL=1024, NUM_PROMPTS=320) the bare baseline benchmark takes ~70 min and a profile run with torch profiler takes ~3 h. The current defaults are calibrated for the smoke workload and silently break the long-workload path:

- Every EXPLORE variant aborts mid-flight with `magpie timeout (timeout_sec=2400)`. `explore_search.tested/accepted/rejected` stay empty because the timeout path doesn't carry a measurement, so the `compute_plateau_explore` judge can't tell "no progress because variants are timing out" from "no progress because we just haven't found a winner yet" — IR-6 is the only remaining gate, but it doesn't fire until 3h of wall-clock remain.
- `baseline.py` / `profile.py` / `integrate_patch.py` / `_grid_runner.py` and the shipped `baseline_sglang.yaml` / `profile_sglang.yaml` all have a 2400 s (40 min) default. Slow workloads can't fit a single benchmark run inside that envelope.

## Root cause

Two layers, one root cause: **the static 2400 s default was tuned for fast smoke benchmarks and was never raised for production-scale workloads**.

1. EXPLORE: `_compute_explore_variant_timeout` exists to anchor the cap on `baseline_runtime_sec × (kill_ratio + safety_margin)` — but only after baseline lands. Pre-baseline path uses a hardcoded floor.
2. Other actions (baseline / profile / integrate / `_grid_runner`): no auto-derive at all — pure constants.

## Fix shape

- **EXPLORE auto-derive** — already shipped on this branch (commits `866d494` PR-A, `1ea8eec` PR-C, `6069bef` PR-B): hard cap = `baseline_runtime_sec × (kill_ratio + safety_margin)`, clamped to `[2400, 14400]` s, with `--explore-variant-timeout-sec` and `--explore-variant-timeout-safety-margin` operator overrides.
- **Static bump for non-EXPLORE actions** — commit `e3d8e9f` raises:
  - `baseline.py` `DEFAULT_TIMEOUT 2400→7800`, `COLD_START 3600→9000`
  - `integrate_patch.py` `DEFAULT_VARIANT_TIMEOUT 2400→7800`
  - `profile.py` `DEFAULT_TIMEOUT 2400→14400`
  - `_grid_runner.py` `_VARIANT_TIMEOUT_SEC_DEFAULT 2400→7800`
  - `baseline_sglang.yaml` `timeout_seconds 2400→7200`
  - `profile_sglang.yaml` `timeout_seconds 2400→14400`

The EXPLORE auto-derive and the static bumps are complementary, not redundant: explore's cap follows measured baseline runtime; the others run *before* baseline lands or in different phases where there is no analogous anchor.

## Reproducer

Any session with `--isl 1024 --osl 1024 --conc 64 --tp 1 --precision bf16` against a 30B+ dense model on MI300X. Symptom: every EXPLORE variant prints `magpie timeout (timeout_sec=2400)` with `explore_search.tested = {}`, then IR-6 force-exits EXPLORE with `cumulative_gain = 0`.

## Verification

`inference_optimizer/tests/` 242 tests pass; new tests in `test_explore_executor.py` cover the auto-derive helper at floor / ceiling / kill-ratio / safety-margin edges and integration end-to-end through `run_grid`.

rodosingh:
Stacked on top of #320. Re-target to `main` once #320 merges (or GitHub will auto-retarget on branch deletion).

Closes #323.

## Summary

Five commits unblocking the slow-workload (Qwen3-32B-class) optimization path:

1. `866d494` — auto-derive explore variant timeout from baseline runtime
2. `1ea8eec` — let operators tune the auto-derive headroom (`--explore-variant-timeout-safety-margin`)
3. `6069bef` — opt-in roofline-categorized variant filter (`--explore-roofline-hard-gate`)
4. `e3d8e9f` — raise default action timeouts for slow workloads (baseline / profile / integrate / yamls)
5. `b456c45` — pin GEAK to the `save-and-test-diff-fallthrough` fix branch (upstream PR https://github.com/AMD-AGI/GEAK/pull/244, also open against GEAK's `main`); revert with `GEAK_REF=v3.2.0 bash install.sh`

## Why

On Qwen3-32B TP=1 BF16 CONC=64 ISL/OSL=1024 NUM_PROMPTS=320:

- A bare baseline run takes ~70 min. The 2400 s (40 min) default cap on EXPLORE variants kills every variant before it produces a measurement, and the kill path doesn't carry a tput value, so the plateau judge can't tell "no progress because variants are timing out" from "no progress because no winner yet". IR-6's wall-clock gate is the only thing that eventually moves the optimizer to KERNEL — at which point ~3 h of EXPLORE budget has been spent on zero usable measurements.
- `baseline.py` / `integrate_patch.py` / `profile.py` / `_grid_runner.py` and the shipped Magpie YAML defaults shared the same 2400 s constant — every action that runs a benchmark on this workload is in the same trap.

### EXPLORE timeout — auto-derive (commits 1-3)

```python
hard_cap = baseline_runtime_sec × (kill_ratio + safety_margin)
clamped to [DEFAULT_EXPLORE_TIMEOUT_FLOOR_SEC, DEFAULT_EXPLORE_TIMEOUT_CEILING_SEC]
       i.e. [2400, 14400] s
```

- New helper `_compute_explore_variant_timeout(...)` in `explore.py` consumes the Coordinator-injected `baseline_runtime_sec` and `explore_overtime_kill_ratio` (already plumbed). Pre-baseline path uses the floor; post-baseline scales with measured runtime.
- New SharedState fields: `explore_variant_timeout_sec_override` (hard pin), `explore_variant_timeout_safety_margin` (headroom), `explore_roofline_hard_gate` (opt-in filter). All Coordinator-injected into every explore task's `params`.
- New CLI flags + env mirrors: `--explore-variant-timeout-sec` / `INFERENCE_OPTIMIZER_EXPLORE_VARIANT_TIMEOUT_SEC`, `--explore-variant-timeout-safety-margin` / `INFERENCE_OPTIMIZER_EXPLORE_VARIANT_TIMEOUT_SAFETY_MARGIN`, `--explore-roofline-hard-gate` / `INFERENCE_OPTIMIZER_EXPLORE_ROOFLINE_HARD_GATE`.
- New module `_explore_roofline_filter.py` with a flag→roofline-direction categorization table, opt-in (off by default). Variants whose every target direction is saturated above the threshold are dropped before dispatch with `reason='roofline_saturated'`. Soft `--roofline-saturation-advisory` prompt hint is unchanged.

### Non-EXPLORE timeouts — static bumps (commit 4)

| File | Field | 2400 → |
|---|---|---|
| `baseline.py` | `BASELINE_DEFAULT_TIMEOUT_SEC` | **7800** (130 min, WARM) |
| `baseline.py` | `BASELINE_COLD_START_TIMEOUT_SEC` | 3600 → **9000** (150 min, COLD) |
| `integrate_patch.py` | `DEFAULT_VARIANT_TIMEOUT_SEC` | **7800** |
| `profile.py` | `PROFILE_DEFAULT_TIMEOUT_SEC` | **14400** (4 h) |
| `_grid_runner.py` | `_VARIANT_TIMEOUT_SEC_DEFAULT` | **7800** |
| `baseline_sglang.yaml` | `timeout_seconds` | **7200** (120 min) |
| `profile_sglang.yaml` | `timeout_seconds` | **14400** (4 h) |

Pairs with the EXPLORE auto-derive: explore variants get a measurement-anchored cap; the actions that run *before* baseline lands or in different phases get a generous static cap matching the documented Qwen3-32B TP=1 cost.

## Test plan

- 242 tests across `test_explore_executor.py`, `test_explore_roofline_filter.py`, `test_grid_runner.py`, `test_baseline_self_loop_policy.py`, `test_baseline_param_overrides.py`, `test_integrate_patch_executor.py`, `test_profile_and_kernel_handlers.py`, `test_profile_trace_validator.py` — all pass (clean env).
- New tests cover: `_compute_explore_variant_timeout` floor / ceiling / kill-ratio clamping / safety-margin override / explicit-override-wins / auto-derive end-to-end through `run_grid`; `categorize_variant` for single / multi / unknown / env-only flag patterns; `filter_variants_by_roofline` no-op / partial / full saturation / threshold override; executor end-to-end with the gate on / off.
- Backward compatibility: old `state.json` files round-trip cleanly (new SharedState fields default to 0 / 0.5 / False = legacy behaviour). Floor on auto-derive preserves the old 2400 s default for any path without baseline data.

## Out of scope

- The earlier draft included some commented-out lines in `kernel-agent/tools/kernel_optimization.py` (debug experiment trimming the GEAK prompt). Those got lost during stash juggling and aren't included here. If they're needed they should land as a deliberate, properly-removed-not-commented-out PR.

### DIFF
--- a/inference_optimizer/cli.py
+++ b/inference_optimizer/cli.py
@@ -673,6 +673,22 @@ def _seed_shared_state(
     except (TypeError, ValueError):
         explore_variant_timeout_sec_override = 0
 
+    # ``--explore-variant-timeout-safety-margin`` mirror. Headroom as a
+    # fraction of baseline_runtime_sec on top of the soft kill ratio when
+    # the auto-derive path computes the cap. Negative values clamp to 0
+    # (which collapses the hard cap onto the soft kill, no headroom).
+    explore_variant_timeout_safety_margin_raw = getattr(
+        args, "explore_variant_timeout_safety_margin", None,
+    )
+    try:
+        explore_variant_timeout_safety_margin = max(
+            0.0,
+            float(explore_variant_timeout_safety_margin_raw)
+            if explore_variant_timeout_safety_margin_raw is not None else 0.5,
+        )
+    except (TypeError, ValueError):
+        explore_variant_timeout_safety_margin = 0.5
+
     state = SharedState(
         session_id=session_id,
         claw_session_id=(os.environ.get("CLAW_SESSION_ID") or "").strip(),
@@ -715,6 +731,7 @@ def _seed_shared_state(
             getattr(args, "gain_driven_kernel_opt", False),
         ),
         explore_variant_timeout_sec_override=explore_variant_timeout_sec_override,
+        explore_variant_timeout_safety_margin=explore_variant_timeout_safety_margin,
     )
     state.save(session_dir)
     return state
@@ -4673,10 +4690,27 @@ def _build_parser() -> argparse.ArgumentParser:
         help="Pin the per-variant hard timeout (seconds) inside the "
              "EXPLORE phase. ``0`` (default) auto-derives from "
              "``baseline_runtime_sec * (--explore-overtime-kill-ratio + "
-             "0.5)`` once baseline lands, with a 2400-14400 s range "
-             "guard. Set to a positive integer to pin (CI smoke runs / "
-             "debugging). Env: "
+             "--explore-variant-timeout-safety-margin)`` once baseline "
+             "lands, with a 2400-14400 s range guard. Set to a positive "
+             "integer to pin (CI smoke runs / debugging). Env: "
              "INFERENCE_OPTIMIZER_EXPLORE_VARIANT_TIMEOUT_SEC.",
+    )
+    opt.add_argument(
+        "--explore-variant-timeout-safety-margin",
+        dest="explore_variant_timeout_safety_margin",
+        type=float,
+        default=_env_float_or(
+            0.5, "INFERENCE_OPTIMIZER_EXPLORE_VARIANT_TIMEOUT_SAFETY_MARGIN",
+        ),
+        help="Headroom (as a fraction of baseline_runtime_sec) added on "
+             "top of --explore-overtime-kill-ratio when the EXPLORE hard "
+             "cap is auto-derived. Default 0.5 (≈ 50%% of baseline as "
+             "buffer for variant cold starts: torch.compile AOTI compile, "
+             "fresh aiter shapes, spec-decoding draft load). Bump for "
+             "workloads with heavy compile cost; lower to tighten the "
+             "backstop. No effect when --explore-variant-timeout-sec is "
+             "set to a positive value. Env: "
+             "INFERENCE_OPTIMIZER_EXPLORE_VARIANT_TIMEOUT_SAFETY_MARGIN.",
     )
     # ------------------------------------------------------------------
     # drop scoreboard
--- a/inference_optimizer/cli.py
+++ b/inference_optimizer/cli.py
@@ -658,6 +658,21 @@ def _seed_shared_state(
     except (TypeError, ValueError):
         explore_overtime_kill_ratio = 1.10
 
+    # ``--explore-variant-timeout-sec`` mirror. ``0`` (default) lets the
+    # ExploreExecutor auto-derive the cap from baseline_runtime_sec; any
+    # positive value pins it (CI smoke / debug).
+    explore_variant_timeout_raw = getattr(
+        args, "explore_variant_timeout_sec", None,
+    )
+    try:
+        explore_variant_timeout_sec_override = max(
+            0,
+            int(explore_variant_timeout_raw)
+            if explore_variant_timeout_raw is not None else 0,
+        )
+    except (TypeError, ValueError):
+        explore_variant_timeout_sec_override = 0
+
     state = SharedState(
         session_id=session_id,
         claw_session_id=(os.environ.get("CLAW_SESSION_ID") or "").strip(),
@@ -699,6 +714,7 @@ def _seed_shared_state(
         gain_driven_kernel_opt=bool(
             getattr(args, "gain_driven_kernel_opt", False),
         ),
+        explore_variant_timeout_sec_override=explore_variant_timeout_sec_override,
     )
     state.save(session_dir)
     return state
@@ -4602,6 +4618,15 @@ def _build_parser() -> argparse.ArgumentParser:
             return float(raw)
         except (TypeError, ValueError):
             return float(default)
+
+    def _env_int_or(default: int, env_var: str) -> int:
+        raw = os.environ.get(env_var, "").strip()
+        if not raw:
+            return int(default)
+        try:
+            return int(raw)
+        except (TypeError, ValueError):
+            return int(default)
     opt.add_argument(
         "--explore-overtime-kill-ratio",
         dest="explore_overtime_kill_ratio",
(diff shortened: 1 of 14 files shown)


ZhengGong-amd:
Two things I'd like addressed before merge, and one nit:

kernel-agent/scripts/install.sh — GEAK_REF pinned to a branch (blocker). Moving the default from v3.2.0 (immutable tag) to fix/save-and-test-diff-fallthrough (mutable branch) breaks installer reproducibility — a future force-push / rebase upstream silently changes what every fresh install gets. Please either (a) pin to the commit SHA of the current tip of that branch (and add a TODO to swap back to a tag once GEAK PR #244 lands), or (b) keep v3.2.0 as default and require operators to opt in via GEAK_REF=... until #244 merges. Also: the comment block right above still says "Pin GEAK to the first release that ships RAG MCP..." — please update it to match the new default and document the temporary nature.

coordinator.py — duplicated inject block (blocker-ish). The new variant_timeout_sec_override / variant_timeout_safety_margin / roofline_hard_gate injection appears twice (~L4638 and ~L4801) in near-identical form. Please extract a small _inject_explore_runtime_params(params) helper and call it from both sites — otherwise the next field added here will inevitably get added to only one branch.

Nit (non-blocking): _explore_roofline_filter._FLAG_TO_DIRECTIONS is a regex table that will rot every time sglang adds a flag. The module docstring already explains the conservative "uncategorized → keep" policy, but please add one explicit sentence documenting the ownership / update protocol (e.g. "new sglang flags MUST be added here or they default to keep") so future contributors don't miss it.

Re-request review once 1 and 2 are addressed.

rodosingh:
Thanks for the detailed feedback. Addressed all three:

1. **`install.sh` GEAK_REF** (`a43125b1`) — went with option (a). Default is now the SHA `ec61bdbdb151904ec187a8d89518afb969c53737` (current tip of the fix branch), so reproducibility is back without giving up auto-fix-delivery on fresh installs. Comment block rewritten to reference [GEAK PR #244](https://github.com/AMD-AGI/GEAK/pull/244) and carries a `TODO(post-GEAK-PR-244)` to swap back to a tag once that ships. The catch was that `git clone --branch` doesn't accept SHAs, so `ensure_geak()` got a small SHA-vs-ref dispatch — a bash regex detects a 7-40 hex SHA and falls back to `init + remote add + fetch --depth 1 + checkout FETCH_HEAD` (relies on GitHub's `uploadpack.allowReachableSHA1InWant`, on by default). Tags/branches keep the original `git clone --depth 1 --branch` path. Sanity-checked the regex against `ec61bdbd…` / `v3.2.0` / `fix/some-branch` and ran `bash -n` over the result.

2. **coordinator.py duplicated inject** (`e1518160`) — fair point, the `_handle_delegate` copy was already drifting (no per-step comments, slightly terser). Extracted `Coordinator._inject_explore_runtime_params(params)` just before `_materialize_approved_proposal`; both call sites collapse to a single line. Landed it as its own commit on top rather than amending the original auto-derive: the duplication built up cumulatively across three commits (auto-derive + safety-margin + roofline-hard-gate each added a field to both sites), so a separate refactor commit reads cleaner than re-writing-three-histories.

3. **`_FLAG_TO_DIRECTIONS` ownership** (`e373d479`, nit) — added a four-line *Ownership / update protocol* paragraph just above the table calling out that new sglang flags MUST be added there or they fall into the `uncategorized → keep` bucket and the gate is a no-op for them. Flag rot drifts toward "keep" (safe, but defeats the gate).

Branch rebased on `main`; PR base retargeted to `main` directly (#320 merged earlier today). Tests in `test_explore_executor.py` / `test_explore_roofline_filter.py` / `test_coordinator_runtime.py` / materialize+delegate tests all green.

ZhengGong-amd (approved):
valid modification to these two points. Approved.

PR #329, opened 2026-05-27
support rayjob env

weilei0120:
Support fetching the env required by the RayJob from the prompt, but don't use the sandbox's own env


ZhengGong-amd (approved):
Useful revision to multinode mode. Approved

PR #339, opened 2026-05-27
fix(orch): GEAK budget tracks GEAK_RUN_MODE + adaptive batch parallelism

### LINKED ISSUE #337: Orchestrator GEAK budget default forces quick-mode timing even when GEAK_RUN_MODE=full
## Symptom

`install.sh` defaults `GEAK_RUN_MODE=full` and the kernel-agent driver picks 130 min as the default `geak-budget-min` for full mode (PR #301). But on the orchestrator path every GEAK attempt still gets a 90-min budget, which is below the 130-min full-mode threshold, so GEAK silently runs in quick-mode timing even when the rest of the stack (installer / driver / yaml) is in full mode.

Concretely: the GEAK harness reads its budget from the orchestrator-supplied wrapper timeout (130 × 60 + 180 in full mode) but with the legacy default it gets (90 × 60 + 180), and the orchestrator-side cap is the binding one.

## Root cause

`inference_optimizer/orchestrator/kernel_request_handlers.py`:

```python
_DEFAULT_GEAK_BUDGET_MINUTES = 90.0

def _geak_budget_minutes(payload: dict) -> float:
    return float(
        payload.get("geak_budget_min")
        or os.environ.get("HYPERLOOM_GEAK_BUDGET_MIN", _DEFAULT_GEAK_BUDGET_MINUTES)
    )
```

PR #301 fixed only the **kernel-agent** layer (`kernel_optimization.py:2570`, `parallel_e2e_runner.py:300`) where the analogous default tracks `GEAK_RUN_MODE` and returns 70 / 130. The orchestrator has its own constant that never got updated, so unless an operator manually exports `HYPERLOOM_GEAK_BUDGET_MIN=130` or passes `geak_budget_min` in the LLM payload, the orchestrator clamp wins.

## Fix shape

Replace the constant with a tiny helper that mirrors the kernel-agent default:

```python
def _default_geak_budget_minutes() -> float:
    raw = (os.environ.get("GEAK_RUN_MODE") or "").strip().lower()
    return 70.0 if raw == "quick" else 130.0
```

and route `_geak_budget_minutes` through it. Override knobs (`HYPERLOOM_GEAK_BUDGET_MIN`, payload `geak_budget_min`) stay unchanged.

## Reproducer

Any session that doesn't explicitly pass `geak_budget_min` or set `HYPERLOOM_GEAK_BUDGET_MIN`. Inspect the GEAK driver log: `geak_budget_min=90` even though `GEAK_RUN_MODE=full` is in the environment and `geak/.../geak_agent.log` shows `mode: full`.

### LINKED ISSUE #338: Kernel batch parallelism default (8) admits more siblings than Ray can schedule on smaller pods
## Symptom

`_run_optimization_batch` admits up to 8 concurrent `_run_kernel_backend_sequence` coroutines by default (sized for an MI300X / MI355X full node = 8 GPU). On smaller pods — 4-GPU labs, partial-node CI shards, single-GPU dev boxes — the asyncio semaphore lets all 8 siblings start, but Ray can only schedule `floor(visible_gpus / num_gpus_per_task)` of them at a time. The rest stack against the GPU lock, and a kernel that *would* finish quickly waits behind a stuck GEAK on the semaphore for many minutes.

We've seen this in the Qwen3-30B-A3B-Base session 20260522T093903Z (3-hour serialization burned on a single hot kernel) and most recently on the Qwen3-32B run when GEAK preprocess hung — fast OOB siblings couldn't even start until the GEAK semaphore slot freed up.

## Root cause

`inference_optimizer/orchestrator/kernel_request_handlers.py`:

```python
_DEFAULT_KERNEL_BATCH_PARALLEL = 8

# inside _run_optimization_batch
max_parallel = int(
    payload.get("max_parallel")
    or os.environ.get("KERNEL_OPT_MAX_PARALLEL", _DEFAULT_KERNEL_BATCH_PARALLEL)
)
```

The constant is calibrated for a full MI300X node and there is no path that adapts it down. Operators *can* pin via `KERNEL_OPT_MAX_PARALLEL=N`, but that's manual / per-deployment and easy to forget.

## Fix shape

Keep the 8 as the *cap* and as the fallback when torch can't tell us the visible-GPU count, but compute the active default as

```python
min(cap, torch.cuda.device_count() // KERNEL_AGENT_NUM_GPUS)
```

with `KERNEL_AGENT_NUM_GPUS` read from the env (set by the kernel-agent submitter; defaults to 1 for compute kernels). Falls back to the legacy 8 on torch-import failure / `device_count() == 0` (CI / mocks / pre-driver init). Operator override via `KERNEL_OPT_MAX_PARALLEL` stays.

## Reproducer

Any pod with `<` 8 visible GPUs that runs `run_optimization` against a TraceLens snapshot with 3+ reusable kernels. Symptom: per-kernel logs `acquired sem` slot for all candidates immediately, then most of them block for ~budget-min waiting on Ray's GPU resource even though Ray's `cluster_resources()['GPU']` is `n < 8`.

rodosingh:
Closes #337.
Closes #338.

## Summary

Two orchestrator-side defaults in `kernel_request_handlers.py` were silently overriding what the kernel-agent installer / driver already do. PR #301 fixed only the kernel-agent layer of the GEAK budget; this is the matching orchestrator fix, plus a sibling fix for the static batch-parallelism constant.

### `_DEFAULT_GEAK_BUDGET_MINUTES = 90.0` (#337)

`install.sh` defaults `GEAK_RUN_MODE=full`. PR #301 made the kernel-agent layer (`kernel_optimization.py`, `parallel_e2e_runner.py`) return 70 / 130 based on that env. The orchestrator has its own constant pinned at 90, which is below the 130-min full-mode threshold, so unless an operator explicitly exports `HYPERLOOM_GEAK_BUDGET_MIN=130` or passes `geak_budget_min` in the LLM payload, the orchestrator clamp wins and GEAK runs in quick-mode timing.

### `_DEFAULT_KERNEL_BATCH_PARALLEL = 8` (#338)

Calibrated for a full MI300X / MI355X node. On smaller pods (4-GPU labs, partial-node CI shards) the asyncio semaphore lets all 8 siblings start, but Ray can only schedule `floor(visible_gpus / num_gpus_per_task)` of them at a time. The rest stack against the GPU lock, so a fast kernel that *would* finish quickly waits behind a stuck GEAK on the semaphore for many minutes. We've seen this in the Qwen3-30B-A3B-Base session 20260522T093903Z (3-hour serialization on a single hot kernel) and on the Qwen3-32B GEAK-preprocess-hang case from this week.

## What

| Symbol | Old | New |
|---|---|---|
| `_DEFAULT_GEAK_BUDGET_MINUTES = 90.0` | constant 90 | `_default_geak_budget_minutes()` -> 70 / 130 based on `$GEAK_RUN_MODE` |
| `_DEFAULT_KERNEL_BATCH_PARALLEL = 8` | constant 8 | constant retained as cap + torch-unavailable fallback; active value computed by `_default_kernel_batch_parallel()` -> `min(cap, visible_gpus // KERNEL_AGENT_NUM_GPUS)` |

Override knobs are unchanged: payload `geak_budget_min`, `HYPERLOOM_GEAK_BUDGET_MIN`, payload `max_parallel`, `KERNEL_OPT_MAX_PARALLEL`.

The legacy hard-coded constants stay defined: `_DEFAULT_KERNEL_BATCH_PARALLEL = 8` is still the cap and fallback, and the GEAK helper falls through to 130 when `GEAK_RUN_MODE` is unset (which is the same as the install-time default), so the only real behaviour change is that the orchestrator stops *clamping* full-mode GEAK below the kernel-agent's full-mode budget.

## Test plan

- `inference_optimizer/tests/test_kernel_request_handlers_units.py` (52 tests, +23 new)
  - `TestDefaultGeakBudgetMinutes` — env unset / empty / `full` / `FULL` / `  full  ` / `garbage` / `quick` / `QUICK` / `  quick  ` (parametrized).
  - `TestGeakBudgetMinutes` — payload override wins, env override beats default, helper fall-through, empty-env-string fall-through (plugs a pre-existing `float("")` edge case).
  - `TestDefaultKernelBatchParallel` — 8/1, 4/1, 8/4, 4/2, 16/1, 1/4 GPU-count / per-task combinations; per-task unset / invalid; zero visible GPUs; torch-failure fallback.
- `inference_optimizer/tests/test_profile_and_kernel_handlers.py` — renamed `test_optimization_wrapper_timeout_sec_geak_default_90min` -> `_full_mode_130min`, added a `_quick_mode_70min` sibling. The pre-existing `_DEFAULT_KERNEL_BATCH_PARALLEL == 8` assertion still holds (constant retained as cap / fallback).

All 153 tests in those files pass except 3 pre-existing unrelated failures on `main` (`backend_ladder_prefers_keep_over_higher_micro_non_keep`, `backend_ladder_falls_back_to_highest_micro_when_no_keep`, `batch_candidates_default_min_gpu_pct_matches_sharedstate_gate` — unchanged on this branch vs. `main`, separate concern).

### DIFF
--- a/inference_optimizer/orchestrator/kernel_request_handlers.py
+++ b/inference_optimizer/orchestrator/kernel_request_handlers.py
@@ -36,6 +36,7 @@ can monkey-patch in tests.
 from __future__ import annotations
 
 import asyncio
+import functools
 import importlib.util
 import json
 import logging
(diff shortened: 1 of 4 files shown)


ZhengGong-amd:
One small thing I'd like before merge:

Cache _default_kernel_batch_parallel() (and ideally _default_geak_budget_minutes()). Both are invoked on the hot path (every _run_optimization_batch / every _geak_budget_minutes call) and the inputs (torch.cuda.device_count(), KERNEL_AGENT_NUM_GPUS, GEAK_RUN_MODE) don't change inside a session. Please wrap them with functools.lru_cache(maxsize=1) (or compute once at module import) to avoid a per-batch driver query.

rodosingh:
Done in `a7649b9c`. Both `_default_geak_budget_minutes()` and `_default_kernel_batch_parallel()` now wear `@functools.lru_cache(maxsize=1)` — the env reads (`$GEAK_RUN_MODE`, `$KERNEL_AGENT_NUM_GPUS`) are set at process start and the `torch.cuda.device_count()` driver query is the real cost we want to skip on every batch dispatch.

The catch was that the unit tests monkeypatch these between cases, so a first call would prime the cache and any subsequent monkeypatched call would see the stale value. Added an autouse `_clear_kernel_request_handler_caches` fixture in `inference_optimizer/tests/conftest.py` that calls `cache_clear()` on both helpers before every test. One existing test (`test_falls_through_to_helper_when_no_overrides`) was re-monkeypatching env mid-test (full → quick within a single body); split it via `parametrize` so each value runs as its own pytest item with a fresh cache.

Rebased on `main`. 159 tests across `test_kernel_request_handlers_units.py` + `test_profile_and_kernel_handlers.py` pass clean.

ZhengGong-amd (approved):
Small refinement. Approved

PR #342, opened 2026-05-27
fix(orchestrator): accept trace_analyze cache for kernel gate + propagate explicit workload flags

### LINKED ISSUE #341: Local mode: KERNEL phase never dispatches GEAK (last_select_kernels vs last_trace_analyze gate) + TP variants leak past visible-device scope
## Summary

Two independently confirmed defects in `inference_optimizer` cause local-mode runs to **finish with `cumulative_gain=0%`, never dispatching GEAK kernel-opt** on machines where the run cannot use all 8 GPUs.

Confirmed against `origin/main` HEAD `a21ec40` on `vllm/vllm-openai-rocm:v0.19.0` / MI350X (`gfx950`) with `--tp 4 --ep 1 --conc 64 --isl 1024 --osl 1024 --precision fp4 --kernel-claude --critic-mock --robustness-mock` and `ROCR_VISIBLE_DEVICES=4,5,6,7`.

---

### Bug A — KERNEL phase deadlock from `last_select_kernels` vs `last_trace_analyze`

**Symptom.** After PRELUDE→EXPLORE→KERNEL, the brain enters an infinite loop emitting `select_kernels` requests; the response is always served from a stale cache and **`SharedState.last_select_kernels` stays `{}`**. `kernel_opt_attempts` stays at `0`, GEAK is never dispatched, run hits `time_exhausted` with 0% gain.

**Root cause.** The bootstrap roofline at PRELUDE→EXPLORE writes only `last_trace_analyze` via `record_trace_analyze()` in `inference_optimizer/orchestrator/shared_state.py` (~`2510-2618`). The KERNEL-phase TODO checklist in `Coordinator._required_next_step()` (`inference_optimizer/orchestrator/coordinator.py:3102-3118`) reads **only** `last_select_kernels`, so it (incorrectly) injects \"emit select_kernels; do not run kernel_opt\" into the brain prompt forever. Every re-emission hits the cache short-circuit at `coordinator.py:5923-5926` (sets `cache_hit_source=\"shared_state_cache\"`); the cache writeback at `coordinator.py:6018-6024` is gated on `cache_hit_source is None`, so `record_select_kernels()` is **never** called.

**Hard evidence (live run `gpt-oss-120b-vllm-20260527_140935`, branch=main `a21ec40`).**
- Stuck window: 15:15:58–16:50:52 UTC, **57** `select_kernels` requests, **0** of any other kind, **0** PolicyGate denials, every response tagged `\"source\":\"shared_state_cache\"`.
- `state.json`: `last_select_kernels={}`, `kernel_opt_attempts={}`, `phase=CLOSE`, `stop_reason=time_exhausted`, `cumulative_gain=0.00`, `cumulative_gain_validated=0.00`.
- Same fingerprint observed in prior run `gpt-oss-120b-vllm-20260527_123014` (tick=56) → not a one-off.

### Bug B — Per-variant `TP` not reconciled against visible-device scope

**Symptom.** With `--tp 4` requested and 4 visible GPUs (`ROCR_VISIBLE_DEVICES=4,5,6,7`), EXPLORE/SWEEP variants nonetheless launch vLLM with `TP=8`, aborting **5/5 EXPLORE** + **9/9 SWEEP** variants in run `_140935`:
\`\`\`
AssertionError: local_world_size (8) must be less than or equal to the number of visible devices (4)
\`\`\`

**Root cause.** Explicit CLI `--tp/--ep/--conc` flags are not propagated into the per-variant executor env, so the grid yaml defaults dominate. `baseline_config.with_envs.yaml` correctly expands `ROCR_VISIBLE_DEVICES` but the per-variant `config.yaml` ends up with `TP=8` against the unchanged ROCR scope; vLLM init fails. Even if Bug A is fixed, EXPLORE/SWEEP will keep aborting on partial-node deployments (i.e. anything < 8 visible GPUs on an 8-GPU box).

---

## Repro

\`\`\`bash
# inside vllm/vllm-openai-rocm:v0.19.0 with Hyperloom main + GEAK fix/final_report_optimized_codes + TraceLens-internal release/hyperloom_integration_0.3.1
source bootstrap_vllm.env.sh && source \$LOCAL_SETUP_ENV && source \$KERNEL_AGENT_ENV
ROCR_VISIBLE_DEVICES=4,5,6,7 inference_optimizer --verbose optimize \\
  --model /models/openai-gpt-oss-120b --framework vllm --gpu-type mi355x \\
  --model-class moe_swa --tp 4 --ep 1 --conc 64 --isl 1024 --osl 1024 \\
  --precision fp4 --target-gain 12 --max-hours 4 --tick-interval-sec 30 \\
  --kernel-claude --critic-mock --robustness-mock
\`\`\`
Expected: GEAK kernel-opt dispatches after KERNEL transition; some integrate attempts land.
Actual: KERNEL hangs in `select_kernels` cache loop until budget exhausted; every EXPLORE/SWEEP variant aborts on `local_world_size`.

## Suggested fix

PR linked below (#TBD) implements minimal surgical fixes:
- `coordinator.py:3102` — TODO 3/5 accepts `last_trace_analyze` as satisfying the analyze cache.
- `cli.py` — explicit `--tp / --ep / --conc` flow into executor envs without clobbering YAML defaults.
- Regression tests in `tests/test_required_step_gates.py` + `tests/test_cli_workload_envs.py` (35 passed locally).

## Environment

- Container: `vllm/vllm-openai-rocm:v0.19.0`
- GPU: 8× MI350X (`gfx950`), 4 visible to run (`ROCR_VISIBLE_DEVICES=4,5,6,7`)
- Hyperloom: `main` @ `a21ec40`
- GEAK: `fix/final_report_optimized_codes` (+ PR #244 applied)
- TraceLens-internal: `release/hyperloom_integration_0.3.1` + `config_vllm_v0.19.0.patch`
- Run artifacts: `/workspace/runs/gpt-oss-120b-vllm-20260527_140935`, `/workspace/runs/gpt-oss-120b-vllm-20260527_123014`

iraj465:
## Summary

Two surgical fixes that together unblock GEAK kernel-opt dispatch on partial-node local-mode runs. Closes #341.

### Bug A — KERNEL gate deadlock
`Coordinator._required_next_step()` only checked `last_select_kernels`, while the bootstrap roofline populates `last_trace_analyze`. Combined with the `shared_state_cache` short-circuit in `_cached_kernel_request` (which writes back only when `cache_hit_source is None`), the brain was forced into an infinite `select_kernels` loop that never advanced state. Live evidence: 57 `select_kernels` requests / 0 of any other kind / `last_select_kernels={}` across a 95-min stuck window.

Fix: accept `last_trace_analyze` as satisfying the analyze cache in `coordinator.py` (~L3102).

### Bug B — TP leak past visible-device scope
Explicit CLI `--tp/--ep/--conc` weren't propagated into per-variant executor envs; the grid yaml's TP=8 default dominated even with only 4 visible GPUs, so every EXPLORE/SWEEP variant aborted with `local_world_size (8) must be <= visible devices (4)`.

Fix: `cli.py` now exports explicit single-node `--tp/--conc/--ep` to executor envs without clobbering YAML defaults.

## Files changed
- `inference_optimizer/orchestrator/coordinator.py`
- `inference_optimizer/cli.py`
- `inference_optimizer/tests/test_required_step_gates.py` (regression)
- `inference_optimizer/tests/test_cli_workload_envs.py` (regression, new file)

## Test plan
- [x] `pytest inference_optimizer/tests/test_required_step_gates.py inference_optimizer/tests/test_cli_workload_envs.py` → **35 passed** locally
- [ ] Reviewer: local-mode dry-run with `--tp 4 ROCR_VISIBLE_DEVICES=4,5,6,7` on a partial-node box; confirm `kernel_opt_attempts > 0` and EXPLORE variants no longer hit `local_world_size`.
- [ ] Reviewer: full local-mode run on 8-GPU MI300X to confirm no regression in the happy-path.

## Risk
- Bug A fix is additive (only widens what counts as a satisfied cache); cannot create a new stuck state worse than before.
- Bug B fix only sets env vars when the flag is explicitly passed; YAML defaults still win in absence.


iraj465:
Pushed second commit `3cb4517` addressing two defects exposed by the live `gpt-oss-120b-vllm-20260527_174712` session (full audit at `/home/sapmajum/hyperloom_work/logs/geak_dispatch_audit.md` — verdict: wedged raylet meant **GEAK was never invoked**; 12/12 backend attempts died in `ray.init(address="auto")` after the 30 s `global_state_accessor.cc:500` retry budget).

**Defect 1 — `kernel_optimization.make_proposal` (lines 2495–2496, audit §Defect 1).**
`compile_passed = bool(best)` is False both when "we compiled and it failed" and when "we never had a usable backend attempt to compile from". The legacy `reasons=["compile failed"]` masked the latter — operators saw a fake compile regression in `state.json:last_kernel_opt.reasons` and chased a phantom. We now branch on `artifact_error` (already populated by `build_verification` as `"no usable backend attempt"` when `usable=[]`) and emit `reasons=["backend dispatch failed: <err>"]` for the dispatch-side path. The legacy string is preserved for the real compile-failure branch.

**Defect 2 — `backends/{geak,oob}_submit.py` (lines 262–293, audit §Defect 2).**
`ensure_ray_cluster()` already existed in `backends/ray_runtime.py:28–50` but was only invoked from `parallel_e2e_runner.py:151`. Both submit paths went straight to `quiet_ray_init()` and burned a full 30 s of ray-retry budget per attempt when the cluster was wedged. We now call `ensure_ray_cluster(num_gpus=..., log_path=output_dir/"ray_lifecycle.log")` first — safe no-op when `ray status` is healthy; runs `ray start --head` otherwise — so we either recover or fail fast. Failure `stderr_tail` also carries the raylet-zombie diagnostic hint so the next operator doesn't have to spelunk.

**Tests.** 7 new regression tests:
- `kernel-agent/tools/test_kernel_optimization_verification.py`: 3 tests pinning down dispatch-vs-compile split in `make_proposal` (covers `artifact_error` populated + `best_attempt_id==""`, real artifact-lookup failure path, empty `artifact_error` fallback).
- `kernel-agent/tools/test_backend_submit_dispatch_gate.py`: 4 tests verifying `ensure_ray_cluster` runs before `run_via_ray` in both submitters and short-circuits to the diagnostic envelope (no `run_via_ray` call) when the cluster cannot be brought up.

Smoke (`pytest inference_optimizer/tests/test_required_step_gates.py inference_optimizer/tests/test_cli_workload_envs.py kernel-agent/tools/test_kernel_optimization_verification.py kernel-agent/tools/test_backend_submit_dispatch_gate.py`): **98 passed** (35 previous + 7 new + 56 existing).

Next: container-side Ray restart + relaunch optimize on this branch to verify GEAK actually lands in `optimization_stack` with `cumulative_gain_validated > 0`.

chaojhou (approved):

PR #347, opened 2026-05-28
Add power managment

umarinkovic:
- Description: adds power management as a knob
- Tests: added its own suite of tests
- Breaking changes: no


umarinkovic:
Failure seems unrelated to the changes; `torch module missing`

PR #351, opened 2026-05-28
Wire dual TraceLens install (public + internal) across bootstrap, run…

Ahmedhasssan-aig:
TraceLens requires two editable installs from separate repos -- the public AMD-AGI/TraceLens base package and AMD-AGI/TraceLens-internal for skills, patches, and the perf-report CLI. This commit updates the full stack to support both.

Bootstrap and install changes:
- local_setup.sh: clone both repos (TraceLens + TraceLens-internal) into source-mirrors as fallback; auto-detect existing container checkouts at /workspace/TraceLens and /workspace/TraceLens-internal; write both TRACELENS_PKG_ROOT and TRACELENS_ROOT to local-setup.env.sh.
- kernel-agent/scripts/install.sh: ensure_tracelens runs pip install -e on both checkouts; exports TRACELENS_PKG_ROOT in kernel-agent.env.sh.
- README: document the two-repo container install steps under Local Mode.

Orchestrator prompt changes:
- tracelens_skill_runner.py: add tracelens_root (public) and tracelens_internal_root parameters to build_orchestrator_prompt() and run_tracelens_skill(); prompt now shows both paths.
- tracelens_analysis.py: add --tracelens-pkg-root CLI arg (env TRACELENS_PKG_ROOT); pass to run_tracelens_skill().

Default paths updated across:
- hyperloom-env.sh, setup_env.sh.example, run.sh, ci/prompt_template.md, robustness-agent/sources/local_probe.py, SKILL docs, and tests.


tsrikris:
Fix aims to address issue https://github.com/AMD-AGI/Hyperloom/issues/353. Pls check to see if this PR covers all of the scenarios in Hyperloom.

rodosingh:
@tsrikris does this also covers https://github.com/AMD-AGI/Hyperloom/issues/352 ?

tsrikris:
@rodosingh this is orthogonal to that issue. This has to do with the open source repo.

chaojhou (commented):
The previous design let Hyperloom run without AMD-AGI/TraceLens-internal (OSS-only report: impact score, no roofline), gated by TRACELENS_INSTALL_INTERNAL. This PR makes TraceLens-internal effectively mandatory at both install and runtime, but the opt-out isn't honored consistently:

inference_optimizer/scripts/local_setup.sh — TRACELENS_INSTALL_INTERNAL default flips 0 → 1. Opt-out (=0) is honored here (skips the clone). ✅
kernel-agent/scripts/install.sh (ensure_tracelens) — now dies when the internal checkout is missing, and does not check TRACELENS_INSTALL_INTERNAL. ❌
kernel-agent/tools/tracelens_analysis.py (main()) — now raises FileNotFoundError if tracelens_internal_root doesn't exist, also unconditionally. This runs on every trace_analyze request. ❌

tsrikris:
@chaojhou could you pls ask an engineer from your team to look at this PR and use it as a baseline to modify on top of? We're not sure how the local setup is meant to be executed or how you plan to do the open source vs internal implementations. Not sure we have the context to make the opt-in work for the internal systems basically.

tsrikris:
@chaojhou pushed in a potential solution! Requesting to assign an engineer from your team to modify this implementation since we are not sure of the exact requirements here.

TRACELENS_INSTALL_INTERNAL (default 1) is the single gate for all TraceLens-internal functionality: local_setup.sh skips the clone, kernel-agent/scripts/install.sh skips the install and omits internal exports from the env file, and tracelens_analysis.py suppresses the missing-directory error, clears TL_EXTENSION, and passes None downstream so prompts render in OSS-only mode. Setting it to 0 at any entry point is sufficient — the flag propagates through the env file and is checked consistently everywhere, so no path can accidentally re-enable the internal extension.

In case of issues:
Pre-fix: v0.4-329-g91645921 (commit 91645921 — "Merge remote PR branch and reconcile TraceLens env naming")
Post-fix (current HEAD): v0.4-330-g1a6df9cb (commit 1a6df9cb — "Potential resolution to tracelens_internal_root issue")

Ahmedhasssan-aig:
I have tested the PR by setting up the Hyperloom locally . I was able to set it up and run the optimization loop. Here is the summary of my experiment:

1. The wire works end-to-end. All three env vars (TRACELENS_ROOT, TRACELENS_INTERNAL_ROOT, TL_EXTENSION=TraceLens_internal) propagated correctly through bootstrap → local-setup.env.sh → kernel-agent.env.sh → optimizer runtime. The optimization loop ran through all phases (PRELUDE → FRAMEWORK_PR → EXPLORE) without any TraceLens-related failures.

2. The .env.template backslash bug was the only breakage. The old TRACELENS_INTERNAL_ROOT=\ placeholder was treated as a real path and caused local_setup.sh to die. Fixed by sanitizing placeholder values in local_setup.sh, install.sh, and cli.py — once cleared, dual TraceLens cloned and installed cleanly on the first try.

BaoYunkai:
@Ahmedhasssan-aig  Thanks for this branch,  it's been really useful as a reference for us.

We're planning to open a new branch on top of yours to make a few modifications, for example: removing the internal toggle, switching to TRACELENS_INTERNAL_ROOT (passed via env) as the single signal for the decision, adjusting the default paths, and so on.

xiaofei-zheng (approved):

PR #402, opened 2026-06-03
fix(baseline): eliminate cold-start warmup artifact via server_lifecycle double-run

xiaofei-zheng:
## Problem

The `baseline` action is the **first** step of the optimization flow, so the inference server is always freshly booted for the model under test. The original code ran the benchmark **once** and took that throughput as `baseline_tput`. The first benchmark window is contaminated by one-time cold-start costs (CUDA/HIP graph capture, allocator warmup, GPU clock ramp, lazy kernel JIT). The hardcoded client-side `--num-warmups` (`2 * CONC`) is far too small to absorb them, so the baseline is under-reported.

Every later optimization round runs against an already-hot server, so `gain = final/baseline - 1` gets inflated into a fictitious "improvement". In `hyperloom_models_jun1.csv` this surfaced as `warmup`-tagged rows with up to **+1635%** bogus gains (e.g. MaralGPT-Maral-7B `270.9 -> 4701.6`).

## Fix

Run the benchmark **twice against the same persistent server** via Magpie's `server_lifecycle` reuse protocol:

- **Round 1 (warmup)** — boots + runs the server, pays every cold-start cost. Throughput **discarded** (surfaced as `warmup_round_tput` for auditability).
- **Round 2 (measured)** — re-attaches to the now-hot server (client only, no restart). Its throughput is the clean baseline.

Because round 2 never restarts the server, this removes **all** cold-start contamination, not just on-disk JIT caches.

### Safety / scope
- **Eligibility-gated**: single-node only (server_lifecycle is local-only), the resolved script must be a Magpie built-in that implements the `MAGPIE_RUN_PHASE` server/client split, profiler off. Otherwise falls back to the legacy single round. The gate is on the *phase-aware script*, not the framework: model-specific InferenceX scripts (e.g. `dsr1_fp8_mi300x.sh`) also fall back.
- **Frameworks**: `vllm_*` / `sglang_*` today; **`atom_*` included** per AMD-AGI/Magpie#34 (atom ships the same phase-split generic script; atom-as-a-framework already requires that PR, so the allowlist entry can never outrun Magpie support).
- **Overtime-kill anchor**: the double-run success result reports round 1's full server-boot + client wall-clock as `subprocess_runtime_sec` (the value the Coordinator promotes to `baseline_runtime_sec` for the explore soft-kill deadline), since explore variants each restart the server. Round 2's client-only reuse time is kept under `measure_round_runtime_sec` for transparency. (Avoids false KILLED_OVERTIME on normal variants.)
- **Defensive `finally` teardown** guarantees no persistent server leaks on any failure path (idempotent; happy path is handled by Magpie's `cleanup=true` on round 2).
- **Escape hatch**: `INFERENCE_OPTIMIZER_BASELINE_DOUBLE_RUN=0` reverts to single round.
- Changes live entirely in `baseline.py`.

## Real-model validation (vLLM, MI300x, TP=1 CONC=64 ISL/OSL=1024)

5 models, bf16 + fp8, each a full double-run. Hot (new) baseline is consistently above the discarded cold (old) value, and round 2 is ~2x faster (server reuse confirmed):

| model | precision | cold (old) | hot (new) | artifact |
|---|---|---|---|---|
| Qwen2.5-3B | bf16 | 9115.5 | 10101.3 | +10.8% |
| Qwen3-4B-Instruct | bf16 | 5704.0 | 6137.6 | +7.6% |
| Llama-3-8B-Instruct | bf16 | 4673.5 | 5407.1 | +15.7% |
| Llama-3.1-8B | bf16 | 4680.4 | 5413.9 | +15.7% |
| Qwen3-8B | fp8 | 4348.9 | 4852.5 | +11.6% |

Note: on vLLM the artifact is ~8-16% (graphs/compile happen at boot, before `/health`). The CSV's extreme +1635% cases were sglang/fp8 (lazy aiter JIT compile caught in the first benchmark window); the same fix removes those too since round 2 always reuses a fully-warm server. The sglang extreme was **not** re-validated here (vLLM-only environment).

## Test plan

- [x] New unit tests `test_baseline_warmup_double_run.py` (7): cold->hot via lifecycle, single-round when disabled, single-round when script not built-in, warmup-failure short-circuit, **atom engages double-run**, **overtime anchor = round-1 full run**, teardown removes state files.
- [x] Full suite: 3353 passed (pre-existing failures are unrelated missing test deps `respx` / web_tools).
- [x] 5 real vLLM models validated end-to-end (server boot, two rounds, reuse, teardown — no GPU leaks).

## Dependency
- `atom_*` lifecycle support depends on **AMD-AGI/Magpie#34** (adds the phase-aware atom scripts + atom to `MAGPIE_BUILTIN_SCRIPTS`). vllm/sglang need no Magpie change.

### DIFF
--- a/inference_optimizer/orchestrator/action_executors/baseline.py
+++ b/inference_optimizer/orchestrator/action_executors/baseline.py
@@ -38,17 +38,24 @@ import importlib.util
 import json
 import logging
 import os
+import signal
 import subprocess
 import time
 from pathlib import Path
 from typing import Any
+
+import yaml
 
 from ...compat.payload_aliases import read_extra_server_args
 from ...paths import asset_root
 from ...session_paths import runs_dir
 from ..sub_agent_runner import RunnerContext
 from ._grid_runner import sanitize_result_dir, sanitize_script_name
-from ._subprocess_kill import run_with_session_kill
+from ._subprocess_kill import (
+    _process_group_alive,
+    _signal_group,
+    run_with_session_kill,
+)
 from ._workload_envs import (
     default_baseline_config,
     materialize_config_with_envs,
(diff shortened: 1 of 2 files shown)


ZhengGong-amd (approved):
Limited modifications. Approved

PR #408, opened 2026-06-03
fix(kernel-opt): harden GEAK routing against invalid kernel ids

xiaofei-zheng:
## Summary
- Fix GEAK backend startup by importing `ensure_ray_cluster` in `geak_submit.py` (it was called but never imported, raising `NameError` on every GEAK attempt that reached the backend).
- Stop invalid/hallucinated `kernel_id` values from crashing kernel optimization: `find_candidate` no longer raises `KeyError`; it canonicalizes safe aliases (`kn001`/`rn010` → `k001`/`k010`) and skips non-routable candidates instead of dispatching a backend.
- Canonicalize aliased ids against the full `hot ∪ skipped` candidate set when `hot_kernels` is empty, so rejection lands on the real `k00x` rather than accumulating hallucinated aliases in `rejected_kernel_ids`. Non-unique / non-routable operator names (e.g. `aten::mm`) and pure hallucinations are never guessed.
- Surface skipped TraceLens candidates (`skipped_kernels_top`) in the orchestration prompt and require `kernel_id` to be copied verbatim from reusable candidate ids, so the LLM stops echoing operator names from `analysis.md`.

## Test plan
- [x] `python3 -m pytest inference_optimizer/tests/ -q -k "prompt or trace_analyze or shared_state or kernel_request or roofline" --ignore=inference_optimizer/tests/test_local_kb_requirements.py --ignore=inference_optimizer/tests/test_recipe_kb_dispatcher.py --ignore=inference_optimizer/tests/test_remote_client.py -p no:cacheprovider` → 715 passed, 1 skipped
- [x] `python3 -m pytest kernel-agent/tools/test_kernel_id_mismatch.py kernel-agent/tools/test_geak_submit_imports.py kernel-agent/tools/test_geak_param_resolution.py -q` → 31 passed
- [x] Manual `run_optimization_handler` check on empty `hot_kernels` + `skipped_kernels` artifacts: `kn001`/`k001` reject as `k001`; `aten::mm` and pure hallucinations stay unguessed.

### DIFF
--- a/inference_optimizer/orchestrator/kernel_request_handlers.py
+++ b/inference_optimizer/orchestrator/kernel_request_handlers.py
@@ -1209,6 +1209,22 @@ async def run_optimization_handler(
             single_payload["kernel_id"] = _reconcile_kernel_id(
                 single_payload.get("kernel_id"), candidates,
             )
+        else:
+            # No routable hot candidate (the common TraceLens-failure case
+            # where every candidate is non-routable and lives only in
+            # ``skipped_kernels``). The downstream reusable-native guard
+            # rejects this kernel anyway, but it keys the rejection off the
+            # payload id -- so canonicalize an aliased id (``kn001`` ->
+            # ``k001``) against the full candidate set first, so the
+            # rejection lands on the real ``k00x`` instead of accumulating
+            # hallucinated aliases in ``rejected_kernel_ids``. A pure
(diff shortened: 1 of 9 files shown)


ZhengGong-amd (approved):
Geak function fix. Approved

PR #409, opened 2026-06-03
feat(kb): stamp config.json architectures/model_type as recipe tags

xiaofei-zheng:
## Summary
- Lift `architectures` and `model_type` from the model weights' `config.json` and stamp them into the recipe-snapshot `extras` on every KB write (T0 anchor + KEEP/REVERT/CLOSE amend), so a fine-tuned model's recipe records the same architecture identity as the base model it derives from.
- `cli._load_model_config_tags`: soft-degrade loader (missing / invalid / non-dict `config.json` -> `{}`); normalises `architectures` to a string list and drops empty fields.
- `SharedState`: new `model_architectures` / `model_type` fields (round-trip via the existing field filter; resume rehydrates from `state.json`).
- Write-side only: warm-start still matches on the exact 5-tuple, so this records the tags without changing recipe reuse yet.

## Test plan
- [x] `test_model_config_tags.py`: loader happy-path + soft-degrade matrix (8) + T0 stamp/skip (2)
- [x] `test_coordinator_kb_writes.py`: amend stamp/skip (2)
- [x] Regression: cortex_t0 / model_arch advisory / shared_state evolution / resume / warm_replay (61 + 79 passed)

### DIFF
--- a/inference_optimizer/cli.py
+++ b/inference_optimizer/cli.py
@@ -937,6 +937,64 @@ def _load_model_arch(workspace_root: Path, model_name: str) -> dict:
     return data
 
 
+def _load_model_config_tags(model_path: str) -> dict:
+    """Best-effort loader for KB architecture tags from ``config.json``.
+
+    Reads ``<model_path>/config.json`` (the HF weights dir) and lifts the
+    two architecture-identity fields the recipe-snapshot KB stamps as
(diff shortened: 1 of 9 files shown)


ZhengGong-amd (approved):
limited modification, approved

PR #430, opened 2026-06-04
Run stability: USER_DATA_PATH write-guard, install flock, sglang watchdog, launch-info session discovery

xiaofei-zheng:
## Summary
Fixes for recurring baseline/session failures on MI300X (root-caused from many baseline_failed runs).

- **Harden USER_DATA_PATH path resolution**: collapse scattered `os.environ.get("USER_DATA_PATH", "/workspace/hyperloom")` onto `paths.workspace_root()` (loud one-shot warning only when unset); kernel-agent standalone tools get a local `_paths.py` helper; GEMM-tuning wrapper honors `$INFERENCEX_PATH` instead of hardcoding `/hyperloom/InferenceX`; CI helpers prefer `USER_DATA_PATH`.
- **manifest dependency-escape warning**: warn only when `MAGPIE_DIR`/`INFERENCEX_PATH` escape `USER_DATA_PATH` into a *pod-local* root (`/workspace`, `/tmp`, `/root`) — the real "artefacts vanish on pod recycle" case; persistent WekaFS shared checkouts do not warn.
- **Serialize installer source-mirror writes**: `flock` on `$HYPERLOOM_RUNTIME_DIR/.install.lock` in both `install.sh` entrypoints (with `HYPERLOOM_INSTALL_LOCK_HELD` to avoid the inference_optimizer->kernel-agent self-deadlock); loud `USER_DATA_PATH`-unset warning across 4 scripts.
- **Inject sglang watchdog timeout**: single choke point in `materialize_config_with_envs` adds `--watchdog-timeout 1800` (configurable via `SGLANG_WATCHDOG_TIMEOUT`) for sglang only, never doubling a user value, never touching `--context-length`. Prevents the first-request aiter `mha_batch_prefill` JIT from tripping the 300s scheduler watchdog -> SIGQUIT -> baseline_failed.
- **Fail fast when model context is too small**: do **not** stretch context via `--context-length`; instead read `config.json` max position length and, when `maxpos < ISL + OSL + headroom` (default headroom 512, env `HYPERLOOM_CONTEXT_HEADROOM_TOKENS`), persist `stop_reason=model_context_window_too_small` plus `stop_detail` to `state.json` and `reports/final.{json,md}`, then exit non-zero. This turns stale 2048-ctx checkpoints into a clear terminal reason instead of hundreds of 400s.
- **Self-heal stale MAGPIE_PYTHON**: validate `$MAGPIE_PYTHON` can `import Magpie`; if not (install.sh bakes it in before Magpie is pip-installed, e.g. `/usr/bin/python3`), ignore it and auto-detect. Avoids ModuleNotFoundError on every benchmark.
- **Use launch-info for session discovery**: SKILL/launcher and `robustness_monitor.sh.example` read `session_dir` from `--launch-info-file` JSON (or the `HYPERLOOM_LAUNCH` sentinel); removed the unreliable `ls ... *T*Z | tail -1` guess; monitor fails loudly instead of falling back to `/workspace/hyperloom`.
- **Show stop detail in final.md**: reports now include `Stop detail` when present, so context-window exits explain the exact maxpos/required-token mismatch in the human-readable report.

## Test plan
- [x] Focused pytest suite (path guard / install / watchdog / baseline params / session layout / kb root / startup robustness / magpie-python / context-window / kernel-agent _paths) -> 167 passed
- [x] bash -n on all install/local_setup/preflight/monitor scripts; git diff --check clean
- [x] Local 1h smoke run (Llama-3.2-3B, USER_DATA_PATH=/wekafs/xiaofei/sessions): all artifacts under USER_DATA_PATH, zero writes to /workspace/hyperloom; launch-info session discovery OK; server_args watchdog_timeout=1800.0; MAGPIE_PYTHON self-heal fired; manifest guard did not false-warn on the WekaFS InferenceX checkout; baseline succeeded at 6800.9 tok/s/GPU and the 1h run ended with expected time_exhausted.

## Notes / out of scope
- Lower stack / image: aiter cold JIT cost and `hipcc` mid-build crash during the optimization phase (watchdog mitigates baseline-warmup only).
- This PR intentionally does **not** inject `--context-length`. Small/stale context models fail fast with `model_context_window_too_small` instead of forcing RoPE extrapolation or risking CUDA errors on learned-position models.

### DIFF
--- a/ci/inferenceX_parser.py
+++ b/ci/inferenceX_parser.py
@@ -361,7 +361,7 @@ def merge_model_config(
         "inferencex_path": defaults.get("inferencex_path") or (get_nfs_root() + "/InferenceX"),
         "oob_path": defaults.get("oob_path") or (get_nfs_root() + "/OOB"),
         "tracelens_root": defaults.get("tracelens_root") or (get_nfs_root() + "/TraceLens-internal"),
-        "result_dir": defaults.get("result_dir", "/workspace/hyperloom"),
+        "result_dir": defaults.get("result_dir") or os.environ.get("USER_DATA_PATH") or "/workspace/hyperloom",
         "inferenceX_benchmarks": ifx_benchmarks,
         "inferenceX_api_name": model_cfg.get("inferenceX_api_name", ""),
         "inferenceX_key": model_cfg.get("inferenceX_key", ""),
--- a/ci/publish_artifacts.py
+++ b/ci/publish_artifacts.py
@@ -62,7 +62,7 @@ def _host_reachable(url: str) -> tuple[bool, str]:
 
 def _build_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(description=__doc__)
-    parser.add_argument("--task-dir", default=os.environ.get("HYPERLOOM_RESULT_DIR", "/workspace/hyperloom"))
+    parser.add_argument("--task-dir", default=os.environ.get("HYPERLOOM_RESULT_DIR") or os.environ.get("USER_DATA_PATH") or "/workspace/hyperloom")
     parser.add_argument("--out-dir", default="")
     parser.add_argument("--model", default="")
     parser.add_argument("--display-name", default="")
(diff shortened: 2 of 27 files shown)


ZhengGong-amd (approved):
checked, approved

PR #436, opened 2026-06-04
fix(kernel): make kernel-reactor max_turns env-overridable (HYPERLOOM_KERNEL_MAX_TURNS, default 40)

iraj465:
Problem: the kernel reactor ClaudeBackend default max_turns (4/5) errors 'Reached maximum number of turns' before GEAK is dispatched. Evidence: kernel_e2e.log -> 'Reached maximum number of turns (4)'. Fix: make it env-overridable via HYPERLOOM_KERNEL_MAX_TURNS (default 40), reusing the module-level os import. (Item #5 test_command->geak.submit() is already on main.)


chaojhou (changes requested):
## Review: env-overridable kernel `max_turns`

**TL;DR:** The intent is sound and low-risk — it fixes a real failure (the kernel reactor exhausting its turn budget with *"Reached maximum number of turns"* before GEAK is dispatched). However, it shouldn't merge as-is: it currently conflicts because `main` moved the patched code, and there are a couple of small robustness/consistency nits plus one verification question.

### 1. Blocking — the patch targets code that no longer lives in `cli.py`

The diff edits `inference_optimizer/cli.py` around line 804 (`_build_backends`). On current `main`, `_build_backends` was refactored **out of `cli.py` into `inference_optimizer/cli_backends.py`** and re-imported back into `cli.py`. The hardcoded value this PR targets now lives in `cli_backends.py` and is still `=5`:

```python
# inference_optimizer/cli_backends.py
    if not no_kernel:
        if kernel_codex:
            backends["kernel"] = CodexBackend(model=codex_model)
        else:
            backends["kernel"] = ClaudeBackend(model=claude_model, max_turns_default=5)
    return backends
```

So the fix is **still needed** (the `=5` hardcode persists on `main`), but the `cli.py` hunk is now obsolete — this is the source of the merge conflict. **Please rebase onto latest `main` and move the change to `cli_backends.py`.** `cli_backends.py` already imports `os` at module level, so the "reuse the module-level `os` import" approach still holds there.

### 2. Substantive — does the fix cover the failing path?

- **The default kernel backend is now Codex, not Claude.** `--kernel-codex` defaults to `True` in `cli.py`, so the default run builds `CodexBackend(model=codex_model)` with no turn override; the `ClaudeBackend(max_turns_default=...)` branch only runs under `--kernel-claude`. This PR therefore only helps the `--kernel-claude` path and leaves the default (Codex) turn budget untouched. Is that intended? If the e2e failures also occur on the Codex path, the default case won't be fixed.
- **Evidence says `(4)`, but the cli sets `5`.** The log is *"Reached maximum number of turns (4)"*, yet the kernel `ClaudeBackend` in `_build_backends` is constructed with `max_turns_default=5` (the bare dataclass default is `4`, see `orchestrator/backends/claude.py`). A `(4)` suggests the failing call may not be using the backend built in `_build_backends`, or that the per-tick `run(max_turns=...)` overrides `max_turns_default` (`max_turns or self.max_turns_default`). Worth confirming the override actually propagates to the path that produced `(4)` — otherwise the bump won't take effect where it's needed.

### 3. Minor / nits

- **Env-parse robustness.** The existing precedent guards empty-string env values: `int(os.environ.get("INFERENCE_OPTIMIZER_SPECIALIST_MAX_TURNS", "8") or "8")`. This PR uses `int(os.environ.get("HYPERLOOM_KERNEL_MAX_TURNS", "40"))` without the `or "40"`, so `HYPERLOOM_KERNEL_MAX_TURNS=""` raises `ValueError` and crashes backend construction. Please add the `or "40"` guard (and optionally clamp to a positive int).
- **Env-var naming.** The existing knob is `INFERENCE_OPTIMIZER_SPECIALIST_MAX_TURNS`; this introduces `HYPERLOOM_KERNEL_MAX_TURNS`. Both are fine individually, but mixing prefixes is inconsistent — consider aligning the convention (or noting the deliberate move to a `HYPERLOOM_` prefix).
- **No test.** A small test asserting the env override flows into `max_turns_default` would match the repo's existing coverage style (e.g. `test_claude_backend.py`). Optional for a change this small.
- **Default 40 vs 5** is a large jump, but since it's a per-tick cap (not a target) it's acceptable; only the token/cost ceiling rises.

### Suggested rebased change (`cli_backends.py`)

```python
    if not no_kernel:
        if kernel_codex:
            backends["kernel"] = CodexBackend(model=codex_model)
        else:
            # Kernel reactor needs many turns to read analysis.md, pick hot
            # kernels, and dispatch kernel_optimization to GEAK; the small
            # default trips "Reached maximum number of turns" before GEAK runs.
            kernel_max_turns = int(
                os.environ.get("HYPERLOOM_KERNEL_MAX_TURNS", "40") or "40"
            )
            backends["kernel"] = ClaudeBackend(
                model=claude_model, max_turns_default=kernel_max_turns
            )
    return backends
```

**Net:** direction approved, requesting changes — rebase + relocate to `cli_backends.py` (resolves the conflict), add the `or "40"` guard, and confirm point #2 (Codex-default coverage + that the override reaches the `(4)` path).

chaojhou:
Direction is right and minimal — making the kernel-reactor `max_turns` env-overridable (default 40) targets a real failure (`Reached maximum number of turns` before GEAK is dispatched), CI is green, and the branch is now rebased onto `main` (no more conflict). One blocking item remains on the single changed line.

**Blocking: env parse has no fallback and no lower bound** (`inference_optimizer/cli_backends.py:125-127`)

```python
kernel_max_turns = int(
    os.environ.get("HYPERLOOM_KERNEL_MAX_TURNS", "40")
)
```

1. **Empty / non-numeric → CLI startup crash.** `os.environ.get` only falls back to `"40"` when the var is *unset*; if it is set but empty or non-numeric (common in K8s/Helm manifests, e.g. `HYPERLOOM_KERNEL_MAX_TURNS=""`), `int()` raises `ValueError` inside `_build_backends` and aborts CLI startup.
2. **`<= 0` re-creates the bug this PR fixes.** No minimum check: a `0`/negative value sets `max_turns_default=0`, and since the coordinator passes `max_turns=0` to mean "use the backend default", the reactor effectively runs with 0 turns — back to "no turns to dispatch GEAK".

Suggested fix:

```python
raw = os.environ.get("HYPERLOOM_KERNEL_MAX_TURNS", "40")
try:
    kernel_max_turns = int(raw)
    if kernel_max_turns <= 0:
        raise ValueError
except ValueError:
    logging.warning("invalid HYPERLOOM_KERNEL_MAX_TURNS=%r; using 40", raw)
    kernel_max_turns = 40
```

Minor (non-blocking):
- The `CodexBackend` (kernel-codex) path isn't covered — fine to scope to the Claude path here; align later if desired.
- Consider a one-line test for the env override + invalid-value fallback (`tests/test_claude_backend.py` already exercises `max_turns_default`).

With the parse hardening (fallback + `<=0` floor) this is good to merge.

PR #441, opened 2026-06-04
Run stability follow-ups: stop_reason vocab, manifest resolve guard, monitor launch-info wait

xiaofei-zheng:
## Summary

Post-merge review of #430 surfaced four edge/contract issues. Each is fixed with a reproducing test (written first, fails before / passes after).

- **stop_reason vocab contract** (`cli.py`, `phase_state.py`, `robustness_monitor.sh.example`): the context-window fail-fast wrote an off-vocab `model_context_window_too_small` via raw attribute assignment, bypassing `set_stop_reason()` / `STOP_REASON_VOCAB` (Inv-8.3). The monitor's vocab-based terminal check (and any `set_stop_reason` round-trip) mis-handled it; if `final.md` failed to write, the monitor could even loop re-resuming a fail-fast. Now registered in the vocab, written via the validated setter, and mirrored in the monitor's offline fallback.
- **manifest resolve robustness** (`manifest.py`): the dependency-escape guard's `Path.resolve(strict=False)` raises `RuntimeError("Symlink loop")` on CPython 3.10 for a self-referential symlink and escaped provenance capture, breaking the "never raise out" contract. Now caught at both resolve sites.
- **monitor launch-info race** (`robustness_monitor.sh.example`): the monitor exited 2 immediately when `LAUNCH_INFO_FILE` was set but not yet flushed (monitor racing the optimizer). Now polls a bounded window (`LAUNCH_INFO_WAIT_SEC`, default 60s); explicit session dir still resolves instantly and an unset launch-info still fails fast.
- **preflight advice direction** (`cli.py`): told operators to *raise* `HYPERLOOM_CONTEXT_HEADROOM_TOKENS` to relax a too-conservative gate, but headroom is added to `required`, so raising is stricter. Fixed to *lower*.

## Test plan
- [x] Reproducing test added per fix (verified red before fix, green after)
- [x] `test_context_window_preflight.py` + `test_manifest_units.py` + `test_robustness_monitor.py` (new) + `test_phase_state_plateau.py` pass
- [x] Full suite: 3777 passed, 1 skipped (2m10s)
- [x] `bash -n` clean on the monitor script

### DIFF
--- a/inference_optimizer/cli.py
+++ b/inference_optimizer/cli.py
@@ -1149,8 +1149,9 @@ def _preflight_context_window(args: argparse.Namespace, session_dir: Path) -> bo
         f"(ISL={isl} + OSL={osl} + headroom={headroom}). The workload exceeds "
         f"the model context window; every request would 400. Refusing to run "
         f"(no --context-length override by policy). Lower ISL/OSL for this "
-        f"model, or raise {_CONTEXT_HEADROOM_ENV} if the headroom is too "
-        f"conservative."
+        f"model, or lower {_CONTEXT_HEADROOM_ENV} if the headroom is too "
+        f"conservative (it is added to `required`, so raising it makes "
+        f"admission stricter, not looser)."
     )
     # Persist the stop reason so CI / the robustness monitor read it from
     # state.json + reports/final.json instead of grepping the run log.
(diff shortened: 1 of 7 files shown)


ZhengGong-amd (approved):
Stability update, Approved.

PR #452, opened 2026-06-05
Add daily top8 model inference-optimization CI 

chennyiiis:
## Summary
- New `.github/workflows/inference-optimization-daily.yml`: runs a fixed InferenceX-comparison model set once per day, each model getting up to `--max-hours=24`, pulling the latest commit into the sandbox like `optimize-submit.yml`.
- Fixed pool in `ci/candidates/inferencex_daily_fixed.json` (10 models: DeepSeek R1, Kimi K2.6, MiniMax M2.5/M2.7, GLM-5/5.1, Qwen3-Next, Qwen3.6, gpt-oss-120b, MiMo V2.5-Pro). All rows are registered and pre-staged under `/wekafs/models` in `core42-hyperloom`, so SaFE register hits `local_path` (no re-download).

## Validation (local)
- `py_compile` + linter clean; workflow YAML parses; matrix emits 10 rows (GLM x2 with tp16/nodes2/rayjob_image, the other 8 tp8 unchanged).
- Dry-run: `--tp16 --nodes2` auto-forces `mode=claw` and injects the topology; guard rejects `--tp16 --nodes1` (exit 2); `nodes=1` yields an empty suffix (no behavior change for single-node).

## Test plan
- [ ] Confirm the 8 single-node models still submit normally (`mode=local`, tp8).


chennyiiis:
## Cross-Repo Prompt Cleanup Notes

This PR now has a paired SaFE backend PR: AMD-AGI/Primus-SaFE#593.

Hyperloom-side changes in this branch / PR context:

- `a124ca18` stopped forcing the InferenceX checkout path from CI by default.
- `71f12d18` stopped forcing OOB and TraceLens source paths into SaFE optimization task submissions by default.
- `3548ef1a` simplified `ci/prompt_prefix.txt`:
  - removed redundant `USER_DATA_PATH` export and mkdir;
  - removed prompt-prefix provenance file writes (`hyperloom_source_commit.txt`, `source_commits.json`);
  - removed extra prompt-prefix env exports that belong outside this prelude;
  - kept sandbox safety guardrails around broad `find`, unsafe `kill -9`, and long blocking shell calls;
  - added a clear “run this shell prelude” instruction.
- `46941cef` moved the daily fixed-model inference optimization schedule to 00:00 Beijing time / 16:00 UTC.

Current local follow-up in Hyperloom:

- `ci/prompt_prefix.txt` also has the new single-session safety rule locally:
  - after first launch, never start a new `optimize --model` / “Launch a New Optimization”;
  - terminal `state.json` `stop_reason` is final;
  - unexpected crash recovery must use `optimize --resume` in the same session dir.
- That same rule has already been added to SaFE-generated prompts in AMD-AGI/Primus-SaFE#593.

SaFE-side paired changes in AMD-AGI/Primus-SaFE#593:

- SaFE no longer defaults empty OOB / TraceLens request fields to `/hyperloom/OOB` or `/hyperloom/TraceLens-internal`.
- SaFE no longer prints `OOB path:` / `TraceLens path:` into generated optimization prompts.
- `oobPath` and `tracelensRoot` remain accepted fields for compatibility; they are just no longer silently defaulted into prompts.
- Prompt-builder tests were updated for the removed path lines and the new single-session rule.

Why both sides are needed:

- Hyperloom not sending default tool paths is not enough by itself, because SaFE previously filled empty values with its own defaults.
- The paired SaFE change prevents those stale defaults from reappearing in generated prompts, so sandbox-side Hyperloom setup / `install.sh` owns optional tool path discovery consistently.

chennyiiis:
Update pushed: `79735155 Fix session stop reason recovery.`

Changes included:
- `session.stop_reason` now falls back to the terminal CLOSE phase reason when top-level `state.stop_reason` is empty, and also when top-level is generic `time_exhausted` but CLOSE has a more specific reason such as `sweep_budget_exhausted`.
- Added `sweep_budget_exhausted` to `STOP_REASON_VOCAB` so the coordinator can preserve this reason at top-level instead of downgrading it.
- Added regression coverage for both cases: empty top-level stop_reason and generic top-level `time_exhausted` with specific CLOSE reason.
- Removed the proxy prefix from the default SGLang submit image in `ci/optimize_submit.py`; `_default_sglang_image()` now returns `primussafe/sglang:v0.5.11-rocm720-mi30x-profilerfix`.

Verification:
- `python -m pytest inference_optimizer/tests/test_breakdown_smoke.py::test_session_stop_reason_falls_back_to_close_phase inference_optimizer/tests/test_breakdown_smoke.py::test_session_stop_reason_prefers_specific_close_reason`
- `_default_sglang_image()` prints `primussafe/sglang:v0.5.11-rocm720-mi30x-profilerfix`.

xiaofei-zheng (approved):

PR #483, opened 2026-06-07
fix: inference-optimizer stability fixes (sglang OOM, model gate, patcher drift, pin deps)

xiaofei-zheng:
## Summary

Fixes 4 confirmed production failure modes discovered from 7 failed sessions after 2025-06-05 21:00 BJT:

- **sglang OOM on large-context models**: Inject `--context-length` cap into `EXTRA_SGLANG_ARGS` so models with huge `max_position_embeddings` (e.g. Mistral-Nemo 1M tokens) don't allocate 127GB workspace_buffer and crash. vllm/atom unaffected (they already pass `--max-model-len`).
- **Multimodal model fail-fast gate**: Whitelist-based preflight rejects non-text-generation models before expensive server boot. Explicit vision/multimodal signals (architecture denylist, model_type denylist, `vision_config`/`image_token_id` keys) are checked first, then text-generation markers (`ForCausalLM`/`LMHeadModel`) allow passage. Validated against 4,336 real model configs (4,209 allowed, 126 rejected, 0 errors).
- **Magpie patcher upstream-aware**: When Magpie upstream already uses atomic copy (`_copy_benchmark_script_atomic`), the patcher returns True (no-op) instead of failing with "legacy block not found".
- **Pin Magpie/InferenceX to commit SHAs**: Replicates the GEAK pin pattern (`git_fetch_pinned` helper). Eliminates version drift from upstream HEAD changes. Magpie patch step downgraded from fail-loud to fail-soft (expected no-op with pinned SHA).

## Test plan

- [x] `test_sglang_context_length_cap.py` — 24 tests covering cap logic, existing flag preservation, vllm/atom no-op, unreadable max_pos
- [x] `test_multimodal_gate.py` — 21 tests covering whitelist, denylist, vision keys, missing config, preflight persistence
- [x] `test_magpie_patcher.py` — 50 tests covering legacy patch, upstream-atomic no-op, region scoping, read-only deployment
- [x] `test_local_setup_script.py` — 3 new tests for pinned ref verification
- [x] All 141 focused regression tests pass
- [x] Real sglang server launch with Mistral-Nemo-12B (max_pos=1024000) confirms `context_length=8192`
- [x] Real model scan: `/wekafs/models` 4,336 configs classified correctly
- [x] Known problem models (`thomaskuo-gemma-3-12b-it-heretic`, `vanta-research-atom-v1-preview-12b`) correctly rejected
- [x] `bash -n` + `py_compile` on all modified files

### DIFF
--- a/inference_optimizer/cli.py
+++ b/inference_optimizer/cli.py
@@ -831,6 +831,52 @@ def _load_model_arch(workspace_root: Path, model_name: str) -> dict:
     return data
 
 
+def _load_model_config_dict(model_path: str) -> dict | None:
+    """Best-effort parse of ``<model_path>/config.json`` into a dict.
+
+    Shared soft-degrade reader used by every config.json-derived preflight /
+    KB-tag loader (never blocks launch): returns the parsed mapping, or
+    ``None`` when ``config.json`` is missing / unreadable / invalid-JSON /
+    not a JSON object. Callers decide how to treat ``None`` (the KB-tag loader
+    degrades to ``{}``; the unsupported-model gate degrades to "do not block").
+    """
+    if not model_path:
+        return None
+    cfg_path = Path(model_path) / "config.json"
+    try:
+        raw = cfg_path.read_text(encoding="utf-8")
+    except FileNotFoundError:
+        return None
+    except OSError as exc:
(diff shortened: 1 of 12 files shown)


ZhengGong-amd (approved):
Tested by xiaofei, approved

PR #459, opened 2026-06-05
fix(integrate): invalidate aiter cpp_itfs runtime cache so good kernels aren't discarded

Umangatamd:
## Summary

`integrate` silently discards genuinely-good aiter **cpp_itfs / runtime-compiled** kernels (e.g. `paged_attention`, `csrc/cpp_itfs/pa/pa_kernels.cuh`). `apply_kernel_patch` lands the optimized source and invalidates the `@compile_ops` JIT cache (`<aiter>/jit/build`), but **never invalidates the cpp_itfs runtime cache** at `$HOME/.aiter/build/pa_ragged_*` — where the *served* kernel actually lives. The re-baseline therefore measures the stale **pristine** binary, scores ~0%, and the kernel is flagged `NEEDS_REVIEW` / `REVERT`.

Impact: a kernel that genuinely delivers **~+2.5%** E2E throughput re-baselined at **−0.17% (≈0%)** and would have been thrown away.

Closes #458

## Root cause

- The cpp_itfs cache folder name is `get_default_func_name(md_name, params)` = `pa_ragged_<md5(params)>` — a hash of **kernel parameters, NOT source content** — so pristine and patched builds **collide on the same directory**.
- `compile_template_op` rebuilds only when `lib.so` is missing (`not_built`), so a stale `lib.so` is dlopened regardless of source edits. With `AITER_REBUILD` unset, aiter's own `rm -rf $BUILD_DIR/*` never fires.
- `setup.py develop` (the aiter rebuild command) builds AOT/static ops; `pa_ragged` is **runtime-built** on first call, so `setup.py develop` cannot refresh this kernel class even in principle.
- The existing `_invalidate_aiter_jit_build` move only covers `@compile_ops` (`jit/build`); nothing in Hyperloom touched `$HOME/.aiter/build` or set `AITER_REBUILD`.

## Fix

[...]

PR #485, opened 2026-06-07
fix(integrate): target-type cache-invalidation registry (generalizes #459) + DEFER_TO_E2E gate

### LINKED ISSUE #458: integrate silently discards good aiter cpp_itfs / runtime-compiled kernels (cpp_itfs runtime cache not invalidated)
## Summary

Hyperloom's `integrate` step **silently discards genuinely-good aiter `cpp_itfs` / runtime-compiled kernels** (e.g. `paged_attention`, `csrc/cpp_itfs/pa/pa_kernels.cuh`). `apply_kernel_patch` correctly lands the optimized source and invalidates the `@compile_ops` JIT cache (`<aiter>/jit/build`), but it **never invalidates the cpp_itfs runtime cache** at `$HOME/.aiter/build/pa_ragged_*` — which is where the *served* kernel actually lives. The re-baseline therefore measures the stale **pristine** binary, scores ~0%, and the kernel is flagged `NEEDS_REVIEW` / `REVERT` even though it is a real win.

## Impact

For a kernel that genuinely delivers **~+2.5%** E2E serving throughput, as-shipped `integrate` measured **−0.17% (≈0%)**. With the default `±keep_threshold_pct = 1.0` band, that is a `NEEDS_REVIEW` (not `KEEP`) verdict → the good GEAK-optimized kernel is **discarded / flagged for review and never shipped**. This affects the entire class of aiter `cpp_itfs` runtime-compiled kernels, not just `pa_ragged`.

## Root cause

Flow: `integrate_handler` → `_maybe_apply_kernel_patch` → `apply_kernel_patch.apply_kernel_patch(...)` → `BaselineExecutor` re-baseline → KEEP/REVERT on `gain_pct`.

What `apply_kernel_patch` does for an `aiter/csrc/*.cuh` target:
- `_detect_strategy` selects `rebuild_command = ["/opt/venv/bin/python", "setup.py", "develop"]` for aiter.
- Copies the optimized source over the target.
- `_clear_python_kernel_caches` runs **only for `.py` targets** → skipped for `.cuh`; even when it runs it clears `__pycache__`/triton/inductor, **never `$HOME/.aiter/build`**.
- `_invalidate_aiter_jit_build` moves `<aiter>/jit/build` aside (the `@compile_ops` cache only). **It does NOT touch `$HOME/.aiter/build/pa_ragged_*`.**
- Runs `setup.py develop`.

Why the served kernel stays stale (see `csrc/cpp_itfs/utils.py` in aiter):
- `BUILD_DIR = $HOME/.aiter/build`. With `AITER_REBUILD` unset, aiter's own `rm -rf $BUILD_DIR/*` (utils.py ~L60) never fires.
- The cache folder name is `get_default_func_name(md_name, params)` = `pa_ragged_<md5(params)>` — a hash of **kernel parameters (gqa_ratio/head_size/dtype/…), NOT source content**. So pristine and patched builds **collide on the same directory** (`pa_ragged_2619e6f9b515d89233333a5e48e5f445`).
- `compile_template_op` rebuilds **iff `not_built(folder)`** (i.e. `lib.so` missing). An existing stale `lib.so` is dlopened regardless of source edits.
- `setup.py develop` rebuilds AOT/static ops; `pa_ragged` is **runtime-built** by `compile_template_op` on the first `paged_attention_ragged` call, so `setup.py develop` **cannot (re)build this kernel class** even in principle — the rebuild step is structurally misaligned for cpp_itfs kernels.

A repo-wide grep of Hyperloom for `\.aiter/build|pa_ragged|AITER_REBUILD|cpp_itfs` finds **zero** code hits (only a comment in `tracelens_analysis.py`). Nothing in Hyperloom clears the cpp_itfs runtime cache or sets `AITER_REBUILD`.

## Evidence (RUN2)

- **As-shipped integrate (no cache clear):** served `pa_ragged` `lib.so` md5 `af459837` was **byte-identical before AND after** integrate (mtime unchanged), and the re-baseline server logged **no `start build`** line → `compile_template_op` saw `not_built()==False` and dlopened the stale pristine `.so`. Measured **−0.17%**.
- **After clearing the cpp_itfs cache:** a fresh build (`fbd84cb8`, distinct from pristine, with `start build` present) measured **+2.49%** in the cleanest drift-canceled A/B pair — i.e. ≈ the hand-measured **+2.5%**. Interleaved drift-canceled mean: **+1.54% tput / −1.85% TPOT**.
- The `@compile_ops` `jit/build` cache *was* invalidated (99 → 0 entries), confirming the gap is specifically the **cpp_itfs runtime cache**.

Full diagnosis + artifacts: `/wekafs/upandey/det-exp/qwen14b/hl_integrate_RUN2/diag/DIAGNOSIS.md` (plus `observations.log`, `logs/`, `results/`, `integrate/`).

## Suggested fix / actionable items

1. **Invalidate the cpp_itfs runtime cache for cpp_itfs / runtime-compiled targets.** During apply (detect by target under `csrc/cpp_itfs/` and/or a runtime-compiled `.cuh`), also remove `$HOME/.aiter/build/pa_ragged_*` — scoped to the impacted kernel/module where possible (e.g. by the `MD_NAME` of the `cpp_itfs` driver), falling back to the cpp_itfs build root the scheduler uses — **or** launch the re-baseline server with `AITER_REBUILD=1` so aiter wipes `BUILD_DIR` on import. Non-cpp_itfs targets must keep current behavior.
2. **Gate the KEEP/REVERT decision on a VERIFIED fresh rebuild.** Before trusting the re-baseline number, assert a real rebuild happened (changed served `lib.so` md5, or a fresh `start build` in the re-baseline server log). If not, flag/abort instead of scoring a stale binary.
3. **Generalize cache-invalidation by target type** so future runtime-compiled kernels are covered, since `setup.py develop` alone cannot refresh them.

## Affected code

- `kernel-agent/tools/apply_kernel_patch.py` — `apply_kernel_patch()`, `_invalidate_aiter_jit_build()` (the existing `@compile_ops`-only invalidation), `_clear_python_kernel_caches()` (`.py`-only), `_detect_strategy()` (sets `setup.py develop` for aiter).
- `inference_optimizer/orchestrator/kernel_request_handlers.py` — `integrate_handler()`, `_maybe_apply_kernel_patch()`, and the `gain_pct` → KEEP/REVERT/NEEDS_REVIEW decision.
- Reference (aiter, for cache semantics): `csrc/cpp_itfs/utils.py` (`BUILD_DIR`, `AITER_REBUILD`, `get_default_func_name`, `not_built`, `compile_template_op`, `start build`/`finish build` logging) and `csrc/cpp_itfs/pa/pa_ragged.py` (`MD_NAME = "pa_ragged"`).

### LINKED ISSUE #484: Generalize integrate-side cache-invalidation by target type (extend #458/#459 beyond aiter cpp_itfs to Triton/@compile_ops/inductor) + make E2E integrate gain authoritative
## Background
PR #459 (Closes #458) fixed integrate silently discarding good **aiter cpp_itfs** kernels by invalidating their runtime cache (`$HOME/.aiter/build`) before the E2E re-baseline. That fix is **specific to cpp_itfs**. The same stale-compiled-cache failure mode exists for **other toolchains**, and there is a second, orthogonal way a real win gets discarded.

## Problem 1 — stale compiled cache is toolchain-specific
The integrate re-baseline can score the **PRE-patch** kernel whenever the served kernel is runtime-compiled and the patched source reuses a stale compiled artifact:
- **Triton** (`@triton.jit`, e.g. the editable sglang `fused_moe`): the served `.hsaco` lives in `TRITON_CACHE_DIR` (`~/.triton/cache`). If it isn't invalidated, the re-baseline serves the old kernel.
- **aiter @compile_ops**: `<aiter>/jit/build/module_*` (there is already a `_invalidate_aiter_jit_build`).
- **aiter cpp_itfs**: `$HOME/.aiter/build` (#459).
- **torch inductor**: `TORCHINDUCTOR_CACHE_DIR` / `~/.cache/torch/inductor` / `/tmp/torchinductor_*`.

These are handled (or not) ad-hoc. We want **one dispatch** keyed by target type → toolchain that declares, per toolchain: {cache locations to invalidate, rebuild trigger/env, fresh-build verification}, fed by `_detect_strategy`. #459's cpp_itfs functions should be refactored into this registry **without regressing** its behavior or tests.

## Problem 2 — a stale/artifact-corrupted micro-score hard-kills a real win
A low/inconclusive micro-speedup self-score hard-drops a candidate to `NEEDS_REVIEW` **before** E2E (`kernel_optimization.make_proposal`; KEEP-only promotion in `shared_state`). When the micro-score is a measurement artifact (see the adjacent GEAK patch-capture bug where a real ~1.7x read as 1.0x), a genuinely-good, high-impact kernel is silently discarded.

We want a **`DEFER_TO_E2E`** disposition: a candidate that is high-impact (roofline/time-share) AND correctness-PASSED but has an inconclusive/low micro-score is routed to `integrate_handler` so the **E2E `gain_pct` is authoritative** for KEEP/REVERT — bounded by the existing `max_e2e_attempts` ledger so expensive E2E runs don't blow up.

## Proposed fix
1. **`CACHE_INVALIDATION_REGISTRY`** in `kernel-agent/tools/apply_kernel_patch.py` keyed by toolchain (aiter @compile_ops, aiter cpp_itfs, Triton, torch inductor), each with detect/invalidate/restore/verify/rebuild_env. Refactor #459's cpp_itfs funcs into the registry (preserve public names + tests). `integrate_handler` drives it (`rebuild_env_for_apply_result` + `verify_rebuilt_for_apply_result`); strict no-op off the cache paths.
2. **`DEFER_TO_E2E`** in `make_proposal` + `shared_state` promotion, OFF by default, bounded by `max_e2e_attempts`.

This makes integrate robust against stale compiled caches for **any** runtime-compiled kernel (incl. the editable Triton `fused_moe`) and prevents a self-measurement artifact from discarding a real win. PR to follow (references #458/#459).

iraj465:
Closes #484 · builds on #459 (Closes #458)

## Summary
Two integrate-side robustness fixes so a genuinely-good kernel is never silently discarded:
- **Part B** — generalize #459's aiter-cpp_itfs cache invalidation into a **target-type → toolchain registry** (aiter @compile_ops, aiter cpp_itfs, **Triton**, **torch inductor**), and make `integrate_handler` drive it so the E2E re-baseline always recompiles the patched kernel (no stale-binary scoring) — including the editable Triton `fused_moe`.
- **Part C** — a **`DEFER_TO_E2E`** disposition so a high-impact, correctness-PASSED candidate with an inconclusive/low micro-score is routed to integrate (E2E `gain_pct` authoritative) instead of being hard-killed as `NEEDS_REVIEW`.

## Root cause
1. #459 invalidates only `cpp_itfs` (`$HOME/.aiter/build`). For a **Triton** kernel (sglang `fused_moe`), the served `.hsaco` lives in `TRITON_CACHE_DIR`; if the integrate re-baseline reuses it, it scores the PRE-patch kernel.
2. A low/inconclusive **micro**-score hard-drops a candidate before E2E. When that score is a measurement artifact (cf. the GEAK patch-capture bug where a real ~1.7x read 1.0x), a real win is discarded.
3. (Loop-closer) `build_verification` only trusted GEAK `status ∈ {complete,succeeded,ok}`; GEAK finalizes multi-round runs as `incremental_after_round_<N>`, so a **verified** GEAK win was scored `correctness=missing → NEEDS_REVIEW` and never reached integrate.

## Fix
- `apply_kernel_patch.py`: `CACHE_INVALIDATION_REGISTRY` keyed by toolchain; entries `{detect, invalidate, restore, verify, rebuild_env, gates_keep}`. Adds Triton (`TRITON_CACHE_DIR` move-aside + fresh-`.hsaco` verify) and torch-inductor entries; **refactors #459's cpp_itfs funcs into the registry, preserving their public names + return shapes** (so #459's tests stay green). `toolchain_for` / `rebuild_env_for_apply_result` / `verify_rebuilt_for_apply_result`. Strict no-op (byte-for-byte) off the cache paths.
- `kernel_request_handlers.py::integrate_handler`: registry-driven rebuild env (scoped + always restored) + verified-rebuild KEEP gate (single-node), generalizing #459.
- `kernel_optimization.py`: `make_proposal(..., high_impact=…)` returns `DEFER_TO_E2E` for high-impact + correctness-passed + inconclusive/low micro (OFF by default; carries `fallback_decision`); `build_verification` trusts GEAK finalized statuses / verified speedup as correctness.
- `shared_state.py`: route `KEEP` **and** `DEFER_TO_E2E` to the integrate queue, bounded by the existing `max_e2e_attempts` ledger (`_defer_e2e_budget_exhausted`).

## Test plan (all green)
- `test_apply_kernel_patch_cpp_itfs_invalidation.py` (**#459, 21 — unchanged, stays green after the registry refactor**)
- `test_cache_invalidation_registry.py` (26 — 4 toolchains: detect/invalidate/restore/verify + Triton apply→revert E2E + strict no-op)
- `test_defer_to_e2e_proposal.py` (21) + `test_defer_to_e2e_disposition.py` (11) — defer fires only for high-impact+passed+inconclusive; KEEP / hard-fail / low-impact unchanged; bounded by max_e2e_attempts
- `test_geak_correctness_status.py` (9) — finalized-status / verified-speedup trust; failed-status + trust-disabled stay conservative
- Full `inference_optimizer/tests` suite green except 7 pre-existing/environmental failures (WekaFS mount absent, ray-head, process-reaping) confirmed unrelated via `git stash`.

## End-to-end validation (fully hands-off, bf16 gpt-oss-120B, TP=4, prefill-heavy ISL=4096/OSL=64, CONC=64)
With the GEAK patch-capture fix + this PR, the loop closes with **no human apply**:
- GEAK self-scored the real `micro_speedup = 1.7098x` → `last_kernel_opt.decision = KEEP` (correctness via `geak_assumed_pass`).
- `integrate_handler` ran hands-off: `apply_result.toolchain = triton`, `triton_cache_backup{is_triton:true}`, `cache_clear` removed `~/.triton/cache` + inductor dir → E2E re-baseline **+21.68%** (746.15 → 907.94 tok/s/gpu); `optimization_stack` non-empty.
- Independent CONC=128 standalone cross-check: **+13.04%** (926.7 → 1047.6 output tok/s).
- Provenance: integrated config md5 == GEAK `optimized_codes` (`a6749912…`); serve logs show the `get_default_config` path with **zero** `E=128,N=720` JSON loads.

### DIFF
--- a/inference_optimizer/orchestrator/kernel_request_handlers.py
+++ b/inference_optimizer/orchestrator/kernel_request_handlers.py
@@ -1619,6 +1619,30 @@ async def integrate_handler(
     )
     ctx = RunnerContext(task=fake_task, lease=None)
 
+    # GH #458 + cache-invalidation registry: when apply_kernel_patch moved a
+    # toolchain's compiled/runtime cache aside (aiter cpp_itfs, Triton, torch
+    # inductor, ...), the re-baseline server must (re)compile the patched
+    # kernel from clean state -- otherwise it can dlopen/serve the stale
+    # PRE-patch binary and integrate would score the wrong kernel (the observed
+    # -0.17% on a real +2.5% paged_attention win; the same risk for the Triton
+    # fused_moe win). The registry tells us which rebuild env vars to set for
+    # the detected toolchain (e.g. AITER_REBUILD=1 for cpp_itfs; empty when the
+    # move-aside alone forces recompilation, as for Triton/inductor). Scoped to
+    # this integrate and ALWAYS restored, so every target the registry doesn't
+    # touch is byte-for-byte unaffected.
+    apply_tool = _load_apply_tool()
+    rebuild_env = apply_tool.rebuild_env_for_apply_result(apply_result)
+    _prev_rebuild_env = {k: os.environ.get(k) for k in rebuild_env}
+    for _env_k, _env_v in rebuild_env.items():
+        os.environ[_env_k] = _env_v
+
+    def _restore_rebuild_env() -> None:
+        for _env_k, _prev in _prev_rebuild_env.items():
+            if _prev is None:
+                os.environ.pop(_env_k, None)
+            else:
+                os.environ[_env_k] = _prev
+
     # Multi-node: apply_kernel_patch has just fanned the new source
     # files to every pod. sglang must be FULLY restarted (not resume-
     # pathed) so it re-imports the patched modules; otherwise the
@@ -1649,6 +1673,7 @@ async def integrate_handler(
             ctx.extra = {**(getattr(ctx, "extra", None) or {}),
                          "mn_round_restarted": True}
         except ServerRestartFailed as exc:
+            _restore_rebuild_env()
             revert_result = _maybe_revert_kernel_patch(apply_result)
             return {
                 "status": "failed",
@@ -1674,6 +1699,10 @@ async def integrate_handler(
             "apply_result": apply_result,
             "revert_result": revert_result,
         }
+    finally:
+        # Restore the rebuild env on every path once the re-baseline server has
+        # been launched, so the env override never leaks past this integrate.
+        _restore_rebuild_env()
 
     if not is_valid_measurement(bench_result):
         revert_result = _maybe_revert_kernel_patch(apply_result)
@@ -1687,6 +1716,41 @@ async def integrate_handler(
             "apply_result": apply_result,
             "revert_result": revert_result,
         }
+
+    # GH #458 (point 2), generalized by the cache-invalidation registry: don't
+    # score a stale binary. For any toolchain whose cache the apply moved aside
+    # (cpp_itfs runtime lib.so, Triton .hsaco, inductor artifact), a re-baseline
+    # that reused the stale binary would silently measure the PRE-patch kernel
+    # (the observed -0.17% on a real +2.5% paged_attention win; the same risk
+    # for the Triton fused_moe win). Before trusting gain_pct, assert a real
+    # rebuild landed: verify_rebuilt_for_apply_result() runs the detected
+    # toolchain's fresh-build check (a freshly-compiled artifact newer than the
+    # invalidation) and is a strict no-op (verified=True) when no gating cache
+    # was invalidated -- so this gate is byte-for-byte off the cache paths.
+    #
+    # Single-node only: in multi-node the served cache lives on the serving
+    # pod, not this sandbox, so the rebuild env on the pod restart is the
+    # mechanism and the sandbox-local check is skipped to avoid false aborts.
+    rebuild_check: HandlerResult = {"verified": True, "status": "skipped"}
+    if not is_multi_node():
+        rebuild_check = apply_tool.verify_rebuilt_for_apply_result(apply_result)
+        if not rebuild_check.get("verified", True):
+            revert_result = _maybe_revert_kernel_patch(apply_result)
+            return {
+                "status": "failed",
+                "error_class": "kernel_cache_rebuild_not_verified",
+                "error": (
+                    "re-baseline did not produce a freshly-built kernel cache "
+                    "artifact; refusing to score a possibly-stale binary"
+                ),
+                "decision": "NEEDS_REVIEW",
+                "kernel_id": kernel_id,
+                "patch_path": patch_path,
+                "target_file": payload.get("target_file") or payload.get("source_file"),
+                "apply_result": apply_result,
+                "revert_result": revert_result,
+                "rebuild_check": rebuild_check,
+            }
 
     new_tput = float(bench_result.get("output_throughput") or 0.0)
     gain_pct = ((new_tput - base_tput) / base_tput * 100.0) if base_tput > 0 else 0.0
@@ -1714,6 +1778,7 @@ async def integrate_handler(
         "extra_sglang_args": extra_args,
         "apply_result": apply_result,
         "revert_result": revert_result,
+        "rebuild_check": rebuild_check,
     }
 
 
--- a/inference_optimizer/orchestrator/shared_state.py
+++ b/inference_optimizer/orchestrator/shared_state.py
@@ -75,6 +75,12 @@ _DEFAULT_HOT_KERNEL_MIN_GPU_PCT = 3.0
 # a noisy trace with 15 reusable rows, only the top-N by gpu_pct are
 # enforced -- avoids the LLM stalling on dozens of tiny kernels.
 _DEFAULT_HOT_KERNEL_GATE_TOP_N = 5
+# PART C (DEFER_TO_E2E): the integrate E2E budget per kernel. Mirrors
+# ``record_kernel_integrate_result``'s ``max_attempts`` default so a
+# DEFER_TO_E2E candidate is routed to integrate only while the same E2E
+# budget allows; once exhausted it drops out of the integrate queue and
+# falls back to its non-deferred disposition.
+_DEFAULT_MAX_E2E_ATTEMPTS = 3
 
 # Per-action audit history cap. ``<action>_attempts`` lists keep the most
 # recent N entries (full audit trail — both successes and failures) so
@@ -993,9 +999,11 @@ class SharedState:
         entry["attempts"] = int(entry.get("attempts", 0)) + 1
         if decision == "PARTIAL":
             entry["partial_count"] = int(entry.get("partial_count", 0)) + 1
-        elif decision == "KEEP":
-            # A successful attempt resets streaks; a future regression
-            # should not be auto-retired on stale history.
+        elif decision in {"KEEP", "DEFER_TO_E2E"}:
+            # A KEEP -- or a DEFER_TO_E2E (correctness passed, high-impact,
+            # routed to integrate so the E2E gain_pct decides) -- resets the
+            # streaks so a promising kernel is not auto-retired on stale
+            # PARTIAL/failure history.
             entry["partial_count"] = 0
             entry["failure_count"] = 0
         if is_infra_failure:
@@ -1007,21 +1015,29 @@ class SharedState:
         entry["last_source_file"] = source_file
         entry["last_ts"] = ts
         entry["history"] = history
+        # PART C: surface the DEFER_TO_E2E disposition on the ledger entry so
+        # state inspectors / prompt rendering can see a candidate is awaiting
+        # an authoritative E2E verdict (and what it falls back to).
+        if decision == "DEFER_TO_E2E":
+            entry["deferred_to_e2e"] = True
+            entry["defer_fallback_decision"] = str(proposal.get("fallback_decision", ""))
 
         # Overwrite policy for last_kernel_opt:
-        #   * KEEP always wins (highest micro bubbles up).
-        #   * Non-KEEP only writes when there's no pending KEEP to protect.
+        #   * KEEP / DEFER_TO_E2E always win (both route to integrate).
+        #   * A non-promoting result only writes when there's no pending
+        #     KEEP / DEFER_TO_E2E to protect (PART C: a late REVERT/PARTIAL
+        #     sibling must not clobber a pending deferred candidate either).
         prev = self.last_kernel_opt or {}
         prev_decision = str(prev.get("decision", "")).upper()
         prev_kid = str(prev.get("kernel_id", ""))
         integrated_ids = self._kernel_ids_in_optimization_stack()
         prev_pending = (
-            prev_decision == "KEEP"
+            prev_decision in {"KEEP", "DEFER_TO_E2E"}
             and bool(prev_kid)
             and prev_kid not in (self.rejected_kernel_ids or [])
             and prev_kid not in integrated_ids
         )
-        if decision == "KEEP" or not prev_pending:
+        if decision in {"KEEP", "DEFER_TO_E2E"} or not prev_pending:
             self.last_kernel_opt = {
                 "kernel_id": kernel_id,
                 "decision": decision,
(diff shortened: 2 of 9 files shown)


chaojhou (changes requested):
## Review: target-type cache-invalidation registry + DEFER_TO_E2E

**TL;DR:** High-quality engineering — the registry abstraction is clean and reversible, 88 tests are green, and there's hands-off E2E evidence (+21.68%). I'm on board with the direction. But I don't think it should merge as-is: there's one structural issue (the PR is stacked on a non-`main` base and carries #459 verbatim) plus a few design/scope/correctness items that need an author decision. **Requesting changes** — mostly to resolve scope/merge-path and to confirm a couple of breadth/correctness boundaries, not because the core logic is buggy.

### 1. (Blocking) Base is not `main`, and this PR carries #459 verbatim
- `baseRefName = fix/kernel-agent-image-agnostic-source-and-harness` (not `main`) — this is a stacked PR.
- Evidence: `test_apply_kernel_patch_cpp_itfs_invalidation.py` (#459's 416-line test) is **ADDED** in this diff, and the cpp_itfs block in `apply_kernel_patch.py` (~280 lines) is **byte-for-byte identical to #459**. So ~700 of the 2674 lines here are the already-open #459.
- Risk: duplicate review with the still-open #459, easy divergence between two cpp_itfs implementations, and merge-ordering confusion.
- **Please settle the merge path first**: does #459 land on `main` first and this PR rebase onto `main` (dropping the carried #459)? Or does #485 supersede #459? Until that's decided I've treated the #459 portion as already-reviewed and focused below on #485's new Part B generalization + Part C.

Positive note: `_invalidate_caches_for_apply` restores already-moved caches on failure (`for done_entry, done_rec in reversed(succeeded): done_entry.restore(done_rec)`), which fixes the cross-entry partial-rollback gap I raised on #459. 👍

### 2. Part B — cache-invalidation registry (architecture approved; 2 substantive points)
The `ToolchainCacheEntry` + `CACHE_INVALIDATION_REGISTRY` design is good: each entry self-gates, `requires_compiled` separates rebuild-path vs source-cache toolchains, `gates_keep` controls the integrate KEEP gate, and `rebuild_env_for_apply_result` / `verify_rebuilt_for_apply_result` drive integrate. Using a plain class (not `@dataclass`) to dodge the `spec_from_file_location` + `from __future__ import annotations` crash is a nice catch.

**B-1 (main): Triton/inductor detection is too broad → whole-cache nuke + weak verify.**

```python
_TRITON_SOURCE_MARKERS = ("@triton.jit", "import triton", "from triton", "triton.jit",
                          "tl.load", "tl.store", "tl.program_id", "tl.arange")
```

- Markers like `import triton` / `tl.load` match very widely: many sglang/vllm `.py` files merely import triton or use `tl.*` without being the kernel under patch. If such a file is the patch target, `_target_is_triton` → True → the **entire `$TRITON_CACHE_DIR` (`~/.triton/cache`) is moved aside**. Same for inductor (`torch.compile` is extremely common).
- Consequences: correctness-safe, but the re-baseline becomes a **whole-model cold Triton/inductor recompile**, materially raising per-integrate cost; and `verify_triton_rebuilt` / `_verify_moved_cache_rebuilt` only checks for *any* fresh-mtime file in the recreated cache, so it proves "Triton compiled something," not "the patched kernel recompiled" (cpp_itfs is scoped by `md_name`; the granularity is asymmetric, and the "fresh-.hsaco verify" framing slightly oversells it).
- Suggestion: tighten Triton detection to `@triton.jit` (a definition) or the path markers, so a bare `import triton` doesn't trigger a full nuke; and document that Triton/inductor are whole-cache invalidation + coarse verify.
- Also: `_inductor_cache_dirs()` includes `/tmp/torchinductor_<user>` — moving that aside on a shared pod can disrupt *other* concurrent `torch.compile` processes on the node, not just this re-baseline. Please confirm isolation.

**B-2 (minor): rebuild env still mutates process-global `os.environ`** (save/set/`finally`-restore are all correct). Same concurrency caveat as #459: within the set→restore window any other server subprocess launched would inherit it. Fine if integrate is serialized.

### 3. Part C — DEFER_TO_E2E + GEAK status recognition (approved; 3 to confirm)

**C-1 (correctness, please confirm): `build_verification` trusts `geak_status.startswith("incremental")` + `verified_speedup >= 1.0` as correctness.**

```python
_geak_finalized = (geak_status in {"complete","completed","succeeded","success","ok",
                   "auto_finalized","finalized"} or geak_status.startswith("incremental"))
_geak_verified_ok = geak_verified is not None and geak_verified >= 1.0
if _geak_finalized or _geak_verified_ok:
    correctness_signal = True; correctness_source = "geak_assumed_pass"
```

- `startswith("incremental")` is broad: is there any non-terminal / non-success `incremental_*` status (interrupted / in-progress)? If so it would be mis-read as correctness-passed. Suggest tightening to the `incremental_after_round_` prefix.
- More importantly: `geak_assumed_pass` treats GEAK's *self-reported* `verified_speedup` as correctness, while integrate E2E mainly measures **throughput (`gain_pct`)**, not numerical correctness. Please confirm the integrate/E2E path still has an **independent accuracy/correctness gate** — otherwise a "fast but numerically wrong" kernel could be KEPT.

**C-2 (behavior change, note): `micro_speedup <= 1.0` changes from hard-REVERT to DEFER_TO_E2E for high-impact candidates.** A genuine micro-regression (e.g. 0.8x) on a high-impact kernel is now routed to an expensive E2E re-baseline (fallback=REVERT). It's bounded by `max_e2e_attempts` (default 3) and OFF by default (requires `high_impact` detection), so it's acceptable — but it's a real behavior change worth calling out in the description.

**C-3: routing/budget impl is clean.** `shared_state` routes KEEP and DEFER_TO_E2E through the same integrate queue, bounds defer via the existing ledger (`_defer_e2e_budget_exhausted`), and protects `last_kernel_opt` from clobber. Looks correct.

### 4. Undocumented per-kernel special-case: `_specialize_fused_moe_target`
`kernel_optimization.main()` adds:

```python
# Fix #1: for the autonomously-surfaced fused_moe candidate, route GEAK at the
# tile-config SELECTION site with the real gpt-oss-120b MoE shapes + M>=4096 ...
resolved_source = _specialize_fused_moe_target(args, candidate, log_path)
```

This is a fused_moe + gpt-oss-120b-specific GEAK routing special-case injected into the generic `kernel_optimization` main path. It is **not listed in the PR summary's fix set**, and it contradicts this PR's own stated philosophy ("general mechanism keyed on target type, so future kernels are covered without per-kernel special-casing"). Please confirm whether it should be split out / generalized, or at minimum documented with a note that it's a strict no-op for non-fused_moe / non-gpt-oss targets.

### 5. Tests
Coverage is strong: registry (26 — Triton/inductor detect/invalidate/restore/verify + apply→revert + strict no-op), DEFER proposal (21) / disposition (11), GEAK status (9), and #459 cpp_itfs (21, still green after the refactor). Suggested additions: a negative case for the over-broad Triton detection (a non-kernel file that merely `import triton` should not trigger a whole-cache nuke), and a non-terminal `incremental_*` status staying conservative.

---

**Net (requesting changes), to resolve before merge:** (1) merge path — retarget to `main` + de-dup vs #459; (2) Triton/inductor detection breadth / whole-cache nuke / verify granularity; (3) the `_specialize_fused_moe_target` special-case (keep/split + document); (4) GEAK `incremental*` / assumed-pass correctness boundary. (1) and (3) are scope/process; (2) and (4) are design/correctness. The architecture itself is good — these are about scope and a few breadth/correctness boundaries.

chaojhou:
Re-reviewed — the PR is unchanged since my earlier **Changes requested** (single commit `17fe8f7b`, base still `fix/kernel-agent-image-agnostic-source-and-harness`, so PR-CI is **SKIPPED** and no real CI has run). The architecture (toolchain `CACHE_INVALIDATION_REGISTRY` + reversible failure rollback) is good, but there's now a **confirmed runtime blocker** plus the still-open items below.

### Blocking — confirmed `NameError` on this branch
`kernel_optimization.py` calls `_specialize_fused_moe_target(args, candidate, log_path)` (around L2742), but `def _specialize_fused_moe_target` is **not defined anywhere in the branch tree** (verified with `git grep` on the PR head). Every non-`dry_run` kernel optimization will raise `NameError` before backend dispatch. This is almost certainly a side effect of the stacked, non-`main` base — the definition is expected from `main` or a sibling PR that isn't in this stack.

Relatedly, `test_command` is built but not forwarded into `geak.submit(...)` (no `test_command=` in the file), so GEAK falls back to a fabricated benchmark path.

### Top action — settle the merge path
- Base is `fix/kernel-agent-image-agnostic-source-and-harness`, not `main`, and this branch carries **#459 verbatim (~700 lines)**. Retarget onto `main` (dropping the duplicated #459) so (a) we don't double-review #459, (b) the missing `_specialize_fused_moe_target` / `test_command` wiring resolves against real `main`, and (c) real CI runs instead of `SKIPPED`.

### Design / correctness (carried from the earlier review)
1. **Triton/inductor detection too broad.** Markers like `import triton` / `tl.load` / `torch.compile` match many non-target files; a hit moves the whole `$TRITON_CACHE_DIR` / inductor cache aside (whole-model cold recompile), and `verify_*` only checks for *any* fresh artifact, not the patched kernel — asymmetric vs the `md_name`-scoped cpp_itfs verify. Tighten Triton detection to `@triton.jit` (a definition).
2. **GEAK assumed-pass correctness boundary.** `build_verification` trusts `geak_status.startswith("incremental")` + `verified_speedup >= 1.0` as correctness. Please confirm the integrate/E2E path still has an **independent accuracy gate** (E2E mainly measures throughput), so a "fast but numerically wrong" kernel can't be KEPT; and tighten `startswith("incremental")` to the terminal `incremental_after_round_` prefix.
3. **`_specialize_fused_moe_target` is an undocumented per-kernel (fused_moe / gpt-oss) special-case** injected into the generic path — contradicts the PR's "general mechanism by target type" framing. Split out or document as a strict no-op for other targets.
4. **cpp_itfs verify can false-abort** when invalidation was `skipped` (no cache existed) — unlike Triton/inductor which treat an empty move-aside as a verified no-op; a first-time cpp_itfs patch can be rejected with `kernel_cache_rebuild_not_verified`.

**Net:** still **Changes requested**. The registry design is solid, but the branch currently can't run (confirmed `NameError`). Please (1) retarget to `main` + de-dup #459 so real CI runs and the missing defs resolve, then (2) address Triton breadth, the assumed-pass accuracy boundary, the `_specialize_fused_moe_target` special-case, and the cpp_itfs skipped-verify false-abort.

PR #509, opened 2026-06-09
fix: inject hf-overrides for Command-A vLLM model load

jatseng-ai:
## Summary

- Inject `--hf-overrides {"architectures":["Cohere2MoeForCausalLM"],"model_type":"cohere2"}` for Command-A models at YAML materialization time (vLLM only)
- Prevents vLLM from resolving Command-A checkpoints via `Cohere2VisionForConditionalGeneration`, which hits `CONFIG_MAPPING["cohere2_moe"]` KeyError on transformers builds that do not register `cohere2_moe`
- Skips injection when the operator already pinned `--hf-overrides` (explore variants can re-test the vision path)

This is the upstream equivalent of prestart Patch 1 from FDE ops bring-up on MI300X (command-a-plus-05-2026-fp8, vLLM v0.21.0). Instead of monkey-patching installed `transformers` at container start, Hyperloom pins the correct architecture at the production choke point (`materialize_config_with_envs`), matching the existing per-model workaround pattern (e.g. MiMo-V2 attention backend).

## Test plan

- [x] `pytest inference_optimizer/tests/test_cohere2_hf_overrides.py` — 8 passed
- [ ] Validate Command-A baseline load on MI300X without prestart Patch 1/4 hf-overrides injection
- [ ] Confirm explore variants with explicit `--hf-overrides` are not overridden


chaojhou (commented):
## Review — #509 `fix: inject hf-overrides for Command-A vLLM model load`

Clean, well-scoped per-model workaround (mirrors the existing MiMo-V2 / Gemma2 injection pattern at the `materialize_config_with_envs` choke point). +182/−0, CI green, 8 tests. One thing to confirm before merge, plus a minor note.

### 1. (confirm) The `"command-a"` substring match may be too broad vs MoE-vs-dense variants
`inject_vllm_cohere2_hf_overrides` triggers on `if "command-a" in Path(model_path).name.lower()` and then pins the **MoE** architecture `Cohere2MoeForCausalLM` (+ `model_type=cohere2`). But Cohere's Command-A family includes **dense** checkpoints (e.g. `command-a-03-2025` → `Cohere2ForCausalLM`, not MoE), which do **not** hit the `cohere2_moe` KeyError. If a dense `command-a-*` checkpoint matches this substring, force-pinning `Cohere2MoeForCausalLM` would mis-resolve a dense model as MoE → wrong load / crash — the opposite of the intended fix.

The PR targets `command-a-plus-05-2026-fp8` (the MoE "plus" variant) specifically. Please either confirm every `command-a*` checkpoint Hyperloom runs is MoE, or narrow the trigger (e.g. require `command-a-plus`, or read `architectures`/`model_type` from the checkpoint config and only override when it's the MoE/vision-misresolution case). As written, a dense Command-A run would regress.

### 2. (minor) The override JSON must stay space-free
`_COHERE2_HF_OVERRIDES` keeps the JSON compact (`{"architectures":[...],"model_type":"cohere2"}`) so it's a single CLI token after `--hf-overrides` once `merge_server_args` whitespace-tokenizes. That's correct here — just a fragility note: any future edit that introduces a space into the value would split it into two args and break the load.

### Verified / positives
- **vLLM-only gating** is correct: returns unchanged unless `server_args_env_name(framework) == "EXTRA_VLLM_ARGS"` (sglang / others skip).
- **Operator-wins skip** is correct: `_HF_OVERRIDES_RE = r"--hf-overrides(?:[=\s]|$)"` detects an already-pinned override (and won't false-match a longer flag like `--hf-overrides-foo`), so explore variants re-testing the vision path are not overridden.
- Reuses `merge_server_args` for the append (consistent with the other injections); placed at the same `materialize_config_with_envs` choke point as the attention-backend / MoE-runner workarounds.
- 8 tests cover inject / skip-when-pinned / non-command-a / non-vLLM paths.

### Net
Solid targeted fix. Gate on item 1 — confirm the `"command-a"` match doesn't catch dense Command-A variants (which would be mis-pinned to MoE); item 2 is a fragility note.

Issue #506, opened 2026-06-09
trace_analyze fails on sglang traces: bundled TraceLens splitter can't parse sglang step[...] markers (upstream: TraceLens#723)


Umangatamd:
## Summary
Hyperloom's `trace_analyze` (roofline) step **hard-fails on sglang workloads** because the TraceLens trace splitter it depends on only recognizes **vLLM-style** markers (`iteration` / `_dummy_run`) and does not parse **sglang** annotations (`step[DECODE bs=…]`, `step[EXTEND bs=… toks=…]`). The splitter finds 0 steady-state chunks and raises `trace_split_no_steady_state`, so Hyperloom gets no trace-anchored kernel shape and **kernel-opt (GEAK) never dispatches**.

This is an **upstream TraceLens bug** — filed at **AMD-AGI/TraceLens#723** (`TraceUtils/split_inference_trace_annotation.py` → `main()`, the `ANNOTATION_PATTERN`/`RUNTIME_EVENT_PATTERN` lists + `get_iter_details_from_name()`) — but it's raised here too because it breaks the Hyperloom pipeline end-to-end on sglang, and Hyperloom likely wants its own guard/fallback rather than silently degrading.

## Impact (observed)
Qwen3-30B-A3B local run (sglang, bf16, TP=1): `trace_analyze_failed` on every attempt → `kernel_opt_attempts={}` → GEAK never ran the whole session.

## Evidence
- TraceLens task `tl-7a1dc250`: `RuntimeError: trace_split_no_steady_state`; logs `Found 0 execution steps (iteration) events`, `Found 0 _dummy_run events`.
- The trace itself is valid + sglang-annotated: 251× `step[DECODE bs=64]` + 5× `step[EXTEND …]` (= 256 = num_steps), 0 vLLM markers. Independent parse of the same trace is sane (8,133 GPU kernels, 2,006 ms device time; `fused_moe_kernel` 66% full / 57% decode) — so only the splitter's marker-matching is wrong, not the trace.

## Suggested fix
- Upstream: teach the TraceLens splitter sglang `step[...]` markers and/or add a whole-window fallback (tracked in TraceLens#723); then pin/bump the TraceLens dependency.
- Hyperloom-side: add a guard so `trace_analyze` fails loudly (or falls back) when the bundled splitter can't parse the framework's annotations, instead of silently dropping into degraded mode.

## Related
[...]

PR #515, opened 2026-06-09
fix(kernel-agent): recover high-GPU-time 'other'-bucket kernels dropped by analysis.md-only extraction (Closes #514)

### LINKED ISSUE #514: Candidate extractor drops high-GPU-time 'other'-bucket kernels (analysis.md-only path) → dominant editable kernel never reaches GEAK
## Summary
Hyperloom builds kernel-opt candidates **exclusively** from the `reasoning-candidate` / P-item blocks of TraceLens `analysis.md`. When TraceLens files a kernel under the un-roofline'd **`other`** category (no `reasoning-candidate` block) — as it does for the Triton fused-MoE GEMM, the **#1 kernel at ~67% GPU time** — that kernel is **silently dropped**: it appears in neither `hot_kernels` nor `skipped_kernels`, and GEAK never gets a target.

## Root cause (verified)
- `kernel-agent/tools/tracelens_skill_runner.py:612-685` — `parse_analysis_md()` only walks the `<!-- reasoning-candidate ... -->` / P-item `**Data:**` blocks (`_split_data_blocks` keeps `tier == "compute"` blocks). An `other`-bucket op has no such block, so it never becomes a candidate.
- `kernel-agent/tools/tracelens_analysis.py:2837-2842` — when `analysis.md` yields no candidates the driver **refuses** any `priority_data` / `category_data` / CSV fallback ("analysis.md is the single source of truth") and raises. So there is no path by which an `other`-bucket op is recovered.

Net effect: an op can be the dominant GPU consumer yet be absent from both output lists because it was categorized `other` (no roofline finding / efficiency unknown).

## Impact
On AMD MoE sglang runs the dominant **editable** kernel — the Triton fused-MoE GEMM under `/sgl-workspace/sglang/`, ~67% of GPU time — is dropped. `classify_patchability()` would mark it `reusable_native_kernel=True` (Triton source under a reusable framework root) and route it to GEAK, but it never gets the chance. The single highest-value kernel-opt target is lost for the whole run.

## Suggested fix (Hyperloom-side defense-in-depth)
Keep the `analysis.md` path **primary**. Add a narrow fallback in the candidate-extraction path so that a **high-GPU-time `other`-bucket op with no `reasoning-candidate` block** still becomes a candidate, using the per-op ranking Hyperloom already has (`ops_summary.csv` / `unified_perf_summary.csv` / `priority_data.json` from the TraceLens output dir). The recovered op flows through the existing `_finalize_candidates()` → `classify_patchability()` so it is routed to GEAK iff it is a reusable native kernel. The fallback only fires for high-time ops that are missing from the `analysis.md` candidates (gated by a configurable GPU-time threshold), so the normal path is unchanged.

Add a test: a synthetic `analysis.md` whose `reasoning-candidate` blocks miss a high-time op, plus a per-op ranking sidecar that lists it → the fallback surfaces it as a candidate (and `classify_patchability` marks a Triton kernel under `/sgl-workspace/sglang/` reusable).

## Related
- The **TraceLens-side categorization gap** (filing the kernel under `other` instead of a roofline'd category) is being fixed separately in the TraceLens repo; this Hyperloom fix is the defense-in-depth fallback for older/again-regressed TraceLens builds.
- Sibling trace→shape pipeline blockers: Hyperloom #506 / upstream AMD-AGI/TraceLens#723 (sglang `step[...]` splitter markers).

Umangatamd:
## Summary
`parse_analysis_md` builds kernel candidates **only** from analysis.md `reasoning-candidate` (P-item) blocks (`kernel-agent/tools/tracelens_skill_runner.py:612-685`), and the driver **refuses** any `priority_data`/`category_data`/CSV fallback (`kernel-agent/tools/tracelens_analysis.py:2837-2842`, "analysis.md is the single source of truth"). TraceLens never emits a `reasoning-candidate` block for a kernel it files under the un-roofline'd **`other`** category, so a dominant editable kernel — the Triton fused-MoE GEMM at **~67% GPU time** — lands in neither `hot_kernels` nor `skipped_kernels` and **GEAK never gets a target** (see #514).

This adds a Hyperloom-side **defense-in-depth** fallback. `recover_other_bucket_candidates()` reads the per-op ranking TraceLens already writes (`ops_summary.csv` / `unified_perf_summary.csv` / `priority_data.json`, schema-tolerant) and surfaces a HIGH-GPU-time `other`-bucket op that is **missing** from the analysis.md candidates as a raw candidate, so it flows through the existing `_finalize_candidates()` → `classify_patchability()` path and is routed to GEAK iff it is a reusable native kernel (e.g. a Triton kernel under `/sgl-workspace/sglang/` → `reusable_native_kernel=True`).

**analysis.md stays primary**: the fallback only fires for ops with no `reasoning-candidate` block that exceed a GPU-time threshold (env `HYPERLOOM_OTHER_BUCKET_MIN_GPU_PCT`, default 10%); it is a no-op when the sidecars are absent or nothing qualifies.

## Changes
- `kernel-agent/tools/tracelens_analysis.py`:
  - `load_ops_ranking()` + schema-tolerant CSV/JSON record parsing (GPU-time in ms/us, percent, category, name across common header/field spellings).
  - `recover_other_bucket_candidates()` — the gated fallback (missing-from-candidates ∧ other-bucket ∧ ≥ threshold), ranked by sidecar GPU time.
  - Driver wiring after `parse_analysis_md` (`report_source` becomes `analysis.md+other_bucket_fallback` when it fires).
- `kernel-agent/tools/test_other_bucket_candidate_fallback.py`: 14 tests — sidecar parsing (ops_summary.csv / unified_perf_summary.csv / priority_data.json), the gate (surface / already-a-candidate / below-threshold / roofline'd-category / env override / no-sidecar), the **headline scenario** (synthetic analysis.md missing the high-time op → fallback surfaces it), and end-to-end routing (recovered candidate → `_finalize_candidates` → `classify_patchability` reusable=True for a Triton sglang kernel).

## Test plan
- [x] `python -m pytest kernel-agent/tools/test_other_bucket_candidate_fallback.py -q` → 14 passed
- [x] `python -m pytest kernel-agent/tools/test_tracelens_csv.py -q` → 149 passed (no regression to the analysis.md-only contract / T2 sidecar-isolation tests)
- [ ] CI

Note: `kernel-agent/tools/test_aiter_source_autoresolve.py` has 6 failures in this sandbox that **reproduce on clean `origin/main`** (they probe the real `/sgl-workspace/aiter` device-source layout); they are pre-existing/environmental and unrelated to this change.

Closes #514

## Related
The **TraceLens-side categorization gap** (filing the kernel under `other` instead of a roofline'd category) is fixed separately upstream in the TraceLens repo; this is the Hyperloom defense-in-depth net. Sibling trace→shape pipeline blockers: Hyperloom #506 / upstream AMD-AGI/TraceLens#723.

### DIFF
--- a/kernel-agent/tools/test_other_bucket_candidate_fallback.py
+++ b/kernel-agent/tools/test_other_bucket_candidate_fallback.py
@@ -17,6 +17,8 @@ import sys
 import textwrap
 from pathlib import Path
 
+import pytest
+
 # tools/ is not a package — stick its dir on sys.path so we can import.
 _TOOL_DIR = Path(__file__).resolve().parent
 if str(_TOOL_DIR) not in sys.path:
@@ -93,7 +95,7 @@ def test_recover_surfaces_high_time_other_bucket_op(tmp_path):
         _ops_summary_csv(
             "fused_moe_kernel,other,6700.0\n"
             "aten::mm,GEMM,2300.0\n"
-            "rmsnorm,elementwise,1000.0\n"
+            "rmsnorm,elementwise,500.0\n"  # ~5%, below the 10% floor
         ),
     )
     # analysis.md already surfaced aten::mm; the 67% other-bucket op was dropped.
@@ -106,7 +108,8 @@ def test_recover_surfaces_high_time_other_bucket_op(tmp_path):
     assert cand["candidate_source"] == "other_bucket_fallback"
     assert cand["source_file"] == ""            # resolved later in _finalize
     assert cand["duration_us"] == 6700.0 * 1000.0
-    assert round(cand["gpu_pct"], 0) == 67.0
+    # gpu_pct is over the full ranking total (6700+2300+500 = 9500 ms): ~70.5%.
+    assert round(cand["gpu_pct"], 0) == 71.0
 
 
 def test_recover_skips_op_already_in_candidates(tmp_path):
@@ -114,15 +117,18 @@ def test_recover_skips_op_already_in_candidates(tmp_path):
         tmp_path / "ops_summary.csv",
         _ops_summary_csv("fused_moe_kernel,other,6700.0\naten::mm,GEMM,3300.0\n"),
     )
-    # The op is already an analysis.md candidate -> not duplicated.
+    # The op already surfaced in analysis.md (fused_moe_kernel) is NOT duplicated;
+    # the broadened gate still recovers the high-time GEMM that was missing.
     recovered = tla.recover_other_bucket_candidates(
         tmp_path, [{"name": "fused_moe_kernel"}], top_k=10,
     )
-    assert recovered == []
+    names = {c["name"] for c in recovered}
+    assert "fused_moe_kernel" not in names           # dedup against existing
+    assert names == {"aten::mm"}                      # broadened gate recovers it
 
 
 def test_recover_skips_below_threshold(tmp_path):
-    # other-bucket op at only ~5% of GPU time, default threshold is 10%.
+    # op at only ~5% of GPU time, default threshold is 10%.
     _write(
         tmp_path / "ops_summary.csv",
         _ops_summary_csv("tiny_other,other,500.0\naten::mm,GEMM,9500.0\n"),
@@ -133,9 +139,10 @@ def test_recover_skips_below_threshold(tmp_path):
     assert recovered == []
 
 
-def test_recover_skips_rooflined_category(tmp_path):
-    # A high-time GEMM op missing from candidates is NOT recovered: it is not
-    # an "other"-bucket op (it would have had a reasoning-candidate block).
+def test_recover_surfaces_high_time_non_other_category(tmp_path):
+    # Gate broadened (#515): a high-time op in a *rooflined* category that is
+    # nonetheless missing from analysis.md candidates IS recovered (previously
+    # this was skipped because it was not an "other"-bucket op).
     _write(
         tmp_path / "ops_summary.csv",
         _ops_summary_csv("big_gemm,GEMM,9000.0\nrmsnorm,elementwise,1000.0\n"),
@@ -143,7 +150,8 @@ def test_recover_skips_rooflined_category(tmp_path):
     recovered = tla.recover_other_bucket_candidates(
         tmp_path, [], top_k=10,
     )
-    assert recovered == []
+    names = {c["name"] for c in recovered}
+    assert "big_gemm" in names
 
 
 def test_recover_threshold_env_override(tmp_path, monkeypatch):
@@ -210,7 +218,7 @@ def test_synthetic_analysis_md_missing_high_time_op_is_recovered(tmp_path):
         _ops_summary_csv(
             "fused_moe_kernel,other,6700.0\n"
             "aten::mm,GEMM,2300.0\n"
-            "rmsnorm,elementwise,1000.0\n"
+            "rmsnorm,elementwise,500.0\n"  # ~5%, below the 10% floor
         ),
     )
     recovered = tla.recover_other_bucket_candidates(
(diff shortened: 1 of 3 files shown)


chaojhou:
**Dead code after the gate broadening (commit 2).**

Confirmed on the PR branch (`fix/tracelens-other-bucket-candidate-fallback`): `_is_other_like_category()` (defined at `tracelens_analysis.py:1770`) and its backing `_OTHER_LIKE_CATEGORIES` set (`:1562`) are **defined but never called** anywhere in the file. They were the gate predicate in commit 1 ("other-bucket only"); commit 2 broadened the recovery gate to "any high-GPU-time op missing from existing candidates", so the predicate is now orphaned.

Suggest deleting both `_is_other_like_category` and `_OTHER_LIKE_CATEGORIES` (plus the now-unused `normalize_upstream_category` reference there, if it isn't used elsewhere) so the code doesn't carry a misleading "other-bucket only" signal that no longer matches the actual behavior. Alternatively, if scoping back to other-like categories is intended, re-wire the predicate into `recover_other_bucket_candidates`.

tsrikris:
@Umangatamd the right solution here is to flag missing perf models to the TraceLens team through a ticket with a sample trace. I'd not have a fallback path unless absolutely necessary. Would suggest waiting on this PR since resolving the issue at the source will bypass the need for this.

Umangatamd:
Extended with the **shape-capture** companion (commit 75db75cf) that closes the last fused-MoE → GEAK dispatch gap.

This branch already recovers the dominant fused-MoE kernel as a candidate, but it was emitted with `shapes: []` (the kernel arrives as a dimensionless pybind built-in), so `_validate_kernel_shape_and_paths` rejected the geak→claude→codex ladder with `empty_kernel_shape` ~20ms after dispatch — no harness, no patch, `cumulative_gain_validated=0`.

`_finalize_candidates` now back-fills any empty-shaped fused-MoE candidate from TraceLens's deterministic `perf_report_csvs/ops_unique_args.csv` (the gate/up + down grouped-GEMM operands) and tags `shape_provenance=torch_trace`, so the candidate clears the gate. Scoped to the fused-MoE op pattern.

**CPU-validated on the existing `tl-f68b48f1` trace** (Qwen3-30B-A3B, conc 64, ISL/OSL 1024): the candidate now carries
`(15360,2048)/(128,1536,2048)/(122880,1536) bf16` (gate/up) + `(122880,768)/(128,2048,768)/(15360,8,2048) bf16` (down), matching the hand-written GEAK harness shapes, and `_validate_kernel_shape_and_paths` returns pass (vs `empty_kernel_shape` on the empty control).

Source-of-truth companion: AMD-AGI/TraceLens#727 (Closes AMD-AGI/TraceLens#731) recovers the same operands in `build_operation_metrics`. `analysis.md` is LLM-rendered and not a deterministic shape carrier, so this Hyperloom-side back-fill is what guarantees the gate passes on every run.

chaojhou (approved):

tsrikris:
@chaojhou can you pls revert this PR? This is not acceptable to the TraceLens team

CC: @pprabhan , @Umangatamd , @devalshahamd, @arsalanfrq-a11y

Umangatamd:
Agreed — `_is_other_like_category` and `_OTHER_LIKE_CATEGORIES` are dead after the commit-2 gate broadening (the gate in `recover_other_bucket_candidates` now only checks GPU %, lines 1948–1951). Dropping the "other-bucket only" scoping was intentional so `MoE_fused` with a null roofline is eligible, so I'll delete both rather than re-wire them. `normalize_upstream_category` stays — it's still used at line 1425. Since this already merged, I'll do it as a small follow-up cleanup PR.

Umangatamd:
Fully agree on fixing at the source — that's TraceLens #727, which adds the roofline + reasoning-candidate so this op is surfaced upstream. This path is intentionally a gated defense-in-depth net, not a replacement: analysis.md stays the primary source, it only fires for an op with no reasoning-candidate block above the GPU-% floor (default 10%, env-tunable via `HYPERLOOM_OTHER_BUCKET_MIN_GPU_PCT`), and `classify_patchability` still rejects vendor/native kernels downstream so nothing non-patchable reaches the kernel-opt backend. I'll file the TraceLens ticket with a sample trace, and once #727 lands and bakes we can revisit tightening or retiring this fallback.

lishuoshuo-amd:
Hi @Umangatamd , since some additional code was incorporated after the https://github.com/AMD-AGI/Hyperloom/pull/515 merge, directly reverting might cause some conflicts. Could you please submit a new pull request to resolve this issue? If so, I would greatly appreciate it.

PR #529, opened 2026-06-11
feat(breakdown): author-time recorder for session_breakdown with collector fallback

haishuok0525:
## Summary
Record session_breakdown facts at author time (where they are produced) instead of re-walking heterogeneous artifacts at export time. This surfaces failures and pruned data that the on-disk collectors could not see.

## Details
- **recorder package**: atomic, multi-producer-safe fragment spool (singleton + item shapes, idempotent keys); the read-side assembler concatenates/overlays fragments into the breakdown sections.
- **geak/oob invocations**: instrument `record_kernel_opt` to capture per-backend attempts from the in-process result, including backend-level failures and pre-dispatch / infra failures that never reach `optimization_attempts.jsonl` (the collector's only source) — fixes the invisible-failure problem.
- **critic/robustness**: record each iteration/signal as an item at author time so the robustness backend's workdir pruning no longer drops history; composed back into the `critic_robustness` singleton at assembly.
- **summaries**: mirror `kernel_optimization_summary`, `conc_sweep_summary`, `kernel_roofline`, and `specialist_runs` into the recorder at their write sites.
- **exporter.build**: merge fragments with a fragment-preferred overlay and collector fallback (no fragments => identical to before). Adds an `INFERENCE_OPTIMIZER_BREAKDOWN_DISABLE_RECORDER` kill-switch; recorder errors never poison the export.

## Notes
- The previous PR #528 was closed (not merged); this PR re-submits the change.

## Test plan
- [ ] With no fragments, export output is identical to the previous behavior (collector fallback path)
- [ ] Backend-level / pre-dispatch failures are visible in the breakdown
- [ ] Robustness history is preserved after workdir pruning
- [ ] Setting the kill-switch falls back to collector-only behavior
- [ ] Export is unaffected when the recorder raises


ZhengGong-amd (approved):
All code reviewed, Tested by Haishuo, Approved

xiaofei-zheng (approved):

PR #537, opened 2026-06-11
fix(kernel-agent): fallback on duration for zero-impact TraceLens candidates

### LINKED ISSUE #434: Candidate-export semantics: hot_kernels from analysis.md (idle-gate + capture stochasticity → empty)
# [Hyperloom] ISSUE — candidate-export semantics: hot_kernels come from analysis.md, not summary.json; impact gate + capture stochasticity yield empty candidates

- **Type:** ISSUE (analysis robustness + docs)
- **Target repo:** AMD-AGI/Hyperloom  (`kernel-agent` analysis/export)
- **Status:** PREPARED ON DISK — not filed (gh not authenticated).

## Problem
Two coupled, under-documented behaviors make `hot_kernels` come back empty even when the trace clearly has hot GEMMs:

1. **Source of truth.** Hot kernels are derived from `analysis.md` (`parse_analysis_md → kernel_candidates.json`), **not** from the `summary.json` sidecar. On idle-bound or cuda-graph-folded traces, nothing clears the per-item **impact gate**, so `analysis.md` lists items with `impact_score = 0.0` and they are dropped → empty `hot_kernels`, even though `summary.json` shows real GPU time.
2. **Trace-capture stochasticity.** The same config produced **0 / 2 / 10** hot_kernels across repeats — per-step annotation capture is racy, so candidate count is non-deterministic.

## Evidence (hard, on disk)
`/home/sapmajum/PROJECTS/runs/gptoss120b_20260602_194008/geak_gemm/kernel-agent/runs/geak_gemm_222059/parallel_e2e_summary.json` → `selected_kernel`:
- `"trace_report_path": ".../tracelens/analysis.md"` (candidates parsed from analysis.md)
- `"impact_score": 0.0`, `"impact_score_high": 0.0`, `"impact_score_low": 0.0` while `"duration_us": 6564.0`, `"efficiency_percent": 69.25`, `"percent_of_total": 0.55` — a real, non-trivial GEMM whose impact_score is nonetheless 0.0 (gate not cleared).
- Across runs: `num_hot_kernels` = **10** (`geak_gemm_222059`) vs **1** (`geak_gemm2..5`) vs **0** (cuda-graph runs, see item #6) for materially similar configs.

## Fix (recommended)
- **Document** the contract: candidates = `parse_analysis_md(analysis.md)` filtered by the impact gate; `summary.json` is a sidecar and is NOT the candidate source.
- **Harden the gate:** when every item's `impact_score` is 0.0 but `summary.json`/`gpu_pct` shows real device time, fall back to ranking by measured GPU time (or relax the gate) rather than returning empty.
- **Reduce stochasticity:** profile enough steps / aggregate across the capture window so candidate extraction is deterministic; warn when candidate count varies across the capture repeats.

BaoYunkai:
## Description

This PR fixes issue #434 in Hyperloom's TraceLens candidate export path.

When TraceLens `analysis.md` reports concrete GPU time for candidates but every impact score collapses to zero, Hyperloom now keeps `analysis.md` as the candidate source and ranks those candidates by measured `duration_us` as a conservative fallback. The high-idle gate still wins: high-idle traces continue to return empty `hot_kernels[]` and route to params/backends instead of kernel rewriting.

The change also preserves main's #514 sidecar recovery behavior: recovered candidates remain a separate path and are merged with duration-ranked analysis candidates without letting sidecar rows trigger the #434 zero-impact warning.

## Linked issue(s)

Closes #434

## Tests

Added/updated tests for:
- zero-impact `analysis.md` candidates falling back to `duration_us` order
- positive-impact candidates preserving TraceLens report order
- sidecar `other_bucket_fallback` rows not triggering the `analysis.md` zero-impact warning
- recovered candidates being preserved when duration fallback fires

Commands run:
- `python3 -m pytest kernel-agent/tools/test_tracelens_csv.py -q` → 160 passed
- Bugbot review loop: 5 rounds total; fixed 2 findings; final round found no bugs

## Breaking changes

No.


tsrikris:
@BaoYunkai This feature seems like a fallback when the agent fails. I would prefer it if the issue is tacked at the source, i.e report the bug to TraceLens so we can ensure this doesn't show up in the  Hyperloom pipeline.

PR #546, opened 2026-06-12
[feat] Long-run durability + autonomous-orchestration overhaul

ZhengGong-amd:
## Summary

This branch lands two large, related bodies of work plus supporting cleanup:

1. **Long-run durability** — make a *single* session able to run continuously for days (up to a 14-day ceiling) without leaking capacity, losing accepted wins, growing storage unbounded, or being killed by stale crash history.
2. **Loosened, more-autonomous orchestration** — demote dozens of hard mechanical guards to *advisory* signals so the Orchestration LLM (not the Coordinator) makes the strategic calls, and converge the specialist surface onto a single freeform/domains scope model.

A short, bounded run (`--max-hours ≤ 24`) is intended to behave **exactly as before**: every long-run behavior is gated on cyclic mode + a long/unbounded budget. (The temporary 1/4 wall-clock scaling used for short-run state-transition testing has been reverted — all thresholds are back to their production values.)

---

## 1. Long-run durability (M1–M13)

The work is organized as thirteen merged cut-ins; each removes a specific failure mode that would otherwise end, stall, or do collateral damage during a multi-day run. See `HyperloomV2-long-run-cutin.drawio` for where each lands in the tick/phase loop.

**M1 — Cyclic long-run engine (default-on).** `SWEEP` loops back to `FRAMEWORK_PR` (or `EXPLORE`) as a new *macro-cycle* while budget remains and the run has not globally converged, instead of the legacy chain dead-ending at `SWEEP → CLOSE` and ignoring newly available PRs. Cycle-scoped idempotency keys keep a new cycle from dedup-matching prior-cycle tasks. Default cycle window `DEFAULT_CYCLE_HOURS = 6.0` (env override `INFERENCE_OPTIMIZER_CYCLE_HOURS`). Set `INFERENCE_OPTIMIZER_CYCLIC_PHASES=0` to force the legacy linear chain.

**M2 — Decaying acceptance curve.** KEEP / stack-stable / convergence thresholds ride `threshold(N) = 0.1 + 0.9/N` (`N = macro_cycle + 1`). `N=1` reproduces the legacy 1.0% / 0.5% bar (×2 multi-node); later cycles relax the bar so late, lower-margin wins can still land. Convergence uses the *same* curve, so a run only converges once even the relaxed bar is unmet for N consecutive cycles — a fixed bar would otherwise never converge.

**M3 — Non-terminal leverage routing.** `no_more_leverage` is split into `explore_no_more_leverage` (EXPLORE → KERNEL) and `kernel_no_more_leverage` (KERNEL → SWEEP); both are phase-exit reasons only, removed from `STOP_REASON_VOCAB` and the CLI resume gate. EXPLORE plateau now switches lever to KERNEL instead of ending the run early (it skips straight to SWEEP only when there is no kernel lever, e.g. baseline-only).

**M4 — Phase wall-clock caps + 14-day ceiling.** Unbounded runs (`max_minutes == 0`) cap at the 14-day container lifetime (`DEFAULT_LONGRUN_MAX_MINUTES`). New absolute per-phase cap `= min(effective_session·pct, ceil(24h·pct))` (`PHASE_ABSOLUTE_CAP_REFERENCE_MINUTES = 24h`): short runs bind on the session term (zero regression), long runs bind on the 24h reference so no single phase monopolizes the run.

**M5 — Crash sliding-window + emergency stop.** Emergency stop keys off a trailing 24h crash-rate window (`crash_timestamps` / `recent_crash_count()`) instead of the never-decaying cumulative `crash_count`, so a multi-day run or resume is no longer killed by old crashes while still catching a genuine crash storm.

**M6 — GPU reset opt-in + scoped.** `recover` previously could issue `--gpureset --gpu=all` and hit other tenants. It now resets only the session's `ROCR_VISIBLE_DEVICES` cards (else SKIP). The GPU reset is allowed by default on the exclusive-node assumption; set `HYPERLOOM_RECOVER_ALLOW_GPU_RESET=0` on shared nodes.

**M7 — Gain-based explore dedup unseal.** KEEP'd variants stay de-duped forever, but a prior non-promoting variant (REVERT / KEEP_UNSTABLE / no_promote) unblocks for re-test once its measured gain clears the current decayed KEEP bar. Infra failures (KILLED_OVERTIME / FAILED) stay blocked. An **Acceptance-threshold advisory** surfaces the current bar plus which earlier rejects are now re-testable, so the LLM can choose what to re-propose.

**M8 — LRU disk guard.** The maintenance tick adds an advisory disk guard: when the session partition runs low it LRU-trims per-task `runs/` workspaces (keeping the most recent per action) and warns on an oversized `state.json`, never touching durable optimization state. DB retention prunes events/tasks below the resume anchor (`db_maintenance.py`).

**M9 — Dispatcher re-scan.** The dispatcher re-scans its queue while awaiting in-flight tasks and spawns fitting tasks as GPU lanes free, so a queued GPU task can no longer starve behind a long-running one and leave GPUs idle.

**M10 — State lock + commit scope.** `CORE_STATE_FIELDS` are locked against LLM mutation (the LLM cannot forge cycle/phase state); KEEP integration uses a patch-scoped `git add` so a KEEP can't fold unrelated working-tree files into the win commit.

**M11 — PolicyGate holes closed.** Closed a GPU-ceiling skip and a kernel-propose bypass: kernel-owned actions are `request{target_agent="kernel"}`-only (never propose/delegate), and bench is gated (`policy.py` / `action_surfaces.py`).

**M12 — Single-node server-probe off-switch.** `--robustness-disable-server-probe` (`--no-robustness-disable-server-probe` to re-enable) stops a transient server restart from reading as "unreachable" and tripping a false stop on single-node runs.

**M13 — AVO trajectory review (default-on).** A deterministic trajectory reviewer surfaces exhausted directions to avoid and an under-exploited bottleneck to push on a plateau. The **LLM trajectory reviewer is now default-on** (`INFERENCE_OPTIMIZER_TRAJECTORY_LLM_REVIEW`, commit bfdfef4c) — on a plateau the Coordinator enqueues the trajectory-digest specialist by default to redirect a stuck LLM toward fresh directions; set the env var falsy to opt out. It reuses the specialist dispatch path.

**Supporting durability.** A structured profiler digest (saturation mix, cross-snapshot delta, hot kernels, suggested lever) is injected into the orchestration prompt. KEEP patches are committed so a later REVERT checkout can't wipe prior wins; append-only telemetry ledgers are bounded; warm-replay gain is anchored on single-round (not hot) tput for a fair comparison. Periodic reaper reaps expired serving + GPU leases and orphaned specialist subprocesses; bounded LLM retry/backoff; cycle-boundary soft restart (memory compaction, lease reap, cache clear, server deep-clean); watchdog reclaim of orphaned running tasks.

### Why it runs for 14 days (and stops cleanly)

1. **Bounded time** — per-phase + run deadlines (M4); short runs untouched.
2. **Bounded resources** — disk LRU-trimmed (M8); GPUs never idle (M9).
3. **Not killed by the past** — 24h crash rate, not lifetime count (M5).
4. **Small blast radius** — GPU reset opt-in/scoped (M6); probe false-positives off (M12); PolicyGate holes closed (M11).
5. **Keeps making progress** — bar decays to converge (M2); rejects re-open (M7); plateau switches lever (M3); AVO redirects the LLM (M13).
6. **Loops without rot** — `SWEEP → FRAMEWORK_PR` (M1); state locked + commits scoped (M10).

Net: a long run converges or hits a deadline — it never crashes out, fills disk, idles GPUs, harms neighbouring tenants, or loops forever.

---

## 2. Loosened / autonomous orchestration

- **R1 is the single LLM phase-rejection source.** Removed the explore-grid silent reroute, per-variant Critic gate, Coordinator-enforced next-step checklist, depth/assess throttles, deferred-proposal queue, `wait_for_auto_roofline` dispatch deny, and the dead escape hatch.
- **Hard guards → advisory.** Plateau judgments, critic *strategy* rejects, robustness ActionLadder strategic intents, explore roofline/grid heuristics, and `action_registry` metadata are now advisory rather than blocking; phase budget defaults rebalanced.
- **Orchestration as a persistent multi-turn conversation** with strengthened ReAct feedback; prompt made descriptive rather than prescriptive. EXPLORE↔KERNEL phase interleave is available but **env-gated and off by default** (`INFERENCE_OPTIMIZER_PHASE_INTERLEAVE`).
- **`recover` is robustness-delegate-only** — the Orchestration propose chain was removed.
- **Watchdog** server-liveness monitoring + fast-fail mechanism.

### Specialist surface convergence

- Add `scope` / `mode` / `bench` / `lane` dispatch params; a universal patch-safety gate; scope-based cross-domain Critic.
- **Absorb `dynamic_specialist` into `scope=freeform`** and **cross-domain into `scope=domains`**, deleting the bespoke wave path and `dynamic_action` mechanical critic floor — cross-domain quality is now the LLM Critic's call.
- Freeform becomes a co-equal scope with a bounded transient auto-retry and a GPU-lane throttle; `run_bench` enabled for bench-capable specialists.
- **FRAMEWORK_PR authoring track**: specialists write source patches; patches are grounded against the framework source tree (hallucinated paths rejected with an actionable advisory at both author and apply time) and `git-apply` auto-detects the `-p0..-p8` strip level for forward/reverse application.

---

## 3. Other changes

- **CLI**: text-coercible multimodal models run on the text path with a warning; list/tuple `extra_args` / `extra_server_args` coerced into space-joined shell tokens.
- **Workload env** configuration reworked for vLLM compatibility.
- **robustness-agent**: removed the retired pre-M1 monitoring stack (checks/monitors/providers/conductor/models + their tests).
- **breakdown**: removed `claw_mirror` / `legacy_collectors` and the legacy-writer test sites.
- **CodeQL**: cleared all outstanding code-scanning alerts (uninitialized var, dead code, NaN compare, wrong-arguments, explicit list concat).
- **Cleanup**: removed dead orchestrator/analysis-keyword/usage-parse helpers, translated residual Chinese strings to English, and trimmed process-description comments across the tree. Consolidated planning docs.

---

## 4. Testing

New / updated suites covering the above:

- `test_longrun_phase0..3.py`, `test_longrun_refinements.py`, `test_phase_force_exit.py` — cyclic loop, decaying acceptance, phase caps, force-exit routing, dedup unlock, crash window, disk guard.
- `test_specialist_dispatch_params.py`, `test_specialist_lifecycle.py`, `test_specialist_concurrent_dispatch.py`, `test_per_domain_prompts.py`, `test_specialist_patch_safety.py` — freeform/domains scope, patch safety, GPU-lane throttle.
- `test_warm_replay.py`, `test_phase_state_plateau.py`, `test_recover_executor.py`, `test_multimodal_gate.py`, `test_objective.py`, `test_cli_robustness_options.py` — supporting behavior.

```bash
pip install -e .[test]
pytest
ruff check .
mypy inference_optimizer kernel-agent robustness-agent
```

## 5. Compatibility & rollout notes

- **Zero-regression intent for short runs**: `--max-hours ≤ 24` keeps legacy time-wall, phase ratios, and thresholds.
- **New default-on behaviors**: cyclic phases (`INFERENCE_OPTIMIZER_CYCLIC_PHASES=0` to opt out), `recover` GPU reset (`HYPERLOOM_RECOVER_ALLOW_GPU_RESET=0` to opt out on shared nodes), and the AVO LLM trajectory reviewer (`INFERENCE_OPTIMIZER_TRAJECTORY_LLM_REVIEW` falsy to opt out).
- **Off-by-default knobs**: EXPLORE↔KERNEL phase interleave (`INFERENCE_OPTIMIZER_PHASE_INTERLEAVE`).
- **Stop-reason vocab changed**: `no_more_leverage` is retired in favor of `global_converged` and the per-phase `*_no_more_leverage` exit reasons — any external consumer of `STOP_REASON_VOCAB` or the resume gate should be updated.
</content>
</invoke>


chaojhou (commented):
## Full-mode review — #546 `Long-run durability + autonomous-orchestration overhaul`

Scope: **+4953 / −4085 across 110 files, ~90 commits.** This bundles ≥5 independent concerns (long-run durability R1–R7, orchestration loosening, pre-M1 robustness-stack removal, trajectory review, db-maintenance, freeform-specialist dispatch). It is effectively un-reviewable line-by-line; I leaned on the strong new test coverage (`test_longrun_phase0-3` + refinements, ~1.4k lines) and green unit/CodeQL CI, and focused on the highest-risk deltas below.

### 1. (blocking-ish) GPU-reset default flipped opt-in → opt-out — `recover.py`
`_env_gate_allows_gpureset()` changed from `== "1"` (default **off**) to `not in {0,false,no,off}` (default **on**, "exclusive-node assumption"). Combined with the new autonomous long-run / crash-window recovery, `rocm-smi --gpureset --gpu=all` can now self-fire on any node unless `HYPERLOOM_RECOVER_ALLOW_GPU_RESET=0` is set. On shared / multi-tenant nodes a node-wide reset is destructive to co-tenants. It is co-gated by `force_cleanup`, which softens but does not remove the risk for autonomous recovery paths.

**Ask:** keep opt-in, OR only enable when node-exclusivity is positively detected (not assumed), OR at minimum emit a loud one-time startup WARN that reset-on-recovery is armed. Please confirm long-run autonomous recovery never sets `force_cleanup=True` implicitly on shared infra.

### 2. (must-do before merge) Branch is behind `main`
`origin/main` is **not** an ancestor of this head (main just advanced to `3e12b7db`). Merge/rebase onto current main and re-run CI before landing; `mergeStateStatus=BLOCKED` is review-required, but the staleness matters given the breadth.

### 3. CodeQL findings
- **`coordinator.py:8147` `except BaseException`** — intentional (carries `# noqa: BLE001 — mirror gather(return_exceptions=True)`); it only captures the *already-completed* child task's result via `atask.result()`, so it cannot swallow an outer cancellation. **Acceptable as-is** — please annotate/dismiss the alert rather than narrowing, since narrowing to `Exception` would drop child `CancelledError` capture.
- **`coordinator.py:750` empty `except: pass`** and **`test_longrun_phase1.py:292` unused `intent`** — trivial; add a one-line comment / drop the dead local to clear the scanner.

### 4. (verified, no action) pre-M1 robustness-stack removal (~2k lines)
Deletion of `checks/`, `conductor.py`, `models.py`, `monitors/`, `providers/` is clean — no dangling imports remain (residual `conductor.db` references are the SQLite event-DB path in the retained `signals/` + `sources/` stack, not the deleted module). Green CI corroborates.

### Recommendation
Approvable in principle — durability work + tests look solid — but **gate on (1) the GPU-reset default decision and (2) rebase onto main.** Given the size, strongly consider splitting the robustness-stack removal and the orchestration-loosening into their own PRs for rollback granularity, even if landed back-to-back.

chaojhou (commented):
## Full-mode review — deep pass (core implementation)

Read the substantive deltas end-to-end (`coordinator.py` +976, `shared_state.py` +324, `policy.py`, `backends/{base,claude}.py`, `integrate_patch.py`, and the new `db_maintenance.py` / `trajectory_reviewer.py`). New findings below; the first comment's GPU-reset + rebase + CodeQL items still stand.

### New findings

**1. (medium) Second shared-node blast-radius default-on — cycle soft-restart server kill**
`_run_cycle_soft_restart` → `_restart_inference_servers()` → `_grid_runner._kill_stale_servers()` does a `/proc` sweep that kills lingering vLLM/SGLang/atom processes at every macro-cycle boundary, **default ON** (`is_cyclic_phases_enabled()` is default-on; opt-out via `INFERENCE_OPTIMIZER_DISABLE_CYCLE_SERVER_RESTART`). This is the same "exclusive-node assumption" as the GPU-reset flip in comment #1. It's narrowed (only fires on long-run macro-cycle loopbacks, never short bounded runs), but on a shared/multi-tenant node a co-tenant's pattern-matching inference server could be killed.
**Ask:** confirm `_kill_stale_servers()` excludes *all* non-owned PIDs (not just our own process group — a co-tenant's vLLM is also outside our PG), or gate this default behind the same positive node-exclusivity detection you settle on for GPU reset. The two together define the node-exclusivity contract; please make it one explicit, documented signal rather than two independent "assume exclusive" defaults.

**2. (low) Unbounded runs are now silently bounded to 14 days**
`run()` previously set `deadline = None` for unbounded runs; it now always sets `deadline = now + DEFAULT_LONGRUN_MAX_MINUTES` (`= 14*24*60`). A sane safety net, but it's a behavior change — an operator who previously ran truly-unbounded will now see a hard `time_exhausted` at 14d. Please confirm intended and surface it in the long-run docs/CLI help.

**3. (low-medium) `_maybe_force_stalled_domain_specialist` drops the escalation on a failed dispatch**
The per-anchor counter is zeroed via `note_specialist_dispatched(anchor)` *before* `await self._handle_intent(...)`. If `_handle_intent` raises (it's caught + `continue`), the stall counter is already reset but no specialist ran and no completion will reset it again — so the domain stays uncovered for another full `FORCE_STALLED_*_ROUNDS` window. Suggest zeroing only after a successful enqueue (or restoring the prior value on failure).

**4. (nit) `_reap_dispatched_task` single-element `zip([...], [...])` loop**
The intentional "preserve the original post-`gather` body verbatim" wrapper (one-element zip so `continue` acts as early-return) is correct but reads as dead-loop machinery. Worth a follow-up flatten to a straight-line body now that it's a per-task method.

### Verified correct / positive (so it's on record as reviewed)
- **`policy.py` is net-hardening, not loosening:** locks the cyclic counters (`macro_cycle`, `gain_at_cycle_start`, `no_gain_cycle_streak`, …) into `CORE_STATE_FIELDS` so `update_state` can't forge convergence/acceptance state; closes the `propose_action` channel for kernel-owned actions; and routes `scope='freeform'` + bench specialists through the same GPU-pool ceiling (previously a hole).
- **Retry/backoff (R6) is done right:** `RetryPolicy.from_env` validates finite/≥0 with per-field fallback (the exact robustness #436 was missing), and `claude.py` wires `retry_on=(TimeoutError, BackendError, ConnectionError, OSError)` — notably **not** `CancelledError` — with a per-attempt `wait_for` cap.
- **warm_replay measurement-basis fix is a real correctness win:** gain / `current_best.tput` now anchor on the single-fresh-server round (`warmup_round_tput`), matching the basis explore/sweep variants are judged against, with the hot value retained separately as `hot_tput`.
- **`db_maintenance.py` retention invariants are sound:** events pruned strictly below `min(min_cursor, max_seq - keep_recent)` (resume-safe); tasks pruning never touches queued/running/failed/needs_manual_review. Bounded-growth caps across `shared_state` ledgers are consistent and advisory-only on derived totals.
- The `except BaseException` at `coordinator.py:8147` (CodeQL) is confirmed intentional and safe — it only captures an already-completed child task's result and cannot swallow an outer cancellation.

### Net
The engineering quality of the durability layer is high and well-tested. Remaining merge gates: **(a) the GPU-reset + server-kill node-exclusivity contract (finding 1 here + #1 in the first comment), (b) rebase onto current main, (c) the trivial CodeQL/cleanup items.** Findings 2–4 are confirm/cleanup, not blockers.

xiaofei-zheng (approved):

PR #547, opened 2026-06-12
fix(inference): dedup vLLM attention-backend (#520), preserve GEAK tunnel URL (#521), allow custom orchestration model (#340)

### LINKED ISSUE #340: Local Hyperloom Setup is AMD Network Cluster Bound
### Target release

v0.4

### What kind of issue is this?

Suggestion / Feature idea

### Related area

General / Not sure

### Description

Do we have a plan to make the Hyperloom local setup independent of the AMD network? Currently, the local setup requires an AMD-API-key which is bind to Hyperloom, which does not work on remote clusters such as Vultr or TensorWave. It only works on the Alola cluster, which is sometimes fully occupied, making it difficult to get a reservation. If there is a way to make the local setup work on non-AMD network clusters, it would be very helpful for development.

### LINKED ISSUE #520: Duplicate --attention-backend flags in launch command cause EngineCoreProc initialization failure
**Context:** discovered while running the inference_optimizer on `command-a-plus-05-2026-fp8` (FP8, TP=8) at ISL=100k/OSL=1k on AMD MI300X, vLLM `v0.21.0+rocm722`, image `vllm/vllm-openai-rocm:v0.21.0`, restart 34 (2026-06-08). Node-access/Slurm issues excluded.

On 2026-06-08, multiple baseline attempts failed with:

```
RuntimeError: Engine core initialization failed. See root cause above. Failed core proc(s): {}
```

followed by NCCL heartbeat monitor `Broken pipe` errors on all 8 ranks. Root cause in `benchmark_stderr.log`: `--attention-backend` appears **twice** in the `vllm serve` command:

```bash
vllm serve ... --attention-backend ROCM_AITER_FA --attention-backend ROCM_AITER_FA
# also seen: ... --attention-backend ROCM_AITER_FA --attention-backend ROCM_FLASH
```

**How duplication occurs:**

`EXTRA_VLLM_ARGS` in `baseline_vllm.yaml` may include `--attention-backend ROCM_AITER_FA` (required for long-context ROCm baselines). Hyperloom also appends `--attention-backend` from sweep/kernel `extra_args` or a config merge step. The two injections are not deduplicated before constructing the final `vllm serve` invocation.

vLLM v0.21.0 does not tolerate duplicate/conflicting `--attention-backend` flags — crash propagates through `EngineCoreClient` → `wait_for_engine_startup` → `RuntimeError`, then NCCL `Broken pipe` on all ranks.

**Log evidence (2026-06-08T21:50, session `20260608T214843Z`):**
```
benchmark_stderr.log:
  setsid vllm serve ... --attention-backend ROCM_AITER_FA --attention-backend ROCM_AITER_FA

server.log:
  [rank4]: Failed to check "should dump" flag on TCPStore: Broken pipe
  (EngineCore pid=4255) Exception: WorkerProc initialization failed
  (APIServer pid=3951) RuntimeError: Engine core initialization failed.
```

**Fix:**

Before constructing the final `vllm serve` invocation, deduplicate CLI args:
- If `--attention-backend` appears in both `EXTRA_VLLM_ARGS` and `extra_args`, keep only one (prefer `extra_args`).
- General pre-launch dedup pass for other conflicting flags (`--gpu-memory-utilization`, `--max-model-len`).

**Related:** #465 (Final summary hides EngineCore error behind benign WARNs — same failure mode, different layer)

**Impact:** P0 — baseline/sweep runs fail; burns session budget on NCCL crash loops.

### LINKED ISSUE #521: GEAK container has no network path to LLM API → KERNEL phase burns budget on recovery loops, zero patches
**Context:** discovered while running the inference_optimizer on `command-a-plus-05-2026-fp8` (FP8, TP=8) at ISL=100k/OSL=1k on AMD MI300X, vLLM `v0.21.0+rocm722`, image `vllm/vllm-openai-rocm:v0.21.0`, restarts 33 (2026-06-05) and 34 (2026-06-09). Node-access/Slurm issues excluded.

In every run that reached the KERNEL phase (restarts 33 and 34), **zero kernel patches were generated or applied**. The journal shows 44 consecutive `kind=other, change=recover, outcome=KEEP` entries — KERNEL runs 1–3 hours with no mutations.

**Root cause:**

The GEAK kernel agent (`geak-run` container) cannot reach the AMD LLM gateway (`core42.primus-safe.amd.com`). `geak_run.log`:

```
minisweagent.models.amd_claude.AmdClaudeModel._query_api ... raised APIConnectionError: Connection error.
(repeated with 4s → 8s → 16s → 32s → 60s backoff)
```

**Why orchestrator works but GEAK does not:**

The Hyperloom orchestrator reaches the gateway via a host-local reverse tunnel (port 18444), configured in `cli.py`, `~/.claude/config.json`, and `Hyperloom/.env`.

GEAK runs in a **separate container** with its own network namespace. `AmdClaudeModel` reads `~/.config/mini-swe-agent/.env` pointing at the gateway host directly. Gateway IP is not routable from TW compute nodes.

**Evidence:**
- `geak_run.log`: `APIConnectionError` on every `_query_api` call
- Journal: 44 KERNEL `recover/KEEP` entries, zero `kernel_*` or `patch` entries
- `patches/` empty in every session that reached KERNEL
- restart33: 11 KERNEL recover entries, 0 patches
- restart34: 33 KERNEL recover entries, 0 patches

**Fix options (preference order):**

1. **Inject tunnel URL at GEAK launch:** pass `OPENAI_BASE_URL=https://127.0.0.1:18444/api/v1/llm-proxy/v1` + API key as env vars; map port 18444 into GEAK container (`--network host` or `-p 18444:18444`).
2. **Run GEAK with `--network host`** so it inherits the host tunnel binding.
3. **Write tunnel URL into GEAK config** before Hyperloom spawns the container.

**Related:** #467 (roofline eats budget before KERNEL — separate, but both block optimization outcomes), #469 (PolicyGate / `HYPERLOOM_KERNEL_AGENT_ROOT` — different failure mode), #470 (`geak_prompt_patcher` drift — unrelated to network)

**Impact:** P0 — blocks all kernel optimization; primary reason no throughput improvement despite full-budget runs.

chennyiiis:
## Summary

Three related fixes for running the inference_optimizer reliably and off the AMD network.

### #520 — Duplicate `--attention-backend` crashes EngineCoreProc (P0)
vLLM v0.21.0 hard-errors on a duplicate/conflicting `--attention-backend`, crashing `EngineCoreProc` on all ranks (then NCCL `Broken pipe`). The flag is injected by both the operator's YAML `EXTRA_VLLM_ARGS` base **and** a sweep/kernel variant's `extra_server_args`, and `merge_server_args` keeps both tokens by design (sglang tolerates repeats as last-wins).
- Add `dedup_vllm_server_args()` in `_grid_runner.py`: collapse repeated single-value flags (`--attention-backend`, `--gpu-memory-utilization`, `--max-model-len`, ...) to **last-wins** for vllm/atom only; no-op for sglang. Unknown flags / store-true switches / positionals are preserved verbatim and in order.
- Wire it at the single `materialize_config_with_envs` choke point every benchmark path (baseline/profile/sweep/explore/framework_pr/conc_sweep) funnels through, so the variant override survives and the duplicate never reaches `vllm serve`.

### #521 — GEAK has no network path to the LLM gateway (P0)
GEAK runs in a separate network namespace that frequently cannot route to the gateway directly, while the orchestrator reaches it via a host-local reverse tunnel. Preflight previously force-overwrote `GEAK_BASE_URL`/`OOB_BASE_URL` back to the direct gateway URL, so pointing GEAK at the tunnel was impossible.
- Preflight now **preserves an intentional operator override** for `OOB_BASE_URL`/`GEAK_BASE_URL`/`LLM_API_BASE`, while still force-rewriting the known-stale legacy `127.0.0.1:4002` auth-proxy URL (`_is_stale_proxy_url`).
- An unset alias still defaults to the gateway, so existing single-network deployments are unchanged.

### #340 — Local setup is AMD-network bound (feature)
The orchestration model is hard-gated to AMD's `opus-4-7`/`opus-4-6`, which non-AMD gateways (Vultr / TensorWave / self-hosted) do not serve.
- Add opt-out `INFERENCE_OPTIMIZER_ALLOW_CUSTOM_ORCH_MODEL=1`: skip the static allowlist and validate the chosen model against the gateway `/models` catalog instead (a typo still fails; under opt-out there is no AMD opus-4-6 fallback). Default behaviour is unchanged.
- Document the non-AMD path and the GEAK tunnel override in `.env.template`.

### Drive-by test fix
- `test_baseline_executor_pins_magpie_inferencex_path` went red on `main` after PR #536 (mirror InferenceX to local disk -> `MAGPIE_INFERENCEX_PATH` now points at the mirror). Isolate the #210 env-inheritance contract by disabling the mirror in that test; the mirror default keeps its own coverage in `test_baseline_warmup_double_run.py`.

## Test plan
Validated on a vLLM ROCm container (vllm `0.19.0+rocm721`, pytest 9.0.2):
- `test_grid_runner.py` — 75 passed (incl. 10 new `TestDedupVllmServerArgs`)
- `test_preflight_auth_override.py` — new #521 tunnel-preservation + stale-rewrite + #340 custom-model opt-out cases pass
- `test_baseline_param_overrides.py` — 32 passed (was 1 red on main, now green)
- Linters clean.

## Notes
- Pure orchestration logic (string/env handling), no GPU dependency.
- The drive-by test fix is a follow-up to the already-merged PR #536, included here only because it surfaced during validation.

Closes #520
Closes #521
Closes #340


ZhengGong-amd (approved):
Modification validated. Approved

PR #549, opened 2026-06-12
fix(report): surface real baseline_failed root cause, not benign WARN (#465)

### LINKED ISSUE #465: Final summary blames [WARN] modeling_cohere2.py not found instead of the real EngineCore error
**Context:** discovered while running the inference_optimizer on `command-a-plus-05-2026-fp8` (FP8, TP=8) at ISL=100k/OSL=1k on AMD MI300X, vLLM `v0.21.0+rocm722`, image `vllm/vllm-openai-rocm:v0.21.0`, across restarts 30–33 (2026-06-04/05). Node-access/Slurm/SSH-tunnel issues excluded by request.

On `baseline_failed`, the top-level summary highlighted:
```
[WARN] Error: [Errno 2] No such file or directory: '/app/.../transformers/models/cohere2/modeling_cohere2.py'
```
This WARN is harmless (the model resolves `Cohere2MoeForCausalLM` fine). The actual failure (NCCL/HIP OOM, and upstream of that, the FLEX slowness) was buried in a per-attempt `runs/baseline/<hash>/benchmark_vllm_*/server.log`. This misdirected debugging onto a non-issue.

**Fix:** propagate the actual EngineCore / worker exception from the per-attempt server.log into the top-level `Final summary` and `final.json`.

**Impact:** P1 — sends operators chasing red herrings; costly when each iteration is a node-hour.

haishuok0525:
## Summary

Fixes #465.

On `baseline_failed`, the end-of-run summary (`final.json`, the Markdown report, and the stdout block) headlined a **benign upstream WARN** — e.g. transformers' `modeling_cohere2.py` `[Errno 2] No such file or directory` — instead of the **real terminal engine/worker fault** (OOM during `ncclCommInitRank`, `EngineCore failed to start`, etc.). Operators saw a misleading top-level error and had to dig through `server.log` to find the actual cause.

This PR surfaces the real root cause and demotes the benign WARN.

## Changes

- **`report.py` — `failure_summary` block**: on `baseline_failed`, read the most recent failed `runs/baseline/<task_id>/result.json` and lift its `error` / `error_class` into a compact `failure_summary`. When that error is missing or matches a known-benign WARN, fall back to scanning the attempt's `server.log` terminal marker via `server_log_death_excerpt`. Adds a coarse `root_cause_type` classifier (`oom` / `benchmark_timeout` / `engine_core_init` / `worker_crash` / `unknown`), plus `last_attempt_id`, `server_log` (session-relative), and `suppressed_benign`.
- **`report.py` — highlight denylist**: benign upstream WARNs are filtered out of event highlights so they can never become the top-level highlight. Full text still lives in the per-attempt logs.
- **`_format_md`**: render a `Root cause` + `Server log` line on failure runs.
- **`cli.py` — `_print_final_summary`**: reads `failure_summary` from `final.json` and prints `root_cause` / `server_log` on `baseline_failed`.

All new behaviour is gated on `stop_reason == "baseline_failed"` and is best-effort (never raises), so success/other-stop runs and legacy callers are unaffected.

## Test plan

- [x] New unit tests in `inference_optimizer/tests/test_baseline_failure_summary.py` (9 cases): result-error path, benign-WARN suppression + `server.log` fallback, latest-attempt selection, gating on stop reason / `session_dir`, classifier, highlight denylist, and Markdown rendering.
- [x] Existing `test_roofline_comparison_pipeline.py` (32 cases) still pass — `_build_summary_dict` signature change is backward compatible.
- [ ] CI on Linux (local Windows cannot import the package due to the Unix-only `fcntl` dependency in the import chain; tests were run locally with an `fcntl` stub).

### DIFF
--- a/inference_optimizer/cli.py
+++ b/inference_optimizer/cli.py
@@ -1810,17 +1810,42 @@ def _default_target_summary(args: argparse.Namespace) -> str:
     return f"Optimize {Path(args.model).name} for up to {args.max_hours}h (no target)."
 
 
-def _print_final_summary(state: SharedState, stop_reason: str) -> None:
+def _read_failure_summary(session_dir: Path) -> dict | None:
+    """Read ``reports/final.json``'s ``failure_summary`` block, if present.
+
+    Best-effort: returns ``None`` when the file is missing/unreadable or the
+    block is absent (e.g. non-failure runs). Used to surface the real terminal
+    root cause in the end-of-run summary on ``baseline_failed`` (#465).
+    """
+    try:
+        from .session_paths import reports_dir
+        final_json = reports_dir(session_dir) / "final.json"
+        data = json.loads(final_json.read_text(encoding="utf-8"))
+    except (OSError, json.JSONDecodeError, ValueError):
+        return None
+    fs = data.get("failure_summary") if isinstance(data, dict) else None
+    return fs if isinstance(fs, dict) else None
+
+
+def _print_final_summary(
+    state: SharedState,
+    stop_reason: str,
+    session_dir: Path | None = None,
+) -> None:
     """Print the end-of-run summary block to stdout.
 
     Reports the stop reason, session id, model, baseline throughput, the
     per-round (informational) cumulative gain, the validated cumulative gain
     (with a staleness warning when the optimization stack grew after the last
     validation), the current best config, pruned families, and crash count.
+    On ``baseline_failed`` it also surfaces the real terminal root cause from
+    ``reports/final.json`` instead of a benign upstream WARN (#465).
 
     Args:
         state (SharedState): The final shared state after the run completes.
         stop_reason (str): Why the run stopped (e.g. ``"target_reached"``).
+        session_dir (Path | None): Session root, used to read the
+            ``failure_summary`` block on failure runs.
 
     Returns:
         None
@@ -1831,6 +1856,19 @@ def _print_final_summary(state: SharedState, stop_reason: str) -> None:
     print(f"  session_id           : {state.session_id}")
     print(f"  model                : {state.model_name}")
     print(f"  baseline_tput        : {state.baseline_tput:.1f} tok/s/GPU")
+    if session_dir is not None and stop_reason == "baseline_failed":
+        failure_summary = _read_failure_summary(session_dir)
+        if failure_summary and failure_summary.get("root_cause"):
+            print(
+                f"  root_cause           : "
+                f"[{failure_summary.get('root_cause_type', 'unknown')}] "
+                f"{failure_summary.get('root_cause')}"
+            )
+            if failure_summary.get("server_log"):
+                print(
+                    f"  server_log           : "
+                    f"{failure_summary.get('server_log')}"
+                )
     print(
         f"  cumulative_gain      : {state.cumulative_gain:.2f}% "
         f"(per-round sum — informational)"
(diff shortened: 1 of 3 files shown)


chaojhou (commented):
## Full-mode review — #549 `fix(report): surface real baseline_failed root cause (#465)`

Read all three files end-to-end (`report.py` +248, `cli.py` +40, the +224-line test). The core logic is clean, fail-soft, and well unit-tested. Two questions need confirming before merge because they decide whether the fix actually fires end-to-end on a real run, plus two smaller notes.

### 1. (must confirm) The exact-match on `"baseline_failed"` looks like it misses the sibling `prelude_baseline_failed` reason
Both gates key on the exact string:
- `report.py:198` — `if str(getattr(state, "stop_reason", "")) != "baseline_failed": return None`
- `cli.py` — `if session_dir is not None and stop_reason == "baseline_failed":`

But the same underlying condition (baseline_failure_streak ≥ 3 in PRELUDE) has **two** terminal markers:
- `coordinator.py:7047` sets `set_stop_reason("baseline_failed")` from the result handler.
- `phase_state.exit_terminal_prelude` routes PRELUDE→CLOSE with reason `prelude_baseline_failed`.

`set_stop_reason` overwrites `self.stop_reason` unconditionally (any in-vocab value wins), so the **persisted** terminal reason depends on the call ordering between those two. If a run finalizes as `prelude_baseline_failed`, both gates fall through → no `failure_summary` is built and the CLI prints no root cause, i.e. the fix is silently bypassed for (at least some of) the exact scenario #465 targets. The test suite only exercises `stop_reason="baseline_failed"` (+ one `time_exhausted`); `prelude_baseline_failed` is uncovered.

**Recommend:** match a small set rather than a literal, e.g. `_BASELINE_FAILED_REASONS = frozenset({"baseline_failed", "prelude_baseline_failed"})`, in both `report.py` and `cli.py`. If you've confirmed the persisted reason is always normalized to `baseline_failed` on every baseline-failure path, please say so and add a `prelude_baseline_failed` test to lock it.

### 2. (must confirm) The CLI half depends on `reports/final.json` already existing
`cli._read_failure_summary` reads `reports/final.json`, which is written by the ReportExecutor (CLOSE-phase sequencer). In `_run_optimize`'s `finally`, the **non-sequencer** branch (`close_sequence_done == False`) writes `final.md` via `write_minimal_final_report` but **not** `final.json`. So if a baseline-failure terminal path tears down without running the CLOSE sequencer (crash/early-abort), `final.json` won't exist and `_read_failure_summary` returns `None` → the CLI prints nothing even when the reason string matches.

**Recommend:** confirm the baseline-failed terminal path always runs the report (CLOSE sequencer) before `_print_final_summary`; otherwise have the safety-net branch also emit `final.json` (it already imports `_build_summary_dict`) so the CLI summary works on the abort path too.

### 3. (minor) `_BENIGN_FAILURE_PATTERNS` is a single hardcoded model-specific literal
`("modeling_cohere2.py",)` is brittle: the next benign upstream WARN needs a code edit, and a *real* fault whose only surviving line references that file would be suppressed (mitigated by the server.log fallback + the summary-only `_highlight_is_benign` check, both good). Consider an env/config-extensible list and an inline link to the upstream issue so the "why cohere2" context survives.

### 4. (nit) Source-of-truth coupling
`cli.py` re-reads `report.py`'s JSON output instead of rebuilding the summary — good (single source of truth), just note it's fully coupled to item 2.

### Positives (reviewed, on record)
- Strong, targeted unit coverage: benign-line partition keeps the real fault, `_classify_root_cause_type` enum, server.log fallback, latest-failed-attempt selection, `last_action_failures` fallback, and `_format_md` wiring. Tests drive the real persistence path via `record_action_attempt`/`record_action_failure`.
- Over-suppression is well-guarded: `_highlight_is_benign` judges the one-line summary only (payload-buried benign mentions don't suppress a real-fault headline), and `_partition_benign_lines` drops only matching lines.
- Fail-soft throughout (`_build_failure_summary` returns `None` on any error so the report still writes); `server_log` rendered session-relative; `json` already imported in `cli.py`; CI green.

### Net
Logic and tests are solid. The blocker is item 1 (reason-string coverage) and confirming item 2 (final.json presence on the failure path) — both determine whether the user-visible fix actually appears for #465's real terminal paths. Items 3–4 are follow-ups.

xiaofei-zheng (approved):

PR #556, opened 2026-06-12
Track GEAK main branch instead of pinned commit

### LINKED ISSUE #555: Use the latest GEAK for bug fixes
### Target release

None

### What kind of issue is this?

Task / Chore

### Related area

None

### Description

Point GEAK to the latest main for bug fixes and faster resolution.

sdubagun-amd:
## Summary
- Change `GEAK_REF` default in `kernel-agent/scripts/install.sh` from the PR #244 commit SHA (`ec61bdbd...`) to the `main` branch, so fresh installs pick up the latest upstream GEAK tip at install time.
- Update the surrounding comment to note the reproducibility trade-off (branch tracking is not reproducible; operators can still pin with `GEAK_REF=<tag|branch|sha>`).
- Update `kernel-agent/SKILL.md` to reflect the new default (`main`, was the stale `v3.2.0`).

The existing install logic already handles a branch ref: a non-SHA `GEAK_REF` falls to the `git clone --depth 1 --branch "$GEAK_REF"` path, so no clone-logic change is needed.

Closes #555

## Test plan
- [ ] `kernel-agent/tests/test_kernel_agent.py` still passes (asserts the `GEAK_REF="${GEAK_REF:-` override pattern, not the exact value)
- [ ] Run `kernel-agent/scripts/install.sh` and confirm GEAK clones at `main` HEAD

### DIFF
--- a/kernel-agent/SKILL.md
+++ b/kernel-agent/SKILL.md
@@ -78,7 +78,7 @@ bash "$REPO_ROOT/kernel-agent/scripts/install.sh"
   `TraceLens_generate_perf_report_pytorch_inference --help`
   (Hyperloom is inference-only since v0.4; the training-mode CLI is no
   longer accepted)
-- GEAK CLI from `GEAK_REF` (default `v3.2.0`) +
+- GEAK CLI from `GEAK_REF` (default `main`) +
   `${HYPERLOOM_RUNTIME_DIR}/geak-config/local.yaml` (model resolution:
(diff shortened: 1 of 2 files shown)


sdubagun-amd:
Closing in favor of #579, which pins GEAK to the v3.2.1 **release** tag rather than tracking the moving `main` branch. We're going with the reproducible release-pin approach instead of branch tracking. See #578 for the replacement issue.

PR #552, opened 2026-06-12
[feat] Explore optimization — warm rebench, measurement, phase flags, and launcher docs

ZhengGong-amd:
## Summary

This branch sharpens the EXPLORE pipeline's measurement fidelity and brings the
phase-disable flags (`--no-explore` / `--no-framework`) up to parity with
`--no-kernel`, both in the orchestration prompt and in PolicyGate enforcement.
It also trims the inference optimizer launcher skill so agents load only the
operational context they need.

The changes are grouped into three themes:

- **Explore measurement quality**: warm, baseline-comparable per-variant rebench
  via server reuse, plus a rough throughput estimate for variants reaped at the
  overtime deadline.
- **Phase-flag symmetry**: `--no-explore` / `--no-framework` now affect prompt
  visibility, SESSION CONTEXT, and the startup banner like `--no-kernel` does,
  and `--no-explore` is enforced fail-closed so the phase-interleave grey
  channel cannot reintroduce explore work into KERNEL.
- **Launcher skill maintainability**: `SKILL.md` is reduced to a thin launcher /
  monitor entrypoint, with detailed operational references and helper scripts
  under `inference_optimizer/launcher/`.

### 1. `a425acf9` feat(explore): estimate rough output tput for killed-overtime variants

A variant reaped at the soft overtime deadline never writes a
`benchmark_report`, so it previously surfaced no throughput at all. This parses
the inference engine's periodic `server.log` throughput logs (sglang
`gen throughput (token/s)` / vllm `Avg generation throughput`) and averages the
steady-state samples into a rough output tok/s estimate.

- Informational only: the variant stays `failed` / `killed_overtime`, and the
  estimate never feeds winner selection or gain math (`output_throughput` stays
  `None`).
- Surfaced via `VariantResult.estimated_output_throughput`, the explore
  `KILLED_OVERTIME` rows / `per_variant_outcomes`, `conc_sweep` serialization,
  and a `nonfatal_warnings` tag.
- New module `action_executors/benchmark_result.py`; tests in
  `test_benchmark_result.py` and `test_grid_runner_kill_and_branches_unit.py`.

### 2. `73e54459` feat(explore): warm-gated rebench via server reuse

Makes the per-variant validate a warm, baseline-comparable measurement:

- Round 1 runs cold and keeps the server alive (`cleanup=false`); only variants
  whose cold gain clears `keep_threshold` earn a round 2 that reuses the hot
  server (`cleanup=true`) for the headline throughput. This aligns the rebench
[...]

PR #558, opened 2026-06-13
feat(tracelens): deterministic (no-LLM) trace analysis route

xiaofei-zheng:
## Summary

Adds a fully deterministic, no-LLM trace-analysis route to the kernel agent
(`HYPERLOOM_TRACE_ANALYSIS_ROUTE=deterministic`) that reproduces the hot-kernel
set the LLM agent route extracts, but by reading TraceLens' structured
`priority_data.json` / `*_metrics.json` / `gpu_timeline.csv` directly — saving
LLM round-trips while keeping hot-kernel data (source file, GPU%, shapes,
efficiency) consistent with the agent route.

### Deterministic route
- New `--analysis-route deterministic|agent` switch; deterministic runs the
  TraceLens Python toolchain (perf report → orchestrator_prepare → category
  scripts → priority_data) and extracts hot kernels directly, with a fail-loud
  guard on a corrupt/failed pipeline.
- `deterministic_extract_hot_kernels`: priority-finding members joined to
  metrics rows by (name, time) + an "other"-bucket pass that recovers dominant
  ops (e.g. Triton fused_moe) the efficiency-model categories miss; sorted by
  GPU time.
- Source resolution shared with the agent route via `_finalize_candidates`
  (pybind/aiter/sgl launcher → device-source promotion).

### Correctness fixes in this branch
- Resolve "other"-bucket ops to their **definition site** (grep on the real op
  name) instead of the launcher wrapper — fixes `fused_moe` resolving to
  `fused_moe.py` instead of `fused_moe_triton_kernels.py`.
- Normalize **graph-captured / synthetic op names** (`hipGraphLaunch->void …`,
  `… .kd (Synthetic Op)`, embedded `_Z` mangled symbols) before keyword
  extraction; previously graph-mode traces yielded **zero** hot kernels.
- Log "other"-bucket ops skipped for an unresolvable source (vendor CK/Tensile
  `.so`-only kernels) so missing hot kernels are observable, not silent.
- Align `gpu_pct` denominator across both routes to whole-trace GPU time.
- Stack-validation KEEP decision now uses the incremental gain over
  `current_best` (not total over baseline), preventing zero/negative-value
  patch stacks from being kept.
- `phase_interleave` defaults OFF (env opt-in); `--no-explore` narrows KERNEL
  extras to `integrate_patch`.

### Human-readable analysis.md (deterministic, no LLM)
- Executive Summary + **System-Level Signals** (idle / exposed comm / memcpy,
  from `gpu_timeline.csv`).
- Top Hot Kernels + per-P-item tables now include **GPU%** (real impact share)
  and **Source File** (resolved editable file, distinct from the launcher
  Kernel Path).

## Test plan
- [x] `kernel-agent/tools/test_tracelens_csv.py` (187 passed) incl. new
  graph-name normalization, skip-logging, System-Level Signals, GPU%/source
  rendering tests
- [x] Orchestrator suites: `test_sweep_phase_auto.py`,
  `test_kernel_integrate_and_report.py`, `test_phase_state_machine.py`,
  `test_kernel_optimization_verification.py` (376 passed combined)
- [x] Validated on 46 real sessions: 0 crashes, 19 graph kernels recovered
- [x] 30B-A3B: `fused_moe → fused_moe_triton_kernels.py` resolves correctly


chaojhou (commented):
## Review — #558 `feat(tracelens): deterministic (no-LLM) trace analysis route`

Read the substantive diffs end-to-end: `tracelens_analysis.py` (+1000 deterministic route), `coordinator.py` (+486 kernel-integration overhaul), `_server_patcher.py`, `kernel_request_handlers.py`, `shared_state.py`, `phase_state.py`, plus a Bugbot pass. Strong engineering in each piece, but the PR's scope is much wider than its title and there are a few correctness items.

### 0. (process) Scope is far wider than the title
This is not just a deterministic trace route. It bundles ≥5 independent concerns:
1. Deterministic TraceLens route (`tracelens_analysis.py` +1000) — the stated feature.
2. A kernel-integration orchestration overhaul (`coordinator.py` +486: crash-safe stack validation, IR-3 auto-integrate, impact-first KEEP ordering, incremental-gain KEEP).
3. `phase_state.is_phase_interleave_enabled` default flipped **ON → OFF**.
4. `--no-explore` KERNEL grey-channel enforcement.
5. `_server_patcher` sentinel verification + GEAK pin.

Two of these collide with other in-flight PRs (see #3 below). Please confirm this bundling is intentional; splitting (2) and (3)/(4) out would make rollback and review far safer given the breadth.

### 1. (high — Bugbot, verified) `_apply_atomic` leaves a partially-patched tree on the new post-apply gate
`_server_patcher.py` ~L672-685: the new post-apply check
```python
if not _is_patched(plan):
    log.error(...); return False
```
runs **after** patches in `applied` were written to the live SGLang install, and returns `False` **without reverting them** — unlike the mid-transaction failure path, breaking the `_apply_atomic` atomicity contract. A fuzzy/partial apply that leaves a sentinel missing therefore reports failure while SGLang is mutated on disk; `ensure_sglang_patched_for_tracelens` then runs in unpatched-fallback mode against a tree that is neither clean nor fully patched. If the failure is a genuinely-incompatible patch (not self-healable on retry), the install stays corrupted for later runs. **Fix:** revert the `applied` set before `return False` here, exactly as the mid-apply failure path does.

### 2. (medium) `_auto_enqueue_pending_integrations` dedup set is not crash-safe
`coordinator._auto_integrate_dispatched` is an in-memory instance `set`, and the integrate REQUEST it emits has no idempotency key. On resume (or coordinator restart) before the integrate attempt is recorded in `kernel_integrate_attempts`, the same KEEP'd kernel can be auto-dispatched again → duplicate integrate work. Suggest persisting the dispatched set to `SharedState` or attaching an idempotency key to the bus REQUEST so the dedup survives a restart (the rest of this PR is careful about crash-safety — the stack-validation checkpoints are a good example).

### 3. (medium — cross-PR coordination) interleave default flip + duplicated `--no-explore` work
- `is_phase_interleave_enabled` flips from default-**ON** to default-**OFF**. PR #546 (long-run/cyclic) relies on phase interleave being default-on. Whoever merges second silently changes the other's behavior — please decide the single owner of this default and align.
- The `explore_enabled` parameter added to `llm_proposable_actions_for_with_interleave` / `is_action_llm_proposable_in_phase_with_interleave` is **identical** to the change in PR #552. These two PRs will merge-conflict on the same function; coordinate which one carries it.

### 4. (low-medium) `integrate_handler` `stack_positive_keep` 0.5% floor vs benchmark noise
A kernel integrate now KEEPs when `stack_incremental_gain_pct >= 0.5%` over `current_best` even if the absolute-vs-baseline gain is below `keep_threshold_pct`. 0.5% is within typical E2E benchmark run-to-run noise, so this can bank noise-driven increments. It's backstopped by the 1% combined stack re-validation (`_run_kernel_stack_validation_e2e`), which is good — just confirm the 0.5% floor sits above your measured single-run noise, or document that the stack-validation pass is the real gate.

### 5. (CodeQL) cleanups
- `kernel_request_handlers.py:35` `KERNEL_STACK_VALIDATION_KEEP_THRESHOLD_PCT` flagged unused — it **is** imported and used by `coordinator._run_kernel_stack_validation_e2e`, so not dead, but it's defined in a module that never uses it. Relocate to `coordinator` (or annotate) to clear the scanner. Not a bug.
- `tracelens_analysis.py` empty `except: pass` at L2886 / L2901 / L3196 — add a one-line comment or debug log to clear the alert.
- `test_sweep_phase_auto.py:362` redundant always-true comparison — fix the assertion (a real test bug; the case isn't actually exercising what it claims).

- **Deterministic route is well-built:** `_match_op_by_time` joins priority members to metrics rows with a 5 ms tolerance and returns `{}` past it (no silent mis-association of launcher paths/shapes); the "other"-bucket pass resolves to the **definition site** via grep (not the launcher wrapper) — the `fused_moe` → `fused_moe_triton_kernels.py` fix; `_normalize_profiler_op_name` strips graph-capture/synthetic/`_Z`-mangled wrappers (fixes zero-hot-kernels on graph traces); unresolvable hot ops are logged rather than silently dropped; and the pipeline **fails loud** (`_raise_on_failed_deterministic_pipeline`, `fail_on_corrupt_priority=True`) instead of returning partial hot kernels.
- **gpu_pct denominator aligned to whole-trace GPU time across BOTH routes** — a real consistency fix so a kernel's gpu_pct means its share of total GPU time, not of the top-k sum.
- **Stack validation is crash-safe:** in-progress guards + `pending_stack_validation_*` checkpoints + `_recover_interrupted_stack_validation` (reverts partial applies / resumes a pending result); incremental-gain decision over `current_best` (avoids banked gain carrying a zero-contribution stack); same-target-file dedup; reverts on non-KEEP and on exception.
- impact-first kernel ordering (gpu_pct then micro speedup) and excluding already-attempted kernels so a near-threshold NEEDS_REVIEW doesn't auto-rerun.
- Good new tests (`test_tracelens_csv` +644, `test_kernel_optimization_verification`, `test_kernel_integrate_and_report`, `test_sweep_phase_auto`); CI green.

### Net
The deterministic route and the kernel-integration crash-safety work are solid. Merge gates: **(1) the `_apply_atomic` rollback (high)** and **(3) cross-PR coordination on the interleave default + duplicated `--no-explore` edit**; (2) and (4) are robustness confirmations; (5) is scanner cleanup. Strongly consider splitting the orchestration overhaul out from the trace-analysis feature.

chaojhou (approved):

PR #563, opened 2026-06-15
feat(session-breakdown): record bypass discovery runs in kernel_journey

luochen-amd:
## Summary

Implements **B1** from the Kernel Overview design (`pulse docs/kernel-overview`): make the deterministic / **bypass** TraceLens route (`--analysis-route deterministic`, #558) show up in `session_breakdown.json`'s `kernel_journey.discovery_runs[]` with `source="bypass"`, so the dashboard's discovery-source comparison (tracelens vs bypass, view A2) finally has bypass rows.

### Root cause
The deterministic route from #558 already produces hot kernels through the **same** `trace_analyze` pipeline as the LLM route, and the orchestrator already records a stage-1 discovery run for every `trace_analyze`. But that recorder call was hard-coded to `source="tracelens"`, so bypass runs were mislabeled and the `bypass` series was always empty.

### Changes
- `orchestrator/kernel_request_handlers.py` (`trace_analyze_handler`): label the discovery run by the route that actually ran — prefer the tool's authoritative `orchestrator_mode` (`"deterministic"` → `"bypass"`), falling back to the requested `analysis_route`. Record the route into `scan.analysis_route` for provenance.
- `breakdown/recorder/instrument.py` (`record_kernel_discovery`): decouple the discovery **`source`** label (dashboard group-by) from the version-provenance **`tool`** key (new optional param, defaults to `source` for back-compat). The handler passes `tool="tracelens"` because bypass runs the same TraceLens toolchain — this avoids minting an empty `versions["bypass"]` entry.

Status / duration / hot-kernel field mapping are untouched, so bypass discovery rows are **structurally identical** to tracelens rows (design requirement) across all three states: success, fail-loud, and high-idle empty set.

### `discovery_runs[]` fields a bypass row carries (per `KernelDiscoveryRun`)
| field | value for bypass |
|---|---|
| `source` | `"bypass"` |
| `status` | tool status (`ok` on success / high-idle, `failed` on fail-loud) |
| `ts` | run timestamp |
| `duration_sec` | wall time of the trace_analyze subprocess |
| `scan` | `{splitter_mode, trace_dir, candidates_path, trace_report_path, analysis_route="bypass"}` |
| `hot_kernel_count` | len(hot_kernels) (0 on high-idle / failure) |
| `hot_kernels[]` | `DiscoveredHotKernel` rows (name, gpu_pct, time_ms, bound_type, efficiency_percent, source_file, recommended_backends, ...) |
| `error` | failure text on `status="failed"`, else null |

Tool version provenance lands in top-level `versions["tracelens"]` (not `versions["bypass"]`).

## Test plan
- [x] `inference_optimizer/tests/test_kernel_journey.py` — recorder-level: `source`/`tool` decoupling keeps version provenance under `tracelens` (no `versions["bypass"]`); `tool` defaults to `source` for back-compat. (Runs locally; 14 passed.)
- [x] `inference_optimizer/tests/test_profile_and_kernel_handlers.py` — handler-level: bypass success / fail-loud / high-idle-empty all record `source="bypass"`; agent route stays `source="tracelens"`. (Linux/CI — module imports `fcntl`, not collectable on Windows dev box.)

### DIFF
--- a/inference_optimizer/breakdown/recorder/instrument.py
+++ b/inference_optimizer/breakdown/recorder/instrument.py
@@ -572,6 +572,7 @@ def record_kernel_discovery(
     status: str,
     hot_kernels: list[Any] | None = None,
     scan: dict[str, Any] | None = None,
+    tool: str | None = None,
     tool_root: str | None = None,
     tool_root_env: str | None = None,
     tool_version: str | None = None,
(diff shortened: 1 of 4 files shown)


ZhengGong-amd (approved):
validated, Approved

PR #581, opened 2026-06-16
fix(inference-optimizer): preflight model gates + Gemma2 roofline shape-discovery workaround

lishuoshuo-amd:
Preflight (cli.py): fail fast on statically-detectable incompatible configs so they stop before the heavy server bring-up instead of crashing in engine init / baseline:
- unrecognized model types/arches now include deepseek_v4 and nested text_config.model_type=ministral3 (vLLM registry KeyError).
- new vocab-shape check: config.vocab_size vs safetensors embed/lm_head dim mismatch (sglang weight-load AssertionError).

Gemma2 roofline (_workload_envs.py): skip
--enable-shape-discovery-for-cuda-graph-profile for Gemma2. The TraceLens kernel_shape_profiler activates inside the CUDA-graph capture critical section; Gemma2 forward builds a host torch.tensor(normalizer) there, which raises hipErrorStreamCaptureUnsupported and aborts capture (roofline yields no ceiling, e2e gain stays 0). Verified locally: with shape-discovery the capture crashes on the 2nd bs; without it Gemma2 captures all 52 bs. CUDA graph + profiling are kept.

Adds regression tests for both.

### DIFF
--- a/inference_optimizer/cli.py
+++ b/inference_optimizer/cli.py
@@ -48,6 +48,7 @@ from .cli_backends import (  # noqa: F401 - re-exported for callers/tests
 from .orchestrator.backends import ClaudeBackend
 from .manifest import load_manifest, write_manifest
 from .model_config_utils import (  # noqa: F401 - re-exported for callers/tests
+    GEMMA2_ARCHITECTURES as _GEMMA2_ARCHITECTURES,
     _config_architectures,
     _load_model_config_dict,
     _model_is_gemma2,
@@ -1188,14 +1189,19 @@ _UNREGISTERED_CUSTOM_CONFIG_TYPES = frozenset({"kimi_k2"})
 # type mimo_v2_flash to instantiate a model of type ." + Unknown attention
 # backend TRITON) — the unrecognized arch leaves an empty model type.
 # deepseek_v4: confirmed from DeepSeek-V4-Flash server.log ModelConfig
-# validation failure. ministral3: vLLM registry raises KeyError('ministral3').
+# validation failure.
 _UNRECOGNIZED_MODEL_TYPES = frozenset({
-    "deepseek_v4", "glm4_moe_lite", "mimo_v2_flash", "ministral3",
+    "deepseek_v4", "glm4_moe_lite", "mimo_v2_flash",
 })
 _UNRECOGNIZED_ARCHITECTURES = frozenset({
     "deepseekv4forcausallm", "glm4moeliteforcausallm",
     "mimov2flashforcausallm",
 })
+# ministral3 only fails inside a Mistral3 multimodal wrapper (Surpem-Supertron2
+# server.log: vLLM registry raises KeyError('ministral3') for text_config.
+# model_type). A bare top-level ministral3 is left to the framework to judge, so
+# this is checked only against the nested text_config scope.
+_NESTED_ONLY_UNRECOGNIZED_MODEL_TYPES = frozenset({"ministral3"})
 
 # Phi-3 su/longrope rope_scaling: Phi3Config._rope_scaling_validation() requires
 # rope_scaling to be a dict of EXACTLY 3 keys (type/short_factor/long_factor).
@@ -1208,8 +1214,8 @@ _PHI3_ROPE_TYPES = frozenset({"su", "longrope"})
 # Gemma2 missing hidden_act: sglang's gemma2 runtime reads config.hidden_act
 # unconditionally. Some checkpoints ship only hidden_activation (the HF field
 # name), so config.hidden_act is absent and the model crashes with
-# AttributeError in engine init. Hardware-agnostic.
-_GEMMA2_ARCHITECTURES = frozenset({"gemma2forcausallm"})
+# AttributeError in engine init. Hardware-agnostic. ``_GEMMA2_ARCHITECTURES`` is
+# imported from model_config_utils (single source of truth) at module top.
 
 # Quantization formats with no ROCm/AMD runtime path. NVIDIA ModelOpt FP8/NVFP4
 # use vendor-specific scale packing (no sglang ROCm loader); bitsandbytes ships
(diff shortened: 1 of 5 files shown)


chaojhou (commented):
## Review — #581 `fix(inference-optimizer): preflight model gates + Gemma2 roofline shape-discovery workaround`

Read both code changes (`cli.py` preflight gates, `_workload_envs.py` Gemma2 skip) end-to-end. Clean, well-scoped, and the safetensors header parsing is done carefully. One correctness question worth confirming, plus two minor notes.

### 1. (confirm) `_detect_vocab_weight_shape_mismatch` strict `!=` may false-positive on padded vocab
The new gate flags whenever the checkpoint's vocab dim `actual != expected (config.vocab_size)` and **blocks the run before server bring-up**. That's the right catch for the confirmed sglang AssertionError, but the exact-inequality is risky in one direction: many HF checkpoints **pad the embedding / `lm_head` rows up to a rounding multiple** (e.g. config `vocab_size=32000`, `lm_head.weight` shape `[32064, …]`), and sglang/vLLM's `VocabParallelEmbedding` is designed to pad `org_vocab_size`, so those load fine. If sglang tolerates `actual > expected` (padded checkpoint), this strict `!=` will wrongly fail-fast a loadable model.

Please confirm the sglang weight-load semantics for the targeted versions:
- If it requires **exact** equality against `org_vocab_size`, the current `!=` is correct — keep it.
- If it tolerates padding (`actual >= expected`), gate only on `actual < expected` (genuinely too few rows) or allow a small pad tolerance, so padded-vocab models aren't blocked.

Given this is a hard fail-fast gate, a false positive is more costly than a missed catch (the latter just falls back to the old engine-init failure).

### 2. (minor, structural) executor → `cli` import coupling
`_workload_envs.materialize_config_with_envs` now does `from ...cli import _model_is_gemma2`. Pulling the CLI module into a low-level action executor risks import cycles and drags CLI-only deps into the hot path. `_model_is_gemma2` / `_load_model_config_dict` / `_config_architectures` are generic model-config helpers — consider relocating them to a shared util module (e.g. a `model_config` helper) that both `cli` and the executors import, rather than importing *from* `cli`. Not blocking, but it's the kind of coupling that bites later.

### 3. (nit, CodeQL) redundant `import json`
`test_profile_and_kernel_handlers.py:678,700` re-import `json` already imported at module top (line 8). Drop the local re-imports to clear the scanner.

- `_read_safetensors_header` is solid: reads only the 8-byte little-endian length + JSON header, bounds it (`> 64 MB → reject`), never materializes tensor data, and catches `OSError / JSONDecodeError / ValueError / struct.error`. Correct safetensors format handling.
- `_detect_unrecognized_architecture` now also scans the nested `text_config` scope (the `ministral3` nested-config case) — good; `shape[0]` is the right vocab/output dimension for `embed_tokens/wte/word_embeddings/lm_head`.
- The Gemma2 workaround is pragmatic and correctly scoped: it disables only shape-discovery (keeping CUDA graph + profiling) and is verified locally (capture crashes on 2nd bs with it, captures all 52 bs without). Detection handles top-level and nested `text_config`.
- Both fixes ship regression tests; CI green.

### Net
Good fix. The only thing I'd resolve before merge is item 1 — confirm sglang's vocab-equality semantics so the new gate doesn't false-positive on padded-vocab checkpoints; items 2–3 are follow-ups.

ZhengGong-amd (approved):
Approved, minor fix shall be applied before merge.

lishuoshuo-amd:
> ## Review — #581 `fix(inference-optimizer): preflight model gates + Gemma2 roofline shape-discovery workaround`
> Read both code changes (`cli.py` preflight gates, `_workload_envs.py` Gemma2 skip) end-to-end. Clean, well-scoped, and the safetensors header parsing is done carefully. One correctness question worth confirming, plus two minor notes.
> 
> ### 1. (confirm) `_detect_vocab_weight_shape_mismatch` strict `!=` may false-positive on padded vocab
> The new gate flags whenever the checkpoint's vocab dim `actual != expected (config.vocab_size)` and **blocks the run before server bring-up**. That's the right catch for the confirmed sglang AssertionError, but the exact-inequality is risky in one direction: many HF checkpoints **pad the embedding / `lm_head` rows up to a rounding multiple** (e.g. config `vocab_size=32000`, `lm_head.weight` shape `[32064, …]`), and sglang/vLLM's `VocabParallelEmbedding` is designed to pad `org_vocab_size`, so those load fine. If sglang tolerates `actual > expected` (padded checkpoint), this strict `!=` will wrongly fail-fast a loadable model.
> 
> Please confirm the sglang weight-load semantics for the targeted versions:
> 
> * If it requires **exact** equality against `org_vocab_size`, the current `!=` is correct — keep it.
> * If it tolerates padding (`actual >= expected`), gate only on `actual < expected` (genuinely too few rows) or allow a small pad tolerance, so padded-vocab models aren't blocked.
> 
> Given this is a hard fail-fast gate, a false positive is more costly than a missed catch (the latter just falls back to the old engine-init failure).
> 
> ### 2. (minor, structural) executor → `cli` import coupling
> `_workload_envs.materialize_config_with_envs` now does `from ...cli import _model_is_gemma2`. Pulling the CLI module into a low-level action executor risks import cycles and drags CLI-only deps into the hot path. `_model_is_gemma2` / `_load_model_config_dict` / `_config_architectures` are generic model-config helpers — consider relocating them to a shared util module (e.g. a `model_config` helper) that both `cli` and the executors import, rather than importing _from_ `cli`. Not blocking, but it's the kind of coupling that bites later.
> 
> ### 3. (nit, CodeQL) redundant `import json`
> `test_profile_and_kernel_handlers.py:678,700` re-import `json` already imported at module top (line 8). Drop the local re-imports to clear the scanner.
> 
> ### Verified correct / positives
> * `_read_safetensors_header` is solid: reads only the 8-byte little-endian length + JSON header, bounds it (`> 64 MB → reject`), never materializes tensor data, and catches `OSError / JSONDecodeError / ValueError / struct.error`. Correct safetensors format handling.
> * `_detect_unrecognized_architecture` now also scans the nested `text_config` scope (the `ministral3` nested-config case) — good; `shape[0]` is the right vocab/output dimension for `embed_tokens/wte/word_embeddings/lm_head`.
> * The Gemma2 workaround is pragmatic and correctly scoped: it disables only shape-discovery (keeping CUDA graph + profiling) and is verified locally (capture crashes on 2nd bs with it, captures all 52 bs without). Detection handles top-level and nested `text_config`.
> * Both fixes ship regression tests; CI green.
> 
> ### Net
> Good fix. The only thing I'd resolve before merge is item 1 — confirm sglang's vocab-equality semantics so the new gate doesn't false-positive on padded-vocab checkpoints; items 2–3 are follow-ups.

Thanks for the review!

**1. vocab strict `!=` — confirmed correct, keeping it.**

Checked sglang source (`srt/layers/vocab_parallel_embedding.py`): `weight_loader` enforces **strict equality** on the normal path:

```python
assert loaded_weight.shape[output_dim] == self.org_vocab_size  # non-presharded
```

`org_vocab_size == config.vocab_size` (`VocabParallelEmbedding(config.vocab_size, ...)`). sglang pads its *internal buffer* (`fill_(0)`) but requires the loaded weight to equal the unpadded `vocab_size`. So `config=32000` + `lm_head=32064` actually fails in sglang too — our `!=` matches its `==` exactly, no false-positive on loadable checkpoints. (Packed/presharded edges are on the non-vocab dim / a rare flag, not relevant here.)

**2. executor → cli coupling — already done.** Helpers moved to a stdlib-only leaf module `model_config_utils.py`; both `cli` and executors import from it (no `from ...cli import`).

**3. redundant `import json` — already done.** Dropped in a follow-up commit.

Items 2 & 3 are in commits past the version you reviewed.

PR #585, opened 2026-06-16
[STEPPING]  bug fixes, small optimizations, and orchestrator/CLI decoupling

ZhengGong-amd:
This branch bundles a batch of small bug fixes and minor optimizations
discovered while running long optimization sessions, plus a behavior-preserving
decomposition of two God-objects (`Coordinator`, `SharedState`) and the CLI
monolith. Every change ahead of `main` is listed below; nothing is omitted.

**Scope:** 27 logical changes ahead of `main`. The full `inference_optimizer`
suite is green (the only local failures are environment-specific: a dev pod with
multi-node env vars set, and a local untracked `.env`).

> Note: the earlier explore-measurement / `--no-explore` / `--no-framework` /
> launcher-docs work (previously described in this file) has been merged into
> `main` and is intentionally excluded here.

---

## Bug fixes

1. **`fix(integrate_patch)`: target-aware framework-root resolution** (`ff135c2e`)
   Pick the framework source root whose tree actually contains the patch's
   modify/delete targets instead of blindly using the first allowlist entry, so
   a `vllm/...` patch no longer fails by being applied under the `aiter` root.

2. **`fix(baseline)`: resolve the live session dir for the reference recipe** (`851534ce`)
   The baseline executor is a module-level singleton instantiated at import time
   (before `$INFERENCE_OPTIMIZER_CURRENT_SESSION_DIR` is pinned), so its cached
   `session_dir` could resolve to the workspace root instead of the live session
   dir whose `state.json` carries the `reference_*` fields. Prefer the live pin
   when present; otherwise honor the caller-supplied path.

3. **`fix(launcher)`: point `PID_FILE` at the real optimizer, harden monitor resume** (`f83385ad`)
   Two launcher-template defects made the robustness monitor misfire a spurious
   `--resume` seconds after a healthy launch:
   - `operations.md` wrote `echo $! > PID_FILE`, but under `setsid nohup ... &`
     that `$!` is the wrapper PID, which exits immediately. The health-check now
     reconciles `PID_FILE` to the authoritative `.pid` from launch-info (with a
     `pgrep` fallback), and reads JSON via `python3` since `jq` is not present on
     every node.
   - `robustness_monitor.sh.example` hardcoded the resume-log dir to
     `$REPO_ROOT/optimizer_runs`, which may not exist when the run dir lives
     under `$USER_DATA_PATH`; the failed redirect overwrote `PID_FILE` with a
     dead wrapper PID. The resume-log dir is now derived from
     `dirname "$PID_FILE"` (with `mkdir -p`), and the resumed optimizer's real
     PID is captured via `pgrep`.

4. **`fix(robustness-agent)`: read `coordinator.db`, not stale `conductor.db`** (`f0da9ca7`)
   The local fallback probe and session-discovery markers pointed at
   `storage/conductor.db` (the pre-rename component name), but the Coordinator
   writes `storage/coordinator.db`. The probe therefore found no DB / no session
   and silently returned no coordinator events, so robustness signals that
   depend on the event stream went blind. Renamed `conductor_db_path` →
   `coordinator_db_path` and converted every `conductor.db` reference (config,
   local probe, docstring, `SKILL.md`) plus the test fixtures.

5. **`fix(integrate_patch)`: contain patch paths and commit deletion-only KEEPs** (`af32ba93`)
   - *Security:* `_resolve_patch_paths` accepted any existing absolute path in
     the LLM-/specialist-controllable `params.patches` without checking it lived
     inside the specialist workspace — a crafted path could read and `git apply`
     an arbitrary file against the framework tree. Now drops any path that
     resolves outside the workspace/worktree (both sides `resolve()`-d so a
     symlinked workspace still matches).
   - *Durability:* `_patch_touched_paths` only emitted paths that exist
     post-apply, so a pure-deletion KEEP emitted nothing, `_git_commit_kept`
     short-circuited, and a later cycle's REVERT (`git checkout -- .`)
     resurrected the deleted file. Now emits the old path for deletion headers
     so `git add -A` stages the removal, and drops headers matching neither pre
     nor post so `git add` cannot error.

6. **`fix(db_maintenance)`: never prune a still-pending proposal event** (`349398d8`)
   `prune_events` deleted any event with `seq <= delete_below` once all cursors
   advanced past it — content-blind. A proposal event could satisfy that while
   still semantically pending (no `review_verdict` targets it yet); pruning its
   row removed the only durable record, so a late critic verdict arriving after
   resume had nothing to attach to (the proposal was silently lost). Added a
   shared `pending_proposal_seqs()` helper that mirrors the decided-id join in
   `Coordinator.replay_for_resume` and excludes those seqs from the DELETE; the
   inner SELECT filters NULL/empty targets to avoid the `NOT IN (NULL)` trap. A
   cross-check test guards against drift from `replay_for_resume`.

7. **`fix(message_bus)`: allow-list the `kill` topic for `KILL_TASK` broadcasts** (`0d72da8e`)
   The Robustness `KILL_TASK` handler broadcasts on topic `"kill"`, but `"kill"`
   was missing from `TOPIC_ALLOWLIST`, so `append_and_seq` raised
   `ValueError("unknown topic")` and crashed the kill path at runtime. Added
   `"kill"` to the allow-list and guarded against recurrence with a static test
   asserting every literal topic emitted via `Message.new(...)` in
   `coordinator.py` is allow-listed.

8. **`fix(baseline)`: reap orphaned aiter JIT locks before cold start** (`6ef21787`)
   A killed `hipcc` compile leaves its aiter JIT file lock behind, and aiter's
   `FileBaton.wait()` spins forever with no timeout, so every later compile of
   that module hangs (build_count frozen, server unreachable, robustness
   escalated). The existing sweep ran only once at process startup, missing
   locks orphaned mid-session. Now sweeps stale locks before each cold server
   start, gated on "no live compiler process" so a concurrent benchmark's
   in-flight compile on the node-global jit dir is never disturbed (only locks
   are removed; ninja resumes incrementally from surviving `.o`). Shared logic
   consolidated into a new `_aiter_jit` module to avoid a circular import;
   `cli.py` keeps a thin delegating shim with its original signature.

9. **`fix(explore/integrate_patch)`: compact JSON server args + seed patched-eval base config** (`e440b20c`)
   Two defects surfaced on MiniMax-M3-MXFP8 (vllm/mi300x) runs:
   - *JSON args un-bootable:* Magpie expands `vllm serve ... $EXTRA_VLLM_ARGS`
     **unquoted**, so a JSON value carrying the conventional separator space
     (e.g. `--compilation-config '{"full_cuda_graph": true}'`,
     `--speculative-config '{"method": "eagle"}'`) is word-split by the shell and
     the server dies at boot — making spec-decode/MTP and compilation-config
     tuning impossible to evaluate. Added `compact_json_server_args()`
     (brace-matched scan, re-serialises each JSON blob with compact separators;
     string-internal spaces preserved) wired into the single
     `materialize_config_with_envs` choke point right after
     `dedup_vllm_server_args` (vllm/atom only; no-op for sglang / no-JSON /
     malformed). 8 regression tests.
   - *integrate_patch eval on bare defaults:* the `_materialize_approved_proposal`
     integrate_patch branch set only `keep_threshold_pct`, so the patched-eval
     server got empty base args, fell back to `default_baseline_config()` (e.g.
     vLLM block_size=16) and crashed at startup regardless of the patch → every
     integrate_patch falsely REVERTed with "no measurable throughput". Now seeds
     `base_tput` / `base_extra_args` (current best) and `config_path` (baseline
     config carrying the M3 serve flags), mirroring the sweep/explore branches.

10. **`fix(profile)`: preserve embedded JSON quotes in server-arg sanitization** (`44948f55`)
    `_sanitize_profile_server_args` tokenized with POSIX `shlex`, which stripped
    the inner double-quotes of JSON values like
    `--speculative-config {"method":"deepseek_mtp",...}` (yielding an unparseable
    `{method:...}`) — so every profile/roofline server boot failed and the kernel
    phase was starved of a fresh trace shape. Now tokenizes with `posix=False`
    and re-joins with spaces so embedded JSON survives verbatim while unsafe
    flags are still dropped. Regression test added.

---

## Small optimizations & minor features

11. **`feat(reference-script)`: seed baseline server args from a reference recipe** (`f9b7acc7`)
    Add `--reference-script` so an operator (or auto-discovery) can supply an
    InferenceX single-node launch recipe whose static serve flags + whitelisted
    exports seed the baseline server args at lowest priority (EXPLORE still
    overrides). Fixes exotic-arch models (MiniMax-M3 MSA, etc.) dead-ending at
    `baseline_failed` because the generic Magpie `{framework}_{gpu}.sh` omits
    the flags they need and no code path injected them.
    - `reference_script.py`: pair-aware parser (drops `$`-valued flags with
      their value, no orphans), enumerated env whitelist, renderer, tiered
      fail-soft discovery, and a shared model-gate (`models_compatible`) reused
      by the baseline executor so discovery and gating never drift.
    - `SharedState`: `reference_*` fact-layer fields (persist + restore on
      resume); `save()` re-renders a read-only `current_setting.sh` from
      `current_best`, re-feedable via `--reference-script`.
    - `materialize_config_with_envs`: reference base as the lowest-priority
      layer, below the YAML base and per-task `extra_server_args` (empty =
      byte-identical to before, zero degrade).
    - Baseline executor reads the model-gated reference from `SharedState` at
      the single choke point every run funnels through (initial / restart /
      rebaseline), including the multi-node per-round restart path.

12. **`feat(cli)`: enhance reference-recipe resolution logic** (`7dbd2014`)
    `_resolve_reference_recipe` now explicitly prevents auto-discovery when
    `--reference-script` is omitted (run behaves identically to before), and
    adds fallback logic to auto-discover an exact match if the provided script
    is unreadable or yields no flags.

13. **`feat(coordinator)`: aggregate research evidence into the session seen-set** (`a49de84a`)
    Add `_aggregate_research_evidence` to de-dupe PR ids, diffs and NVIDIA
    references across the session so FRAMEWORK_PR / explore lanes avoid
    re-fetching the same references. Fail-soft on partial payloads.

14. **`feat(coordinator)`: wire research-evidence aggregation into the specialist-done hook** (`9291d20d`)
    Call `_aggregate_research_evidence` from the per-domain bookkeeping path so
    any domain reporting a research block (e.g. `pr_intel`) feeds its PR ids,
    diffs and NVIDIA refs into the session-wide seen-set, not just
    `research_scout`. Idempotent and fail-soft.

---

## Refactors (behavior-preserving)

A staged decomposition of two God-objects and the CLI monolith. Every phase was
gated by a behavioral golden (added in phase 0) kept byte-identical, and all
re-exports / call sites were preserved so external imports and `cli._foo(...)` /
`coord._handle_*(...)` test sites are unchanged.

15. **`test(refactor)`: Coordinator behavioral golden safety net (phase 0)** (`be7583f0`)
    Locks down observable Coordinator behavior (persisted `SharedState` + bus
    decision/result topics) under a fixed propose → approve → delegate intent
    sequence — the regression gate for the splits below. Verified deterministic
    and that it trips on injected behavior drift.

16. **`refactor(orchestrator)`: extract `IntentRouter` collaborator (phase 1A)** (`c124933b`)
    Move the 15 intent handlers verbatim out of `Coordinator` into a new
    `IntentRouter` (~890 lines); `coordinator.py` drops 741 lines. Back-reference
    + `__getattr__` delegation; intra-router dispatch routes through
    `self._coord._handle_*` so coordinator-level monkeypatches still take effect.

17. **`refactor(orchestrator)`: extract `ResultRecorder` collaborator (phase 1B)** (`02e0a955`)
    Move the 12 result-recording / fact-synthesis methods verbatim into a new
    `ResultRecorder` (~1,077 lines); `coordinator.py` drops another 918 lines
    (10,535 → 8,876 across 1A+1B). Same transitional pattern; one `@staticmethod`
    keeps a forwarding shim.

18. **`refactor(orchestrator)`: offload `SharedState` behavior to write-owner modules (phase 2)** (`895ae075`)
    Move 21 decision-domain behavior methods out of the 3,753-line `SharedState`
    data record into write-owner modules as free functions (`shared_state_kernel`
    × 16, `shared_state_policy` × 3, `shared_state_phase` × 2). AST-verified
    behavior-preserving; `SharedState` keeps forwarding shims, so all
    `state.record_*(...)` call sites are unchanged. `shared_state.py` drops 360
    lines.

19. **`refactor(cli)`: split model-gate / preflight / gpu / bootstrap out of cli.py (phase 4)** (`0eaa5dc2`)
    Break the 6,201-line `cli.py` monolith into `cli_gpu.py`,
    `cli_model_gate.py`, `cli_preflight.py`, and `cli_bootstrap.py` (functions
    AST-extracted verbatim, constants assigned to their owning module). Clean
    one-way DAG: `cli_gpu ← cli_model_gate ← cli_preflight`, `cli_bootstrap →
    cli_model_gate`. `cli.py` drops 1,389 lines; all moved names re-exported.

20. **`refactor(orchestrator)`: fold `shared_state_policy` back into `policy.py` (phase 6A)** (`0e883fc2`)
    The phase-2 policy satellite was reached only through `SharedState` shims;
    its real decision owner is `PolicyGate`, so move the functions into
    `policy.py` and repoint the shims.

21. **`refactor(orchestrator)`: fold `shared_state_phase` back into `phase_state.py` (phase 6B)** (`bef28355`)
    `record_phase_transition` / `record_lifecycle_event` build their rows via
    `phase_state`'s helpers, so move them into `phase_state.py` (their decision
    domain) and repoint the shims.

22. **`refactor(orchestrator)`: fold `shared_state_kernel` back into `kernel_request_handlers` (phase 6C)** (`37b6ee5b`)
    Move the 16 kernel-decision write-owner functions into
    `kernel_request_handlers.py` (their decision domain) and repoint the shims;
    `shared_state` constants imported lazily to keep the edge one-way and
    cycle-free.

23. **`refactor(cli)`: fold `cli_gpu` + `cli_preflight` back into `cli_model_gate` (phase 6D)** (`14c68331`)
    `cli_gpu` and `cli_preflight` answer the same "can this model run on this
    hardware?" question `cli_model_gate` already owns, so fold both in and
    collapse `cli.py`'s three re-export blocks into one. Two test fixtures
    repoint their `_autodetect_gpu_type` monkeypatch from `cli_gpu` to the real
    call site in `cli_model_gate`.

---

## Documentation

24. **`docs(orchestration)`: clarify `skip_to_close` usage in phase transitions** (`7a6fb52e`)
    Emphasize that `skip_to_close` must not be emitted after a normal SWEEP
    completion (it mislabels the run as `robustness_escalated`); reserve it for
    genuine early-abandonment, and reflect the exception in the prompt builder.

25. **`docs(policy)`: clarify freeform specialist skips `max_turns` by design** (`82dabf99`)
    Expand the comment explaining that the `scope='freeform'` path intentionally
    omits the domain-anchored `max_turns` gate (bounded by task timeout instead)
    so a future reader does not "fix" it by re-adding a turn cap.

26. **`docs(shared_state)`: correct stale `crash_count` attribution comment** (`9b0c67e1`)
    `crash_count` is incremented by the Coordinator (not Robustness, which only
    reads it); fix the misleading attribute comment.

27. **`docs`: delete stale `reference_script.MD`** (`afb28c6c`)
    Remove the obsolete 407-line design doc superseded by the shipped
    `reference_script.py` implementation.

---

## New files

- `inference_optimizer/reference_script.py` — reference-recipe parser / renderer / discovery / model-gate.
- `inference_optimizer/cli_model_gate.py`, `cli_bootstrap.py` — extracted CLI modules.
- `inference_optimizer/orchestrator/intent_router.py`, `result_recorder.py` — extracted Coordinator collaborators.
- `inference_optimizer/orchestrator/action_executors/_aiter_jit.py` — shared aiter JIT lock-sweep helpers.
- Tests: `test_reference_script.py`, `test_aiter_jit.py`, `test_kill_topic_allowlist.py`,
  `test_refactor_golden_breakdown.py` + golden `tests/golden/coordinator_behavior_golden.json`.

---

## Test status

- Full `inference_optimizer` suite passes (5,287+ tests). The behavioral golden
  is kept byte-identical across every refactor phase (only the merge with
  `origin/main` added the new `baseline_eager_fallback` SharedState field).
- The only local failures are environment-specific and reproduce identically on
  the pre-branch code: a dev pod that exports multi-node env vars
  (`MASTER_ADDR` / `WORLD_SIZE`), and a local untracked `.env` that exports
  `TRACELENS_ROOT` (affecting one kernel-agent install-script test).


chaojhou (commented):
## Review — #585 `[STEPPING] bug fixes, small optimizations, and orchestrator/CLI decoupling`

+8.6k/−5.1k, 53 files. Mostly an **extraction refactor** (Coordinator −1938 → `intent_router.py` +917, `result_recorder.py` +1175, `kernel_request_handlers.py` +848; cli −1902 → `cli_bootstrap.py` +543, `cli_model_gate.py` +1382; shared_state −894 to write-owner modules) plus a handful of real bug-fixes/features. CI all green (incl. CodeQL), rebased on post-#596 main (mergeable; BLOCKED = review-required only).

### Verified / positives
- **Right way to do a big extraction:** the PR lands a *behavioral golden safety net first* (`coordinator_behavior_golden.json` + `test_refactor_golden_breakdown.py`, `test_reference_script.py`, etc.) and the full suite is green — that's the correct guardrail for a move-heavy refactor, and the strongest evidence the extraction is behavior-preserving.
- **`_aiter_jit.py` stale-lock reaper is carefully gated:** it deletes only lock files (`lock` / `.ninja_lock` / `lock_*`), never build artifacts; and it only does the unconditional `stale_minutes=0` sweep when `_any_live_compiler()` is **proven False**, falls back to a 5-min mtime gate when liveness is **unknown** (psutil missing), and **skips entirely** when a compiler is alive. The node-global shared-jit-dir concurrency hazard is explicitly handled, and it never raises. Solid fix for the "untimed FileBaton wait hangs the server" class.
- **`message_bus` kill-topic:** adding `"kill"` to `TOPIC_ALLOWLIST` is necessary (otherwise `append_and_seq` rejects the KILL_TASK audit broadcast as unknown-topic) and is covered by `test_kill_topic_allowlist`. The rest of that diff is pure literal reformat.

### Notes / asks
1. **(cleanup, lint becoming enforcing) leftover dangling imports from the split.** CodeQL/quality flags unused imports in `cli.py` (`_AMD_GPU_TYPES`, `_AMD_UNSUPPORTED_*`, `_default_target_summary`, `_parse_conc_sweep_concs`, `_print_kernel_*`, `ClaudeBackend`), `cli_model_gate.py` (`subprocess`), `intent_router.py` (`NoIntentEmitted`). If the `cli.py` ones are intentional **back-compat re-exports** after the module split, add them to `__all__` (or `# noqa: F401`) so they read as deliberate; otherwise drop them. Same for the empty-`except` at `cli_model_gate.py:112` and `kernel_request_handlers.py:3812/3822` — annotate the intentional non-fatal handlers. Worth clearing now since lint.yml is on track to become enforcing.
2. **(minor) `_any_live_compiler` is broadly conservative on shared nodes.** `COMPILER_PROCESS_NAMES` includes generic `clang`/`clang++`/`cc1plus`, so any unrelated C++ build on the node reads as "compiler alive" → the proven-dead fast path won't fire and it falls back to the 5-min mtime gate. That's safe (errs toward not-deleting a live lock), just less effective on busy shared nodes — acceptable, noting it.
3. **(scope) refactor + 2 new features (aiter-lock reap, reference-recipe seeding) + many bug fixes bundled.** Large to review as one unit, but the granular commit history + golden net make it traceable; not a blocker.

### Net
LGTM in approach and the substantive new behavior I checked (`_aiter_jit`, kill-topic) is sound, backed by a golden safety net and green CI. Only cleanups: clear the unused-import / empty-except lint findings (mark intentional re-exports in `__all__`) before lint goes enforcing. No blockers.

ZhengGong-amd:
> ## Review — #585 `[STEPPING] bug fixes, small optimizations, and orchestrator/CLI decoupling`
> +8.6k/−5.1k, 53 files. Mostly an **extraction refactor** (Coordinator −1938 → `intent_router.py` +917, `result_recorder.py` +1175, `kernel_request_handlers.py` +848; cli −1902 → `cli_bootstrap.py` +543, `cli_model_gate.py` +1382; shared_state −894 to write-owner modules) plus a handful of real bug-fixes/features. CI all green (incl. CodeQL), rebased on post-#596 main (mergeable; BLOCKED = review-required only).
> 
> ### Verified / positives
> * **Right way to do a big extraction:** the PR lands a _behavioral golden safety net first_ (`coordinator_behavior_golden.json` + `test_refactor_golden_breakdown.py`, `test_reference_script.py`, etc.) and the full suite is green — that's the correct guardrail for a move-heavy refactor, and the strongest evidence the extraction is behavior-preserving.
> * **`_aiter_jit.py` stale-lock reaper is carefully gated:** it deletes only lock files (`lock` / `.ninja_lock` / `lock_*`), never build artifacts; and it only does the unconditional `stale_minutes=0` sweep when `_any_live_compiler()` is **proven False**, falls back to a 5-min mtime gate when liveness is **unknown** (psutil missing), and **skips entirely** when a compiler is alive. The node-global shared-jit-dir concurrency hazard is explicitly handled, and it never raises. Solid fix for the "untimed FileBaton wait hangs the server" class.
> * **`message_bus` kill-topic:** adding `"kill"` to `TOPIC_ALLOWLIST` is necessary (otherwise `append_and_seq` rejects the KILL_TASK audit broadcast as unknown-topic) and is covered by `test_kill_topic_allowlist`. The rest of that diff is pure literal reformat.
> 
> ### Notes / asks
> 1. **(cleanup, lint becoming enforcing) leftover dangling imports from the split.** CodeQL/quality flags unused imports in `cli.py` (`_AMD_GPU_TYPES`, `_AMD_UNSUPPORTED_*`, `_default_target_summary`, `_parse_conc_sweep_concs`, `_print_kernel_*`, `ClaudeBackend`), `cli_model_gate.py` (`subprocess`), `intent_router.py` (`NoIntentEmitted`). If the `cli.py` ones are intentional **back-compat re-exports** after the module split, add them to `__all__` (or `# noqa: F401`) so they read as deliberate; otherwise drop them. Same for the empty-`except` at `cli_model_gate.py:112` and `kernel_request_handlers.py:3812/3822` — annotate the intentional non-fatal handlers. Worth clearing now since lint.yml is on track to become enforcing.
> 2. **(minor) `_any_live_compiler` is broadly conservative on shared nodes.** `COMPILER_PROCESS_NAMES` includes generic `clang`/`clang++`/`cc1plus`, so any unrelated C++ build on the node reads as "compiler alive" → the proven-dead fast path won't fire and it falls back to the 5-min mtime gate. That's safe (errs toward not-deleting a live lock), just less effective on busy shared nodes — acceptable, noting it.
> 3. **(scope) refactor + 2 new features (aiter-lock reap, reference-recipe seeding) + many bug fixes bundled.** Large to review as one unit, but the granular commit history + golden net make it traceable; not a blocker.
> 
> ### Net
> LGTM in approach and the substantive new behavior I checked (`_aiter_jit`, kill-topic) is sound, backed by a golden safety net and green CI. Only cleanups: clear the unused-import / empty-except lint findings (mark intentional re-exports in `__all__`) before lint goes enforcing. No blockers.

Problem resolved. Please review again.

chaojhou (approved):

PR #587, opened 2026-06-16
feat(recipe_kb): extend canonical identity to 7-tuple + config-donor warm-replay

zhanglei-amd:
## Summary

- Extend recipe canonical_id from 5-tuple to 7-tuple by adding `model_type` and `architectures` dimensions, enabling same-architecture cross-model warm-start reuse (+21~37% empirical gain).
- Implement 4-level fallback strategy: L1 exact → L2 same-arch-class (0.95) → L3 cross-version (0.5) → L4 relative (0.3), with confidence gating warm-replay automatically.
- Config-donor decoupling: when identity match has no replayable `best_config`, borrow champion config from nearest same-architecture sibling; KEEP write-back ensures `best_config` is stamped alongside lessons so future sessions can auto-replay.

## Changes

| Area | Files | What |
|------|-------|------|
| Identity layer | `recipe_snapshot_constants.py`, `canonical_id.py` | 8-segment canonical_id (`inference:model:hw:fw:model_type:arch:fwv:prec`), legacy 6-segment backward compat |
| Local store | `local_store.py` | 7-level directory walk, `list_all_live_recipes()` (bypasses 1000-row cap for mirror ingest) |
| Remote clients | `gbrain_remote_client.py`, `gbrain_ingest.py`, `composite_remote.py` | Contains-match for architectures labels, emit `model_type`/`architectures` tags on gbrain pages |
| T0 warm-start | `cortex_t0.py` | 4-level fallback cascade, `_find_config_donor`, `_cid_matches` for 5→7 resume compat |
| Coordinator | `coordinator.py` | `_extract_kept_best_config` + `_kb_best_config_overrides_for_keep` (stamp best_config on KEEP), tick-level cache for `_read_local_recipe_row`, config-donor replay gate |
| Dispatcher | `dispatcher.py` | 7-tuple label_match decoding |
| Tests | 9 test files | 192 tests passing, new coverage for best_config write-back and architecture matching |

## Backward compatibility

- Legacy 5-tuple canonical_ids (6-segment strings) are auto-padded with `unknown_model_type`/`unknown_arch` defaults on parse.
- Legacy 5-level KB directories are accepted alongside new 7-level directories.
- Resume from old `state.json` with 5-tuple cid works via `_cid_matches` normalization.

## Test plan

- [x] 192 unit tests passing (`pytest inference_optimizer/tests/ -q`)
- [x] End-to-end session test: Qwen2.5-7B-Instruct warm-replay reproduced +1.2% gain from KB `--disable-radix-cache`
- [ ] Verify gbrain mirror ingest with new image (pending gbrain token refresh)
- [ ] L2 fallback validation with a model that has no exact recipe but same model_type sibling

### DIFF
--- a/inference_optimizer/orchestrator/coordinator.py
+++ b/inference_optimizer/orchestrator/coordinator.py
@@ -6367,8 +6367,7 @@ class Coordinator:
 
         best_config: dict[str, Any] = {}
         if args.strip():
-            # RecipeKB + warm-replay readers still key on the legacy field name.
-            best_config["extra_sglang_args"] = args.strip()
+            best_config["extra_server_args"] = args.strip()
         if envs:
             best_config["extra_envs"] = envs
         return best_config
--- a/inference_optimizer/orchestrator/coordinator.py
+++ b/inference_optimizer/orchestrator/coordinator.py
@@ -17,7 +17,7 @@ from concurrent.futures import TimeoutError as FuturesTimeoutError
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import Any, Awaitable, Callable
+from typing import Any, Awaitable, Callable, Mapping
 
 from ..compat.payload_aliases import read_extra_server_args
 from ..recipe_kb import RecipeKB, recipe_canonical_id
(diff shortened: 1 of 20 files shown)


chaojhou (commented):
## Review — #587 `feat(recipe_kb): extend canonical identity to 7-tuple + config-donor warm-replay`

Read the identity layer (`canonical_id.py`, `recipe_snapshot_constants.py`), the T0 cascade + config-donor (`cortex_t0.py`), and the warm-replay gate + KEEP write-back (`coordinator.py`). The design is well thought-out and most of it checks out; the one thing to resolve before merge is the migration story for the existing KB corpus, plus the rebase.

### 1. (high — confirm) 5-tuple-era recipes for the *same model* become unreachable after this upgrade
New sessions now build a 7-tuple cid with **real** `model_type` / `architectures`, e.g.
`inference:llama3:mi300x:sglang:llama:LlamaForCausalLM:0.4.x:fp8`.
A recipe written before this PR has only the 5-tuple identity; when its directory/row is normalized it pads to `…:unknown_model_type:unknown_arch:…`. The cascade then can't reach it for the **same model**:
- **L1** is an exact string cid → miss (real arch ≠ `unknown_arch`).
- **L2 / L3** build `label_match` but explicitly drop `unknown_model_type` / `unknown_arch` (`v not in (...)`), so a stored row whose arch slug *is* `unknown_arch` won't match the arch-keyed search.
- Only **L4** (dispatcher relative) can surface it — at `conf=0.3`, which is below the `0.7` replay gate, i.e. advisory-only.

Net: after this lands, the entire pre-existing recipe corpus (same-model, 5-tuple era) silently drops from confident warm-start (L1) to advisory-only until each recipe is re-written with model_type/arch. The KEEP write-back will re-stamp over time, but the immediate effect is a warm-start regression on day one.

Please confirm one of: (a) a 5→7 re-key migration of the existing local store **and** the gbrain/cortex remotes, or (b) a same-model fallback tier that matches a stored `unknown_arch` row by the (model+hw+fw+fwv+prec) 5-tuple at a replay-worthy confidence. If neither exists, the "+21–37% gain" is real for new corpus but the upgrade costs existing hits.

### 2. (blocking) `mergeStateStatus = DIRTY` — rebase onto `main` before merge.

- **Confidence gate is sound** (`_maybe_enqueue_warm_replay`): it now gates on the donor *transfer* confidence (`config_confidence`) vs `min_conf=0.7`, decoupled from the identity-match confidence. L1 (1.0) / L2 (0.95) auto-replay; **L3 (0.5) / L4 (0.3) stay advisory-only and never auto-apply a borrowed config** — exactly the documented intent. The cross-version risk is correctly not auto-replayed.
- **Config-donor is stack-safe:** `_find_config_donor` constrains siblings to the same `hardware + framework + model_type + arch (+ precision)(+ framework_version)`, so the borrowed champion config always matches HW + precision + architecture — only the model differs. Priors/replay decoupling (priors from the identity match, replay config from the donor) is clean.
- **KEEP write-back is conservative:** `_kb_best_config_overrides_for_keep` stamps `best_config` only when the live row has no replayable config OR the new tput ≥ live tput (champion-stays-best); `_read_local_recipe_row` is tick-cached and the cache is correctly invalidated after `put_recipe`.
- **Legacy cid parsing is positionally correct:** the 6→8-segment (`cid_to_path_components`) and 5→7-level (`canonical_id_for_path`) paths insert `model_type` / `architectures` defaults between `framework` and `framework_version`, matching the new segment order. `_architectures_slug` sorts + `+`-joins for determinism.

### Minor
- `cid_to_path_components`: the error-message expression `f"got {len(parts) if len(parts) != 8 else parts}"` is a dead/confusing branch — when `len(parts)==8` the guard above never raises, so the `else parts` arm is unreachable. Simplify to `got {len(parts)}`.
- CodeQL flags an empty `except` at `cortex_t0.py:166`; commit `8aa651a7` claims to comment it — confirm the annotation actually clears the alert (the bot comment predates that commit).

### Net
The gate, donor constraints, and KEEP write-back are solid. Gate on **(1)** the existing-corpus migration/fallback (otherwise day-one warm-start regression for same-model legacy recipes) and **(2)** the rebase. Minor cleanups are optional.

chaojhou (commented):
## Re-review (latest HEAD `3372d5f4`, rebased) — follow-up on prior findings

Pulled the rebased branch. Most of my prior review is addressed; one item is **not fully resolved and is now actually more load-bearing**.

### Resolved ✅
- **Rebase (prev #2):** branch is rebased onto current `main`; `mergeStateStatus` is no longer `DIRTY` (just review-required).
- **Wildcard label-matching (prev #1, the right fix):** `4a849c1d` adds `_arch_contains` / `_model_type_matches`, and both treat `None / empty / DEFAULT_*_SLUG / "none"` on **either side** as a wildcard. So a legacy recipe whose arch/model_type slug is `unknown_*` still matches a new session's real-arch L2/L3 query (and vice-versa). This correctly closes the "L2/L3 drop `unknown_arch` → legacy unreachable" gap I raised — good, and bidirectional.
- **Minor cleanups:** `cid_to_path_components` error-message dead branch fixed (`4a849c1d`); empty-`except` at `cortex_t0.py` annotated (`a9348f67`); KEEP write-back now stamps canonical `extra_server_args` (avoids the legacy-field DeprecationWarning).

### Still open (now the key blocker) ❌ — the referenced migration script does not exist
`3372d5f4` removes the in-place legacy compat (6→8 padding, 5-level dir acceptance, `_cid_matches`) and its message tells operators to *"use the backfill_7tuple.py migration script"* before upgrading. **That script is not in this PR or anywhere in the repo** — I searched the whole tree; the only migration-shaped file is `scripts/migrate-labels.sh`, which is an unrelated GitHub **labels/milestone** tool (from #478), not a recipe-KB 5→7 directory migration.

Consequence, given `local_store._walk_recipe_dirs` now hard-gates on depth (`if len(rel_parts) != 7: skip`):
- On upgrade, every pre-existing **5-level** recipe directory is silently skipped (debug log only) → the entire prior local KB corpus goes invisible.
- With the in-place padding compat also removed, there is no longer any runtime bridge — so without a migration tool that physically relocates `model/hw/fw/fwv/prec` → `model/hw/fw/unknown_model_type/unknown_arch/fwv/prec`, this is exactly the "upgrade-day amnesia" the commit claims to prevent (and the wildcard matching can't help, because those dirs never reach `_matches_labels` — they're dropped at the depth gate first).

**Ask:** ship the `backfill_7tuple.py` (or whatever the real migration is) in this PR, wire it for the remote gbrain/cortex corpus too if those are keyed the same way, and fix the commit message to point at the real path. If a migration already exists elsewhere (different repo/runbook), link it and confirm it covers both local + remote; otherwise this should block merge for any deployment with an existing KB.

### Net
Direction is right and the wildcard matching is a clean fix. The one remaining gate is the **missing migration script** that the legacy-compat removal now depends on — without it, existing-KB deployments lose all prior warm-start hits on upgrade.

chaojhou (approved):

PR #616, opened 2026-06-18
fix(inference): correct GEAK dispatch attribution

### LINKED ISSUE #601: GEAK dispatch rejected by --backends serialization bug ("unsupported backend(s): ['geak']")
## Summary

Three runs rejected GEAK with a self-contradictory error:

```
ValueError: unsupported backend(s): ['geak'] (allowed: ['claude', 'codex', 'cursor', 'forge', 'geak'] ...)
```

— it lists `geak` as allowed while refusing it. The contradiction is the diagnostic: the validator compares each token against the allow-list `{geak, claude, codex, cursor, forge}`, and the token here is the **literal string** `"['geak']"` — a Python list that was stringified upstream and passed to `--backends` as one opaque value instead of the bare string `geak`. `"['geak']"` is correctly not in the allow-list, so it is rejected, and the error prints the stringified list back verbatim. This is a serialization/dispatch bug in how the `--backends` argument is built, **not** a version or stale-build problem.

## Not a version issue

An earlier read attributed this to an outdated GEAK build (a "v3.2.0 vs v3.2.1" mismatch). The artifacts disprove that:
- Per-run version records (`versions … geak`) carry git short-hashes, not semver tags; all three rejected runs ran GEAK commit `c0a1f93`.
- `c0a1f93` **is** the v3.2.1 release (`c0a1f937ca74789ec54bde4daae17db684acb33a`, the pin from #578) — so these runs were on the *intended* version, not a stale one.
- Other runs in the same cohort on this same commit did **not** hit the error, so the build is identical and not the discriminating factor.
- The allow-list already includes `geak`, yet the dispatch is rejected — only a malformed token explains that.

Refreshing/re-pinning GEAK would not fix these runs.

## Root cause

- The rejection is raised in `kernel-agent/tools/kernel_optimization.py`, `_parse_backends` (the `raise ValueError(f"unsupported backend(s): …")` at line 568). **The validator is correct and should stay.**
- The real fix is upstream at the call site that builds `--backends`: it hands over `str(["geak"])` → `"['geak']"` instead of `"geak"`. Trace the payload's `backends` field through the dispatch in `inference_optimizer/orchestrator/kernel_request_handlers.py` and ensure a list is `",".join(...)`-ed (or a single element unwrapped) before it reaches the subprocess command line.

## Suggested fix

Build the `--backends` value as a bare backend name (or a properly comma-joined string), never the `repr()` of a list. Optionally make `_parse_backends` tolerant of a stringified list as defense-in-depth.

## Evidence

Signature: `error` = `ValueError: unsupported backend(s): ['geak'] (...)`. All three runs use GEAK commit `c0a1f93` (= v3.2.1).

```
/wekafs/users/d2d10ce6177d46ed841e1f205b74b86f/ChaoticNeutrals-Domain-Fusion-L3-8B/20260616T142856Z
/wekafs/users/d2d10ce6177d46ed841e1f205b74b86f/ChaoticNeutrals-Templar_v1_8B/20260616T142344Z
/wekafs/users/d2d10ce6177d46ed841e1f205b74b86f/chargoddard-Meta-Llama-3-8B-InitializedEmbeds/20260616T142347Z
```

_Part of a 5-issue set from a GEAK-dispatch review of the 24h-before-2026-06-17 cohort (74 runs, 0 GEAK patches). Verified against run artifacts and the deployed build on 2026-06-17. Supersedes the "outdated GEAK version" framing; related to #578 (which confirms c0a1f93 = v3.2.1)._

### LINKED ISSUE #602: Non-GEAK attempts (e.g. Claude) recorded under the GEAK invocation bucket
## Summary

Kernel-optimization attempts can be filed under the **GEAK** reporting bucket even when a different optimizer actually ran them. In one run, a Claude attempt that timed out (~63 min) was recorded as a GEAK failure. This makes GEAK look worse than it is and muddies any cross-optimizer comparison.

## Impact

1 run in the 24h-before-2026-06-17 cohort (reporting/metrics correctness; severity low, but it contaminates GEAK success/failure numbers used elsewhere).

## Root cause

- `inference_optimizer/breakdown/recorder/instrument.py`, `_invocation_section` (~line 423) maps an optimizer name to its reporting bucket correctly. The bug is the **fallback**: when no per-optimizer attempt detail is available, the record defaults to the GEAK bucket — `section = _invocation_section(backend) or "geak_invocations"` (~line 531), and the backend itself defaults via `backend or "geak"` (~line 536).
- So a pre-dispatch/infra failure whose backend wasn't resolved silently lands in `geak_invocations`.

## Suggested fix

File the attempt under whichever optimizer actually ran (the name is available on the attempt/command, e.g. the `--backends` value). Only use a neutral/"unknown" bucket when the optimizer truly can't be determined — never default silently to GEAK.

## Evidence

Signature: the `geak_invocations` record has `backend: "geak"` but its `error` is a `TimeoutExpired` on a subprocess invoked with `--backends claude` (`error_class` = `subtask_exception`) — a Claude attempt mislabeled as GEAK.

```
/wekafs/users/d2d10ce6177d46ed841e1f205b74b86f/bunnycore-QwQen-3B-LCoT-R1/20260616T030624Z
```

_Part of a 5-issue set from a GEAK-dispatch review of the 24h-before-2026-06-17 cohort (74 runs, 0 GEAK patches). Verified against run artifacts and the deployed build on 2026-06-17._

xiaofei-zheng:
## Summary
- Fix #601 by normalizing `backends` payloads before building `--backends`, with defense-in-depth parsing for stringified backend lists.
- Fix #602 by preserving the actual dispatched backend on timeout/pre-dispatch failures and stopping unattributable failures from defaulting into GEAK metrics.
- Add a per-kernel backend-ladder budget so fallback execution stays within one kernel budget when an earlier backend hangs.

## Test plan
- [x] `python -m pytest inference_optimizer/tests/test_geak_dispatch_routing.py -q`
- [x] `python -m pytest inference_optimizer/tests/test_decision_framework.py inference_optimizer/tests/test_profile_and_kernel_handlers.py -q`
- [x] `python -m pytest inference_optimizer/tests/test_kernel_journey.py inference_optimizer/tests/test_forge_lane.py inference_optimizer/tests/test_kernel_shape_validation.py inference_optimizer/tests/test_breakdown_smoke.py inference_optimizer/tests/test_geak_dispatch_routing.py kernel-agent/tools/test_forge_submit.py -q`

Fixes #601.
Fixes #602.

### DIFF
--- a/inference_optimizer/breakdown/recorder/instrument.py
+++ b/inference_optimizer/breakdown/recorder/instrument.py
@@ -528,12 +528,21 @@ def record_kernel_invocations(
         if not failed:
             return
         backend = str(result.get("backend") or "").lower()
-        section = _invocation_section(backend) or "geak_invocations"
+        section = _invocation_section(backend)
+        if section is None:
+            # The optimizer that ran could not be determined (a genuine
+            # pre-dispatch / gating failure where no backend subprocess
+            # launched, or an ambiguous multi-backend value). Do NOT fabricate a
+            # GEAK invocation — that contaminates GEAK's success/failure tally
+            # with failures it never owned. The failure stays visible via the
+            # kernel_dispatch / kernel_backend_result journey lanes. See
+            # Hyperloom#602.
+            return
         payload = {
             "kernel_id": kid,
             "attempt_id": "",
             "run_id": run_id,
-            "backend": backend or "geak",
+            "backend": backend,
             "decision": "FAILED",
             "status": status or "failed",
             "error": result.get("error") or err_class or None,
@@ -1079,7 +1088,11 @@ def record_kernel_backend_result(
         failed = status in _FAILED_STATUSES or (decision == "REVERT" and bool(err_class))
         if not failed:
             return
-        backend = str(result.get("backend") or "").lower() or "geak"
+        # Never default an unattributable failure to GEAK; a genuine pre-dispatch
+        # gating failure (no backend launched) is recorded honestly as
+        # "unknown" so the kernel_journey timeline does not inflate GEAK's
+        # failure count with attempts it never owned. See Hyperloom#602.
+        backend = str(result.get("backend") or "").lower() or "unknown"
         payload = {
             "kernel_id": kid,
             "attempt_id": "",
@@ -1105,13 +1118,14 @@ def record_kernel_backend_result(
             payload,
             key=f"{kid}-predispatch",
         )
-        record_tool_version(
-            session_dir,
-            tool=backend,
-            root=str(result_meta.get("root_dir") or "") or None,
-            version=str(result_meta.get("version") or "") or None,
-            producer=producer,
-        )
+        if backend != "unknown":
+            record_tool_version(
+                session_dir,
+                tool=backend,
+                root=str(result_meta.get("root_dir") or "") or None,
+                version=str(result_meta.get("version") or "") or None,
+                producer=producer,
+            )
     except Exception:  # noqa: BLE001
         log.debug("record_kernel_backend_result failed", exc_info=True)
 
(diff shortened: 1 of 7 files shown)


ZhengGong-amd (approved):
Validated, approved

PR #621, opened 2026-06-18
Fix cu-mapping resolver so HL dispatches aiter MoE kernels to GEAK

iraj465:
## Summary
Makes the cu-mapping (`op_to_source.json`) resolver actually resolve aiter kernels to their editable `.cu`, so HL's patchability gate stops dropping every aiter op and the HL→TraceLens→GEAK loop reaches GEAK. Stacked on `cu_mapping_test` (Tharun's cu-resolve WIP); this PR is the 2 fix commits on top.

Root cause: the committed `op_to_source.json` stores each op's source NESTED as `{sglang,vllm}->{device_kernel:{kernel_source_path}}`, but `OpResolver` only emits sources when `status=="resolved"` and reads `abs_path`/`abs_path_{sglang,vllm}` (single) or `routes[].match` (dispatch). With no bridge, every op resolved to empty → HL's filter (`parallel_e2e_runner.py:248`, `source_file && Path(...).exists()`) rejected all aiter kernels → GEAK dispatched nothing.

## Changes (`kernel-agent/tools/tracelens_analysis.py` + `data/op_to_source.json`)
- **Normalizer** (`_normalize_op_entry` in `load_mapping()`): bridges the nested data schema → resolver schema for all 104 ops (single vs dispatch by distinct device kernel). Idempotent; single chokepoint every consumer goes through.
- **`aiter::fmoe_g1u1` entry**: the 42.36% GLM-4.7-FP8 MoE kernel (pertoken route → `asm_fmoe.cu`), previously absent.
- **Robustness (general, not model-specific):**
  - `match = glob.escape(dk.strip())` for auto-generated dispatch routes — 24 device-kernel names contain fnmatch metachars `[ ] ? *` that would silently fail self-match or over-match; strip also fixes a trailing-space name the resolver strips on lookup.
  - Strip a trailing ` (prefill|decode|prefilldecode|mixed)` phase suffix before the exact `mapping.get()` lookup — TraceLens annotates candidate names with the phase, which otherwise caused a nondeterministic miss → kernel skipped.
  - Removed a stray `@lru_cache` on `_normalize_op_entry` (unhashable dict arg).

## Verification
On the existing GLM-4.7-FP8 roofline trace (no re-serve):
- `hot_kernels` 0 → 5; `5/5` survive HL's patchable filter.
- `aiter::fmoe_g1u1` @ 42.36% resolves to `/sgl-workspace/aiter/csrc/py_itfs_cu/asm_fmoe.cu` (`op_to_source_status=resolved`), with faithful traced geometry (w1 `(161,512,5120) fp8`, w2 `(161,5120,256) fp8`).
- Unit: 78/78 dispatch routes self-resolve; synthetic `kernel[2]` metachar resolves; phase-suffixed `aiter::fmoe_g1u1 (prefill)`/`(decode)` resolve; whole table 92/104 resolved.
- End-to-end: full `tracelens_analysis` through the LLM orchestrator passes; HL then dispatches GEAK on the fmoe kernel, and GEAK autonomously builds its harness on the exact traced shapes `(7680,5120,256,161,9,32)` (`harness_shapes_source=user_task:production`) — not synthetic.

## Test plan
- [ ] `tracelens_analysis` on a GLM-4.7-FP8 trace yields `hot_kernels>0` with aiter MoE resolved to a real `.cu`
- [ ] resolver unit checks: dispatch self-resolve, metachar route, phase-suffix lookup
- [ ] HL kernel-opt dispatches GEAK (`kernel_attempts>0`) on the resolved kernel

### DIFF
--- a/kernel-agent/tools/data/op_to_source.json
+++ b/kernel-agent/tools/data/op_to_source.json
@@ -1879,5 +1879,21 @@
       }
     },
     "sglang": {}
+  },
+  "aiter::fmoe_g1u1": {
+    "kind": "single",
+    "patchable": true,
+    "python_launcher_path": [
+      "aiter/fused_moe.py(367): fused_moestage"
+    ],
+    "sglang": {
+      "aiter::fmoe_bf16_pertokenFp8_g1u1_vs_silu_1tg_ps_32x256": {
+        "kernel_kind": "aiter_asm_kernel",
+        "kernel_source_line": "5",
+        "kernel_source_path": "/sgl-workspace/aiter/csrc/py_itfs_cu/asm_fmoe.cu",
+        "patchable": true
+      }
+    },
+    "vllm": {}
   }
 }
--- a/kernel-agent/tools/tracelens_analysis.py
+++ b/kernel-agent/tools/tracelens_analysis.py
@@ -180,19 +180,95 @@ class OpResolution:
         self.stamp_onto(item)
 
 
+def _normalize_op_entry(entry: dict[str, Any]) -> dict[str, Any]:
+    """Bridge the data schema (nested ``{fw:{device_kernel:{kernel_source_path}}}``)
+    to the schema the resolver consumes (``status``/``abs_path``/``routes``).
+
+    The committed ``op_to_source.json`` stores each op's editable source NESTED as
+    ``sglang``/``vllm`` -> ``{<device_kernel_name>: {kernel_source_path, patchable}}``,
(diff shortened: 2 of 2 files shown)


iraj465:
Landed the 2 cu-mapping fix commits (7e0db39, 7625c99) directly onto fix/tracelens_geak_interface/cu_mapping via clean fast-forward from 04d77e1 (no conflicts). Continue from that branch; closing this stacked PR.

PR #626, opened 2026-06-18
Hyperloom Interface for TraceLens-GEAK: `.py` → `.cu` resolver Update

tsrikris:
## Summary

Replaces the heuristic, per-family launcher→`.cu` promotion chain in the TraceLens→GEAK
interface with a single **dict-first resolver** driven by a curated ground-truth dictionary
(`op_to_source.json`). Each CPU op from a trace is now resolved to its editable device source
via the dictionary; composite/multi-source ops fan out into one GEAK run per `.cu`, dispatch
ops are disambiguated by the trace's device kernel symbol, and the legacy aiter/sgl-kernel
promotion helpers (plus their tests) are deleted. Net: **5 files changed, +2473 / -492**.

Authors: @tsrikris, @devalshahamd, @kyle-hoffmeyer, @iraj465 , @rodosingh 

Required: https://github.com/AMD-AGI/Hyperloom/pull/new/feature/tracelens_release_v0.7.0

## Why

| Pain point | Old | New |
|---|---|---|
| Launcher→device mapping | Hardcoded per-family allowlists (`_AITER_COMPILE_OPS_PROMOTIONS`, `_SGL_KERNEL_DEVICE_PROMOTIONS`) grown one op at a time; wrapper rewrites that compiled but had zero E2E effect | A single curated `op_to_source.json` ground-truth dictionary (104 ops) read by one resolver |
| Multi-kernel / composite ops | Collapsed to one source path; a fused op was optimized as a single candidate | Fan-out: N editable sources → N candidates, `duration_us` split evenly so `gpu_pct` isn't inflated |
| Dispatch ops (one CPU op → many device kernels) | No way to pick the right `__global__` | Matched against the trace's device `Kernel Name` column; prompt names the exact device symbol to focus the rewrite |
| Patchability verdict | Inferred from markers/library/root heuristics on every op | Curated `patchable` + editable-source check is authoritative; heuristics only run on dictionary misses |

## Architecture

```
trace report (analysis.md)
   │  CPU op name + .py launcher + "Kernel Name" column
   ▼
tracelens_skill_runner._row_to_candidate
   │  _parse_kernel_name_cell → device_kernel_name(s)
   ▼
tracelens_analysis._finalize_candidates
   │
   ├─ _expand_op_fanout ──► resolve_op_source(name, framework, device_kernel_name)
   │                              │  OpResolver + load_mapping(data/op_to_source.json)
   │                              ▼
   │                        OpResolution {single | dispatch | composite}
   │                              │  leaf_resolutions() → 1 leaf per editable .cu
   │                              ▼
   │                        N candidates (duration split)
   │
   ├─ routable verdict ──► OpResolution.apply_to(item)   # source_file ← curated .cu
   │                       classify_patchability()        # honors op_to_source verdict
   │
   └─ miss / unresolved ─► legacy .py-launcher + grep/pybind pipeline (unchanged)
                                   │
                                   ▼
                  kernel_optimization.build_prompt
                     device_kernel_name + kernel_sources in metadata
                     ">>> DEVICE KERNEL FOCUS <<<" block names the __global__
```

### Data model — `data/op_to_source.json`

The dictionary is keyed on the **CPU op name** exactly as it appears in the trace's
`Operation` column. Each entry carries an op-level `kind` and `patchable`, then one or both
serving-framework containers (`vllm` / `sglang`) that map a **device kernel symbol glob** to a
leaf record. Routability is *derived* — there is no stored `status`; an op is routable iff its
resolved leaf has `patchable == true` and an **editable source**: native device code
(`.cu`/`.cuh`/`.hip`) **or** a repo-resident Triton/TileLang `.py` kernel. The only `.py`
exclusion is *generated* Triton (`triton_inductor_generated`, or any `torchinductor` / `/tmp/`
path), which is regenerated code rather than an editable repo source. The leaf's `kernel_kind`
(`triton`/`tilelang`/...) does not otherwise restrict editability.

```jsonc
"<cpu_op_name>": {
  "kind": "single" | "dispatch" | "composite",
  "patchable": true | false,                 // op-level curated verdict
  "sglang": {                                 // and/or "vllm": { ... }
    "<device_kernel_symbol_glob>": {
      "kernel_source_path": "/abs/path/to/file.cu" | null,
      "kernel_source_line": 123 | null,
      "kernel_kind": "ck_gemm" | "aiter_quant" | "rocm_paged_attention"
                   | "triton" | "vendored_collective" | "aten" | ...,
      "patchable": true | false               // leaf-level verdict
    }
  }
}
```

Meaning of `kind`:

- **single** — one editable device kernel; the device symbol is informational, the lone leaf
  is always selected.
- **dispatch** — one CPU op can launch *one of several* device kernels; exactly one runs per
  call, so the trace's device symbol selects which leaf (and whether it is editable at all).
- **composite** — one CPU op fires *several* sub-kernels in sequence and they **all run**; the
  op fans out into one candidate per distinct editable `.cu` (its full blast radius).

### `OpResolution` / `OpResolver`

```python
@dataclass
class Leaf:
    device_kernel_name: str
    kernel_source_path: str | None
    kernel_source_line: int | None
    kernel_kind: str
    patchable: bool

@dataclass
class OpResolution:
    op_name: str
    framework: str            # "vllm" | "sglang" (trace origin)
    kind: str                 # "single" | "dispatch" | "composite"
    patchable: bool           # op-level curated verdict
    matched: bool             # dict hit (False → legacy fallback)
    leaves: list[Leaf]        # device-kernel leaves selected for this op

    def leaf_resolutions(self) -> list[Leaf]:
        """Editable leaves only, de-duplicated by .cu path — one GEAK run each."""
        return _select_sources(self.leaves)

    def apply_to(self, item: dict) -> None:
        """Stamp the curated .cu + symbol onto a candidate dict."""
        leaf = item["_leaf"]
        item["source_file"]        = leaf.kernel_source_path
        item["source_line"]        = leaf.kernel_source_line
        item["device_kernel_name"] = leaf.device_kernel_name
        item["kernel_kind"]        = leaf.kernel_kind
        item["resolution_method"]  = "op_to_source"
        item["reusable_native_kernel"] = leaf.patchable
```

The resolver itself is a pure dictionary walk — no I/O on the hot path beyond one cached parse:

```python
def resolve_op_source(name, framework, device_kernel_name):
    mapping = load_mapping()                       # cached parse of data/op_to_source.json
    entry = mapping.get(name)
    if entry is None:
        return OpResolution(op_name=name, matched=False, leaves=[])   # → legacy fallback

    fw = entry.get(framework) or _coalesce_frameworks(entry)          # vllm/sglang container
    kind = entry["kind"]

    if kind == "single":
        leaves = _single(fw)                       # the lone symbol entry
    elif kind == "dispatch":
        leaves = _dispatch(fw, device_kernel_name) # symbol glob that matches the trace
    else:  # composite
        leaves = _composite(fw)                    # every sub-kernel (all run)

    return OpResolution(
        op_name=name, framework=framework, kind=kind,
        patchable=entry["patchable"], matched=True, leaves=leaves,
    )


def _single(fw):                       # one leaf, symbol is informational
    sym, rec = next(iter(fw.items()))
    return [_leaf(sym, rec)]


def _dispatch(fw, device_kernel_name): # pick the one route the trace actually ran
    for sym_glob, rec in fw.items():
        if fnmatch(device_kernel_name or "", sym_glob):
            return [_leaf(sym_glob, rec)]
    return []                          # no match → unresolved → legacy fallback


def _composite(fw):                    # all sub-kernels fire → full blast radius
    return [_leaf(sym, rec) for sym, rec in fw.items()]


def _select_sources(leaves):           # editable + deduped, one GEAK run per .cu
    seen, out = set(), []
    for l in leaves:
        if _is_editable_source(l) and l.kernel_source_path not in seen:
            seen.add(l.kernel_source_path)
            out.append(l)
    return out


def _is_editable_source(leaf):
    p = leaf.kernel_source_path
    if not (leaf.patchable and p):
        return False
    low = p.lower()
    if low.endswith((".cu", ".cuh", ".hip")):   # native device code
        return True
    if low.endswith(".py"):                     # repo Triton / TileLang kernel
        if leaf.kernel_kind == "triton_inductor_generated":
            return False                        # regenerated, not an editable source
        if "torchinductor" in p or p.startswith("/tmp/"):
            return False
        return True
    return False

```

### Fan-out and `duration_us` splitting

`_expand_op_fanout` is the bridge between one trace row and the GEAK candidate list. A
`single` op (or a `dispatch` op whose route resolved to one `.cu`) yields exactly one
candidate. A `composite` op with N distinct editable `.cu` files yields **N candidates**, and
the row's measured time is split evenly so the per-candidate `gpu_pct` is not N× inflated:

```python
def _expand_op_fanout(item):
    res = resolve_op_source(item["name"], item["framework"], item.get("device_kernel_name"))
    if not res.matched:
        return None                                # caller runs legacy launcher/grep/pybind

    leaves = res.leaf_resolutions()
    if not leaves:                                 # matched but nothing editable
        item["reusable_native_kernel"] = False     # non-rewritable, explicit skip_reason
        item["skip_reason"] = f"op_to_source: {res.kind} → no editable .cu"
        return [item]

    n = len(leaves)
    out = []
    for leaf in leaves:
        c = dict(item)
        c["_leaf"] = leaf
        c["duration_us"] = item["duration_us"] / n  # split so gpu_pct isn't inflated
        res.apply_to(c)
        out.append(c)
    return out
```

### Patchability short-circuit

`classify_patchability` no longer re-derives a verdict from name/library/root markers when the
op was resolved via the dictionary. If `resolution_method == "op_to_source"`, it honors the
curated leaf verdict directly; the marker/library/root heuristics only run for dictionary
misses routed through the legacy pipeline.

## Worked examples

All four examples below are real ops captured in sglang serving DeepSeek-R1 / GLM-4 on MI300X traces. The
`.cu` paths are the audited ground truth.

### Example A — `single` (editable)

A plain FP8 GEMM. The launcher `.py` is many-to-many with `.cu`, but the op resolves directly
to one CK source; the device symbol only confirms it.

Trace row (`gemm_metrics.json`):

```
Operation : aiter::gemm_a8w8_bpreshuffle
Launcher  : aiter/ops/gemm_op_a8w8.py(583): gemm_a8w8_bpreshuffle
Kernel    : void ck::kernel_gemm_xdl_cshuffle_v3_multi_d_b_preshuffle_2lds<...>
Library   : AITER   Time: 28.377 ms   Count: 92
```

Dictionary entry:

```jsonc
"aiter::gemm_a8w8_bpreshuffle": {
  "kind": "single", "patchable": true,
  "sglang": {
    "*kernel_gemm_xdl_cshuffle_v3_multi_d_b_preshuffle*": {
      "kernel_source_path": "/sgl-workspace/aiter/csrc/ck_gemm_a8w8_bpreshuffle/gemm_a8w8_bpreshuffle.cu",
      "kernel_source_line": 1,
      "kernel_kind": "ck_gemm", "patchable": true
    }
  }
}
```

Result: 1 candidate, `source_file = .../gemm_a8w8_bpreshuffle.cu`, `device_kernel_name`
stamped, routable to GEAK.

### Example B — `dispatch` (symbol disambiguation)

`vllm::unified_attention_with_output` is the canonical dispatch op: the same CPU op launches a
CK ROCm paged-attention kernel **or** one of several Triton kernels depending on the runtime
path. Only the device `Kernel Name` column can disambiguate which one ran.

Dictionary entry:

```jsonc
"vllm::unified_attention_with_output": {
  "kind": "dispatch", "patchable": true,
  "vllm": {
    "*paged_attention_ll4mi_QKV_mfma16_kernel*": {
      "kernel_source_path": "/wekafs/tsrikris/vllm/csrc/rocm/attention.cu",
      "kernel_source_line": 700,
      "kernel_kind": "rocm_paged_attention", "patchable": true
    },
    "*_fwd_kernel*":                 { "kernel_source_path": null, "kernel_kind": "triton", "patchable": false },
    "*kernel_unified_attention*":    { "kernel_source_path": null, "kernel_kind": "triton", "patchable": false },
    "*kernel_paged_attentiond*":     { "kernel_source_path": null, "kernel_kind": "triton", "patchable": false }
  }
}
```

Resolution:
- Trace symbol `paged_attention_ll4mi_QKV_mfma16_kernel` → matches route 1 → editable
  `csrc/rocm/attention.cu`, routable; prompt's `>>> DEVICE KERNEL FOCUS <<<` block names that
  exact `__global__`.
- Trace symbol `_fwd_kernel` (Triton prefix-prefill path) → matches a route whose leaf has a
  `null` `kernel_source_path` and `patchable: false` → non-editable, explicit skip rather than a
  fabricated `.cu`. (Note this leaf is non-routable because it has no on-disk source, *not*
  because it is Triton — a Triton route with a real repo `.py` path is editable.)

### Example C — `composite` (fan-out, mixed patchability)

The fused-MoE op fires **two** device kernels in sequence and both run on every invocation: an
AITER per-token FP8 quant kernel, then the Triton MoE GEMM kernel.

Trace row (`other_metrics.json`):

```
Operation : sglang_profiler::fused_moe_triton_kernels_invoke_fused_moe_kernel
Launcher  : .../moe/moe_runner/triton_utils/fused_moe.py(391): _fused_moe_kernel_sequence
Kernel    : Kernel 1: _ZN5aiter37dynamic_per_token_scaled_quant_kernel...
            Kernel 2: fused_moe_kernel
Library   : Triton   Time: 276.541 ms   Count: 2759
```

Dictionary entry:

```jsonc
"sglang_profiler::fused_moe_triton_kernels_invoke_fused_moe_kernel": {
  "kind": "composite", "patchable": true,
  "sglang": {
    "*dynamic_per_token_scaled_quant_kernel*": {
      "kernel_source_path": "/sgl-workspace/aiter/csrc/kernels/quant_kernels.cu",
      "kernel_source_line": 1,
      "kernel_kind": "aiter_quant", "patchable": true
    },
    "fused_moe_kernel": {
      "kernel_source_path": null,
      "kernel_kind": "triton", "patchable": false
    }
  }
}
```

Resolution: `_composite` returns both leaves; `_select_sources` keeps only the editable one
(`quant_kernels.cu`). The op fans out into **1 routable GEAK candidate** (the quant kernel,
`duration_us` halved) while the MoE GEMM leaf is dropped because its `kernel_source_path` is
`null` and `patchable: false` — not because it is Triton. (Had that leaf carried a real repo
`.py` Triton source, it would itself be editable and the row would fan out into two candidates
with the time split evenly.)

### Example D — `single`, non-rewritable (vendored collective)

The custom all-reduce dominates communication time but has no editable device source — it is a
vendored QuickReduce collective. The dictionary records this authoritatively so no heuristic
re-derivation and no fabricated path is ever produced.

Trace row (`customcollective_metrics.json`):

```
Operation : sgl_kernel::qr_all_reduce
Launcher  : sgl_kernel/allreduce.py(78): qr_all_reduce
Kernel    : void quickreduce::allreduce_prototype_twoshot<...CodecQ8<__half,8>...>
Time      : 173.35 ms   Count: 185
```

Dictionary entry:

```jsonc
"sgl_kernel::qr_all_reduce": {
  "kind": "single", "patchable": false,
  "sglang": {
    "*quickreduce::allreduce_prototype_twoshot*": {
      "kernel_source_path": null,
      "kernel_kind": "vendored_collective", "patchable": false
    }
  }
}
```

Result: matched, `leaf_resolutions()` empty → `reusable_native_kernel = false`,
`skip_reason = "op_to_source: single → no editable .cu"`. The op is reported, not optimized.
(`aten::copy_`/`aten::add` resolve the same way with `kernel_kind == "aten"`.)

### Python (4 files)

| File | Change |
|---|---|
| `kernel-agent/tools/tracelens_analysis.py` | New dict-first resolver: `OpResolution` dataclass, `OpResolver` (`_single`/`_dispatch`/`_composite`, `_select_sources`, `_is_editable_source`), `resolve_op_source`, cached `load_mapping`. `_expand_op_fanout` splits multi-source ops into one candidate per `.cu`. `_finalize_candidates` rewritten to prefer the curated verdict and fall back to the legacy launcher/grep/pybind chain only on a miss/unresolved route. `classify_patchability` short-circuits on the `op_to_source` verdict. Deleted the legacy `upgrade_aiter_compile_ops_launcher` / `upgrade_sgl_kernel_launcher` helpers and their `_AITER_*` / `_SGL_KERNEL_*` tables. |
| `kernel-agent/tools/tracelens_skill_runner.py` | `_parse_kernel_name_cell` parses the TraceLens `Kernel Name` column (splits `Kernel 1: a<br>Kernel 2: b`, strips labels/placeholders); `_row_to_candidate` stamps `device_kernel_name` + `device_kernel_names` on each candidate. |
| `kernel-agent/tools/kernel_optimization.py` | `build_kernel_metadata` carries `device_kernel_name` + `kernel_sources` (sibling `.cu` context); `build_prompt` emits a `>>> DEVICE KERNEL FOCUS <<<` block naming the exact device symbol when an op was resolved via `op_to_source`. |
| `kernel-agent/tools/test_aiter_compile_ops_promotion.py` | Deleted (286 lines) — covered the removed PR-K aiter `@compile_ops` promotion path. |

### Data (1 file)

| File | Change |
|---|---|
| `kernel-agent/tools/data/op_to_source.json` | New curated ground-truth dictionary (+1893 lines, 104 ops). Per op: `kind` (`single`/`dispatch`/`composite`), `patchable`, and nested `vllm`/`sglang` containers mapping device kernel symbol → `{kernel_source_path, kernel_source_line, kernel_kind, patchable}`. Routability is derived (no stored `status`). |

## Net diff

```
5 files changed, 2473 insertions(+), 492 deletions(-)
```


tsrikris:
CC: @xiaofei-zheng, @chaojhou this is an interface update based on studies conducted by the TraceLens and GEAK team. Pls consider this a draft PR until we release the corresponding TraceLens version

chaojhou:
﻿**No unit tests for the new dict-first resolver.**

This PR adds ~525 lines of new branching logic in `tracelens_analysis.py` — `resolve_op_source`, `OpResolver._single` / `_dispatch` / `_composite`, `_select_sources`, `OpResolution.leaf_resolutions`, `_expand_op_fanout`, the `classify_patchability` op_to_source short-circuit, and the legacy-fallback wiring in `_finalize_candidates` — but ships no tests for any of it, and at the same time deletes the 286-line `test_aiter_compile_ops_promotion.py`.

This resolver is now the single routing path that decides how *every* kernel candidate reaches GEAK, so the regression risk is high. Please add unit coverage for at least:

- `single` with multiple editable sources -> N fan-out leaves (one per `.cu`).
- `dispatch`: trace symbol matches an editable route / matches a non-editable route / no `device_kernel_name` (unresolved).
- `composite`: mixed patchability -> only editable leaves kept; all-non-editable -> `non_rewritable`.
- `_select_sources` container selection (vllm vs sglang; on-disk tie-break; framework hint).
- `_expand_op_fanout` duration split across N leaves (and the single-leaf no-split case).
- `classify_patchability` short-circuit: `op_to_source` patchable True -> routable, False -> reason; `None`/miss -> falls through to legacy heuristics.
- Dictionary miss / unresolved -> legacy launcher/grep/pybind fallback still fires.

tsrikris:
@chaojhou  can you pls take a look?

xiaofei-zheng:
Code review findings:

1. **Framework is not threaded into the dict-first resolver (high risk).**

`tracelens_analysis.py` accepts `--framework`, and `trace_analyze_handler` passes it, but `_expand_op_fanout()` ignores `args.framework` and only reads `HYPERLOOM_FRAMEWORK`:

```python
framework = os.environ.get("HYPERLOOM_FRAMEWORK", "").strip().lower() or None
```

If that env var is unset, `_select_sources()` falls back to source presence / default ordering and can prefer `sglang` sources even for a `vllm` trace in a dual-framework image. That can route GEAK to the wrong editable source, which is exactly the class of issue this PR is trying to eliminate.

Suggested fix: pass `args.framework` into `_finalize_candidates()` / `_expand_op_fanout()` and use that explicit value for `resolve_op_source()`. Please add a regression where the same op has both `vllm` and `sglang` entries with different sources and `framework="vllm"` selects the vLLM source.

2. **Dispatch kernel matching is too permissive (medium risk).**

`_kernel_matches()` currently does broad substring matching:

```python
return trace_name == key or trace_name in key or key in trace_name
```

For dispatch ops, this can match the wrong route when symbols share prefixes/substrings or when names are truncated. It also differs from the PR description, which describes glob/fnmatch-style matching.

Suggested fix: prefer exact match, or use an explicit glob contract with escaped literal symbols, and treat multiple matches as ambiguous/unresolved rather than taking the first. Please add a regression with two route names where one contains the other and verify the resolver does not pick the wrong route.

3. **`.py` routability contract needs clarification (medium/low).**

`_is_editable_source()` treats non-inductor `.py` files as editable/routable:

```python
if low.endswith(".py"):
    if kernel_kind == "triton_inductor_generated":
        return False
    if "torchinductor" in path or path.startswith("/tmp/"):
        return False
    return True
```

But the PR text repeatedly says routable means editable `.cu/.cuh/.hip` and ?one GEAK run per `.cu`?. If repo Triton `.py` is intentionally supported, the PR docs and metadata contract should say so explicitly. If not, this should be restricted to native source extensions.

CI is green and the overall direction looks good, but I would fix at least (1) before merging because it can route candidates to the wrong framework source.

tsrikris:
@xiaofei-zheng Thank you for your comments! Addressed 1 and 3. We will keep 2 in mind as we scan more traces during next week's larger model triage.

xiaofei-zheng (approved):

PR #630, opened 2026-06-19
feat(forge): GBrain KB integration + compiled fellows default-on

zhanglei-amd:
## Summary

- Forward GBrain credentials (GBRAIN_BASE_URL/GBRAIN_TOKEN/KERNELFORGE_GBRAIN_ENABLED) to forge-loop subprocess, enabling cross-KB knowledge injection from KernelForge + GEAK + PTAO into Forge fellows
- Enable compiled fellows (hip/ck/aiter/hipblaslt/flydsl) by default; opt-out via FORGE_DISABLE_COMPILED_FELLOWS=1 — fixes RCA 0616 root cause G' where non-triton kernel candidates were silently skipped
- Fix .pth editable root discovery for in-place worktree builds
- Fix forge harness spin: param typing, key union, baseline gate
- Inject claude CLI path, API timeout, TLS skip into forge child env

## Key changes

| File | Change |
|---|---|
| `kernel-agent/tools/backends/forge_submit.py` | `_apply_fellow_env()` GBrain env forwarding, compiled fellows default-on, claude CLI injection |
| `kernel-agent/tools/test_forge_submit.py` | Updated tests for FORGE_DISABLE_COMPILED_FELLOWS opt-out |

## Test plan

- [x] Unit tests updated and passing for compiled fellows routing
- [x] E2E validated via Claw sessions: Qwen3-30B-A3B achieved 1.64x fused_moe kernel speedup
- [x] RCA 0616 problem models (calme-3b, testcat-v04) now enter KERNEL phase (previously stuck at zero output)
- [x] GBrain KB endpoint confirmed wired up in sandbox logs

### DIFF
--- a/inference_optimizer/SKILL.md
+++ b/inference_optimizer/SKILL.md
@@ -13,164 +13,1138 @@ globs:
 
 # Inference Optimizer Skill
 
-You are the **launcher and monitor**. The optimizer itself is the Python
-`inference_optimizer` runtime under this repository. Do not manually optimize in
-chat unless debugging; launch the CLI, poll persisted state, and report
+You are the launcher and monitor. The optimizer itself is the Python
+`inference_optimizer` runtime under this repository. Do not manually optimize
+inside chat unless debugging; launch the CLI, poll persisted state, and report
 objective progress.
-
-## Quick Start
-
-The whole job is three steps:
-
-1. **Install** (IR-2): run `inference_optimizer/scripts/install.sh`, then source
-   the generated `runtime/kernel-agent.env.sh` in the same shell.
-2. **Launch**: verify GPU cleanliness (IR-1), then start
-   `inference_optimizer --verbose optimize` under `setsid nohup` with
-   `--launch-info-file`.
-3. **Monitor & report**: learn `SESSION_DIR` from launch-info, poll
-   `state.json` every 300s (unless debugging startup), surface lifecycle lines,
-   and resume only unexpected crashes via same-session
-   `optimize --resume --resume-from "$SESSION_DIR"`.
-
-A user request to optimize a model is approval to run install + launch on a
-fresh node; do not stop for an extra confirmation (the one exception:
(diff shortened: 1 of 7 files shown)


chaojhou:
﻿**Compile-only autogen driver fabricates correctness signals and uses binary size as a perf proxy.**

For HIP/CK kernels with no `test_command`, autogen falls back to the compile-only driver. It only runs `hipcc -c` and never executes the kernel, yet on a successful compile it prints:

```python
print("SNR: 120.00 dB")
print("allclose: True")
...
synthetic_ms = obj_size / 1000.0 if obj_size > 0 else 1000.0
print(f"wall_ms: {synthetic_ms:.4f}")
```

Three concerns, amplified by compiled fellows now being default-on (this path becomes the default for most MI300X `hip_cpp` candidates):

1. **Fabricated `SNR: 120 dB` / `allclose: True`** — the kernel was never run, but a correctness pass is emitted. This misleads anything that trusts the signal, including the new `_baseline_correctness_ok` gate (which keys on `"snr:"` / `"allclose: true"`) and the SBD / sub-gain telemetry.
2. **`obj_size` is not a runtime-perf proxy** — `-O3` object size doesn't correlate with speed, so the IterationLoop optimizes toward "smaller `.o`" and can KEEP edits that are slower or numerically different (correctness is never actually checked).
3. **All real validation is deferred to E2E** — so a large share of `hip_cpp` candidates will report `improved=true` on a compile+shrink loop and burn E2E budget on non-improvements (consistent with forge's ~0 verified E2E sub-gain this week).

I understand this is an intentional trade-off (the docstring notes E2E is the real gate), but please don't fabricate a pass: have the compile-only path emit an explicit `correctness: UNVERIFIED` / compile-only marker instead of `allclose: True` / `SNR: 120`, so keep-decisions, the baseline gate, and telemetry can distinguish "actually verified" from "only compiled".

xiaofei-zheng (approved):

PR #639, opened 2026-06-20
feat(breakdown): mirror manifest + session breakdown to claw sync root

haishuok0525:
## Summary

The session collector discovers a run's artifacts **MinIO-first**, under the key prefix that primus-claw syncs from its workspace root (typically `/workspace`). Previously hyperloom only mirrored `session_breakdown.json` into the `USER_DATA_PATH/hyperloom/<sid>` subtree, which is **not** part of the brain-synced workspace subtree. As a result the collector's MinIO-first lookup missed the artifacts and had to fall back to WekaFS.

This PR adds a redundant mirror so the breakdown artifacts reliably land in **both** places.

## Changes

- Keep the existing `mirror_breakdown_to_claw_storage()` copy into `USER_DATA_PATH/hyperloom/<sid>` (unchanged source-of-truth behaviour).
- Add `mirror_session_artifacts_to_claw_storage()` which copies **both** `manifest.json` and `session_breakdown.json` into the claw sync root (`CLAW_WORKSPACE_PATH`, default `/workspace`), so the snapshot the collector reads is complete and self-describing (the manifest carries `claw_session_id`).
- Invoke both mirrors from the CLI `finally` block so they run once at end-of-run regardless of how optimization exits.
- Mirroring is **best-effort and non-fatal**: WekaFS under `USER_DATA_PATH` remains the source of truth, and any copy failure is logged rather than raised.

## Test plan

- [x] `inference_optimizer/tests/test_cli_breakdown_claw_mirror.py` covers both mirrors: success, no-op outside the sandbox, missing files, and overwrite.
- [ ] End-to-end: run a hyperloom marathon and confirm the collector resolves the breakdown MinIO-first (no WekaFS fallback) and the global service shows the breakdown.

### DIFF
--- a/inference_optimizer/breakdown/claw_mirror.py
+++ b/inference_optimizer/breakdown/claw_mirror.py
@@ -1,34 +1,53 @@
-"""Mirror ``session_breakdown.json`` into the claw-synced subtree.
+"""Mirror session artifacts into claw-synced storage.
 
-Hyperloom's canonical write lands in ``<session_dir>/`` which under the
-default ``per_model_ts`` layout is ``<workspace>/<model>/<ts>/``. The
-claw sandbox checkpoint sync only watches the ``hyperloom/`` subtree
-under ``$USER_DATA_PATH``, so without an explicit mirror the canonical
-artifact is lost the moment the sandbox is reaped.
+Hyperloom's canonical write lands in ``session_dir`` which, under the
+default ``per_model_ts`` layout, is ``<USER_DATA_PATH>/<model>/<ts>/``.
+In a Primus-Claw sandbox the brain only uploads ``/workspace`` to object
+storage (``syncWorkspaceToS3`` walks ``/workspace`` and maps a file at
+``/workspace/<rel>`` to the S3 key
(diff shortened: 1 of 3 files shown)


haishuok0525:
Closing as superseded. `main` has since refactored the breakdown module and added `inference_optimizer/breakdown/session_package.py`, which already implements (and goes beyond) this PR's goal:

- `PACKAGE_GLOBS` includes both `session_breakdown.json` and `manifest.json`.
- It writes those files loose under the Claw sync root `/workspace` (e.g. `/workspace/session_breakdown.json`, `/workspace/manifest.json`), plus a zip bundle and a package manifest.
- It is anchored at `/workspace` regardless of `USER_DATA_PATH`, exactly so Claw's sandbox sync ships the artifacts to S3 / MinIO for the session collector to discover MinIO-first.
- It is wired into `cli.py` (both the `finally` safety-net and the success path via `package_session_artifacts`).

The `claw_mirror.py` this PR modified was deleted in that refactor, so there is nothing left to merge.

PR #650, opened 2026-06-22
Fix vLLM patch fallback and server args propagation

chennyiiis:
## Summary
- Fix vLLM TraceLens patch resolution for rc/dev/local build versions, with same-minor and nearest-lower fallback plus explicit degraded-profile warnings.
- Add a unified `--server-args` CLI surface and route operator server flags into vLLM/SGLang/Atom YAML envs across baseline, profile, explore, and sweep.
- Preserve explicit `--max-model-len` / `$MAX_MODEL_LEN` and accept comma `$CONC` ladders without crashing downstream int parsing.

## Test plan
- Added regression tests for vLLM patch fallback version matching.
- Added CLI/workload env tests for server arg routing, `CONC` ladder handling, and max model len precedence.
- Previously validated patch discovery and sandbox patch application on remote vLLM 0.19 / 0.22 images without touching `/wekafs`.

### DIFF
--- a/inference_optimizer/SKILL.md
+++ b/inference_optimizer/SKILL.md
@@ -677,6 +677,15 @@ sanitized `*.sh` name; overrides the gpu_type auto-pick) and
 that changes params after a failure — the agent must retry with
 identical params and the run terminates after 3 consecutive failures.
 
+Operator server flags have one supported CLI entry point:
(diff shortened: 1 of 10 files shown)


chennyiiis:
Closed because this PR targeted the wrong base branch (ix/suppress-failfast-leaderboard). Recreated against main in #652.

PR #655, opened 2026-06-22
[feature] Add diffusion model optimization support

avjves:
Hello!

I'll start by saying this PR is not ready for merging or even tested with LLMs yet. I do not expect it to be merged as-is. I'm basically opening this PR for discussion so I can hopefully get some opinions from the maintainers before I invest time into finalizing and testing this. This could have been an issue, but it's easier to show working code via a PR :)

### Motivation

Hyperloom is LLM-only: every framework so far (sglang/vllm/atom) is an online, server-based LLM engine and a lot of behavior is designed specifically around these assumptions. These assumptions include the online server, decoding loop, `lm-eval` accuracy gating and parameter sweeping.

I would like to add support for diffusion models (text-to-image, text-to-video etc), so we can use Hyperloom to analyze and optimize these as well. 

This PR (in conjunction with a separate PR to Magpie, will open soon) adds initial support. I've ran it e2e already, so it works with diffusion models already, using the xDiT framework.

### How?

Rather than scattering `if framework == "xdit"` checks across the codebase, this PR introduces a small registry as the single source of truth for the current framework and routes to the relevant code paths based on this. The LLM path remains unchanged. This allows cleaner integration and the ability to add more frameworks in the future (I plan on adding support to vLLM-Omni and SGL-Diffusion as well)

### Main changes

Each framework now has a few capabilities:
- serving - whether the framework uses an online vs offline server
- launch - Build the invoke command dynamically vs a given verbatim command
- modality - text vs diffusion
- accuracy gate - `lm_eval` vs new `image_diff`
- search_knobs - what's worth sweeping. This is a big difference between LLMs and diffusion; the initial requirements already specify all the input parameters. There's no need to sweep any parameters with diffusion models.

Based on the capabilities, different code paths are taken. This is not just LLM vs Diffusion change, as some diffusion frameworks are also different from one another (online vs offline, for example).

----

As stated above, this is not yet fully tested. What do you think, would something like this work for Hyperloom? Any changes in approach that I should do? I'm open for all changes :)


avjves:
Ah, my copy of Hyperloom was a bit outdated 😄 I will need to fix these conflicts first.

PR #671, opened 2026-06-23
Fix/shuoshuo/569 vllm ep flag

lishuoshuo-amd:


ZhengGong-amd (approved):
change validated, approved

PR #681, opened 2026-06-23
fix(kernel-agent): reconstruct full source from patch when no complete artifact found

### LINKED ISSUE #680: FULL_BENCHMARK verifier defers patch-only kernel wins (artifact_source='missing')
## Summary
When a GEAK/OOB attempt's best artifact is a **unified diff** with no complete-source counterpart and no fenced code block in the logs, `_select_source_artifact` (kernel-agent/tools/kernel_optimization.py) returns `artifact_source='missing'` → `verification_status='deferred'`. A real, **correctness-passing** kernel win is then clamped to unverified and dropped by the orchestrator.

## Repro (observed)
DeepSeek-R1-0528, SGLang, MI300x, TP8, fp8, ISL/OSL 1024 — editable Triton MoE (`sglang::inplace_fused_experts`). GEAK's winning `asm-kernel-rewrite` produced **+8.4% kernel (1.0836x), CORRECTNESS: PASS**, but verification recorded:
```
"artifact_error": "no complete .py source artifact found; tried: .../asm-kernel-rewrite/patch_0.patch, .../slot_7/3rdparty/amd/profiling/torch_profiler.patch",
"artifact_source": "missing", "artifact_valid": false, "verification_status": "deferred",
"micro_speedup": 1.083567, "compile_passed": true
```

## Root cause
`_select_source_artifact` has two source tiers: (1) suffix-match a complete file, (2) scrape a fenced block from text/patch. A unified diff is neither. The full-source worktree DOES exist (`results/round_1/<task>/worktrees/slot_<M>/.../<file>.py`), but `_geak_best_worktree()` only maps best-patch paths whose parent dir is named `parallel_<M>`. The current GEAK layout writes `<task-label>/patch_N.patch` + sibling `worktrees/slot_<M>/`, so the mapping returns `None` and the complete source is never located. (Also, `geak_latest_patch` pointed at an unrelated `torch_profiler.patch`.)

## Fix
PR #PLACEHOLDER: add a universal final fallback — apply the patch to the original kernel (`git apply` across strip levels, then `patch`) to reconstruct the optimized source deterministically. Layout/strategy/suffix-agnostic; additive (runs only after existing tiers fail), so working paths are unchanged. Regression test included; validated against the real DeepSeek artifact (reconstructs the full 1248-line optimized file → verifies).

## Impact
Any workload where the winning strategy emits a patch-only artifact (common for asm/ck/flydsl rewrites) silently loses verified kernel wins. Generalization-relevant for the multi-workload CI.

iraj465:
Fixes #680.

## Summary
- A GEAK/OOB attempt whose best artifact is a **unified diff** (no full-source `.py`/`.cu`, no fenced block) previously verified as `artifact_source='missing'` → `verification_status='deferred'`, clamping a real correctness-passing kernel win to unverified.
- Add a universal final fallback in `_select_source_artifact`: apply the patch to the original kernel (`git apply` across strip levels, then `patch`) to **reconstruct the complete optimized source deterministically** — the same apply deployment uses.

## Why this is the right fix (not overfitted)
- **Universal**: depends only on the patch + original source — independent of worktree / `optimized_versions` naming, strategy, or suffix (`.py`/`.cu`/`.cuh`). (The earlier `parallel_<M>` worktree mapping in `_geak_best_worktree` is stale vs the current `<task-label>/` + `worktrees/slot_<M>/` layout; rather than chase layout strings, reconstruct from the two artifacts that always exist.)
- **Non-breaking / additive**: runs **only after** the existing tiers (full-source suffix-match, fenced-block extraction) fail. When today's path already resolves a source, behavior is byte-identical. It only converts a previously-`missing` case into a verified one.
- **Faithful**: `git apply` / `patch` is exactly how `apply_kernel_patch.py` deploys kernels, so the reconstructed file matches what runs.

## Validation
- New regression test `test_patch_only_winner_reconstructs_full_source` reproduces the DeepSeek-R1 fp8-MoE patch-only winner.
- Existing verification suite unchanged: **76 → 77 pass**; broader kernel-opt/integrate suites: **135 pass**; `ruff` clean.
- Validated against the **real** failing DeepSeek artifact (`asm-kernel-rewrite/patch_0.patch`): reconstructs the full 1248-line optimized source with the asm optimization intact → would verify the +8.4% win instead of deferring.

## Test plan
- [ ] CI: full Python matrix (3.10/3.11) + ruff/pylint/CodeQL
- [ ] Confirm no change to attempts that already resolve a full-source artifact

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,13 @@
+version: 2
+
+build:
+  os: ubuntu-24.04
+  tools:
+    python: "3.10"
+
+sphinx:
+  configuration: docs/conf.py
+
+python:
+  install:
+    - requirements: docs/requirements.txt
--- a/inference_optimizer/orchestrator/kernel_request_handlers.py
+++ b/inference_optimizer/orchestrator/kernel_request_handlers.py
@@ -863,6 +863,10 @@ def _maybe_apply_kernel_patch(
     kid = str(kernel_id or payload.get("kernel_id") or "")
     backup_root = payload.get("backup_root") or (patches_dir(session_dir, kid or "anon") / "backup")
     tool = _load_apply_tool()
+    # Snapshot mode (content-addressed deploy): when a snapshot dir of byte-exact
+    # final files is present, the patch lands atomically across all its files.
+    snapshot_dir = str(payload.get("snapshot_dir") or "").strip() or None
+    repo_root = str(payload.get("kernel_repo") or payload.get("repo") or "").strip() or None
     return tool.apply_kernel_patch(
         patch_path=patch_path,
         target_file=target_file,
@@ -874,6 +878,8 @@ def _maybe_apply_kernel_patch(
         skip_rebuild=bool(payload.get("skip_rebuild", False)),
         allow_unknown_target=bool(payload.get("allow_unknown_target", False)),
         dry_run=bool(payload.get("dry_run_patch", False)),
+        snapshot_dir=snapshot_dir,
+        repo_root=repo_root,
     )
 
 
@@ -985,6 +991,14 @@ def _resolve_integrate_payload(payload: dict, *, session_dir: Path) -> tuple[dic
     last_kernel = state.last_kernel_opt or {}
 
     if kernel_id and str(last_kernel.get("kernel_id") or "") == kernel_id:
+        # Snapshot deploy: prefer the original patch (manifest) + its byte-exact
+        # snapshot dir so the WHOLE multi-file patch lands atomically.
+        if not resolved.get("snapshot_dir") and last_kernel.get("deploy_snapshot_dir"):
+            resolved["snapshot_dir"] = str(last_kernel["deploy_snapshot_dir"])
+            if last_kernel.get("deploy_patch_path") and not resolved.get("patch_path"):
+                resolved["patch_path"] = str(last_kernel["deploy_patch_path"])
+            if last_kernel.get("deploy_repo_root") and not resolved.get("kernel_repo"):
+                resolved["kernel_repo"] = str(last_kernel["deploy_repo_root"])
         if not resolved.get("patch_path"):
             artifact = (
                 last_kernel.get("best_artifact_path")
@@ -999,6 +1013,12 @@ def _resolve_integrate_payload(payload: dict, *, session_dir: Path) -> tuple[dic
     # Multi-KEEP queue fallback: ``last_kernel_opt`` holds only the strongest pending KEEP, so pull patch_path/source_file from the per-kernel ledger for other queued KEEPs.
     if kernel_id:
         attempt = (state.kernel_opt_attempts or {}).get(kernel_id) or {}
+        if not resolved.get("snapshot_dir") and attempt.get("last_snapshot_dir"):
+            resolved["snapshot_dir"] = str(attempt["last_snapshot_dir"])
+            if attempt.get("last_deploy_patch_path") and not resolved.get("patch_path"):
+                resolved["patch_path"] = str(attempt["last_deploy_patch_path"])
+            if attempt.get("last_deploy_repo_root") and not resolved.get("kernel_repo"):
+                resolved["kernel_repo"] = str(attempt["last_deploy_repo_root"])
         if not resolved.get("patch_path") and attempt.get("last_artifact_path"):
             resolved["patch_path"] = str(attempt["last_artifact_path"])
         if not resolved.get("source_file") and attempt.get("last_source_file"):
@@ -4798,6 +4818,9 @@ def record_kernel_opt(state, result: dict[str, Any]) -> None:
     except (TypeError, ValueError):
         micro_float = 0.0
     best_artifact_path = str(verification.get("best_artifact_path", "") or "")
+    deploy_snapshot_dir = str(verification.get("deploy_snapshot_dir", "") or "")
+    deploy_patch_path = str(verification.get("deploy_patch_path", "") or "")
+    deploy_repo_root = str(verification.get("deploy_repo_root", "") or "")
     source_file = str(
         result.get("source_file")
         or (result.get("candidate") or {}).get("source_file")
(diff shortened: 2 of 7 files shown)


iraj465:
### Status: clean + green (no conflicts, all CI passing)

- **0 commits behind `main`, 0 conflict markers** — merges cleanly.
- CI: `ruff` ✅ · `pylint` ✅ · `test (3.10)` ✅ · `test (3.11)` ✅ · `CodeQL`/`Analyze` ✅
- Local: verification suite **76 → 77**, broader kernel-opt/integrate suites **135 pass**.
- `MERGEABLE` (BLOCKED = `REVIEW_REQUIRED` only — awaiting a reviewer).

### Data-backed validation against the real failing artifact
Ran the new `_select_source_artifact` against DeepSeek-R1's actual deferred attempt (`geak-44b4abf8` / `asm-kernel-rewrite`):
```
before:  artifact_source="missing", verification_status="deferred"  (micro_speedup=1.0836)
after:   artifact_source="reconstructed_from_patch", artifact_valid=True
         -> reconstructed 1248-line .py == slot worktree source (byte-identical),
            asm optimization (_ALIGN_CACHE/_aligned_for_block_m) intact
```

### Why not overfitted (evidence)
Same run, per-strategy: `asm` 1.0836x (patch only, no `optimized_versions/`), `flydsl` 1.0574x, `tilelang` 1.0336x (also no `optimized_versions/`), `ck` 1.0334x. A fix keyed on `optimized_versions/` or the `parallel_<M>` worktree name would miss `tilelang`/`asm`. Reconstructing from **patch + original** is layout/strategy/suffix-agnostic and additive (only runs after the existing full-source and fenced-block tiers fail), so already-passing attempts are untouched.

xiaofei-zheng:
Found one issue that should be fixed before merge:

`kernel-agent/tools/kernel_optimization.py` adds `_reconstruct_source_from_patch()` and runs backend-produced `.patch` / `.diff` files through:

```bash
git apply -p<N> --unsafe-paths --directory=<extraction_root> <patch>
```

These patch artifacts come from GEAK/OOB/LLM backends, so they should not be treated as trusted input. With `--unsafe-paths`, a malicious or malformed diff header using an absolute path or `..` can escape the temporary extraction directory during this verification-only reconstruction step and write outside the intended workspace.

Suggested fix:
- Remove `--unsafe-paths`.
- Before applying, parse/validate the diff headers and only allow a single target, no absolute paths, no `..`, and a target that maps to the original `target_file` / expected basename.
- Or simplify the implementation to apply the validated patch only to the explicit temporary work file, since the current `git apply --directory=work.parent` path does not actually target `work` directly and the normal `a/<file>` case appears to be handled by the later `patch ... <work>` fallback anyway.
- Add regression tests for absolute-path and `../` patch headers.

CI is green, but the current tests only cover the happy-path reconstruction and do not cover path traversal / unsafe patch headers.

iraj465:
Thanks @xiaofei-zheng — addressed in 81be3de (your suggestions 2–4):

**Dropped `--unsafe-paths`/`--directory` entirely.** Reconstruction now applies the diff ONLY to an explicit temp work file via `patch -p<N> <work>`, which writes solely to the named file and ignores the header path — a hostile header can't redirect the write.

**Header validation (`_patch_targets_only`)** before applying: rejects multi-file diffs, absolute paths, `..` traversal, and any target whose basename ≠ the kernel file (`/dev/null` add/delete side ignored).

**Regression tests added:**
- `test_patch_with_absolute_path_header_is_rejected` (`+++ /etc/cron.d/pwned`) → `missing`, no write
- `test_patch_with_parent_traversal_header_is_rejected` (`+++ b/../../../../tmp/pwned.py`) → `missing`
- `test_patch_targeting_other_basename_is_rejected` → `missing`
- happy-path test now also asserts the artifact stays within `run_dir`

No functional regression: the real DeepSeek GEAK patch still reconstructs (`reconstructed_from_patch`, complete source). Verification suite 77 → 80 pass, ruff clean.

sdubagun-amd:
## Validation: patch reconstruction tested against real GEAK patches from 5 workloads

I exercised this PR's reconstruction path (`_select_source_artifact` → `_reconstruct_source_from_patch`) against the actual `best_patch` artifacts from GEAK runs across **5 production workloads** (deployment `d2d10ce6…`): Solomon-Qwen2.5-0.5B, Vikhr-Llama-3.2-1B, glenn-gemma-2B, Ratnesh-qwen2.5-3B, distil-qwen3-4B.

### Patch-application result (what this PR fixes)
**9/9 GEAK best-patches reconstructed successfully** — spanning both real CUDA/HIP kernels and Python dispatch shims:

| Patch type | Examples | Result |
|---|---|---|
| Dispatch shim (.py) | `aiter_backend.py`, `mha.py`, `pa_decode_opt.py` | ✅ `reconstructed_from_patch` |
| CUDA kernel (.cu) | `rmsnorm_quant_kernels.cu` ×3 | ✅ `reconstructed_from_patch` |
| HIP kernel (.cuh) | `hip_act_and_mul.cuh` (silu_and_mul) | ✅ `reconstructed_from_patch` |

For one case (Ratnesh-3B `silu_and_mul`) I additionally byte-compared the reconstructed source against an independent `git apply` of the same patch — **identical**. And on the pre-PR code, the same inputs returned `artifact_source = missing`. So this PR converts the diff-only failure mode (`missing`) into a correct applied source, deterministically.

### Separate finding (NOT a concern for this PR — flagging for GEAK)
While validating, I checked whether each reconstructed patch would actually reach E2E. Most would not, for reasons outside this PR's scope:
- The **headline verified speedups are scaffolding/harness artifacts**, not kernel wins — e.g. Vikhr-1B's "25.3×" edits `sitecustomize.py` + benchmark files; others edit attention **dispatch shims** rather than device kernels.
- The genuine `.cu` kernel patches target `add_rmsnorm_quant_kernel` (compiled into `module_rmsnorm_quant`), but all 5 models serve **without quantization** → the served path uses `module_rmsnorm` (`rmsnorm2d_fwd`). Different module, so those patches are off the serve path (provable ~0% E2E).
- Exactly one patch (Ratnesh-3B `silu_and_mul`) is on the bf16 serve path, but its verified micro-speedup is only ~1.03×.

**Bottom line:** the patch-application layer this PR implements works correctly and reliably across all patch types. Whether GEAK's chosen optimizations are E2E-relevant is a separate, upstream (GEAK) question, tracked independently.

xiaofei-zheng (approved):

PR #703, opened 2026-06-24
feat(kernel-agent): GEAK_v4 integration follow-up — E2E patchability (apply+remeasure + aiter harness routing) + dispatch context

### LINKED ISSUE #715: aiter harnesses miss AITER_META_DIR routing -> patch-blind ~1.00x / harness-init hang (gap in #597)
## Summary

Auto-generated aiter kernel harnesses (from `kernel-agent/tools/harness_generator.py`) route `GEAK_WORK_DIR` onto `sys.path` but do **not** set `AITER_META_DIR`/`AITER_JIT_DIR`. Because `aiter` is installed **editable** (via a `sys.meta_path` finder), `import aiter` resolves to the ORIGINAL repo regardless of `sys.path` order, and aiter's JIT compiles `AITER_CSRC_DIR = $AITER_META_DIR/csrc`. With `AITER_META_DIR` unset, the JIT builds the **baseline** `csrc/*.cu` — so:

- every measured speedup is ~1.00x and **correctness is blind to the candidate patch** (the classic worktree-bypass failure mode, aiter-specific), and
- on GEAK v3.2.2+ (which added the deterministic `find_aiter_routing_violation` validator), the harness is correctly **rejected**, the `eval_command` fallback then hangs at `harness-init` until the preprocess hard cap (observed: 3600s), and the kernel never optimizes.

This gap was not covered by the GEAK_v4 integration (#597).

## Repro / evidence

- Workload: meta-llama-Llama-3.1-8B-Instruct (SGLang), aiter `dynamic_per_tensor_quant` (`csrc/kernels/quant_kernels.cu`), GEAK v3.2.2 line (PR #299 checkout).
- Agent log (verbatim): `pre-validated bypass REJECTED harness (worktree bypass): ... imports aiter and compiles a csrc/*.cu kernel but never sets AITER_META_DIR from $GEAK_WORK_DIR ... aiter's JIT (AITER_CSRC_DIR=$AITER_META_DIR/csrc) compiles the baseline kernel — every speedup is ~1.00x and correctness is blind to the patch.`
- Then: `[preprocess] preprocess hard cap (3600s) exceeded at stage 'harness-init' (commandment_present=False) -- terminating registry`.
- Validator that detects it: `minisweagent/kernel_languages/contract.py::find_aiter_routing_violation`.

## Why it matters

- Pre-#299 GEAK lacked the validator, so such harnesses were silently accepted and trained the optimizer on a **flat ~1.00x signal** (patch-blind) — a correctness/measurement integrity bug, not just a perf one.
- Post-#299 it's a hard failure (60-min hang -> hard-cap) that blocks any aiter `.cu` kernel from optimizing.

## Fix

In both harness templates in `kernel-agent/tools/harness_generator.py`, set BEFORE `import aiter`:
```python
os.environ.setdefault("AITER_META_DIR", WORK_DIR)                       # source routing
os.environ.setdefault("AITER_JIT_DIR", os.path.join(WORK_DIR, "_aiter_jit"))  # per-worktree build output
```
Properties: generalizable (any aiter kernel/workload), no-op for non-aiter harnesses (only aiter's JIT reads these), `setdefault` respects an operator override, and the per-worktree `AITER_JIT_DIR` prevents source-repo pollution AND parallel `run_pool` slot collisions on a shared `build/`/`.so`/ninja-lock.

Verified: rendered harness now PASSES `find_aiter_routing_violation`; all existing harness-gen tests pass. PR to follow.

iraj465:
## Summary — GEAK_v4 integration follow-up: E2E patchability + dispatch context

Follow-up to the GEAK_v4/PerfSkills integration (#597). It closes three gaps that block a kernel
optimization (GEAK v3 **or** v4) from cleanly turning into an end-to-end serving-throughput number, and
makes the apply→remeasure path a single shared, non-overfit primitive. Four commits:

### 1. Gate-less patch-apply + warm-serve E2E remeasure (`apply_and_bench.py`)
The kernel-opt integrate path couples patch APPLICATION to a POLICY gate (KEEP / REVERT / NEEDS_REVIEW:
dead-band, parity arbitration). For validation/comparison we want pure measurement: apply the exact
patch, warm-serve an A/B, report the delta, revert — **no verdict**. New `kernel-agent/tools/apply_and_bench.py`:
- **Apply** via the existing `apply_kernel_patch()` (the one place that knows HOW to apply any kernel —
  Python overlay OR aiter `.cu` rebuild + jit/cpp_itfs cache invalidation + revert). No reinvented apply.
- **Serve + bench** via the shared `bench_e2e.sh` + `adapters/{sglang,vllm}.sh`.
- **Engagement-proof** (trust, not policy): reports whether the patched kernel was actually on the live
  serving path (aiter JIT rebuild markers / `is tuned on cu_num` hits / source stem in build-load), so a
  ~0% delta is distinguishable from "patch silently didn't engage". Emits NO accept/reject.
- Result: backend-agnostic, kernel-agnostic, **optimizer-agnostic** — GEAK v3 and v4 call the SAME
  primitive (today v3 goes through the integrate gate and v4's overlay can't even apply an aiter `.cu`).
  KEEP/NEEDS_REVIEW stay OUT as optional caller policy. (Design: separate mechanism from policy.)

### 2. aiter harness JIT routing fix (`harness_generator.py`)
Auto-generated aiter harnesses set `GEAK_WORK_DIR` + `sys.path` but not `AITER_META_DIR`/`AITER_JIT_DIR`.
Because aiter is editable (sys.meta_path finder), `import aiter` resolves to the ORIGINAL repo and its
JIT compiles the BASELINE `csrc/*.cu` — patch-blind ~1.00x; on GEAK v3.2.2+ the validator rejects it and
the fallback hangs to the preprocess hard cap. Fix: both harness templates now
`os.environ.setdefault("AITER_META_DIR", WORK_DIR)` + `AITER_JIT_DIR=WORK_DIR/_aiter_jit` before
`import aiter`. Generic (any aiter kernel/workload), no-op for non-aiter, per-worktree JIT dir avoids
source pollution + parallel-slot collisions. (Issue #715.)

### 3. Per-kernel `extra_dispatch_context` → `## WORKLOAD CONTEXT (authoritative)` block (original #703)
Injects the surrounding workload context HL doesn't pass to GEAK — serving config (ISL/OSL/ctx/conc/TP),
E2E/Amdahl framing, neighbouring-kernel/fusion cues, resolved roofline. The serving config resolves the
decode-context ambiguity that otherwise forces harness-gen to GUESS the live shapes (the proven root
cause of kernel wins that don't transfer to E2E). Default-off / byte-identical when absent.

---

### 4. GEAK dispatch is PROMPT-ONLY (no test_command / pre-baked harness / op_test)
HL previously pre-baked a harness for GEAK and/or handed it an HL-resolved op_test via `--test-command`,
plus rendered a hardcoded op→test map into the prompt. That bypassed GEAK's own `preprocess_v3`
orchestrator and could mis-target/mis-shape the kernel (e.g. a hardcoded `test_pa.py` GEMM-flattened the
10-tensor paged-attention arg spec). The canonical baseline run dispatched with NO test_command and
GEAK's preprocess produced the correct harness + 1.2647x. So for `backend=="geak"` HL now passes ONLY
the dispatch prompt (device source + captured serving shapes + WORKLOAD CONTEXT) and lets GEAK's
preprocess discover/build the harness. Forge/OOB keep harness-gen (they have no preprocess of their own).
Commits: `bf6671e` (no pre-baked harness), `714344e`/`8db32d1` (no test_command), `44f9017` (no op_test
in prompt), `11af18b` (refuse to fabricate a GEMM harness for heterogeneous multi-tensor ops).

### 5. Autonomous combined-E2E after the GEAK batch (`kernel_request_handlers.py`)
Once `run_optimization`'s batch finishes optimizing all hot kernels, HL now applies ALL their best
patches together and remeasures end-to-end throughput **itself** — no manual `apply_and_bench` step —
attaching the result to `out["combined_e2e"]`. GEAK-only + opt-in: a no-op (byte-identical) unless
`payload["combined_e2e"]` is set AND the backend is geak AND a servable `model_path` + ≥1 GPU are
present (OOB/forge never trigger it). Applies each kernel's best microbench patch regardless of the
KEEP/REVERT verdict (the gate-less E2E A/B + engagement-proof are the arbiter), reuses the same
`apply_and_bench` primitive, runs off the event loop, and never raises into the batch result. Commit
`f2216a9`. A newcomer runbook is added at `docs/HOWTO_RUN_GEAK_V4_END_TO_END.md`.

## Validated end-to-end
- #703 context block fixed v4 attention's ctx-guess (18401 → real 1024/2048) → turned a 0.9995x no-win
  into a real 1.0096x win.
- `apply_and_bench` applied a v4 Llama quant patch (aiter `.cu` rebuild) → E2E 5300.6→5307.9 tok/s
  (+0.14%, an expected Amdahl-bound outcome on host-bound TP=1), source reverted clean.

## Tests
- `_build_extra_context_block({})`/whitespace → "" (byte-identical prompt proof).
- All harness-gen tests pass (80); rendered aiter harness now PASSES `find_aiter_routing_violation`.
- engagement-proof unit-checked on rebuild vs no-marker logs.

## Linked issues
Closes #715 (aiter AITER_META_DIR harness routing). Addresses #717 (gate-less apply + E2E remeasure).

## Breaking changes
None — additive; default-off context block, new standalone tool, harness env vars are no-ops off the aiter path.


iraj465:
Addressed the CodeQL / code-quality review comments on `apply_and_bench.py` in 7276b7f:
- Removed the unused `shutil` import.
- The three best-effort `except` blocks now carry an explanatory comment on why they are intentionally non-fatal, and the server-teardown handler is narrowed to `(ProcessLookupError, OSError)`:
  - log write: best-effort after the line already went to stdout;
  - server SIGTERM: superseded by the authoritative `pkill` sweep;
  - prebuilt `.so` removal: a non-removable `.so` is still forced to recompile via `AITER_REBUILD=1` + `apply_kernel_patch`'s jit cache invalidation.

No behavior change.

xiaofei-zheng:
I found two issues that should be addressed before merge:

1. `apply_and_bench.py` reintroduces a separate diff-apply path instead of reusing the snapshot deploy machinery from #681.

For `.diff`/`.patch` inputs, this PR applies patches with `git apply`, scopes multi-file diffs by target basename, and reverts using `git checkout --` over the `--numstat` touched paths. That makes the combined E2E result less trustworthy for real GEAK outputs:

- Multi-file optimizer outputs can include companion edits such as generated drivers, headers, or codegen shims. `_scope_diff_to_target()` can drop those sections, so the patched server may not be running the actual optimizer-produced patch set.
- New files, deletes, renames/copies, mode changes, and binary-ish cases are not reverted with the same guarantees as #681's manifest/snapshot path.
- This creates two patch deployment semantics in the codebase: #681's byte-exact atomic snapshot deploy for integration, and this PR's direct `git apply` flow for combined E2E.

Suggestion: make `apply_and_bench.py` use the same #681 `parse_patch_manifest` / `apply_snapshot` / `revert_kernel_patch` path, or otherwise thread the deploy snapshot captured during verification into combined E2E. The E2E A/B should apply the same byte-exact patch contents that integration would deploy.

2. #720 is not resolved by the current pre-run cleanup workaround.

The issue explicitly calls out that blind `pkill -f hipcc|ninja|clang++` is unsafe on shared nodes. The runbook / driver cleanup is useful for single-tenant experiment boxes, but it is not a tenant-safe startup reaper and could kill unrelated compiler or serving work on a shared node. Please keep #720 open and implement attribution-based cleanup (recorded GEAK pgid/session state, cwd/open build dir under this run's workspace, etc.) before treating this as fixed.

CI/docs note: current checks show CodeQL green but the ReadTheDocs status is failing, so that also needs to be accounted for before merge.

iraj465:
Thanks @xiaofei-zheng — both points addressed, plus the ruff + ReadTheDocs failures.

**1. Byte-exact deploy, one mechanism (was: separate `git apply` + scope path).** `apply_and_bench` no longer has a second deploy semantics. For a `.diff` it now:
- reconstructs the **byte-exact** optimized source for **every** file the diff touches by applying the **whole** diff (all hunks, all files — nothing scoped/dropped) in an isolated `git worktree` of the repo (live tree never mutated), then
- deploys each reconstructed file via `apply_kernel_patch` (backup + aiter `.cu` rebuild + cache invalidation + **manifest**) — the same full-source-replace path integration uses — with revert uniformly through `revert_kernel_patch(manifest)`.

`_scope_diff_to_target` / `_git_apply_diff` / the `git checkout` revert are removed. So companion file-sections (generated drivers/headers/codegen shims) are preserved, new/delete/rename go through the manifest path, and the combined-E2E now applies the exact bytes integration would deploy. (commit `4e0c3994`)

**2. #720 stays open; cleanup labeled single-tenant-only.** Agreed — the blind `pkill -f` sweep is not a tenant-safe reaper. I did not mark #720 resolved (it's still OPEN), and I've added explicit warnings in both `run_gvf.sh` and the docs that the sweep is **single-tenant experiment-box only**, must be skipped (`SKIP_PRERUN_CLEANUP=1`) on shared nodes, and that the production fix is the attribution-based reaper (recorded GEAK pgid/session, build-dir-scoped) tracked in #720. (commit `dfd05176`)

**CI:** ruff `E401/E402/E702` + `ruff format` fixed (`bc80180d`); ReadTheDocs was failing because the new HOWTO was orphaned (not in any toctree) + used an unknown `jsonc` Pygments lexer + a stray trailing fence — fixed and verified locally with `sphinx -b html` → `build succeeded`, 0 warnings (`dfd05176`).

iraj465:
**ReadTheDocs status note (needs a maintainer with RTD dashboard access).**

The RTD check is red, but I believe the **doc content is correct** and the failure is **environmental on RTD's side**, not in the markdown:

- I reproduced the build locally with the repo's exact doc deps (`docs/requirements.txt`: sphinx≥7.2 / furo / myst-parser / linkify) and the **strictest** flags `sphinx -b html -W --keep-going` → **`build succeeded`, 0 warnings**.
- The original RTD failure causes I *could* see and fixed: the new HOWTO was orphaned (added it to the `How-to Guides` toctree in `docs/index.rst`), an unknown `jsonc` Pygments lexer (→ `json`), and a stray trailing code fence. (`dfd05176`)
- RTD did trigger a fresh build after the fix (`4148994` ≠ the pre-fix `4148974`) but it still fails for a reason that does **not** reproduce locally. The RTD build log requires dashboard auth, which I don't have, so I can't see the specific env error.

Could a maintainer either share the `4148994` build log (so I can fix the exact env issue) or confirm RTD is informational here? It does not appear in `main`'s required status checks, and #710/#681 pass it, so this looks specific to this PR's RTD build environment rather than the doc itself. Happy to fix immediately once the log is visible.

xiaofei-zheng:
Thanks for the follow-up. The previous diff-scoping concern is largely addressed now: `apply_and_bench.py` no longer drops non-target sections, and the whole diff is reconstructed in an isolated worktree before deploying reconstructed files through `apply_kernel_patch()` / manifest-based revert.

I still see a few issues to address before merge:

1. `apply_and_bench.py` still uses a global server teardown:

```python
pat = "sglang.launch_server" if backend == "sglang" else "vllm.entrypoints"
subprocess.run(["pkill", "-9", "-f", pat], check=False)
```

This can kill unrelated sglang/vLLM servers on the same node. Because autonomous `combined_e2e` calls this same primitive, this is not limited to a manual single-tenant experiment script. Please make server cleanup process-scoped: launch the server in its own process group/session and kill only the spawned process tree. If a broad `pkill` fallback remains, it should be explicit single-tenant opt-in, not default behavior.

2. The new "byte-exact deploy" path still does not fully preserve all patch operations.

The worktree reconstruction is a good improvement for normal modify/add source patches, but pure deletes are currently skipped (`files[rel] = ""` then no deploy), so the live tree keeps files the optimizer patch deleted. Rename/mode-only/binary-style changes also are not equivalent to #681's manifest/snapshot semantics when converted to full-source replacement. If the intent is to claim the E2E A/B applies the exact optimizer patch set, please either reuse #681's `parse_patch_manifest` / `apply_snapshot` / `revert_kernel_patch` snapshot path directly, or add explicit support/tests for delete, rename/copy, chmod/mode-only, and binary cases.

3. Latest-head CI is not ready yet.

At head `7881d481`, I only see `Analyze (python)` passing and the ReadTheDocs status failing. The recent `Tests with Coverage` and `Lint` successes are for older head SHAs and do not cover the later `apply_and_bench` rewrite / docs commits. Please rerun lint + test on the current head and get the RTD status resolved or confirmed non-required with logs.

So the direction looks much better than the earlier version, but I would still hold merge until the server teardown is process-scoped and latest-head CI is green.

iraj465:
Thanks @xiaofei-zheng — all three addressed in `7f28ada1`.

**1. Process-scoped server teardown (no global `pkill`).** The server is now launched with `start_new_session=True` (its own process group), and `_kill_servers` does `os.killpg(pgid, SIGTERM)` → grace → `SIGKILL` on **only that group** — the server tree we spawned. The global `pkill -f sglang/vllm` is gone from the default path; it remains only as an explicit single-tenant opt-in via `APPLY_BENCH_PKILL_SWEEP=1`. Since autonomous `combined_e2e` calls this same primitive, this makes it multi-tenant-safe by default, not just for the manual script.

**2. Patch-op coverage — refuse what full-source replace can't represent.** You're right that delete/rename/copy/mode/binary aren't faithfully expressible as full-source replacement, and the prior silent delete-skip was wrong. Rather than half-apply, `_diff_unsupported_ops` now parses the diff headers and reconstruction **fails loudly** when any of `delete / rename / copy / mode-change / binary` is present, with a message pointing to the integration snapshot path for those cases. A kernel-optimization patch should be modify/add only; if it isn't, the E2E refuses instead of measuring a half-applied tree. (The earlier `files[rel]=""` delete-skip and its downstream branch are removed.)

New `kernel-agent/tools/test_apply_and_bench.py` (7 tests) covers the op gate (modify/add allowed; delete/rename/copy/mode/binary refused; dedup+sorted) and the spread/significance helper — runs without serving/GPU.

**3. Latest-head CI.** Pushed `7f28ada1`; this re-triggers lint + tests on the current head (locally: `ruff check` + `ruff format --check` clean, 159 tests green incl. the new file). RTD: the doc content builds clean locally (`sphinx -b html -W --keep-going` → succeeded, 0 warnings); the remaining red is the environmental RTD-side issue I flagged above — still need a maintainer to share build log `4148994` or confirm it's non-required (it's not in `main`'s required checks; #710/#681 pass it).

iraj465:
Follow-up on the latest-head CI (item 3), with specifics:

- **Lint / Tests-with-Coverage are not auto-triggering on this PR's pushes** (only `Analyze (python)` + ReadTheDocs post statuses here). They run fine on other PRs, so it appears to be a `paths-ignore` / trigger quirk on this branch rather than a failure — and both `lint.yml` ruff/pylint steps are `continue-on-error: true` (advisory), so they don't gate merge regardless.
- I reproduced both locally on the current head `7f28ada1`:
  - `ruff check .` → the only 5 findings are **pre-existing**, in files this PR does not touch (`intent_router.py`, `reference_script.py`, `backends/forge_submit.py`, `test_faithful_dtypes.py`); **all of my changed files are ruff-clean**.
  - `ruff format --check` → the one reformat suggestion (`kernel_request_handlers.py`) is in **pre-existing** lines (1242/1343/1692), not my `combined_e2e` additions; the file was already format-dirty on `main`. My new/edited code is format-clean.
  - `pytest` on the touched suites (apply_and_bench, harness_generator, kernel_request_handlers) → **237 passed**, including the new `test_apply_and_bench.py` (7 tests).

If a maintainer can kick the Lint/Tests-Coverage workflows on this head (or confirm the paths-ignore trigger behavior), they should come back green for this PR's code. RTD remains the one item needing dashboard access (env-side; doc builds clean locally with `-W`).

iraj465:
Pushed `95caa206`:

- **Empty-except (CodeQL/code-quality) fixed:** the process-scoped teardown's best-effort `except` blocks (killpg SIGTERM/SIGKILL already-gone, the non-POSIX kill race, the final reap, and the bench-parse drop) now each carry an explanatory comment — behavior unchanged, no more "empty except" findings. (`bb8e72d6`)
- **Merged latest `main`** into the branch (was 113 behind). One conflict in `kernel_optimization.py` at the GEAK dispatch site: kept the prompt-only behavior (no test_command/op_test copy) **and** adopted main's updated `_apply_geak_env_overrides(args, prompt_file, candidate)` signature. Branch is now 0 behind main. (merge `95caa206`)
- **Local verification on the merged head:** `ruff check` clean on all changed files; **242 tests pass** (handlers + apply_and_bench + harness_generator). pylint isn't available in my env, but ruff's error classes are clean.

Net for the three round-3 items: process-scoped teardown ✅, patch-op coverage gate + tests ✅, and the branch is now current with main with the empty-except findings cleared. RTD remains the one maintainer-access item.

xiaofei-zheng (approved):

Issue #709, opened 2026-06-25
Primus-Safe/Vertex proxy rejects non-streaming Claude calls (400 INVALID_ARGUMENT): GEAK + ProposalScorer fail 100%


sdubagun-amd:
## Summary

The AMD Primus-Safe / VertexGenAI LLM proxy **rejects any non-streaming prediction** to the Claude deployment with a generic `400 BadRequest` / `INVALID_ARGUMENT`. Components that call the proxy **without `stream=True`** fail 100% of the time; components that stream (the Claude Agent SDK specialists) work fine.

This affects **two modules today**:
- **GEAK** (kernel-agent, litellm, non-streaming) — every run dies in preprocess and produces **no patch**.
- **ProposalScorer** (`inference_optimizer/orchestrator/proposal_scorer.py:315`, non-streaming `chat.completions.create`) — fails on **all** panel models (`claude-opus-4-8`, `Kimi-K2.6`, `gemini-3.1-pro`).

Confirmed across **both deployments** (`1b028c…` and `d2d10c…`).

## Error signature

```
litellm.BadRequestError: AnthropicException - {"llmService":"VertexGenAI",
 "deploymentInfo":{"id":"claude-opus-4-8@default","version":"vertex-2023-10-16"},
 "statusCode":"BadRequest",
 "response":{"error":{"code":400,
   "message":"Prediction on deployed model (endpoint_id: ..., deployed_model_id: ...) failed with error: \"Bad Request\".",
   "status":"INVALID_ARGUMENT"}}}
```
Fails instantly (`elapsed ~00:00:00.3`) — a request-validation reject, not a timeout/rate-limit/context overflow.

## Root cause (reproduced)

A standalone repro against the proxy (`POST .../llm-proxy/v1/chat/completions`, model `claude-opus-4-8`) flipping one field at a time:

| Variant | Result |
|---|---|
| `max_tokens=16384` (GEAK params) | **400** |
| `max_completion_tokens=4096` (ProposalScorer params) | **400** |
| no token cap | **400** |
| `max_tokens=1024` (small) | **400** |
| **`max_tokens=16384` + `stream:true`** | **200 ✅** |
| `/v1/messages` native, non-streaming | **400** |
| `openai/`-prefixed model name | 400 (separate "Invalid model name" error) |

[...]

PR #710, opened 2026-06-25
fix(proposal-scorer): stream LLM calls to satisfy Primus-Safe proxy

sdubagun-amd:
## Summary

- Fixes ProposalScorer's 100% failure rate against the Primus-Safe / VertexGenAI proxy.
- The proxy rejects **non-streaming** Claude predictions with an opaque `400 BadRequest / INVALID_ARGUMENT`; only `stream:true` requests return 200. ProposalScorer's `chat.completions.create` was non-streaming, so every panel model (`claude-opus-4-8`, `Kimi-K2.6`, `gemini-3.1-pro`) failed.
- Streams the request, accumulates `delta.content`, and reads usage from the final `include_usage` chunk.

Proof-of-fix for #709. The same root cause also affects **GEAK** (litellm path), addressed separately.

## Root cause (reproduced against the live proxy)

| chat/completions variant | Result |
|---|---|
| `max_tokens=16384` / `max_completion_tokens=4096` / no cap / small cap | **400** |
| **same + `stream:true`** | **200** |
| `/v1/messages` native, non-streaming | 400 |

`stream:true` is the only field that yields 200 — independent of model id (this is why PR #699's `4-7`->`4-8` bump didn't fix it) and token cap.

## Test plan

- [x] Unit tests updated to emulate streaming responses; `test_proposal_scorer.py` + `test_proposal_scorer_coverage_unit.py` pass (32 passed).
- [x] Streaming pattern verified end-to-end against the live proxy (full text accumulated, usage captured from final chunk).
- [ ] Observe a live optimizer run and confirm `ProposalScorer: model=... failed: BadRequestError` warnings stop.

## Note

`asyncio.wait_for` now wraps stream creation rather than the full body read; the per-call timeout still applies to connect/first-chunk. Acceptable for this fix; a full-read deadline can be added later if needed.

### DIFF
--- a/inference_optimizer/orchestrator/proposal_scorer.py
+++ b/inference_optimizer/orchestrator/proposal_scorer.py
@@ -311,29 +311,37 @@ class ProposalScorer:
         full_prompt = f"{prompt}\n\n{_SCORING_INSTRUCTIONS}"
         messages = [{"role": "user", "content": full_prompt}]
         _t0 = time.perf_counter()
+
         # The Primus-Safe/Vertex proxy rejects non-streaming predictions with an
         # opaque 400 INVALID_ARGUMENT; only streamed requests are accepted. Stream
         # and accumulate the deltas, pulling usage from the final chunk.
-        text_parts: list[str] = []
-        usage = None
-        try:
-            stream = await asyncio.wait_for(
-                client.chat.completions.create(
-                    model=model,
-                    messages=messages,
-                    max_completion_tokens=self.max_completion_tokens,
-                    stream=True,
-                    stream_options={"include_usage": True},
-                ),
-                timeout=self.call_timeout_s,
+        #
+        # The deadline must cover BOTH stream creation and the chunk-consumption
+        # loop: a proxy can establish the stream and then stall mid-body, so a
+        # timeout around creation alone would let this hang indefinitely. We wrap
+        # the whole read in a single ``asyncio.wait_for`` (``asyncio.timeout`` is
+        # 3.11+, and this project still targets 3.10).
+        async def _read_stream() -> tuple[list[str], Any]:
+            parts: list[str] = []
+            usage_obj = None
+            stream = await client.chat.completions.create(
+                model=model,
+                messages=messages,
+                max_completion_tokens=self.max_completion_tokens,
+                stream=True,
+                stream_options={"include_usage": True},
             )
             async for chunk in stream:
                 if getattr(chunk, "usage", None) is not None:
-                    usage = chunk.usage
+                    usage_obj = chunk.usage
                 if chunk.choices:
                     delta = chunk.choices[0].delta
                     if delta is not None and delta.content:
-                        text_parts.append(delta.content)
+                        parts.append(delta.content)
+            return parts, usage_obj
+
+        try:
+            text_parts, usage = await asyncio.wait_for(_read_stream(), timeout=self.call_timeout_s)
         except asyncio.TimeoutError as exc:
             raise RuntimeError(f"timed out after {self.call_timeout_s:.0f}s") from exc
         latency_ms = int((time.perf_counter() - _t0) * 1000)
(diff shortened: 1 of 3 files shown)


chennyiiis:
Thanks for the quick turnaround on this — the root-cause analysis is clear and the fix lines up nicely with the repro table in #709. The streaming parse looks robust (`if chunk.choices:` correctly handles the usage-only final chunk, and skipping empty `delta.content` is sensible). Sharing a few observations for consideration before merge:

### High priority

**1. `asyncio.wait_for` no longer covers the full response read.**
After the change, the timeout only wraps `chat.completions.create()` (stream creation), not the `async for chunk in stream:` body. Previously the 120s budget covered the entire request including the body. With the new shape, if the connection is established but chunks stall (e.g. a slower reasoning model like Kimi-K2.6), the call could hang indefinitely. The PR description acknowledges this, but for the production scorer (`call_timeout_s=120`) it is a behavioral regression. Could we wrap the whole `async for` loop with `asyncio.timeout()` (3.11+) or otherwise apply a first-chunk/last-chunk deadline?

### Medium priority

**2. Scope of #709 — GEAK is still affected.**
This PR resolves the ProposalScorer caller, but #709 also lists GEAK (litellm / `$GEAK_CONFIG`) as failing 100% with no patches produced. After merge the ProposalScorer warnings will stop, but the GEAK zero-output issue remains. Could we track the GEAK streaming fix as a follow-up (or note the split/closing criteria on #709)?

**3. `stream_options={"include_usage": True}` has no fallback.**
Token tracing depends on the final usage chunk. If a panel model does not honor `include_usage`, `_trace_scorer_llm_call` silently receives `usage=None` and writes a null-token row (functionally harmless, but trace data is lost). A short note documenting the degradation behavior would help future readers.

**4. Tests do not assert `stream=True`.**
The fake client now accepts the `stream` kwarg, but no test asserts `calls[0]["stream"] is True`. A revert of the streaming flag would still pass the suite — adding that assertion would lock in the fix.

### Low priority

**5. Multi-chunk accumulation is not exercised.**
`_FakeStream` emits a single content chunk plus one usage chunk. The real proxy may split content across several deltas; `"".join(text_parts)` is correct, but a 2+ content-chunk case would guard against regressions.

**6. Coverage-unit fake stream returns `usage=None`.**
`test_proposal_scorer_coverage_unit.py`'s fake stream does not emit a final `include_usage` chunk, so it diverges from the main path and cannot validate the token-row write.

### Not blocking (already flagged in #709)
`CriticAgent` (`critic_agent.py:1232,1269`) and `CodexBackend` (`codex.py:205`) share the same non-streaming pattern. They route to a codex/GPT endpoint today so they are unaffected, but would break if ever repointed at the streaming-required proxy — worth a hardening follow-up.

Overall this looks good to me — approving in spirit, with the timeout coverage (item 1) being the main thing I would suggest addressing before merge. Thanks again!

xiaofei-zheng:
Thanks for fixing the streaming-required proxy path. The change is directionally right, but I think one issue should be addressed before merge:

`asyncio.wait_for()` now only wraps `client.chat.completions.create(...)`, not consumption of the stream body. If the proxy establishes the stream and then stalls while yielding chunks, `_score_one_model()` can hang indefinitely. Previously the timeout covered the full non-streaming request/response.

Please wrap the full streaming read in a deadline, for example with `asyncio.timeout(self.call_timeout_s)` around both stream creation and the `async for chunk in stream` loop (or an equivalent 3.10-compatible helper if needed). It would also be good to add tests that assert:

- `stream=True` is actually passed.
- Multiple content chunks are accumulated correctly.
- A stalled stream body times out.
- The final `include_usage` chunk is handled when present, and missing usage degrades cleanly.

This PR fixes ProposalScorer only; the GEAK/litellm streaming-required proxy issue should stay tracked separately if #709 covers both call sites.

xiaofei-zheng (approved):

PR #727, opened 2026-06-26
fix(gemm-tuning,explore): per-model GEMM shapes, runtime-correct fp8 quant routing, and stop overtime-killing capture-bound explore variants

xiaofei-zheng:
## Summary

Three fixes that unblock real optimization gains, found while investigating why ~half of the daily model-leaderboard runs (heavily small <=2B models) ended at zero gain.

### 1a. GEMM tuning tuned the wrong shapes
`_resolve_forge_untuned_csv` picked a specialist-worktree untuned CSV purely by filename + mtime. In practice that is AITER's checked-in default (`a8w8_blockscale_untuned_gemm.csv`, DeepSeek shapes, K=7168), so dense fp8 sglang sessions tuned shapes unrelated to the model and always reported `improved=0/0, best_speedup=1.000x` while burning ~15 min GPU/session. Now each candidate CSV is validated against the model `hidden_size` (the contraction dim of QKV/gate-up/o, invariant under TP); on mismatch it is skipped so forge derives correct per-model shapes from `config.json`.

### 1b. fp8 quant_type routed to the wrong tuner
Non per-token fp8 defaulted to `quant_type="auto"`, which forge resolves to `a8w8_blockscale`. Blockscale is only valid for checkpoints that ship block-quantized weights. A plain fp16/bf16 model under dynamic `--quantization fp8` uses the per-token a8w8 path. Added `_model_is_block_quantized` (reads `config.json` `quantization_config`) and route: block-quantized -> `blockscale`, otherwise known model -> `per_token`, unknown -> `auto`.

### 2. Explore overtime-kill discarded throughput-positive variants
The decision-round soft deadline anchored on the WARM (client-only) baseline time, which excludes server boot + cuda-graph capture. Config-changing variants (cuda-graph / attention-backend / quantization / kv-cache-dtype ...) cannot reuse the warm server, so their decision round re-captures; wall-clock exceeds `warm_anchor * 1.2` and is `KILLED_OVERTIME` before throughput is measured. On small models, 95% of overtime kills were actually >= baseline (median +35%). `_compute_decision_deadline_sec` now adds the baseline's own cold-minus-warm boot+capture overhead as a conservative allowance; the hard `variant_timeout_sec` cap still bounds genuine runaways.

## Test plan
- [x] `test_kernel_request_handlers_units.py`: untuned-CSV model match/mismatch, legacy fallbacks, fp8 quant routing (plain -> per_token, block -> blockscale, unknown -> auto), `_model_is_block_quantized`
- [x] `test_explore_executor.py`: `_compute_decision_deadline_sec` adds boot+capture allowance, legacy cold mode, disabled, and warm>=cold guard
- [x] Offline-validated forge derives correct per-model shapes (K=2048/8192) once the generic CSV is rejected
- [ ] Note: async executor integration tests require pytest-asyncio (not installed in this env); all added tests are synchronous


ZhengGong-amd (approved):
validated changes. approved

PR #736, opened 2026-06-26
fix(quantize): preserve source model identity after quantization prelude

BaoYunkai:
## Summary
- After the quantization prelude rewrites `--model` to the generic export dir
  `<workspace>/quantization/<model>/quantized` (basename always `quantized`),
  the three naming sites (session dir via `make_session_dir`,
  `SharedState.model_name`, `manifest.json` model_name) all derived the name
  from that basename — collapsing every quantized run to `quantized`. The real
  model name was lost, reports showed `Model: quantized`, and concurrent
  quantized models collided under `<root>/quantized/<ts>`.
- Pin the source identity on `args.model_display_name` in the prelude
  (`<source>-quantized`) and add `resolve_model_display_name()` so all three
  sites prefer it, falling back to the path basename for non-quantized runs.
  `model_path` still records the real export dir unchanged.

## Test plan
- [x] New tests reproduce the collapse and verify the fix (prelude identity,
      `_seed_shared_state`, manifest builder, helper).
- [x] `test_quantization_prelude.py` + `test_cli_bootstrap_coverage_unit.py`
      green; regressions in `test_session_layout` / `test_resume` /
      `test_shared_state_*` / `-k manifest` green.

### DIFF
--- a/inference_optimizer/cli.py
+++ b/inference_optimizer/cli.py
@@ -116,6 +116,7 @@ from .cli_bootstrap import (  # noqa: F401 - re-exported for callers/tests
     _resolve_session_dir_for_summary,
     _seed_shared_state,
     _snapshot_system_prompts,
+    resolve_model_display_name,
 )
 from .orchestrator.action_executors._aiter_jit import (
     AITER_LOCK_STALE_MINUTES,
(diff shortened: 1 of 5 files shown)


BaoYunkai:
Superseded by #739 (consolidated into a single PR).

PR #737, opened 2026-06-26
feat(quark): param-gated quark_quantizer shell over quantization_agent

BaoYunkai:
## Summary
- Add `quark_quantizer`, a thin shell over the existing `quantization_agent`.
  Its only addition is a **parameter-based trigger**: callers pass `enabled`
  (default `False`) to decide whether quantization runs, so an outer
  orchestration LLM has no natural-language path to accidentally activate it.
- When enabled, the shell forwards the NL prompt verbatim to
  `quantization_agent.quantize_via_prompt` (unchanged — it still turns the
  prompt into the Quark CLI and executes); when disabled it is a no-op.
- Repoint the inference_optimizer prelude adapter to call the shell
  (`enabled=True`, since reaching the prelude already means
  `--quantize`/`--quantize-scheme` was set). Prelude position/order, partial
  handling, and `SystemExit(3)` policy are unchanged. `quantization_agent` is
  kept intact.

## Test plan
- [x] Offline shell tests (fake agent seam): disabled no-op, default-disabled,
      delegation + result normalization, sync wrapper, CLI.
- [x] Adapter/prelude tests + the kept `quantization_agent` tests green.

### DIFF
--- a/inference_optimizer/orchestrator/quantization_request_handlers.py
+++ b/inference_optimizer/orchestrator/quantization_request_handlers.py
@@ -1,19 +1,24 @@
-"""Adapter: drive the quantization-agent from inference_optimizer.
+"""Adapter: drive quantization from inference_optimizer via the quark_quantizer shell.
 
-Thin shim between ``cli._run_quantization_prelude`` and the standalone
-``quantization_agent`` package. It builds an effective prompt (source model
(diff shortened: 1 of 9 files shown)


BaoYunkai:
Superseded by #739 (consolidated into a single PR).

PR #768, opened 2026-06-29
feat: honest E2E verdict (paired A/B, promotion) + faithful dispatch (dominant-kernel op-source, real dims, comm-aware select)

iraj465:
## What
Honest E2E verdict + faithful dispatch. All flags **default-off** ⇒ byte-identical until enabled.

## Verdict (E2E A/B)
- `HL_INTEGRATE_PAIRED_AB` / `HL_HONEST_E2E`: replace the **unpaired stored-scalar** `base_tput` with a **paired same-config** measurement (revert patch → re-run the same `BaselineExecutor(ctx)` on the pristine kernel); gain computed from the pair; KEEP re-applies; pairing failure ⇒ NEEDS_REVIEW.
- Generalized **engagement** gate (target bytes must change, content-hash) + **noise band** (`HL_E2E_NOISE_BAND_PCT`); un-engaged KEEP → NEEDS_REVIEW.
- `HL_PROMOTE_VERIFIED_MICRO`: a correctness-passing verified-micro NEEDS_REVIEW can reach the E2E measurement (rank tier) instead of always losing to any KEEP.

## Faithful dispatch (op→source + shapes)
- **Data-driven dominant-kernel resolution**: `load_op_dominant_kernel_map` parses per-device-kernel durations (`kernel_details_summary`) and pins the composite op's single hot source via the existing resolver. Collapses the fan-out generally, both frameworks (verified: `fused_moe`→`fused_moe_triton_kernels.py`, `gdn_attention_core`→`chunk_delta_h.py`). Name-anchored fallback when `kernel_details` is absent.
- Use **real `torch_trace` dims** for benchmark cases even when values are synthesized (was discarded; 298/298 candidates were flagged synthetic).
- `HL_E2E_AWARE_SELECT`: dispatch by `gpu_pct × headroom` and **deprioritize pure-communication** kernels (all_reduce/collectives can't be sped by a kernel rewrite).

Tests: 26 resolver + 96 kernel-opt + 9 shape pass. `py_compile` clean.

### DIFF
--- a/inference_optimizer/orchestrator/kernel_request_handlers.py
+++ b/inference_optimizer/orchestrator/kernel_request_handlers.py
@@ -98,6 +98,14 @@ def _vram_guarded_server_args(extra_args: str) -> str:
     """
     if not _honest_flag("HL_INTEGRATE_VRAM_GUARD"):
         return extra_args
+    # ``--gpu-memory-utilization`` is a vLLM-only flag; sglang REJECTS it
+    # ("unrecognized arguments") and the server exits immediately, failing the
+    # integrate re-baseline for EVERY sglang workload. sglang caps memory via
+    # ``--mem-fraction-static`` (set by its launcher), so it needs no guard here.
+    # Apply the cap only for vLLM; sglang/unknown framework -> strict no-op.
+    framework = (os.environ.get("FRAMEWORK") or "").strip().lower()
+    if framework != "vllm":
+        return extra_args
     if "gpu-memory-utilization" in (extra_args or ""):
         return extra_args
     try:
--- a/inference_optimizer/orchestrator/kernel_request_handlers.py
+++ b/inference_optimizer/orchestrator/kernel_request_handlers.py
@@ -59,8 +59,11 @@ def _honest_flag(specific_env: str) -> bool:
 
     Returns ``True`` when the per-fix env ``specific_env`` is truthy, OR when it
     is unset and the umbrella ``HL_HONEST_E2E`` is truthy. An explicit falsey
-    per-fix value always wins (lets one fix opt out of the umbrella). Off by
-    default so callers are byte-identical to legacy behavior when nothing is set.
+    per-fix value always wins (lets one fix opt out of the umbrella). The umbrella
+    now defaults ON: E2E-faithful verdicts (paired same-config A/B, engagement
+    gate, verified-micro promotion) are the default so a GEAK kernel win is judged
+    by real serving throughput, not a stored-scalar gain. Opt the whole cohort
+    back out with ``HL_HONEST_E2E=0`` (or a single fix via its per-fix env).
 
     Args:
         specific_env: The per-fix environment variable name.
@@ -73,7 +76,7 @@ def _honest_flag(specific_env: str) -> bool:
         return True
     if raw in _FALSEY:
         return False
-    return os.environ.get(_HONEST_E2E_UMBRELLA_ENV, "").strip().lower() in _TRUEY
+    return os.environ.get(_HONEST_E2E_UMBRELLA_ENV, "1").strip().lower() in _TRUEY
 
 
 def _vram_guarded_server_args(extra_args: str) -> str:
--- a/inference_optimizer/orchestrator/kernel_request_handlers.py
+++ b/inference_optimizer/orchestrator/kernel_request_handlers.py
@@ -3704,13 +3704,21 @@ def _run_combined_e2e_sync(
         return {"status": "error", "error": f"apply_and_bench import failed: {exc}"}
 
     cfg = _combined_e2e_serving_config(payload)
+    # Expose exactly the GPUs the serving TP needs. apply_and_bench sets
+    # HIP/ROCR_VISIBLE_DEVICES=gpu, so a hardcoded "0" with tp>1 makes the server
+    # request device ordinals that aren't visible -> "HIP error: invalid device
+    # ordinal" and a baseline_failed A/B. Map the first `tp` visible GPUs (clamped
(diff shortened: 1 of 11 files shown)


xiaofei-zheng:
Blocking issue: Dynamo SSH placement does not get the new GEAK defaults.

This PR changes the Ray/CLI GEAK paths to default to kernel scoring and skip the advisory profiler (`GEAK_SCORE_TARGET=kernel` / `GEAK_SKIP_PROFILE=1`), but `geak_submit.submit()` bypasses those paths entirely when `KERNEL_AGENT_GPU_PLACEMENT=ssh` and calls `run_geak_over_ssh()`.

The pod-side runner in `kernel-agent/tools/backends/ssh_runtime.py` reconstructs the GEAK command without adding `--target kernel`, and it also never sets `GEAK_SKIP_PROFILE=1` by default. That means Dynamo multi-node runs still use the old GEAK scoring/profiler behavior, which bypasses one of the main goals of this PR.

Suggested fix: mirror the Ray/CLI behavior inside `_POD_RUNNER`: set `os.environ.setdefault(GEAK_SKIP_PROFILE, 1)`, read `GEAK_SCORE_TARGET` with default `kernel`, and append `--target <wall|kernel>` when valid. Please add an SSH-runtime regression test that proves the pod-side command includes `--target kernel` by default and still respects explicit overrides.

iraj465:
Good catch — fixed in 819d2f90.

The SSH placement path now mirrors the Ray/CLI GEAK defaults inside the pod-side `_POD_RUNNER` (`ssh_runtime.py`):
- `os.environ.setdefault("GEAK_SKIP_PROFILE", "1")` — advisory profiler off by default on the Dynamo pod.
- reads `GEAK_SCORE_TARGET` with default `kernel` and appends `--target <wall|kernel>` to the reconstructed geak command.

Both keys are already in `SAFE_ENV_KEYS` (`ray_runtime.py:348-349`), so an explicit orchestrator-side `GEAK_SCORE_TARGET` / `GEAK_SKIP_PROFILE` is forwarded over SSH via `_env_prologue()` and wins via `setdefault` / the env read — i.e. defaults applied, explicit overrides respected.

Added the requested SSH-runtime regression tests (`test_ssh_runtime.py`) that **execute the real `_POD_RUNNER`** with a fake `geak` on PATH and assert:
- default → pod command contains `--target kernel` and the geak child sees `GEAK_SKIP_PROFILE=1`;
- explicit `GEAK_SCORE_TARGET=wall` / `GEAK_SKIP_PROFILE=0` → command contains `--target wall` and `SKIP=0`.

`pytest test_ssh_runtime.py` → 9 passed. (Note: the OOB pod runner at the second GPU-env block is intentionally untouched — it doesn't dispatch GEAK.)

xiaofei-zheng:
Blocking issue remains in `parallel_e2e_runner.py`: the GEAK budget was raised to 180 min, but the outer subprocess timeout still uses `backend_budget_min` (default 60 min), so the runner can still kill `kernel_optimization.py` long before GEAK's own `--geak-budget-min` expires.

Current code passes the GEAK-specific budget into the child:

```python
"--geak-budget-min",
str(args.geak_budget_min),
```

but then wraps the child with:

```python
result = run_json(
    cmd,
    env=local_env,
    timeout_s=int(args.backend_budget_min * 60) + 120,
    log_path=log_path,
)
```

With defaults, `backend_budget_min=60` means the outer timeout is ~62 minutes, even though `geak_budget_min=180`. That still reproduces the failure mode this PR is trying to eliminate: GEAK gets hard-killed before its own budget-aware wind-down/final artifact write, so deploy patch collection and combined E2E can be skipped.

Suggested fix: choose the wrapper timeout budget by backend, e.g.

```python
attempt_budget_min = args.geak_budget_min if backend == "geak" else args.backend_budget_min
result = run_json(cmd, env=local_env, timeout_s=int(attempt_budget_min * 60) + 120, log_path=log_path)
```

and add a regression test that `backend="geak"` uses `geak_budget_min` for `run_json(timeout_s=...)`, while OOB backends still use `backend_budget_min`.

I also checked `origin/main`: the previous GEAK full default here was 130 min, not 120 min. The 120 number appears to be the full-mode threshold mentioned in `kernel_optimization.py` (`budget_min >= 120`). Either way, the fix should not depend on the old value; the outer wrapper should use the effective GEAK budget it forwards to the child.

Non-blocking cleanup: the header comment in `kernel_request_handlers.py` still says "Every behavior below is OFF by default", but `_honest_flag()` now defaults the umbrella `HL_HONEST_E2E` to ON. Please update that comment to avoid misleading future readers.

iraj465:
Good catch — fixed in `0083b1d5`.

The outer `run_json()` wrapper was using `backend_budget_min*60 + 120` (~62 min) as its subprocess timeout regardless of which backend ran, so after the 130→180 change it would SIGKILL `kernel_optimization.py` ~62 min in — before GEAK (180 min) finalizes its deploy artifact, which then skips the combined-E2E/integrate A/B. (This is the same failure mode that the salvage/integrate validation hit on GLM-4.7.)

Fix: size the outer timeout to the **actual backend's** budget, plus GEAK's finalize grace:
```python
_effective_budget_min = args.geak_budget_min if backend == "geak" else args.backend_budget_min
result = run_json(cmd, env=local_env, timeout_s=int(_effective_budget_min * 60) + 360, log_path=log_path)
```
`+360s` ≥ GEAK's `finalize_grace (300s) + kill_buffer (60s)`, so the outer wrapper is always strictly larger than the inner geak_submit subprocess timeout (`geak_budget_min*60`). OOB backends are unchanged (still `backend_budget_min`).

xiaofei-zheng (approved):

PR #772, opened 2026-06-30
fix(recipe-kb): repair gbrain SSE de-framing + version-agnostic get_recipe fallback

zhanglei-amd:
## Summary
recipe-KB composite `[gbrain, cortex]` was under-using the gbrain source. Two read-path fixes:

- **gbrain SSE de-framing** (`gbrain_remote_client.py`): parse `text/event-stream` per spec — join all `data:` lines of an event and stop on the blank-line boundary, instead of taking only the first `data:` line / breaking on a bare `}`. The old parser truncated large `get_page` bodies into a `JSONDecodeError` ("bad envelope") and skipped gbrain on every run (large pages on gbrain >= 0.41 are chunked across multiple `data:` lines).
- **version-agnostic `get_recipe` fallback** (`composite_remote.py`): when the version-exact label search misses, retry once with `framework_version` dropped, so a same-model recipe stored under a different version (cortex pins e.g. `0.5.11` while gbrain mirrors the on-disk `0.5.12`, or vice versa) still seeds warm-start. warm-replay's own reproduce/confidence gate validates the borrowed config.

## Why
- gbrain holds the full recipe corpus, but the SSE bug dropped it from the composite for any large page → warm-start fell back to cortex only.
- cortex/gbrain canonical_ids drift on the `framework_version` segment, so a version-exact cross-KB match misses even when both hold the same model (measured ~34% of cortex models exist in gbrain only under a different version).

## Test plan
- New regression tests in `test_composite_remote.py`: version-drift fallback recovers the sibling; no version-agnostic retry fires when the version-exact search already hits.
- Extended `test_gbrain_remote_recipe_client_coverage_unit.py` with multi-`data:`-line SSE coverage.
- `python -m pytest inference_optimizer/tests/test_composite_remote.py inference_optimizer/tests/test_composite_remote_unit.py` → 27 passed.
- E2E: 1-GPU Llama-3.1-8B claw sessions pinned to this branch — no `bad envelope`, warm-start `_field_sources.best_config = gbrain` (gbrain contributes), KG §5e graph-guided knobs populated.

### DIFF
--- a/inference_optimizer/recipe_kb/dispatcher.py
+++ b/inference_optimizer/recipe_kb/dispatcher.py
@@ -775,6 +775,31 @@ class RecipeKB:
                         )
                     )
                     return ranked[0]
+                # Framework version drift should not hide otherwise reusable
+                # recipe pages. Retry without that single dimension only after
+                # both the exact slug and full label-match paths miss.
+                relaxed_labels = dict(labels)
+                relaxed_labels.pop("framework_version", None)
+                rows = self.remote.search(  # type: ignore[union-attr]
+                    label_match=relaxed_labels,
+                    limit=candidate_limit,
+                    prefer=prefer,
+                )
+                if rows:
+                    normalized_rows = [self._normalize_remote_row(r) for r in rows]
+                    ranked = _rerank_by_prefer(normalized_rows, prefer)
+                    self._emit_audit(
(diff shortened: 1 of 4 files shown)


ZhengGong-amd (approved):
Approved,  please resolve the conflicts before merging.

PR #784, opened 2026-07-01
fix(cli): honor phase-budget-pct overrides after KERNEL_AGENT rename

BaoYunkai:
The KERNEL -> KERNEL_AGENT / FRAMEWORK -> FRAMEWORK_AGENT phase rename (33ac6ccc) left stale literals that silently dropped CLI overrides.
- cli: build phase_budget_pct via new _build_phase_budget_pct() using canonical phase_state constants, and actually wire it into the optimize path (previously the real path still used the "KERNEL" literal, so the override was dropped by normalize_budget_pct and fell back to 0.305).
- cli: register --phase-budget-*-pct aliases alongside the legacy --max-minutes-*-pct flags so pipeline callers' spelling is accepted; fix stale help defaults.
- cli: fix resume --no-explore guard comparing cur_phase to legacy "FRAMEWORK" instead of "FRAMEWORK_AGENT".
- schema/tests: refresh stale FRAMEWORK/KERNEL phase-name references.
- tests: add regression coverage incl. a guard that the live optimize path is wired to the helper (catches the "defined but not called" bug).

- Description: what and why

### DIFF
--- a/inference_optimizer/breakdown/legacy_collectors.py
+++ b/inference_optimizer/breakdown/legacy_collectors.py
@@ -37,9 +37,11 @@ _ACTION_PHASE: dict[str, str] = {
     "params": "EXPLORE",
     "backends": "EXPLORE",
     "sweep": "SWEEP",
-    "select_kernels": "KERNEL",
-    "kernel_opt": "KERNEL",
-    "integrate": "KERNEL",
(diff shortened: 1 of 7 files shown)


ZhengGong-amd (approved):
Limited revision, apporved

PR #821, opened 2026-07-09
[Stepping] inference-optimizer runtime hardening (framework guard, kv_cache OOM, conc_sweep recovery, SGLang patch fallback, kernel-agent turns)

zengleixin-amd:
## Summary

Net contribution vs `main` (three-dot PR diff): **15 files, +754 / -12**. Runtime-robustness fixes for the `inference_optimizer`, surfaced during a GLM-5.2 long-horizon run. No behavior change for a healthy run — these harden error classification / reporting / recovery and add two opt-in config knobs, each with regression tests.

> Base note: #801 (bypass analysis + src-layout) is already **squash-merged into `main`**, so this branch is cut fresh from `origin/main` and carries **only** these fixes (not #801's 46-commit history). All 15 touched files are byte-identical between #801 and `main`, so the fixes apply without conflict.

## 1. Framework guard (`EXPECTED_FRAMEWORK`)
Fail fast when a launcher-pinned framework is violated. A stale `switch_to_sglang`-style script that mutates `$FRAMEWORK` otherwise silently runs a different backend than requested (observed: a `FRAMEWORK=vllm` run executed as sglang).
- `cli/__init__.py`: `_enforce_expected_framework()` reads `EXPECTED_FRAMEWORK` / `INFERENCE_OPTIMIZER_EXPECTED_FRAMEWORK`; enforced on **both** the fresh-resolve and resume paths before `os.environ["FRAMEWORK"]` is set.
- Tests: `test_preflight_auth_override.py` (mismatch / match / namespaced var / precedence / unset-noop).

## 2. KV-cache OOM classified distinctly
"Loaded weights leave no GPU memory for the KV cache" is an over-aggressive `--mem-fraction-static`, not a generic crash — so the loop/report can advise "raise mem-fraction" instead of treating it as a hard failure.
- `baseline.py`: `_KV_CACHE_OOM_MARKERS`, matched **before** the fast-exit elapsed gate (KV OOM surfaces minutes in, after weight load).
- `report.py`: a `kv_cache_oom` root-cause bucket checked **before** generic `oom`; markers are **reused from `baseline` (single source of truth)**, so a bare `--mem-fraction-static` appearing in an argv dump is no longer mis-bucketed (Bugbot finding).
- `shared_state.py`: `kv_cache_oom` added to the `stderr_tail` capture set (both recorders).
- Tests: `test_baseline_warmup_double_run.py`, `test_baseline_failure_summary.py`, `test_shared_state_units.py`.

## 3. conc_sweep summary recovery
When `reports/conc_sweep_summary.json` is absent or stale, recover the real result from the `runs/conc_sweep/` workspaces (fixed: the final report showed `successful_pairs=0` while the process had actually succeeded).
- `breakdown/collectors/kernels.py`: import-light recovery (never pulls Magpie/torch in at report time); defensive `parents[2]` base so a shallow session path can't `IndexError`.
- Tests: `test_breakdown_conc_sweep_collectors.py` (recover-absent / recover-newer / keep-a-better-existing-report / empty).

## 4. stop_reason explained in the final report
A human-readable "Why it stopped" line, aligned to the closed `STOP_REASON_VOCAB` (all 34 members) so the mapping can't silently drift.
- `report.py`: `_STOP_REASON_EXPLANATIONS` + `_explain_stop_reason()`; a coverage test imports `STOP_REASON_VOCAB` and asserts every member has a non-empty explanation.
- Tests: `test_report_helpers_coverage_unit.py`.

## 5. SGLang TraceLens patch version fallback
`_resolve_sglang_patches_dir()` falls back to the nearest not-newer same-minor (then cross-minor) patch subdir when no exact `sglang_X_Y_Z` match exists (fixed: a 0.5.14 profile degraded because only older patch dirs shipped). `_apply_atomic`'s `git apply --check` still guards a genuinely incompatible pick.
- `_server_patcher.py`; Tests: `test_server_patcher_sglang_version_fallback.py` (exact / same-minor / cross-minor / only-newer / empty-subdir).

## 6. kernel-agent max_turns configurable
`INFERENCE_OPTIMIZER_KERNEL_AGENT_MAX_TURNS` (default 5, unchanged) lets an operator raise the Claude turn budget to avoid a "Reached maximum number of turns" failure on complex kernel tasks without editing the reactor.
- `cli/backends.py`; Tests: `test_cli_backends_unit.py` (default / override / invalid / <=0).

## Commits
- `5d6782f5` core runtime fixes (sections 1–6).
- `dc9190f7` review + Bugbot follow-ups (STOP_REASON_VOCAB alignment, resume-path guard, kv_cache_oom ordering + mem-fraction specificity, conc_sweep `parents[2]` hardening, added regression tests).

## Test plan
- [x] `pytest` targeted set (the 8 touched test files) — **207 passed**
- [x] Clean branch cut from `origin/main`; PR diff is exactly the 15 fix files (no #801 history)
- [x] No conflict markers; every touched file byte-compiles
- [ ] Full CI on push
- [ ] `ruff check .` / `ruff format --check .`

### DIFF
--- a/src/hyperloom/inference_optimizer/breakdown/collectors/kernels.py
+++ b/src/hyperloom/inference_optimizer/breakdown/collectors/kernels.py
@@ -10,6 +10,7 @@ recorded in ``warnings`` and the section returns a best-effort partial.
 
 from __future__ import annotations
 
+import re
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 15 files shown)


chaojhou (approved):
Approving — solid runtime-hardening PR.

Why approve:
- Each of the 6 fixes is opt-in (env-gated) or a strict error-classification refinement, so healthy runs are unaffected.
- Full CI is green (test 3.10 + 3.11, ruff, pylint, CodeQL, sphinx).
- Multi-model review follow-ups in 75e3d93d are all addressed with regression tests:
  - baseline._classify_subprocess_error now guards stderr_tail=None (the hoist above the elapsed gate could AttributeError on a slow crash with no captured stderr).
  - _load_conc_variant_point uses an explicit is-None check so a legitimate 0.0 output_throughput is not dropped by `a or b`.
  - collect_conc_sweep_summary only falls back to run-workspace recovery when the report has 0 successful pairs (healthy report stays authoritative; skips the runs/ scan on the healthy path).
  - collectors never raise: _safe_mtime() + try/except around the per-task scan.

Optional nits (non-blocking):
- The framework-guard tests live in test_preflight_auth_override.py; a rename would help discoverability.
- report.py imports the private _KV_CACHE_OOM_MARKERS from baseline at call scope; a shared constants module would be a touch cleaner (fine as single-source-of-truth).

PR #861, opened 2026-07-14
Changed priority orders of hardcoded specs and benchmarking script

devalshahamd:
## Problem

On the open-source path, `populate_gpu_arch_json` preferred hyperloom's in-repo achievable spec (`roofline_ceiling.HW_SPECS_ACHIEVABLE`) over a live GPU microbenchmark, so measured MAF was never used when an in-repo spec existed. The in repo spec is based on datasheet specs, whereas the microbenchmark based numbers are preferred for analysis. 

## Change

Swap the priority: run the TraceLens microbenchmark first (measured MAF), and fall back to the hyperloom achievable spec only when the microbenchmark can't produce a usable spec (no idle GPU, non-zero exit, missing output, or all-zero MAF/bandwidth).

- Microbench block wrapped in `try/except`; failures fall through to the hyperloom spec instead of aborting.
- Loud-failure preserved: if the fallback is also unavailable, the original microbench error is re-raised.
- Added tests for both fallback paths (microbench unavailable, zeroed measured spec).

### DIFF
--- a/src/hyperloom/agents/kernel/tests/test_tracelens_arch_benchmark.py
+++ b/src/hyperloom/agents/kernel/tests/test_tracelens_arch_benchmark.py
@@ -173,9 +173,6 @@ def test_populate_gpu_arch_json_runs_microbench_when_missing(
 
     with (
         patch.object(tab, "resolve_arch_json_path", return_value=None),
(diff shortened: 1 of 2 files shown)


mohbasit (approved):
looks good

PR #865, opened 2026-07-14
Default proposal scoring OFF behind --enable-proposal-scoring flag

zengleixin-amd:
## Summary
- Default the advisory proposal scorer **OFF** and gate it behind a new `--enable-proposal-scoring` flag.
- Proposal scoring is advisory-only (never gates, never alters the search) and is OpenAI-compatible only. Defaulting it off avoids non-OpenAI (e.g. Anthropic-only) deployments silently issuing failing scorer calls.
- Preserve existing behavior: `--no-proposal-scoring` still force-disables, and the existing Anthropic-only guard (`_official_anthropic_only()`) is unchanged.
- Precedence in `_build_proposal_scorer`: `--no-proposal-scoring` > default OFF (needs `--enable-proposal-scoring`) > Anthropic-only guard > empty model list.
- Clarify `--proposal-scorer-models` help: it only takes effect together with `--enable-proposal-scoring` and does not enable scoring on its own.
- Not persisted across `--resume` (advisory; re-pass the flag). Documented in the builder docstring and the assembly comment.

## Changes
- `cli/parser.py`: add `--enable-proposal-scoring`; clarify `--proposal-scorer-models` help.
- `cli/backends.py`: `_build_proposal_scorer` returns `None` unless enabled; updated docstring.
- `cli/__init__.py`: update the proposal-scorer assembly comment.
- `tests/test_cli_backends_unit.py`: default-OFF, conflicting-flag precedence, parser-level parsing, and "models without enable stays off" regression tests.

## Test plan
- [x] `pytest src/hyperloom/inference_optimizer/tests/test_cli_backends_unit.py` (28 passed)
- [x] `pytest .../test_proposal_scorer.py .../test_proposal_scorer_coverage_unit.py` (38 passed)
- [x] `py_compile` on all changed modules
- [x] CI green

### DIFF
--- a/src/hyperloom/inference_optimizer/cli/__init__.py
+++ b/src/hyperloom/inference_optimizer/cli/__init__.py
@@ -2156,9 +2156,9 @@ async def _run_optimize(args: argparse.Namespace) -> int:
         # KnowledgePlane facade (None when --degraded-kb).
         knowledge_plane=knowledge_plane,
         # Advisory multi-model specialist-proposal scorer. Disabled by
-        # default: ``None`` unless --enable-proposal-scoring is passed, and
-        # still ``None`` under --no-proposal-scoring, in Anthropic-only
-        # deployments, or with an empty model list. When active it scores
+        # default: ``None`` unless --proposal-scoring is passed (--no-proposal-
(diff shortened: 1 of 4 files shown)


ZhengGong-amd (approved):
approved

PR #913, opened 2026-07-16
Clarify Hyperloom setup workspace flow

xiaofei-zheng:
## Summary
- Document the recommended flow as opening a dedicated agent workspace before installing Hyperloom with `pip install --target .`.
- Clarify that `/hyperloom-setup` creates or updates `.env` in the current workspace and normally only needs to run once.
- Enforce the documented setup option ordering in the setup skill, including Docker before baremetal and the Anthropic URL choices.

## Testing
- Ran `git diff --check`.


ZhengGong-amd (approved):
approved

PR #967, opened 2026-07-20
[CLEAN] chore: remove dormant/legacy code paths + close a serving-priority race

ZhengGong-amd:
Two independent changesets on this branch, plus a clean merge of `origin/main`:

1. **Dormant / legacy cleanup** — delete a set of hardcoded-off, deferred,
   env-gated, and legacy mechanisms end-to-end (code + tests + prompts/docs)
   with no fallback left behind, and flip the FRAMEWORK LLM candidate ranker
   to default-on.
2. **Serving-priority race fix** — probe the Ray `serving_slot` immediately
   before each GPU-specialist admission instead of once per dispatch pass.

Net: `+401 / -1851` across 41 files — mostly deletions.

## 1. Dormant / legacy cleanup

Removed the following, collapsing each to its former default behavior:

- **EXPLORE↔KERNEL phase interleave** — hardcoded `return False` with no
  env/CLI unlock. Deleted `is_phase_interleave_enabled`, the `_INTERLEAVE_*`
  constants and force-exit thresholds, and the dead IR-6 narrowing block;
  collapsed the `*_with_interleave` helpers into the plain per-phase ones and
  updated `gate.py` / `prompt_builder.py` / `render.py` / orchestration &
  critic prompts.
- **Cross-framework `use_llm` ignored arm** — the `cross_framework.py` branch
  that only appended a "deferred" risk note. The honored `audit.py`
  (`fa phase-audit`) `use_llm` path is untouched.
- **`HYPERLOOM_BYPASS_CRITIC`** — operator bypass of the integrate_patch Critic
  gate (both the PolicyGate check and the executor defense). `integrate_patch`
  now *always* requires a permissive Critic verdict.
- **`INFERENCE_OPTIMIZER_SWEEP_RUN_EVAL`** — conc/sweep grids now always inject
  `RUN_EVAL=false` (accuracy eval is concurrency-invariant; the gate still runs
  on explore/baseline).
- **`HYPERLOOM_RECOVER_ALLOW_GPU_RESET`** — the entire `rocm-smi --gpureset`
  recovery path (`recover.py`, preflight arming print, `recover.md`, and the
  robustness `event.py` fields that consumed its result). Recovery is now
  soft-cleanup only.
- **Retired roofline/profile CLI flags** — the six `_RetiredFlag`
  registrations (`--use-roofline-composite` etc.). The `_RetiredFlag` class
  stays (still used by `--enable-proposal-scoring`).
- **Flat legacy session layout** — `INFERENCE_OPTIMIZER_SESSION_LAYOUT` /
  `_layout_mode`. Layout is always per-model timestamp tree.
- **Legacy per-variant-restart conc-sweep path** —
  `INFERENCE_OPTIMIZER_CONC_SWEEP_SINGLE_SERVER=0`. Deleted the gate function,
  the legacy `run_conc_sweep` branch, and the now-unused `_build_grid` /
  `_order_concs_anchor_first`. conc-sweep always uses single-server arm reuse.

**Enabled by default:** the FRAMEWORK LLM candidate ranker now reuses the
orchestration backend's OpenAI-compatible client / credentials, so it is on
whenever orchestration has LLM credentials (still falls back to deterministic
ranking when none are present).

**Kept after evaluation** (verified reachable / load-bearing, out of scope):
`prune_branch` intent, `INFERENCE_OPTIMIZER_STRICT_PATHS`, and
`INFERENCE_OPTIMIZER_STRICT_PHASE`.

`unactionable-methods.info.md` (audit note) is updated to reflect the removals
and the ranker flip.

## 2. Serving-priority race fix

The dispatcher's `_spawn_fitting_queued` computed `serving_slot_busy()` once
per pump pass, so a serving benchmark that started mid-pass could still admit a
competing GPU research specialist. The enable-check
(`ray_serving_priority_enabled()`) stays pass-level (pure env read), but the
actual slot probe now runs immediately before each GPU-specialist admission.
Best-effort semantics are preserved — any probe failure fails open (no pause),
and the physical GPU/serving mutex is still enforced by Ray `num_gpus` +
`serving_slot` regardless. Added dispatcher-level regression tests for the
defer + SQLite lane release, and documented
`INFERENCE_OPTIMIZER_RAY_GPU_PENDING_LIMIT` /
`INFERENCE_OPTIMIZER_RAY_SERVING_PRIORITY`.

## Merge with `origin/main`

Merged `origin/main` with no content conflicts (main's additions land in
regions disjoint from this branch). Post-merge verification:

- None of the removed symbols were reintroduced.
- Main's incoming features compose cleanly and, where they interact, reinforce
  this branch: `PolicyGate.validate_dispatched_task` (dispatch-time PolicyGate
  replay) makes the Critic gate *stricter*, consistent with the
  always-require-verdict change here.
- All touched modules import successfully.

## Validation

- Targeted suites for every touched area pass locally (phase state / force-exit,
  cross-framework, critic-per-variant, integrate_patch coverage, conc_sweep,
  sweep executor, recover executor, framework-agent selection, session layout,
  resume, robustness signals + ladder, whole-machine GPU dispatch).
- Import/compile check across all merged modules is green.
- Full offline suite: the only failures are pre-existing on clean `main`
  (default model-name / gateway-key resolution tests), unrelated to this branch.
- Ruff/mypy are not installed in the working environment; IDE lints are clean.

## Test plan

- [ ] CI green (lint, mypy, full pytest with coverage) on the pushed branch.
- [ ] Confirm no consumer relies on any removed env var / CLI flag.


ZhengGong-amd:
Closed for PR #972

PR #968, opened 2026-07-20
fix(vllm): keep site-packages MAGPIE_PATH off PYTHONPATH

lishuoshuo-amd:
A pip/docker/baremetal install resolves MAGPIE_PATH to the main venv site-packages. Composing it into PYTHONPATH let its torch shadow an isolated vLLM venv's torch, crashing vLLM's C extension with undefined symbol / libtorch_cuda.so.

Drop MAGPIE_PATH from PYTHONPATH when it is a site-packages/dist-packages dir (already importable) at the three composition points (kernel-agent install.sh init + write_env_file, _grid_runner server launch). A source checkout root is still kept; directory uses of MAGPIE_PATH are untouched.


BaoYunkai:
Added the baseline accuracy fix here as requested.

Summary:
- Resolve relative `result_dir` values against the benchmark `output_dir` before exporting `RESULT_DIR` and `EVAL_RESULT_DIR`.
- Preserve `output_dir` and the resolved `result_dir` in successful baseline results.
- Add regression coverage for relative `result_dir` parsing so the accuracy parser no longer resolves the path from the coordinator cwd.

Tested:
- `PYTHONPATH="/wekafs/yunkai/Hyperloom-pr968/src" pytest -q "src/hyperloom/inference_optimizer/tests/test_eval_result_dir_wiring.py" "src/hyperloom/inference_optimizer/tests/test_baseline_param_overrides.py" "src/hyperloom/inference_optimizer/tests/test_accuracy_gate_units.py" "src/hyperloom/inference_optimizer/tests/test_accuracy_keep_gate.py"`
- Result: 86 passed, 1 existing pytest config warning.

chaojhou (approved):

PR #971, opened 2026-07-21
fix(baseline): per-session unique port for server_lifecycle reuse (fixes false baseline_accuracy_failed)

luochen-amd:
## Problem

The baseline cold-start **double-run guard** pins a **fixed default port `8888`** for the persistent inference server (`REUSE_PORT_DEFAULT`). On host-networked spur nodes this collides:

- Round 1 (warmup) launches the server with `cleanup=false` and leaves it running for round 2 to re-attach. When an attempt fails, that server can be **left behind on 8888**.
- The coordinator retries with a **new attempt dir** (new `pid_dir`), so the leftover server has **no matching reuse metadata**, and Magpie aborts the fresh warmup in ~2s with:

  > `Reuse metadata mismatch (missing reuse metadata JSON ...); running server on PORT=8888 is incompatible with this benchmark config or leftover state is stale.`

- A **leaked server from a prior job** (or a co-tenant job) on the same host port triggers the same abort.

Every retry re-uses `8888` and re-collides, so the baseline **never records an accuracy** and the run stops with `baseline_accuracy_failed` — even though the model serves correctly. Observed on `Qwen/Qwen3-0.6B` (vLLM, bf16): the eval actually scores ~0.66 but the gate saw `0.0`.

## Fix

`resolve_lifecycle_params` now **always probes a free ephemeral port** (`_pick_free_port`, via `bind(("", 0))`) for the persistent server instead of the fixed default. Both double-run rounds share the resolved port (via `inject_lifecycle` → `envs.PORT`), so the server bind, Magpie reuse keying, the lm-eval `base_url`, and teardown all agree. A leaked/co-tenant server on the old fixed port can no longer collide.

This is safer than a node-wide process kill (which would also reap co-tenant servers). No new config knobs; a free port is always used.

## Testing

- **Unit**: `test_baseline_warmup_double_run.py` + `test_grid_runner_behavior_lock.py::TestAutoWarmupTeardown` — **44 passed**. Tests that assert a specific port monkeypatch `_pick_free_port` to return `8888`.
- **End-to-end**: a `Qwen3-0.6B` baseline run **passes on the first attempt** — lifecycle uses a unique free port (e.g. `45597`), no `Reuse metadata mismatch`, `baseline_accuracy=0.66`, `tput=13797`, `stop_reason=""` (no more `baseline_accuracy_failed`).

### DIFF
--- a/src/hyperloom/inference_optimizer/tests/test_backend_gating.py
+++ b/src/hyperloom/inference_optimizer/tests/test_backend_gating.py
@@ -126,6 +126,9 @@ def test_lifecycle_delegates_to_bypass_backend(tmp_path, monkeypatch):
     cfg_path = tmp_path / "cfg.yaml"
     cfg_path.write_text(yaml.safe_dump(cfg), encoding="utf-8")
 
+    # Pin the free-port picker so the port assertion is deterministic
+    # (resolve now assigns a per-session free port on the common path).
+    monkeypatch.setattr(sl, "_pick_free_port", lambda: 8888)
     info = sl.resolve_lifecycle_params(cfg_path)
     # bypass now honors the YAML lifecycle block, so a serving framework
     # with profiling off is eligible.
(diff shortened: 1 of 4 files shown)


luochen-amd:
## CI E2E report — ✅

| item | value |
|---|---|
| result | ✅ (phase=Succeeded) |
| model | `Qwen/Qwen3-8B` (dense) |
| resources | 1× GPU, TP=1 |
| PR branch | `fix/baseline-reuse-port-collision` |
| claw_session_id | `135d2616-b72c-4de9-9e9a-7e8efc172b58` |
| spur job / node | `21941` / `crsuse2-m2m-266` |
| queue → dispatch | 0s |
| run time | 51m 3s |
| total | 51m 3s |

[details](https://github.com/AMD-AGI/Hyperloom/pull/971)

chaojhou (approved):

PR #976, opened 2026-07-21
ci: expose explicit failure info for sharded test runs

zengleixin-amd:
## Summary

This PR improves failure visibility for the sharded pytest matrix in `Tests with Coverage` CI. Previously, when a test failed in one of the parallel shards, you had to dig through raw job logs to find out *which* test and *which* shard failed. This change surfaces that information directly in the CI UI.

## What changed

1. **Per-shard error annotations** — When a shard fails, a step now emits GitHub `::error` annotations that expose the exact failing tests (the `FAILED <nodeid> - <reason>` lines extracted from pytest `-rf` output), so failures show up inline on the job without opening logs.
2. **Coverage job summary table** — The coverage aggregation job writes a per-shard pass/fail table to the job summary and lists the failing tests, giving an at-a-glance view of the whole matrix.
3. **Precise "Enforce test success" reporting** — The enforcement step now names the specific failing shard(s) (e.g. `Failed shard(s) for Python 3.11: py3.11-shard3`) instead of a generic "tests failed" message.
4. **Failure log artifact** — On failure, `pytest-output.log` is uploaded as an artifact so the full output is retrievable, with explicit handling for the missing-log case.

## Scope

Changes are limited to `.github/workflows/tests-coverage.yml` (CI configuration only). No production code is affected.

## Verification

This feature was verified on a real CI run using a temporary, intentionally-failing probe test. The run confirmed all four behaviors: shard-level `##[error]` annotations with the extracted `FAILED <nodeid>` lines, the per-shard outcome table in the coverage job summary, and the enforcement step naming the exact failing shard for both Python 3.10 and 3.11. The probe test has since been removed; only the workflow change remains.


zengleixin-amd:
Retargeting this CI shard-failure-visibility work to the v1.0-dev branch. Superseded by #992. Closing this main-targeted PR.

PR #990, opened 2026-07-22
[DO NOT MERGE] TEMP probe A3: precise nodeid + coverage-on-failure

zengleixin-amd:
Throwaway probe. Verifies precise pytest-nodeid extraction (FAILED + fixture ERROR) from the -rfE summary section, and that coverage is still combined/reported when tests fail but all shards reported. DO NOT MERGE; branch deleted after verification.


zengleixin-amd:
Throwaway probe complete. Verified: precise pytest nodeids extracted from the -rfE summary section (FAILED + fixture ERROR, '/'+'.py::' form), and coverage is still combined/reported (TOTAL shown with 'partial or misleading' caveat) when tests fail but all shards reported, while fail_under is skipped and the job stays red. Verified change is in PR #976. Deleting probe branch.

PR #993, opened 2026-07-22
fix(magpie-patcher): extend SGLang client trust patch to mi300x local path

fengshaoyi-amd:
## What & why
The SGLang client `--trust-remote-code` patch only reached:
- `sglang_mi355x.sh` remote + local (via meina's mi355x patcher, b97a2d6), and
- `sglang_mi300x.sh` **remote-direct only** (via the remote-trust patcher).

`sglang_mi300x.sh`'s **local-server client path** (`BENCHMARK_BASE_URL` unset — the baseline `server_lifecycle` reuse path) was never gated, so custom-tokenizer models on that path still failed the client-side tokenizer load.

## Change
Both scripts share **byte-identical** client blocks, so generalise the mi355x-specific patcher into a script-agnostic _apply_sglang_client_trust_patch_atomic / _is_sglang_client_trust_patched and apply it to **both** scripts in `magpie_scripts_patch_status`.
- Remote-direct gate stays required (missing block = drift → fail-loud).
- Local splice is applied only when the script has a local-server path (`--result-dir` marker); a **drifted** local block there still fails loud, while remote-only / reduced layouts skip cleanly.
- MI300X keeps its existing eval-concurrency inline rewrite (unchanged).

Builds on top of @meina's b97a2d6.

## Tests
Added: mi300x local-client trust (full flow), remote-only skip, drifted-local reject, unified idempotency. All trust tests pass. (Local Windows run also shows 6 pre-existing failures in unrelated atomic-copy / file-lock / exec-mode-bit tests — POSIX-only, not touched by this PR.)

### DIFF
--- a/src/hyperloom/inference_optimizer/tests/test_magpie_patcher.py
+++ b/src/hyperloom/inference_optimizer/tests/test_magpie_patcher.py
@@ -302,6 +302,28 @@ def test_sglang_mi355x_local_and_remote_client_trust_is_env_gated(fake_magpie: P
     assert script.read_text(encoding="utf-8") == text
 
 
+def test_sglang_mi300x_local_client_trust_is_env_gated(fake_magpie: Path):
+    # Regression: the client-trust patch must also reach the local-server path
+    # of sglang_mi300x.sh (previously only mi355x got the local splice). Use the
+    # realistic client shape (with a local-server path) for the mi300x script.
+    script = _write_sglang_script(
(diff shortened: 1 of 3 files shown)


fengshaoyi-amd:
Per meina: no separate PR needed — the commit was pushed directly onto v1.0-dev (3b93a4b..09ae78c, rebased on top of #994). Closing this PR; the fix is already on the dev branch.

PR #1019, opened 2026-07-24
fix(ray): default single-node Ray exec on

lishuoshuo-amd:
Restore single-node default to route GPU/serving work through the Ray backend when INFERENCE_OPTIMIZER_RAY_EXEC is unset. Multi-node stays off and pytest keeps the local subprocess path.


ZhengGong-amd (approved):
approved

xiaofei-zheng (approved):

PR #1030, opened 2026-07-25
Reuse Roofline shapes for vLLM block-FP8 tuning

BaoYunkai:
## Summary
- reuse block-FP8 GEMM shapes only from the steady-state trace selected by the existing Roofline/TraceLens pipeline when profile provenance and the normalized effective runtime match
- on stale or missing evidence, invoke the standard RooflineExecutor without overriding its profile config, timeout, benchmark script, backend environment, analysis route, steady-state mode, retry policy, or persistence semantics
- keep shape-specific code limited to M/N/K extraction and provenance validation, with cycle-free effective-runtime fingerprinting and live Coordinator state synchronization

## Test plan
- [x] GEMM handler and trace-record contract tests (223 passed)
- [x] Roofline executor, SharedState, PRELUDE, and profile promotion tests (147 passed, 1 skipped)
- [x] Baseline effective-args and Coordinator integration tests (156 passed)
- [x] `python -m pytest src/hyperloom/inference_optimizer/tests/test_profile_and_kernel_handlers.py -q` (132 passed)
- [x] Fresh CI run: all Python 3.10/3.11 shards, coverage, lint, CodeQL, and docs passed
- [x] CI E2E rerun attempted; infrastructure rejected the workload before launch because QOS `amd-burst-qos` does not exist


BaoYunkai:
Pushed three fixes from a follow-up review of this PR.

**1. Empty steady-state trace scanned the whole CWD.** In the standard-Roofline fallback, `Path("")` normalizes to `Path(".")`, whose `is_dir()` is `True`, so `_extract_vllm_block_fp8_profile_shapes` recursively globbed the process working directory. A Roofline that reported `succeeded` but selected no steady-state trace could therefore pick up shapes from an unrelated or stale trace and return `status: ok` with an empty `source_profile_trace`, feeding wrong shapes into tuning. Reproduced end to end before fixing. Now both the extractor and its caller reject an empty path.

The pre-existing `..._without_trace_is_explicit_failure` test only passed because the repo root happened to contain no matching events, so a regression test now places a matching stale trace in the CWD.

**2. The terminal lifecycle event clobbered the inline Roofline's state.** On the `intent_router` path, `_emit_lifecycle(status="END")` unconditionally saves the live `shared_state`, and `SharedState.save` is a whole-object overwrite. It ran between the handler returning and `_sync_profile_state_after_gemm_roofline`, so the refreshed fields were wiped from disk before the merge could read them: provenance failed, the merge was skipped, and the on-disk trace was destroyed, which also made the next run's reuse check miss and reprofile. The KERNEL auto-entry path was unaffected.

Since the root cause is ordering, the merge now runs before the terminal lifecycle save; the later call in `_handle_gemm_tuning_result` is kept and is idempotent. Verified by mutation testing: removing the fix makes the new regression test fail with the original `GEMM Roofline state sync skipped` warning.

The merge also no longer overwrites `lifecycle`, which previously discarded events the live state had not yet persisted. It now unions rows, deduplicating on `(step, status, ts)` and renumbering `seq`.

**3. A bare `current_best` could inherit a tuned arm's runtime.** The backfill in `current_profile_workload_context` is intentional and stays: `writeback` writes baseline/profile `current_best` entries with no runtime keys at all, so without it the fingerprint would read as an empty runtime and force a reprofile every time. But it was unconditional, so a runtime recorded off a tuned arm could still be reported as active after `current_best` reverted to a bare baseline. The profiled arm is now recorded alongside the workload and the backfill requires a matching arm, reusing the Roofline's existing `roofline_arm` semantics rather than introducing a separate notion. Legacy sessions without the recorded arm keep the previous behavior so they do not reprofile on every run.

Note that adding the field required updating `CORE_STATE_FIELDS` in both `policy/gate.py` and the robustness envelope, which are asserted to be identical.

Full suite: 8077 passed. The two remaining failures (`test_bench_specialist_without_explicit_needs_gpu_is_gated`, `test_aiter_build_e2e_real_rocm`) were confirmed to fail identically on a clean tree.

xiaofei-zheng (approved):

PR #1038, opened 2026-07-28
fix(inference): stop forcing the triton MoE runner onto aiter-only quant schemes

### LINKED ISSUE #1036: [Bug] Auto-injected --moe-runner-backend triton breaks Quark MXFP4 MoE serving on ROCm (Qwen3.5-397B-A17B-MXFP4)
### Target release

Unsure

Local

### Optimization domain

Inference

### Issue type

Crash / failure

### Which phase failed?

Setup / baseline

### Short Summary

 Hyperloom unconditionally injects --moe-runner-backend triton for MoE sglang models on AMD/ROCm. For a Quark W4A4
  MXFP4 MoE checkpoint (Qwen3.5-397B-A17B-MXFP4) this makes the sglang server crash on the first forward pass, so every
  orchestrator run through this path (baseline / roofline / profile / sweep) fails and no TraceLens/roofline snapshot is
  ever produced.

  The same model serves fine when the MoE runner backend is aiter (or left as auto) — which is exactly what GEAK's own
  launch path uses, reaching a valid baseline (~1237 tok/s). So the failure is caused purely by our injected flag.

### Job ID (hosted UI runs only)

_No response_

```shell
- Model: Qwen3.5-397B-A17B-MXFP4 (Quark PTQ, MXFP4 W4A4 MoE)
Environment:
- FRAMEWORK=sglang
- GPU_TYPE=mi355x
- TP=4, CONC=64, ISL=1024, OSL=1024
- PRECISION=mxfp4
- --model-class moe_swa
- --target-gain 50
- --max-hours 24
- InferenceX path: InferenceX
```

### Expected behavior

 Either:
  1. QuarkW4A4MXFp4MoE initializes self.runner for the triton MoE runner backend (so apply_weights works), or
  2. sglang fails fast at startup with a clear message that the triton MoE runner backend is not supported for the Quark
  W4A4 MXFP4 MoE scheme (and, ideally, falls back to a supported backend).

### Actual behavior
  Where the injection happens

  - Function: src/hyperloom/orchestrator/actions/executors/_grid_server_args.py:724 inject_sglang_moe_runner_backend()
  - Default value: DEFAULT_SGLANG_AMD_MOE_RUNNER_BACKEND = "triton" (_grid_server_args.py:716)
  - Call site: src/hyperloom/orchestrator/actions/executors/_workload_envs.py:302 (step 4 of server-arg composition),
  invoked whenever: framework=sglang and GPU is AMD/ROCm and model is MoE and the operator has not already pinned
  --moe-runner-backend.
  - Rationale in code comment: avoid the broken "aiter CK 2-stage fused-MoE JIT build" in some ROCm images.

  What breaks

  With --moe-runner-backend triton injected, the sglang server aborts during decode CUDA-graph capture:

  File ".../quark/schemes/quark_w4a4_mxfp4_moe.py", line 295, in apply_weights
      return self.runner.run(dispatch_output, quant_info)
  AttributeError: 'QuarkW4A4MXFp4MoE' object has no attribute 'runner'

  (The triton MoE runner path never initializes self.runner for the Quark W4A4 MXFP4 scheme — an sglang-side defect, but
  our forced injection is what triggers it, and it's currently unavoidable without hand-editing config.)

  Impact in the observed run: all baseline/roofline/sweep launches failed → 19 benchmark_report.json with
  tracelens_analysis: null, no analysis.md, roofline loop stalled. Only GEAK's separate launch (which does NOT inject
  triton) succeeded.
### Steps to reproduce

 Environment
- docker image: docker.io/primussafe/sglang:v0.5.12-rocm720-mi35x-profilerfix
  - Model: Qwen3.5-397B-A17B-MXFP4 — Quark PTQ, MXFP4 (W4A4) MoE (arch Qwen3_5MoeForConditionalGeneration,
  model_type=qwen3_5_moe; served on the text path)

  Steps to reproduce

  Launch any Quark W4A4 MXFP4 MoE checkpoint on ROCm and force the triton MoE runner:

  python3 -m sglang.launch_server \
    --model-path=/path/to/Qwen3.5-397B-A17B-MXFP4 \
    --host=0.0.0.0 --port=8899 \
    --trust-remote-code \
    --tensor-parallel-size=2 \
    --attention-backend aiter \
    --mem-fraction-static=0.68 \
    --disable-radix-cache \
    --enable-aiter-allreduce-fusion \
    --page-size 16 \
    --context-length 11264 \
    --moe-runner-backend triton      # <-- trigger

  The server aborts during startup / CUDA-graph capture (the scheduler receives SIGQUIT from the child and the process
  tree is killed).

Root cause (analysis)

  sglang/srt/layers/quantization/quark/schemes/quark_w4a4_mxfp4_moe.py apply_weights() calls
  self.runner.run(dispatch_output, quant_info), but self.runner is only ever assigned on the aiter/native code path.
  When the MoE runner backend is triton, the scheme is constructed without a runner attribute, so the first
  apply_weights invocation raises AttributeError. Selecting aiter / auto for --moe-runner-backend initializes
  self.runner correctly and the model serves normally.

### Logs

```shell

```

_No response_

### Data handling

- [x] I confirmed the attached logs/artifacts contain no sensitive information

BaoYunkai:
Fixes #1036

## Summary

Hyperloom injected `--moe-runner-backend triton` for every MoE sglang model on AMD/ROCm without looking at the checkpoint's quant scheme. sglang implements the Quark MXFP4 / W4A4 MoE scheme only on its aiter path, so the forced triton runner builds the scheme without a `runner` attribute and the server dies on the first forward pass (`AttributeError: 'QuarkW4A4MXFp4MoE' object has no attribute 'runner'`). Every orchestrator run through that path — baseline, roofline, profile, sweep — failed, and no TraceLens/roofline snapshot was produced. The same checkpoint serves fine when the backend is left to sglang.

Two layers of fix:

- **Quant-aware gate.** `_model_moe_runner_requires_aiter()` reads `config.json` (top level and nested `text_config`): `quant_method: quark` plus an `mxfp4` / `mx_fp4` / `w4a4` marker, or `fp4` together with the `e8m0` MX scale format (which keeps nvfp4 out). `inject_sglang_moe_runner_backend()` skips injection for those checkpoints. Living inside the injection helper, the gate covers every path: baseline, profile, roofline and the explore/grid runner.
- **One-shot runtime fallback.** The gate can only cover schemes we already know about, so `BaselineExecutor` now detects the missing-runner signature in the failure's error tail, warnings and `server.log`, then retries once with `drop_moe_runner_backend=True`. The strip happens at the end of the server-arg pipeline, after every source is merged (task params, `$INFERENCE_OPTIMIZER_SERVER_ARGS`, the reference recipe, the YAML base), so the flag cannot come back regardless of where it originated. Fires once; a second failure is reported as-is.

Notes:

- An eval-rooted fallback that already forced `RUN_EVAL=false` carries forward into the MoE retry, so the retry neither resurrects the eval that broke nor loses its `accuracy_source` / warning bookkeeping.
- An operator-pinned `--moe-runner-backend` still wins at injection time, but is dropped on the fallback retry — at that point it is the proven cause of death.
- The log-scan used by the eval detector was extracted into a shared `_failure_carries_markers()` helper now used by both detectors.
- The runtime fallback is scoped to `BaselineExecutor` (inherited by profile/roofline). The explore/grid variant loop is covered by the gate only; a single failing variant there does not sink a whole run.

## Test plan

- [x] New: `test_baseline_moe_runner_fallback.py` — signature detection (error tail / `server.log` / negative), end-to-end flagless retry, flag arriving via `$INFERENCE_OPTIMIZER_SERVER_ARGS` and via the reference recipe, operator-pinned flag dropped on retry, eval-disabled state carried into the MoE retry, fires only once, unrelated failures do not retry.
- [x] Extended: `test_moe_runner_backend_injection.py` — Quark 4-bit detection true/false matrix, nested `text_config`, injection and `materialize_config_with_envs` skip, env override cannot resurrect the injection, `drop_moe_runner_backend` strips an inherited flag while leaving other args intact.
- [x] `pytest src/hyperloom/inference_optimizer/tests` — 8130 passed. The 2 failures (`test_bench_specialist_without_explicit_needs_gpu_is_gated`, `test_aiter_build_e2e_real_rocm`) reproduce unchanged on `main`.
- [x] `pytest src/hyperloom/agents` — 2410 passed.
- [x] `ruff check` / `ruff format --check` clean on every touched file.

### DIFF
--- a/src/hyperloom/inference_optimizer/cli/model_gate.py
+++ b/src/hyperloom/inference_optimizer/cli/model_gate.py
@@ -256,6 +256,17 @@ _AMD_UNSUPPORTED_QUANT_ALGOS = frozenset({"nvfp4", "fp4"})
 
 _AMD_UNSUPPORTED_QUANT_METHODS = frozenset({"bitsandbytes", "bnb"})
 
+# Quark PTQ 4-bit MoE schemes (MXFP4 / W4A4) are implemented in sglang only on
+# the aiter/native MoE runner path; the triton runner leaves the scheme without
+# a ``runner`` attribute and the server dies on the first forward pass.
+_NATIVE_MOE_RUNNER_QUANT_METHODS = frozenset({"quark"})
+
+_MXFP4_QUANT_MARKERS = ("mxfp4", "mx_fp4", "w4a4")
+
+# ``fp4`` alone is ambiguous (nvfp4 uses it too); the e8m0 scale format is what
+# makes it MX.
+_MX_SCALE_MARKER = "e8m0"
+
 # Quant methods with a real vLLM/sglang loader. Anything else declared in
 # config.json is a private/third-party format that fails in engine init.
 # bitsandbytes/bnb are listed here but separately gated on AMD.
@@ -666,6 +677,51 @@ def _model_is_moe(model_path: str) -> bool:
         if "moe" in str(cfg.get("model_type") or "").lower():
             return True
         if any("moe" in arch.lower() for arch in _config_architectures(cfg)):
+            return True
+    return False
+
+
+def _model_moe_runner_requires_aiter(model_path: str) -> bool:
+    """Best-effort detect a MoE quant scheme that only the aiter runner serves.
+
+    A Quark PTQ 4-bit MoE checkpoint (MXFP4 / W4A4, e.g.
+    Qwen3.5-397B-A17B-MXFP4) is only wired up on sglang's aiter/native MoE
+    runner. Forcing ``--moe-runner-backend triton`` builds the scheme without a
+    ``runner`` attribute, so the first forward pass raises ``AttributeError``
+    and the server dies during cuda-graph capture. Callers use this to skip the
+    AMD triton injection and let sglang pick the backend itself. Checks the top
+    level and a nested ``text_config``. Soft-degrades to False on any missing /
+    unreadable / invalid config.
+
+    Args:
+        model_path (str): The local model directory containing ``config.json``.
+
+    Returns:
+        bool: ``True`` when the checkpoint carries a Quark 4-bit MoE scheme.
+    """
+    if not model_path:
+        return False
+    data = _load_model_config_dict(model_path)
+    if data is None:
+        return False
+    candidates = [data]
+    nested = data.get("text_config")
+    if isinstance(nested, dict):
+        candidates.append(nested)
+    for cfg in candidates:
+        qc = cfg.get("quantization_config")
+        if not isinstance(qc, dict):
+            continue
+        method = str(qc.get("quant_method") or "").strip().lower()
+        if method not in _NATIVE_MOE_RUNNER_QUANT_METHODS:
+            continue
+        try:
+            blob = json.dumps(qc).lower()
+        except (TypeError, ValueError):
+            blob = str(qc).lower()
+        if any(marker in blob for marker in _MXFP4_QUANT_MARKERS):
+            return True
+        if "fp4" in blob and _MX_SCALE_MARKER in blob:
             return True
     return False
 
--- a/src/hyperloom/inference_optimizer/cli/model_gate.py
+++ b/src/hyperloom/inference_optimizer/cli/model_gate.py
@@ -17,6 +17,7 @@ import sys
 from collections.abc import Callable
 from dataclasses import dataclass
 from pathlib import Path
+from typing import Any
 
 from .. import gpu_types as _gpu_types
 from ..model_config_utils import (  # noqa: F401 - re-exported for callers/tests
@@ -256,16 +257,16 @@ _AMD_UNSUPPORTED_QUANT_ALGOS = frozenset({"nvfp4", "fp4"})
 
 _AMD_UNSUPPORTED_QUANT_METHODS = frozenset({"bitsandbytes", "bnb"})
 
-# Quark PTQ 4-bit MoE schemes (MXFP4 / W4A4) are implemented in sglang only on
-# the aiter/native MoE runner path; the triton runner leaves the scheme without
-# a ``runner`` attribute and the server dies on the first forward pass.
+# Quark PTQ MX-FP4 (W4A4) MoE is implemented in sglang only on its aiter MoE
+# runner; every other backend leaves the scheme without a ``runner`` attribute
+# and the server dies on the first forward pass.
 _NATIVE_MOE_RUNNER_QUANT_METHODS = frozenset({"quark"})
 
-_MXFP4_QUANT_MARKERS = ("mxfp4", "mx_fp4", "w4a4")
+# MX group size, mirroring sglang's ``QuarkConfig._is_mx_fp4`` validation.
+_MX_FP4_GROUP_SIZE = 32
 
-# ``fp4`` alone is ambiguous (nvfp4 uses it too); the e8m0 scale format is what
-# makes it MX.
-_MX_SCALE_MARKER = "e8m0"
+# sglang resolves a layer's quant config from these, most specific first.
+_QUARK_LAYER_CONFIG_KEYS = ("layer_quant_config", "layer_type_quant_config")
 
 # Quant methods with a real vLLM/sglang loader. Anything else declared in
 # config.json is a private/third-party format that fails in engine init.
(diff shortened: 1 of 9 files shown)


BaoYunkai:
Thanks — this review sent me to read sglang 0.5.12 directly instead of reasoning from the traceback, which turned up more than the review asked for. All points addressed in 143a209 except #5, which I believe is a misread; details below.

### 1. No runtime fallback on explore/sweep/grid — fixed, but not with a retry

Worth stating what the grid actually does: it never injects a MoE runner backend. Injection lives solely in `materialize_config_with_envs`, which the grid does not use (it builds variant YAML via `apply_runtime_benchmark_overrides` + `compose_server_args`). So a backend reaches a grid variant only by **inheritance** — from the baseline recipe it was seeded with, or from an authored variant.

That makes a retry the wrong tool: it would burn a second full benchmark to discover something we can decide before launch. `_build_variant_yaml` now drops an inherited `--moe-runner-backend` when the model is one only aiter can serve, and logs a warning naming the variant. Cheaper than a retry and it covers both inheritance routes.

Residual limitation, acknowledged: if detection misses a *new* aiter-only scheme, grid variants still fail with no salvage. I am leaving that as-is — grid variants are individually tolerant of failure (`keep_going_on_failure`), and a baseline always runs first, so the baseline fallback surfaces the problem before a grid ever starts.

### 2. Markers too broad — fixed, but not by keying on Quark

Agreed on the risk, disagreed on the narrowing. Binding the signature to `QuarkW4A4` / `quark/schemes` would make it miss the other aiter-only schemes found under point 4, which fail identically.

Detection now requires a missing-runner marker **and** a MoE-scheme context marker (`create_moe_runner` / `moe_runner` / `_moe.py` / `fused_moe` / `/moe/`) in the same blob. An unrelated `AttributeError` about some other `runner` no longer burns a retry, and the signature stays scheme-agnostic. Two tests pin both halves.

### 3. Quark + operator pin untested — fixed

Added `test_quark_checkpoint_with_operator_pinned_backend_recovers` (Quark MX-FP4 fixture + pinned `--moe-runner-backend triton` → crash → flagless retry, unrelated args preserved) and `test_quark_checkpoint_without_pin_never_gets_the_flag` (gate alone is enough; no crash, no retry).

### 4. Other Quark MoE variants — confirmed, and there are three

Right to ask. Grepping sglang for the `TODO(cwan): refactor other backends` fallthrough finds six `create_moe_runner` implementations; three build a runner only for aiter:

| Scheme | Selected by | Runner |
|---|---|---|
| `QuarkW4A4MXFp4MoE` | checkpoint `quantization_config` | aiter only |
| `QuarkInt4Fp8MoE` | `--quantization quark_int4fp8_moe` | aiter only |
| `Mxfp4DynamicQuantMoEMethod` | `--quantization mxfp4` on a non-mxfp4-serialized checkpoint | aiter only |
| `Fp8MoEMethod`, `CompressedTensorsW8A8Fp8MoE`, `Mxfp4MoEMethod` | — | triton supported |

The last two aiter-only ones are selected **online** through `--quantization`, so a config.json-only gate cannot see them. The gate now also reads `--quantization` from the server args; `mxfp4` counts only when the checkpoint is not mxfp4-serialized, mirroring sglang's own `is_checkpoint_mxfp4_serialized = "mxfp4" in quant_method`.

Separately, detection of the checkpoint scheme now mirrors `QuarkConfig._is_mx_fp4` field for field (fp4 + per_group + group_size 32 + e8m0 scales, weights static, activations dynamic) and walks `layer_quant_config` / `layer_type_quant_config` before `global_quant_config`, the way sglang resolves a MoE layer's config. The previous string matching was a guess; it would have missed a checkpoint whose config never spells "mxfp4", and fired on an nvfp4 one.

### 5. Unconditional `params["extra_server_args"]` sync — not from this PR

This is pre-existing code, not new. The diff adds only the line above it:

```
+            effective_extra_server_args = _remove_moe_runner_backend_arg(effective_extra_server_args)
             if effective_extra_server_args or "extra_server_args" in params:
```

`git diff HEAD~1 HEAD` shows one `+` on the strip and the `if` as unchanged context. The write-back has been there since the cuda-graph eager fallback landed; nothing in this PR changes when it fires. Happy to look at narrowing it separately if it is a real concern, but it is out of scope here.

### 6. Bare `--moe-runner-backend` not stripped — fixed

The value is now optional in the pattern, with a negative lookahead so a following flag is not swallowed as the value. Three cases added to the parametrized test: bare flag at end of string, bare flag alone, bare flag followed by another flag.

### Verification

`pytest src/hyperloom/inference_optimizer/tests` — 8147 passed; the same 2 failures (`test_bench_specialist_without_explicit_needs_gpu_is_gated`, `test_aiter_build_e2e_real_rocm`) reproduce unchanged on `main`. `pytest src/hyperloom/agents` — 2410 passed. `ruff check` / `ruff format --check` clean on every touched file (`_grid_runner.py` has a pre-existing formatting finding on `main`, untouched by this PR).

BaoYunkai:
Follow-up on point 5, since I offered to narrow the write-back separately: having checked, I do not think it should change.

- It is deliberate. The comment states the intent (keep the task envelope aligned with the materialized runtime so Roofline fingerprints record the one-shot eager fallback), and `shared_state.py` does read `params["extra_server_args"]` when building that context.
- It is pinned by an existing test: `test_baseline_param_overrides.py:559` asserts `"--enforce-eager" in ctx.task.params["extra_server_args"]` after an eager fallback.
- It is the established pattern, not a baseline quirk: `profile.py:710` writes `params["extra_server_args"]` the same way.

On the "ctx reused across tasks" concern: `task.params` belongs to a single task and is not shared between tasks. When the same task re-runs, carrying the parameters it actually ran with is exactly what the fingerprint wants.

Happy to revisit if there is a concrete case where this misattributes something.

BaoYunkai:
Addressed in afab2a3a.

1. `group_size: "32"`: no production change. sglang 0.5.12 itself uses the strict comparison `weight_quant.get("group_size") != 32` (and the same for inputs), so normalizing the string in Hyperloom would make the gate claim compatibility that sglang does not recognize. Added a negative test to pin exact upstream parity.

2. Grid online quantization coverage: added tests for both aiter-only online paths:
   - inherited `--moe-runner-backend triton` + variant `--quantization quark_int4fp8_moe`
   - inherited `--moe-runner-backend triton` + variant `--quantization mxfp4` on a non-serialized checkpoint

Both assert the backend is removed while the quantization flag is preserved. The focused suite passes: 59 tests.

xiaofei-zheng (approved):

PR #1040, opened 2026-07-28
fix(eval): stop false-positive install abort on refactored run_lm_eval parser

lishuoshuo-amd:
The pinned InferenceX (a4bb43af) refactored run_lm_eval's arg parser into a single merged `--port|--task|...|--top-p)` case with a `>&2` / `return N` catch-all. The per-flag legacy block the patcher matched no longer exists, so `_apply_run_lm_eval_arg_patch_atomic` returned False, `eval_flag_ok` went False, and (since the fail-loud change) install aborted with exit 5 -- even though no caller passes `--concurrent-requests` and run_lm_eval already reads concurrency from EVAL_CONCURRENT_REQUESTS/CONC, so accuracy eval was never actually blocked.

Two fixes:
- Teach the parser patch to recognise the merged-case shape and splice a dedicated `--concurrent-requests` case before the catch-all (indent-aware), restoring the defence-in-depth belt for the run-time re-copy.
- Align install-time judgement with the run-time entry point via a shared `_eval_concurrency_unblocked` helper: a defence-in-depth patch that merely could not be applied is not fatal on its own -- only a live `run_eval --concurrent-requests` that InferenceX would reject blocks eval.

Tests: 4 new cases covering the merged-case parse/patch, the reproduced false-positive, and the narrowed judgement (unpatchable parser is non-fatal without a live flag, still fatal with one). 183 passed.

### DIFF
--- a/src/hyperloom/inference_optimizer/assets/install.sh
+++ b/src/hyperloom/inference_optimizer/assets/install.sh
@@ -817,11 +817,13 @@ if not status.atomic_ok:
     sys.exit(1)
 if not status.remote_trust_ok:
     sys.exit(2)
-# Redundant --concurrent-requests eval flag could not be stripped from a
-# generic benchmark script (unrecognised shape). Non-fatal: RUN_EVAL=true
-# baselines may abort on InferenceX's 'Unknown parameter', but the baseline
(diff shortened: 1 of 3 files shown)


ZhengGong-amd (approved):
limited revision, approved

PR #1049, opened 2026-07-30
feat(bypass): kernel source resolution + shape recovery for graph/Triton kernels

lishuoshuo-amd:
## Summary
- Bring the bypass (TraceLens-free) kernel analysis path to parity for CUDA-graph and Triton workloads, so hot kernels resolve to editable sources **and** carry a dispatchable shape without patching sglang.
- Commit 1 (`resolve kernel sources by repo scan ...`): repo-scan source resolution by device kernel name, comms/MoE/SDPA/GEMM/Elementwise classification, and graph-mode trace handling (hipGraphLaunch attribution + idle-gate skip on under-recorded graph traces).
- Commit 2 (`recover kernel shapes ...`): a shape-resolution waterfall so graph-replay and Triton direct-launch kernels stop falling to `unresolved`.

## Motivation
On a real Kimi-K3 graph-mode trace, ~45% of kernel executions (63 distinct kernels, including routable hand-written Triton/KDA kernels) had a broken `correlation -> External id -> cpu_op -> Input Dims` chain and resolved to `shape_provenance=unresolved`. The data is not lost, only displaced (capture-time record, launch geometry, or the tile-encoding name), so a general identity-keyed fallback recovers it.

## Shape waterfall (identity-keyed, no name allowlist)
1. `torch_trace` -- own cpu_op Input Dims (precise; unchanged)
2. `capture_backfill` -- same-name kernel's capture-time shape
3. `launch_grid` -- this kernel's launch grid/block geometry
4. `tile_name` -- `BLOCK_SIZE_*` tile parsed from the kernel name
5. `unresolved` -- none of the above

`launch_grid`/`tile_name` are geometry (not operand dims) and are deliberately excluded from the analytical roofline so they cannot poison AI/bound-type estimates.

## Validation
- On the real Kimi-K3 graph trace, the top-15 hot kernels went from 3 `unresolved` to 0 (the 3 became `launch_grid`); the 12 `torch_trace` rows are unchanged (no regression).
- kernel-agent bypass suite: 195 passed (adds waterfall + reader passthrough tests).

## Notes / follow-up (not in this PR)
- These fallbacks make the kernels *visible* with a shape. Whether a `launch_grid`/`tile_name`-level shape should be allowed to satisfy the kernel_opt dispatch gate (`non-empty trace shape REQUIRED`) is left as a separate, opt-in decision, since geometry is coarser than operand shapes.

## Test plan
- [ ] CI green on the bypass / kernel-agent suites

### DIFF
--- a/src/hyperloom/agents/kernel/tests/test_bypass_classify.py
+++ b/src/hyperloom/agents/kernel/tests/test_bypass_classify.py
@@ -1,0 +1,114 @@
+###############################################################################
+# SPDX-FileCopyrightText: 2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+#
+# See LICENSE for license information.
+###############################################################################
+
+"""Unit tests for :mod:`_bypass_classify.classify_kernel` taxonomy coverage."""
+
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+import pytest
+
+sys.path.insert(0, str(Path(__file__).resolve().parents[1] / "tools"))
+
+from _bypass_classify import classify_kernel  # noqa: E402
+
+
+# (device_kernel_name, expected_category) covering real sglang+aiter kernels.
+_CATEGORY_CASES = [
+    # Communication / collective.
+    ("cross_device_reduce_2stage", "Communication"),
+    ("sglang::outplace_all_reduce", "Communication"),
+    ("all_reduce", "Communication"),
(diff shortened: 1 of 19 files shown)


xiaofei-zheng (approved):

tsrikris:
@chaojhou pls revert this PR, seems very messy and unmaintainable. This is a feature request that has to be routed to our team to study and get back to you; not a huge bypassing PR that adds so much code.

PR #1057, opened 2026-07-30
fix(install): stop xDiT quality deps from clobbering the ROCm torch

BaoYunkai:
## Problem
`ensure_xdit_quality_deps()` in `inference_optimizer/assets/install.sh` ran a bare `pip install scikit-image lpips` (no index, no `--no-deps`, no pins). `lpips` declares `torch>=0.4.0`, so pip's resolver pulled a **PyPI CUDA** `torch` (and `triton`) and **replaced the vendor ROCm build** already installed in the shared venv.

The install exited `0` with only a cosmetic `"not importable"` warn. The torch/triton swap was completely silent. Because atom/vllm/sglang on the pod all share one `/opt/venv`, an xDiT-only step run for a `--framework atom` session bricked GPU access machine-wide — every co-tenant then failed at `torch.cuda.is_available()` with `Found no NVIDIA driver`.

## Root cause
An **optional** quality-gate dep install had the power to move the **load-bearing** core (`torch`/`torchvision`/`triton`). The installer has no `--framework` notion and the venv is shared, so "don't run it for non-xDiT" is neither clean nor sufficient — the real invariant is: *this install must never touch the core.*

## Fix
1. **Pin the core (prevention).** `_write_core_constraints()` records the installed exact versions of `torch/torchvision/torchaudio/triton`; the dep install passes them via `pip install -c <constraints>`. The resolver can no longer upgrade/replace them, while `scikit-image`/`lpips` still pull their other (safe) deps — cleaner than a blanket `--no-deps` that would starve `scikit-image` of `scipy` etc.
2. **Tripwire (defense in depth).** `_guard_torch_not_clobbered()` snapshots torch's ROCm/HIP build before the install; if it vanished afterwards it attempts a best-effort rollback to the pinned versions and then **aborts HARD via `die()`** instead of the silent warn that let this ship.

## Tests
`src/hyperloom/inference_optimizer/tests/test_install_xdit_deps_clobber.py` extracts the real shell functions and drives them with a fake `python` whose fake `pip` models the resolver (an unconstrained install swaps the ROCm torch; a constrained one does not):
- `test_constraints_prevent_torch_clobber` — pinned install leaves torch as the ROCm build.
- `test_tripwire_aborts_when_torch_clobbered` — a clobber that slips through causes a hard abort + rollback attempt.
- `test_static_no_unconstrained_install_remains` — the old bare install is gone.

Reverting the fix turns all three red. `bash -n` clean.

### DIFF
--- a/src/hyperloom/inference_optimizer/assets/install.sh
+++ b/src/hyperloom/inference_optimizer/assets/install.sh
@@ -1072,6 +1072,48 @@ _XDIT_QUALITY_DEPS=(
   "lpips:lpips"
 )
 
+# Load-bearing packages an OPTIONAL dep install must never move. On a ROCm pod
+# these are vendor ROCm builds from a private index; letting pip's resolver pull
+# a PyPI (CUDA) torch to satisfy e.g. lpips' `torch>=0.4.0` silently bricks GPU
+# access for EVERY framework sharing this venv (that is exactly how this brick
+# shipped: a CUDA torch replaced the ROCm one, exit 0, no visible error).
+_XDIT_CORE_PINS=(torch torchvision torchaudio triton)
+
+# Write a pip constraints file pinning each installed core package to its exact
+# current version. `pip install -c <file>` then forbids the resolver from moving
+# them, while still letting the requested deps pull their OTHER (safe) deps.
+_write_core_constraints() {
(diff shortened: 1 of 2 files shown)


ZhengGong-amd (approved):
approved

PR #1048, opened 2026-07-30
Remove remote Cortex KB wiring and specialist cortex_kb MCP.

meinali-566:
Drop Critic --cortex-kb-url/CORTEX_KB_URL injection and specialist mcp__cortex_kb__* tools while keeping local Recipe KB, gbrain reads, and PR Monitor. Decouple --degraded-kb from PR Monitor, skip T0 when degraded, and limit IR-3 to PR Monitor probes; update docs and tests.

- Description: what and why
- Linked issue(s): close/fix refs
- Tests: added/updated? commands run?
- Breaking changes: yes/no (details if yes)

PR #1059, opened 2026-07-30
refactor(kb): remove remote Cortex KB end-to-end; rename cortex_* to recipe_*

meinali-566:
Follow-up to #1048, addressing its review feedback. #1048 removed the CLI
wiring for the remote Cortex KB but left the capability reachable, left a dead
policy rule behind, and left a naming surface that no longer described what it
held.

## 1. Remote Cortex KB — removed end to end, not just its wiring

#1048 stopped the CLI from injecting `CORTEX_KB_URL`, but the Critic still read
it: `kb_assess_client.py` built a client from the environment and POSTed to
`{CORTEX_KB_URL}/v2/reasoning/assess`. A value arriving via `.env`, a shell
export, or an inherited subprocess environment still produced outbound traffic
— while the entry had already been deleted from the environment-variables
reference, which is the one place an operator would look.

Deleted the whole path: `kb_assess_client.py`, `DecisionReviewer._inject_kb_assess`
/ `_assess_focus` / `_proposal_levers`, the judge-bundle fields
(`kb_assess_by_proposal`, `kb_assess_trace`), the prompt injection, the Langfuse
`kb_assess` span, and the breakdown recorder parameter.

`CORTEX_KB_URL`, `CORTEX_KB_HTTP_TIMEOUT_SEC` and `CORTEX_KB_ASSESS_INJECT` are
now read nowhere, so the deletion from the env-var doc is accurate rather than
misleading. No `env.pop()` guard was added: nothing reads the variable, so a
guard would be dead code.

Note: `runtime/kb_client.py` is untouched. It talks to the generic `/api/kb/*`
service via `KB_BASE_URL` (only when `kb_mode=live`) and reads no `CORTEX_*`.

## 2. PolicyGate R4 — removed

`KB_WRITE_TOOL_NAMES` was an empty frozenset, so `kb_write_unauthorized` could
never fire while its comment still claimed it guarded KB writes — a rule that
reads as protection but is a no-op. Local Recipe KB writes go through direct
Python calls in `writeback.py` / `proposals.py`, which R4 never covered, so
there is nothing left for it to guard.

Removed the rule, the constant, its three call sites, and the empty-set union in
`ALL_KNOWN_EXTERNAL_TOOL_NAMES`. R5 (`tool_whitelist_role`) is unchanged and
still gates PR Monitor / Web tools. Also dropped a now-stale "PolicyGate R4"
citation from the specialist prompt.

## 3. `cortex_*` → `recipe_*`

With the remote KB gone, these names held a local `RecipeKB` and no longer
referred to anything called Cortex. Renamed across the Python API, the
`--cortex-strict-fingerprint` flag, emitted data, and the on-disk
`<session>/runtime/cortex/` directory. See the CHANGELOG entry for the full
list — **this includes consumer-visible data** (breakdown keys, stop reasons,
source tags), so downstream parsers need updating.

No on-disk migration is provided, and none is needed: that directory holds only
derived bookkeeping. The three T0 snapshots in it are rewritten on every anchor
and duplicate `shared_state.warm_start_recipe`; the queue/audit files
(`.kb_pending.ndjson` and friends) have no writer in the tree at all — the
flusher daemon that wrote them was part of the removed remote-KB push path. The
authoritative recipe store is the local KB root outside the session tree,
mirrored to gbrain.

## Deliberately out of scope: Primus Cortex

`agents/framework/sources/primus_cortex.py` and `PRIMUS_CORTEX_PR_API` are
untouched. They share only the word "Cortex" with the removed KB: Primus Cortex
is the framework-agent's PR-candidate source and the backend behind PR Monitor
(`--pr-monitor-url` defaults to `$PRIMUS_CORTEX_PR_API`, and its client raises
"unsupported PR Monitor URL scheme"). #1048 explicitly keeps PR Monitor, so
removing or renaming Primus Cortex would work against that.

## Testing

`pytest src/hyperloom` on this branch and on the pre-change baseline produce
**identical failure sets** (verified by a two-way `comm` diff): 43 failures in
both, all environmental (26 missing build toolchain, 4 preflight credentials, 3
robustness runtime, etc.) and all in files this PR does not touch. The pass-count
delta is exactly the deliberately deleted `kb_assess` tests.

Note for local runs: `pytest-asyncio` must be installed or ~377 async tests fail
spuriously.


meinali-566:
retest

ZhengGong-amd (approved):
approved,

PR #1072, opened 2026-08-01
fix(benchmark): prevent API key leakage into session artifacts

BaoYunkai:
## Summary
- remove control-plane credentials from benchmark subprocess environments and materialized workload configs
- redact credential values before benchmark failures and diagnostics are persisted
- cover baseline, grid, warmup, bypass, and GEAK execution paths with regression tests

## Test plan
- [x] Run 392 affected benchmark and state tests
- [x] Run IDE lint checks on changed files
- [x] Run `git diff --check`


ZhengGong-amd (approved):
Validated and approved

PR #1080, opened 2026-08-03
Add an operator-supplied workload path and a source-rewrite specialist for iterative pipelines

lishuoshuo-amd:
Runs Hyperloom against an **iterative model pipeline** — a diffusion or
autoregressive rollout rather than a request-serving framework — and lets an
operator point it at their own workload without shipping a framework definition.

Validated end to end on HY-WorldPlay (HunyuanVideo-1.5, 8x MI355X): the loop took
the unmodified checkout from **0.3466 to 2.3157 fps, a 6.68x speedup (+568%)**,
discovering every optimization itself from the source and the profiler. For
reference the customer's hand-tuned result on the same workload and hardware is
+421%; normalized to the same baseline that is 1.806 fps, so this is 28% faster
than the hand-tuned stack.

## What this adds

**An operator-supplied workload path.** `--framework custom` with
`--framework-path` and `--benchmark-scripts-dir` runs a workload that lives
entirely outside this repo. Nothing about it needs a code change here, so
onboarding the next workload is a launch argument rather than a new entry
in `FRAMEWORKS`. The vendored entries stay: `xdit` and `hunyuan_image3` are
still registered and still tested, and retiring either one is a separate
change with its own callers to migrate.

**A rewrite specialist for iterative pipelines.** A diffusion rollout and a
serving framework share almost no optimization surface: one is scheduling,
batching and KV-cache admission; the other is redundant work the loop structure
creates — step-invariant computation repeated every step, collectives that
round-trip through the host to agree on a shape, tables rebuilt and re-uploaded.
`framework_rewrite_specialist` targets the second, and FRAMEWORK routes to it by
framework kind, so a scriptable pipeline is no longer pointed at a hot path that
does not exist there. Every rewrite lands behind a default-off environment switch
with a declared manifest, so each is attributable and composable, and a parity
leg proves the switches-off path still measures the baseline.

**Backend substitution as a first-class lever.** An `aten::` op holding a large
share of device time still dispatches through PyTorch, so the call site can
select a different backend. The roofline evidence now names those ops with their
call sites and states the rule — the kernel body is off-limits, the call site is
not — without prescribing a library. On this workload that is what took attention
from 61% of GPU kernel time at 6% of the bf16 roofline to a vendor kernel, worth
+258% on its own.

## Fixes

The GPU evidence had never once reached a specialist across four sessions, which
hid everything downstream of it. Repairing that surfaced three further defects in
paths that had never run on real data:

- **roofline**: a `status=ok` envelope that lost its payload cached an empty
  snapshot while the analysis sat complete on disk, so `roofline_snapshots`,
  the specialist's evidence section and the kernel phase were all silently empty.
- **policy**: a trace names a frame as `<relative path>(<line>): <function>`;
  checked verbatim that resolves against the process CWD, so once the evidence
  arrived every citation was denied and the specialist carrying it was cancelled
  before it ran. 13 of 20 specialists died this way in one run.
- **framework**: a specialist that never ran was recorded as one that found
  nothing, so five instant failures tripped the plateau nine minutes into a 24h
  budget. And PR discovery, scoped on having a repo URL as well as being
  scriptable, left out the workload that has no upstream repo by construction,
  which spent ten authoring rounds on unrelated serving PRs.
- **integrate**: reverting a candidate by reverse-applying its diff is textual
  and can leave the tree a few lines off HEAD while reporting success. That
  residue was banked as user state by the next auto-stash, conflicted with the
  KEEP that landed in between, and left conflict markers in source — one run
  benchmarked a tree that no longer parsed, another could not stash at all.
  Every KEEP is a commit, so HEAD is the accepted stack: restoring to it is the
  revert.
- **profiling**: the profiled capture was unbounded and blew ROCm's roctracer
  record budget, so the exported trace carried zero kernel events. Bounding the
  window took the trace from 2.18 GB with 0 kernels to 25.9 MB with 62,589.

## Test plan

- [x] 11,985 unit tests pass (one pre-existing failure in a kernel-agent markdown
      renderer, unrelated).
- [x] End-to-end on 8x MI355X: 0.3466 → 2.3157 fps (+568%), quality gate passed,
      parity leg confirms the default-off contract.
- [x] Reproduced from a pristine checkout on a later run: +287% in the first
      6 hours, with the vendor-kernel substitution found independently at hour 2.5.
- [x] The successful session's artifacts replay clean against this branch —
      evidence recovery, every `source_file` it produced accepted by PolicyGate,
      escapes still refused, and none of its progress rows reclassified.


tsrikris:
@lishuoshuo-amd From TraceLens PoV, the ground up feature will be enabled and many of the Hyperloom side changes will need to be reverted. @Ahmedhasssan-aig, @kyle-hoffmeyer will be working on this

ZhengGong-amd (approved):
Approved

PR #1084, opened 2026-08-03
feat(kernel): gate the kernel integrate KEEP on accuracy

chennyiiis:
## Summary

A kernel patch could `KEEP` on E2E throughput alone. Every piece needed to check accuracy was already in place:

- the integrate re-baseline already runs the serving eval;
- `kernel_optimization` already returns `KEEP` with the reason `"kernel artifact ready; E2E/accuracy deferred to integrate"`;
- `record_kernel_integrate_result` already documents a genuine `REVERT` as *"gain below threshold / accuracy regression"*.

But `integrate_handler` only read `output_throughput`. The deferral had no landing site and the measured score was discarded, so nothing in the kernel lane ever checked that a faster kernel still computes the right answer.

### The gate

Grading happens **only for a candidate that already cleared the throughput bar**, and reads the re-baseline's own eval output, so the verdict costs no extra GPU time and a regressing candidate never spends one.

| Outcome | Decision |
|---|---|
| Measured drop > `ACCURACY_THRESHOLD` (5pp) | `REVERT` — hard negative evidence |
| No score, but a positive `baseline_accuracy` exists | `NEEDS_REVIEW` — evidence gap, not a regression |
| No `baseline_accuracy` at all | `KEEP` on throughput only (degraded) |

The gate sits ahead of the optional source-import and paired-A/B passes so a patch that loses accuracy short-circuits before they run. Opt out with `INFERENCE_OPTIMIZER_REQUIRE_KERNEL_ACCURACY=0` (default on), mirroring the existing `INFERENCE_OPTIMIZER_REQUIRE_FRAMEWORK_ACCURACY` knob.

The gate only ever **compares** against `baseline_accuracy` and never writes it, so it is consistent with the `quality_ref_exempt` contract these synthetic re-baselines already carry: *compare, never establish*.

### Where the score actually lives

Reading the score takes two steps, which is worth calling out because the obvious one-liner is wrong here. The baseline double-run deliberately evaluates in the **warmup round only**, and the measured round's parse — rooted at its own slot — cannot see that sibling file, so the returned result carries no `accuracy`. The sibling-accuracy salvage that would recover it is skipped for this synthetic task by `quality_ref_exempt` (`_maybe_stop_on_missing_baseline_accuracy` returns early).

So `_grade_integrate_accuracy` prefers `bench_result["accuracy"]` and falls back to parsing the task workspace, where `parse_eval_results` resolves the warmup round's score via its `measured or result_files` fallback. Without that fallback the gate would see no score on every double-run integrate and downgrade **every** kernel `KEEP` to `NEEDS_REVIEW`. `test_falls_back_to_warmup_round_eval_output` pins this.

### Audit trail

`accuracy` / `baseline_accuracy` / `accuracy_pass` / `accuracy_gate` now land on the integrate result, and `accuracy` / `accuracy_pass` / `decision_reason` on the `kernel_integrate_attempts` ledger. Previously the kernel integrate path kept no accuracy trace outside a log line. A blocking accuracy verdict owns `decision_reason` (`accuracy_regression` / `accuracy_evidence_missing`) so it is not masked by the throughput-side annotations.

## Not in this PR

The discarded-warmup-eval waste still exists in two places (each needs a different fix, and the baseline double-run case is already solved on `main`):

- `_grid_runner._build_variant_yaml` — explore/sweep variants inherit `RUN_EVAL` into their `warmup_round`;
- `baseline.py` multi-node `mn_warmup` — explicitly sets `EVAL_RESULT_DIR` for a pass that is then discarded.

## Test plan

Added 20 tests: 7 driving the full `integrate_handler`, 13 unit tests for the grading matrix (including the warmup-fallback sourcing).

- [x] `test_kernel_integrate_and_report.py` — 30 passed (7 new + 23 pre-existing still green)
- [x] `test_accuracy_gate_units.py` — 56 passed (13 new)
- [x] Regression parity against `main` on the `kernel or integrate or accuracy or baseline or eval` subset: identical 28-failure set, `1671 → 1691` passed (exactly the new tests). Local failures are all Windows-only environment issues.
- [x] No new `ruff check` / `ruff format` findings attributable to this change (the pre-existing findings on `main` are untouched).

### Still needed before merge: one GPU validation run

Everything above mocks the benchmark subprocess, so no real Magpie and no real lm-eval was exercised. The eval artefact path is non-trivial (InferenceX's `run_lm_eval` cleans `$EVAL_RESULT_DIR`, so the parse searches down from `$RESULT_DIR`) and the tests write `results_gsm8k.json` directly, bypassing that plumbing. Please confirm on real hardware:

- [ ] The log line `integrate_handler: accuracy gate pass=... blocked=... new=... baseline=... source=...` reports a real `new=` value and a real `source=`, not `n/a` / `none`.
- [ ] `state.json` → `kernel_integrate_attempts[*].attempts[-1].accuracy` is non-null.
- [ ] A known-good kernel patch still reaches `KEEP` (no false blocking).

The failure mode to watch for is a false `NEEDS_REVIEW`: if the score cannot be resolved on real integrate runs while `baseline_accuracy > 0`, every kernel `KEEP` would be downgraded and the kernel lane would stall. `INFERENCE_OPTIMIZER_REQUIRE_KERNEL_ACCURACY=0` is the immediate escape hatch if that happens.


chennyiiis:
## GPU validation done — the risky path is confirmed working

Validated on an MI355X box (`chenyi-dev-vllm-pshjn-master-0`) with a real single-GPU vLLM serving Qwen3-8B and a **real lm-eval GSM8K run driven through InferenceX's own `benchmark_lib.sh` plumbing** (same `_install_lm_eval_deps` / `_patch_lm_eval` / `--output_path` semantics, lm_eval 0.4.9.2). Measured score: `exact_match,strict-match = 0.825` over 40 samples.

The artefacts landed exactly where the upstream double-run puts them, and nowhere else:

```
<integrate-ws>/warmup_round/eval_output/results_2026-08-03T11-03-55.854383.json
<integrate-ws>/measure_round/            <- no eval output at all
```

### The predicted failure mode reproduced, and the fallback handles it

The concern in the test plan was that the returned result carries no `accuracy` on a double-run, which would downgrade every kernel `KEEP` to `NEEDS_REVIEW`. That is exactly what happens, and it is now visible in the handler's own log:

```
baseline_executor: accuracy eval not found: no results*.json in .../integrate-k_hold/measure_round
integrate_handler: accuracy gate pass=True blocked=False degraded=False new=0.8250 baseline=0.8000
  source=.../integrate-k_hold/warmup_round/eval_output/results_2026-08-03T11-03-55.854383.json
```

The measured round genuinely resolves nothing; the workspace-rooted fallback recovers the real score. A negative control confirms it directly: `parse_eval_results(<ws>/measure_round)` returns `accuracy=None`, while `parse_eval_results(<ws>)` returns `0.825`.

### Checklist from the test plan

- [x] The log line reports a real `new=` value and a real `source=`, not `n/a` / `none` — `new=0.8250`, `source=.../warmup_round/eval_output/results_*.json`.
- [x] `kernel_integrate_attempts[*].attempts[-1].accuracy` is non-null — ledger recorded `{'decision': 'KEEP', 'accuracy': 0.825, 'accuracy_pass': True}`.
- [x] A known-good kernel patch still reaches `KEEP` (no false blocking) — `gain_pct=12.5`, accuracy holds, `decision=KEEP`.

Three end-to-end cases were driven through the real `integrate_handler` with genuine lm-eval artefacts planted in the round slots (only the benchmark subprocess is mocked, so the throughput number is synthetic — the accuracy is not):

| Case | Throughput | Baseline acc | Real acc | Decision | Reason |
|---|---|---|---|---|---|
| accuracy holds | +12.5% | 0.80 | 0.825 | `KEEP` | — |
| accuracy regresses | +12.5% | 0.95 | 0.825 | `REVERT` | `accuracy_regression` (patch reverted) |
| throughput loss | −12.5% | 0.80 | not graded | `REVERT` | no verdict spent on a loser |

### Test suite on Linux

The suite also runs clean on Linux, which closes the gap left by the Windows-only `fcntl` failures reported earlier: **194 passed, 0 failed** (`test_accuracy_gate_units.py` 56, `test_kernel_integrate_and_report.py` 30, plus 108 across `test_baseline_warmup_double_run.py` / `test_baseline_eval_fallback.py` / `test_eval_result_dir_wiring.py`).

### Still not covered

A full orchestrator run (`baseline → kernel_opt → integrate` with a real kernel patch) was not attempted: it needs LLM credentials and hours of GPU time, and reaching `integrate` depends on the kernel agent actually producing an integrable patch. What that would add over the above is the patch-authoring path, not the accuracy gate itself — the gate's inputs and decisions are now validated against real eval output.

chennyiiis:
## Full orchestrated run: completed, but it did not reach the kernel `integrate` path

Ran the full pipeline on the MI355X box (Qwen3-8B, vLLM 0.24.0, TP=1, single GPU, kernel lane enabled, `--no-explore --no-framework-agent` so the budget went to KERNEL). The run finished cleanly — `stop_reason=conc_sweep_done`, phase `CLOSE` — with a real baseline of **6677 tok/s/GPU and a real GSM8K accuracy of 0.9158** (full dataset, not the 40-sample smoke used earlier).

**It did not exercise this PR's code**, and the state makes that unambiguous:

```
runs/integrate/            <- empty
runs/kernel_opt/           <- empty
kernel_integrate_attempts  = 0
kernel_opt_attempts        = 0
last_kernel_opt            = {}
```

No `integrate_handler: accuracy gate ...` line was emitted anywhere. The kernel agent completed its TraceLens analysis successfully (trace split, `ops_summary.csv`, `perf_report.xlsx`, GEAK preflight verdict `ok (no blockers)`) but never produced a kernel artifact to integrate. The single optimization the run landed came from a specialist and went through `integrate_patch`, not the kernel `integrate`:

```
optimization_stack = [{'action': 'integrate_patch', 'source_phase': 'KERNEL_AGENT',
                       'scope': 'source_patch', 'operation_kind': 'param',
                       'extra_server_args': '--gpu-memory-utilization 0.90',
                       'tput': 6501.0}]
```

That is a server-arg change on a path that already had its own accuracy gate, so it says nothing about this change either way. Reaching `integrate_handler` requires the kernel agent to author an integrable kernel patch, which did not happen within this budget.

### What the run *did* independently confirm

The production baseline reproduced the exact mechanism this PR works around, in the orchestrator's own words:

```
baseline_executor: accuracy eval not found: no results*.json in .../runs/baseline/<id>/measure_round
baseline_executor: this attempt's RESULT_DIR had no accuracy, but salvaged a valid
  baseline accuracy=0.9158 from a sibling attempt
  (.../runs/baseline/<id>/warmup_round/benchmark_vllm_20260803_113711/results_2026-08-03T11-43-46.557165.json);
  not stopping the run
```

The measured round resolves nothing, and a *genuine* baseline recovers the score through `_salvage_sibling_baseline_accuracy`. A kernel integrate re-baseline carries `quality_ref_exempt`, so that salvage is skipped for it — which is precisely why `_grade_integrate_accuracy` needs its own workspace-rooted fallback. Note also that the real artefact sat directly in the benchmark workspace (`warmup_round/benchmark_vllm_*/results_*.json`), not under `eval_output/` as in the earlier smoke run; the recursive glob covers both layouts.

### Evidence status for this PR

| Claim | Evidence |
|---|---|
| Score resolves from real lm-eval output in the double-run layout | [Validated](https://github.com/AMD-AGI/Hyperloom/pull/1084#issuecomment-5165593293) against genuine artefacts; negative control confirms the measured round alone sees nothing |
| KEEP / REVERT / no-verdict decisions and ledger recording | Validated end-to-end through the real `integrate_handler` with genuine eval artefacts (3 cases) |
| The measure-round-has-no-accuracy mechanism is real, not theoretical | Confirmed by this production run's own logs |
| Unit + integration suite | 194 passed on Linux, 0 failed |
| A real kernel patch flowing through `integrate_handler` in a live run | **Not covered** — needs the kernel agent to author an integrable patch |

The remaining gap is the patch-authoring path upstream of the gate, not the gate itself. `INFERENCE_OPTIMIZER_REQUIRE_KERNEL_ACCURACY=0` remains the escape hatch if the gate ever misbehaves in production.

ZhengGong-amd (approved):
Approved, plz merge after minor revision

PR #1088, opened 2026-08-03
fix(orchestrator): correct GEAK resume revalidation escape hatch and align plateau-advisory prompt

lishuoshuo-amd:
## Summary

Two fixes for the GEAK 2b revalidation path and the orchestration prompt.

### 1. GEAK resume revalidation escape hatch (`writeback.py`)

The `has_prior_geak_e2e` escape hatch only fired when `geak_result` was empty, but `geak_result` is a persisted `SharedState` field, so on resume it is non-empty. A resume revalidation therefore fell into the material check against the live `current_best` — which already holds the promoted GEAK `accepted_config` — so the fingerprint matched by construction and a real accepted win was mis-judged `no_material`:

- `cumulative_gain_validated` zeroed
- watermark not refilled (next resume repeats the same drop)
- the accepted kernel's `kernel_journey` KEEP reverted to `REVERT` / `geak_no_material_product`, contradicting the headline that still carries its config

Fix: apply the material gate only on the FIRST validation (no prior `geak_e2e` stack entry), where `current_best` is still the pre-KERNEL config. A pre-existing `geak_e2e` entry means an already-proven win, so the resume revalidation short-circuits the material check regardless of the persisted `geak_result`.

Invariant: `geak_e2e` is only appended after `_promote_geak_from_candidate` succeeds, so the first 2b never sees a prior `geak_e2e` entry.

### 2. Plateau-advisory prompt wording (`prompt_builder.py`)

The assembled orchestration prompt claimed the Plateau advisory block never advances the phase, contradicting `orchestration.md`, the `_plateau_advisory_block` docstring, and `machine_state` (an EXPLORE plateau deterministically advances EXPLORE -> KERNEL_AGENT via `explore_no_more_leverage`). Scoped the "informational only" wording to KERNEL/FRAMEWORK and stated the EXPLORE auto-advance explicitly.

## Test plan

- [x] `test_2b_resume_reverify_of_promoted_geak_win_still_promotes` reproduces the fault (fails on unpatched code with gain=0.0, passes after fix)
- [x] `test_2b_no_material_candidate_does_not_promote` still passes (first-validation path unchanged)
- [x] `test_geak_gain_alignment.py`: 38 passed
- [x] prompt_builder wording verified consistent with `orchestration.md` / `conversation.py` docstring / `machine_state` behavior


ZhengGong-amd (approved):
Approved

PR #1097, opened 2026-08-05
Capture Platform Config and warn on CPU settings that aren't optimal 

rpoornac:
Capture platform fingerprint for reproducibility/record keeping. 
Record host CPU tuning state and warn on settings that skew results.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,10 @@ feedback on how to improve Hyperloom by completing the
 - Main agent instructions: [`src/hyperloom/inference_optimizer/SKILL.md`](src/hyperloom/inference_optimizer/SKILL.md)
 - CLI entry point: `python -m hyperloom.inference_optimizer.cli optimize`
 - Operator tools: `python -m hyperloom.inference_optimizer.tools.*`
+- Platform/BIOS audit: `sudo python3 scripts/platform_audit.py` — checks the EPYC
+  tuning knobs (boost, determinism, NPS, APBDIS, DF C-states, SMT) that silently
+  change benchmark results. Runs on the host, not in the container, since the
+  BIOS-only knobs need the BMC. Exits non-zero when a knob is off target.
 - Documentation source: `docs/`
 
 For contribution workflow, testing, and linting, see
--- a/scripts/platform_audit.py
+++ b/scripts/platform_audit.py
@@ -25,21 +25,27 @@ therefore synonym-based rather than a per-OEM table, so it degrades sensibly on
 hardware it has never seen.
 
 Usage:
-    sudo python3 scripts/platform_audit.py            # OS layer + Redfish via minted BMC account
-    sudo python3 scripts/platform_audit.py --no-bmc   # OS layer only, no BMC contact
-    python3 scripts/platform_audit.py --bmc-host H --bmc-user U --bmc-pass P
-    sudo python3 scripts/platform_audit.py --json     # machine-readable; exits non-zero on any FAIL
+    sudo python3 scripts/platform_audit.py --ca-cert /etc/ssl/bmc-ca.pem
+    sudo python3 scripts/platform_audit.py --insecure   # self-signed BMC cert
+    sudo python3 scripts/platform_audit.py --no-bmc     # OS layer only, no BMC contact
+    BMC_PASSWORD=... python3 scripts/platform_audit.py --bmc-host H --bmc-user U --insecure
+    sudo python3 scripts/platform_audit.py --json --insecure
+
+Exit codes (identical in both output modes): 0 all pass, 1 a knob is wrong,
+2 a knob is unresolved, 3 a temporary BMC account could not be revoked.
 """
 
 from __future__ import annotations
 
 import argparse
 import base64
+import getpass
 import glob
 import json
 import os
 import re
 import secrets
+import socket
 import ssl
 import struct
 import subprocess
@@ -95,19 +101,35 @@ KNOBS: dict[str, dict] = {
 NAME_BLOCKLIST = re.compile(r"sev|snp.?support|encrypt", re.I)
 
 
-def sh(cmd: list[str], timeout: int = 30, stdin: str | None = None) -> str:
-    """Run a command, returning stdout ('' on any failure)."""
+def run(cmd: list[str], timeout: int = 30, stdin: str | None = None) -> tuple[bool, str, str]:
+    """Run a command, returning ``(ok, stdout, stderr)``.
+
+    Callers that must not swallow a failure -- anything revoking BMC access --
+    use this rather than :func:`sh`.
+    """
     try:
         r = subprocess.run(cmd, capture_output=True, text=True, timeout=timeout, input=stdin)
-        return r.stdout if r.returncode == 0 else ""
-    except Exception:
-        return ""
+        return r.returncode == 0, r.stdout, r.stderr
+    except Exception as exc:  # noqa: BLE001 — surfaced to the caller as stderr
+        return False, "", str(exc)
+
+
+def _as_root(cmd: list[str]) -> list[str]:
+    return cmd if os.geteuid() == 0 else ["sudo", "-n"] + cmd
+
+
+def sudo_run(cmd: list[str], timeout: int = 30, stdin: str | None = None) -> tuple[bool, str, str]:
+    return run(_as_root(cmd), timeout, stdin)
+
+
+def sh(cmd: list[str], timeout: int = 30, stdin: str | None = None) -> str:
+    """Run a command, returning stdout ('' on any failure)."""
+    ok, out, _ = run(cmd, timeout, stdin)
+    return out if ok else ""
 
 
 def sudo(cmd: list[str], timeout: int = 30, stdin: str | None = None) -> str:
-    if os.geteuid() != 0:
-        cmd = ["sudo", "-n"] + cmd
-    return sh(cmd, timeout, stdin)
+    return sh(_as_root(cmd), timeout, stdin)
 
 
 def read(path: str) -> str:
@@ -131,66 +153,168 @@ def cpu_identity() -> dict:
     sockets = len({read(p) for p in glob.glob(
         "/sys/devices/system/cpu/cpu*/topology/physical_package_id")} - {""})
     nodes = len(glob.glob("/sys/devices/system/node/node[0-9]*"))
-    # EPYC 9005 = Turin, 9004 = Genoa/Bergamo. Governs expected boost ceiling.
-    gen = "unknown"
-    m = re.search(r"EPYC\s+(\d)(\d)", model)
-    if m:
-        gen = {"90": "Turin/Genoa"}.get(m.group(1) + "0", "unknown")
-        if m.group(1) == "9":
-            gen = "EPYC 9005 (Turin)" if m.group(2) == "5" else "EPYC 9004 (Genoa/Bergamo)"
-    return {"model": model, "sockets": max(sockets, 1), "numa_nodes": nodes, "generation": gen}
+    return {
+        "model": model,
+        "sockets": max(sockets, 1),
+        "numa_nodes": nodes,
+        "generation": epyc_generation(model),
+    }
 
 
-def measure_peak_mhz(core: int = 4, seconds: int = 4) -> float:
-    """Pin a spinner to one core and sample its achieved frequency.
+#: Series code -> family. Expected boost ceilings differ per generation, so the
+#: target profile is keyed on this rather than on the GPU installed.
+_EPYC_FAMILIES = {
+    "9005": "Turin",
+    "9004": "Genoa/Bergamo",
+    "8004": "Siena",
+    "7003": "Milan",
+    "7002": "Rome",
+    "7001": "Naples",
+}
 
-    This is what resolves an "Auto" boost setting into a real answer.
+
+def epyc_generation(model: str) -> str:
+    """Map an EPYC model number to its series.
+
+    In a 4-digit EPYC number the *last* digit carries the generation and the
+    first carries the series: 9575F -> 9005 (Turin), 9654 -> 9004 (Genoa).
+    Reading the second digit instead happens to work for 9575F and fails for
+    parts like 9755, so index the digits explicitly.
     """
-    proc = subprocess.Popen(
-        [sys.executable, "-c", f"import time\nt=time.time()+{seconds}\nwhile time.time()<t: pass"]
+    m = re.search(r"EPYC\s+(\d{4})", model)
+    if not m:
+        return "unknown"
+    d = m.group(1)
+    series = f"{d[0]}00{d[3]}"
+    family = _EPYC_FAMILIES.get(series)
+    return f"EPYC {series} ({family})" if family else f"EPYC {series}"
+
+
+def physical_cores() -> list[int]:
+    """One online CPU id per physical core, lowest id first.
+
+    SMT siblings share a core, so sampling both halves of a pair measures the
+    same silicon twice and understates the spread across the part.
+    """
+    first: dict[tuple[str, str], int] = {}
+    for path in glob.glob("/sys/devices/system/cpu/cpu[0-9]*"):
+        m = re.search(r"cpu(\d+)$", path)
+        if not m or read(f"{path}/online") == "0":
+            continue
+        cpu = int(m.group(1))
+        pkg = read(f"{path}/topology/physical_package_id")
+        core = read(f"{path}/topology/core_id")
+        if not pkg or not core:
+            continue
+        key = (pkg, core)
+        if key not in first or cpu < first[key]:
+            first[key] = cpu
+    return sorted(first.values())
+
+
+def sample_cores(count: int = 4) -> list[int]:
+    """Pick ``count`` physical cores spread evenly across the topology.
+
+    Hardcoding ids assumed a part with at least 69 cores; on a smaller SKU the
+    affinity call fails, the sysfs reads come back empty, and a meaningless
+    spread still reaches the determinism verdict.
+    """
+    cores = physical_cores()
+    if len(cores) <= count:
+        return cores
+    step = len(cores) / float(count)
+    return [cores[int(i * step)] for i in range(count)]
+
+
+def _spinner(seconds: int) -> subprocess.Popen:
+    return subprocess.Popen(
+        [sys.executable, "-c", f"import time\nt=time.time()+{seconds}\nwhile time.time()<t: pass"],
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.DEVNULL,
     )
+
+
+def core_freq_mhz(cpu: int) -> float | None:
+    """Current frequency of one specific CPU, or ``None`` if unreadable."""
+    v = read(f"/sys/devices/system/cpu/cpu{cpu}/cpufreq/scaling_cur_freq")
+    if v.isdigit():
+        return float(v) / 1000.0
+    # cpufreq absent (common under some hypervisors): fall back to the
+    # per-processor block in /proc/cpuinfo rather than a global maximum.
+    current = None
+    for line in read("/proc/cpuinfo").splitlines():
+        if line.startswith("processor"):
+            try:
+                current = int(line.split(":")[1])
+            except (IndexError, ValueError):
+                current = None
+        elif line.startswith("cpu MHz") and current == cpu:
+            try:
+                return float(line.split(":")[1])
+            except (IndexError, ValueError):
+                return None
+    return None
+
+
+def measure_peak_mhz(core: int | None = None, seconds: int = 4) -> float | None:
+    """Load one core and sample *that core's* achieved frequency.
+
+    Resolves an "Auto" boost setting into a real answer. Returns ``None`` when
+    the measurement cannot be trusted -- affinity refused, or the core's
+    frequency unreadable -- so the caller reports UNKNOWN instead of a verdict
+    built on a number that never described the pinned core.
+    """
+    cores = sample_cores(1) if core is None else [core]
+    if not cores:
+        return None
+    target = cores[0]
+    proc = _spinner(seconds)
     try:
-        sh(["taskset", "-pc", str(core), str(proc.pid)])
+        try:
+            os.sched_setaffinity(proc.pid, {target})
+        except (OSError, AttributeError):
+            return None
         time.sleep(1.5)
-        peak = 0.0
-        for _ in range(4):
-            freqs = [float(l.split(":")[1]) for l in read("/proc/cpuinfo").splitlines()
-                     if l.startswith("cpu MHz")]
-            if freqs:
-                peak = max(peak, max(freqs))
-            time.sleep(0.4)
-        return peak
+        samples = [f for f in (core_freq_mhz(target) for _ in _sleepy(4, 0.4)) if f]
+        return max(samples) if samples else None
     finally:
         proc.kill()
         proc.wait()
 
 
-def per_core_spread(cores=(4, 20, 40, 68), seconds: int = 3) -> float:
-    """Frequency spread across cores.
+def _sleepy(times: int, gap: float):
+    """Yield ``times`` ticks, sleeping ``gap`` between them."""
+    for i in range(times):
+        if i:
+            time.sleep(gap)
+        yield i
+
+
+def per_core_spread(cores: list[int] | None = None, seconds: int = 3) -> float | None:
+    """Frequency spread across physical cores, or ``None`` if untrustworthy.
 
     Power determinism lets each part run to its own silicon limit, so cores
     differ. Performance determinism pins every core to a common guaranteed
-    ceiling, collapsing the spread toward zero.
+    ceiling, collapsing the spread toward zero. A partial sample cannot tell
+    those apart, so anything short of a reading per loaded core returns None.
     """
+    cores = sample_cores(4) if cores is None else cores
+    if len(cores) < 2:
+        return None
     procs = []
-    for c in cores:
-        p = subprocess.Popen(
-            [sys.executable, "-c", f"import time\nt=time.time()+{seconds}\nwhile time.time()<t: pass"]
-        )
-        sh(["taskset", "-pc", str(c), str(p.pid)])
-        procs.append(p)
     try:
-        time.sleep(1.5)
-        vals = []
         for c in cores:
-            f = read(f"/sys/devices/system/cpu/cpu{c}/cpufreq/scaling_cur_freq")
-            if f:
-                vals.append(float(f) / 1000.0)
-        if not vals:
-            lines = [float(l.split(":")[1]) for l in read("/proc/cpuinfo").splitlines()
-                     if l.startswith("cpu MHz")]
-            vals = sorted(lines)[-len(cores):]
-        return (max(vals) - min(vals)) if vals else 0.0
+            p = _spinner(seconds)
+            procs.append(p)
+            try:
+                os.sched_setaffinity(p.pid, {c})
+            except (OSError, AttributeError):
+                return None
+        time.sleep(1.5)
+        vals = [core_freq_mhz(c) for c in cores]
+        if any(v is None for v in vals):
+            return None
+        return max(vals) - min(vals)
     finally:
         for p in procs:
             p.kill()
@@ -238,23 +362,28 @@ def os_layer(quick: bool = False) -> dict:
 
     peak = measure_peak_mhz()
     spread = per_core_spread()
-    out["peak_mhz"] = round(peak)
-    out["core_spread_mhz"] = round(spread, 1)
+    out["peak_mhz"] = round(peak) if peak is not None else None
+    out["core_spread_mhz"] = round(spread, 1) if spread is not None else None
+    out["sampled_cores"] = sample_cores(4)
 
     ceil_mhz = out["boost_ceiling_mhz"] or 0
-    boosting = bool(ceil_mhz and peak > ceil_mhz * 0.90)
-    out["cpb_effective"] = "Enabled" if boosting or out["cpb_msr"] == "Enabled" else "Disabled"
+    if peak is not None and ceil_mhz:
+        boosting = peak > ceil_mhz * 0.90
+        out["cpb_effective"] = "Enabled" if boosting or out["cpb_msr"] == "Enabled" else "Disabled"
+    else:
+        # No trustworthy measurement: the MSR still answers definitively, and
+        # without it the knob stays UNKNOWN rather than guessed.
+        out["cpb_effective"] = out["cpb_msr"]
 
     # Overshooting the advertised ceiling, or measurable per-core variance,
-    # both indicate the part is running to its own limit rather than a
-    # common guaranteed one.
-    if ceil_mhz:
+    # both indicate the part is running to its own limit rather than a common
+    # guaranteed one. Both readings are required: a missing sample cannot
+    # distinguish a uniform part from a failed measurement.
+    if ceil_mhz and peak is not None and spread is not None:
         if peak > ceil_mhz or spread > 8:
             out["determinism_inferred"] = "Power"
-        elif peak > 0:
-            out["determinism_inferred"] = "Performance (or Power at a uniform bin)"
         else:
-            out["determinism_inferred"] = None
+            out["determinism_inferred"] = "Performance (or Power at a uniform bin)"
     else:
         out["determinism_inferred"] = None
     return out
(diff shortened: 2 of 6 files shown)


xiaofei-zheng (commented):
Reviewed all 6 commits and the full diff, and downloaded `scripts/platform_audit.py` to run its pure functions locally against the verdict logic.

The direction is sound and the engineering judgment behind most of this is good — the gfx arch fix in particular is a real bug with a clean fix and proper test coverage. Three blocking issues below, plus a correction to the stated motivation that I think matters for how this code is maintained going forward.

---

## Blocking

### 1. `verdict()` substring matching makes the determinism knob report PASS unconditionally

`verdict()` matches with `t in v`, and the OS layer's inferred Performance value is the literal string `"Performance (or Power at a uniform bin)"` — which contains the substring `power`, the target for this knob. Verified locally:

```
verdict('determinism', 'Performance (or Power at a uniform bin)')  ->  PASS
verdict('determinism', 'Power')                                    ->  PASS
```

So the one case this check exists to catch — a node running Performance determinism, with every core pinned to a common guaranteed ceiling — is the one case it silently passes. That is worse than not having the check, because it emits a positive verdict.

The root cause is that a human-readable explanatory string is fed straight into the matcher. Suggest having `determinism_inferred` return a normalized value (`"power"` / `"performance"` / `None`) and moving the parenthetical caveat into a separate field used only by the text renderer. While there, tighten `t in v` to an exact match after normalization — no knob in `KNOBS` actually needs substring semantics, and this is the second-order bug it produced.

### 2. `TempBmcAccount.__enter__` can skip revocation entirely, and the exit code will not reflect it

`self.slot` is only assigned after `set_bmc_password` succeeds:

```python
if account_enabled(slot):
    print("WARNING: BMC account ... was still enabled from an earlier run ...", file=sys.stderr)
pw = secrets.token_urlsafe(24)[:16]
sudo(["ipmitool", "user", "set", "name", str(slot), self.user])
if not set_bmc_password(slot, pw):
    return self          # self.slot stays None
self.slot = slot
```

and `__exit__` opens with `if self.slot is None: return False`.

Reading those together: when the sentinel slot is found **already enabled** — i.e. the crashed-or-rebooted case `2b5eba3` specifically added handling for — and the password set then fails, the script prints one stderr line and skips revocation completely. `REVOKE_FAILURES` stays empty, so the exit code is 0/1/2 and CI sees a clean run. An ADMINISTRATOR account with a known password stays reachable on the LAN channel.

This is exactly the failure mode `2b5eba3` set out to eliminate ("the script could not tell the operator when it did not know something").

Suggest assigning `self.slot = slot` as soon as the slot is resolved and before any mutation, so `__exit__` always runs the revoke path. Revoking a slot that was never enabled is idempotent and costs nothing. Separately, finding the sentinel already enabled is itself evidence of a prior credential leak and should influence the exit code, not just print to stderr.

### 3. `account_enabled()` silently reports "not enabled" when the read fails

The read-back at the end of `__exit__` is the trust anchor for the whole revocation ("Trust the read-back, not the exit codes"), but it is built on the failure-swallowing `sudo()`:

```python
def account_enabled(slot: int) -> bool:
    out = sudo(["ipmitool", "channel", "getaccess", "1", str(slot)])
    m = re.search(r"Enable Status\s*:\s*(\w+)", out)
    return bool(m and m.group(1).lower() == "enabled")
```

`sudo()` returns `""` on failure, the regex misses, and the function answers `False` — "not enabled". So any BMC that errors, times out, or formats `channel getaccess` differently produces a **false confirmation that revocation succeeded**. This should use `sudo_run()` and distinguish "read it, disabled" from "could not read", with the latter appended to `REVOKE_FAILURES`.

---

## Motivation: two claims in the commit messages do not hold

Flagging these not to nitpick wording, but because a wrong stated motivation misleads whoever maintains this next, and because the cost/benefit of the 792-line BMC script depends on the justification being accurate.

### `a71292c` — a session-internal A/B is same-machine, so the delta is not affected

The commit argues: *"a throughput delta credited to a code change could equally have been SMT, the NUMA layout, or a frequency governor."* In Hyperloom's session model the baseline and every subsequent trial run on the same node within one session, so BIOS state applies equally to both sides and cancels out of the delta. The stated reason does not describe a failure this project can have.

The reasons that do hold, and that I would put in the commit message instead:

1. **A misconfigured node yields a valid delta but an invalid optimum.** With a wrong governor or boost disabled, the session-internal A/B stays fair, but (a) the absolute throughput reported downstream is low, and (b) the optimizer is searching on a hobbled machine — a CPU-side bottleneck shifts where the real bottleneck sits, so the tuner may select a config that works around a phantom constraint and is not optimal on a correctly configured node. That config then lands in the recipe KB for others. This is the strongest argument for the whole PR and it is independent of same-machine A/B.
2. **Cross-machine comparison genuinely exists via `--compare-against-gpu`.** `target_analysis` pulls an InferenceX reference point into `target_analysis/target_baseline.json`, and `report.py` renders it as the "External baseline" section. That compares our numbers against someone else's hardware, where our own BIOS state directly skews the conclusion.
3. **Cross-session and cross-time.** `final.json` is an archived artifact; the same model measured this month and next month, or after a node reimage, is a real comparison with nothing today to flag that the machine changed. Plus the multi-node gap the PR already documents.

### `bcf4116` — the value does not reach the recipe KB

*"The value reaches the session manifest and the recipe KB, so runs were filed under the wrong silicon."* The manifest half is correct. The KB half does not hold: the KB hardware dimension comes from `kb_hardware_slug(gpu_type, ...)` in `recipe_snapshot_constants.py`, and `gpu_type` originates from the CLI `--gpu-type` argument, not from `detect_gfx_arch()`. The only consumer of `detect_gfx_arch()` is `build_provenance()`, which lands in `session/manifest.py` as the `gfx_arch` field.

The bug is still worth fixing — recording a false fact and suppressing the probe that would have answered correctly are both real — but the blast radius is provenance records, not KB key collisions. Suggest dropping or correcting that clause.

---

## Non-blocking

**The SMT target contradicts this PR's own conclusion.** `087428f` removes the SMT warning with a good argument: *"SMT is on by default on EPYC, so it fired on nearly every node, and an alert that always fires trains readers to ignore the rest of the check."* But `KNOBS["smt"]["target"]` is `["disabled"]`, and I verified `verdict('smt', 'Enabled')` returns `FAIL`. Combined with the README framing this for CI ("Exits non-zero when a knob is off target"), the check is red on nearly every EPYC node from day one and will get `|| true`'d. Either demote SMT to record-only for consistency with the other two call sites, or document why the standard differs here.

**The target profile has no cited basis, and NPS accepts both answers.** The `KNOBS` targets look like the standard EPYC low-latency/HPC recommendation set, but nothing in the file cites a source, a document version, or why that profile suits LLM inference specifically (the host CPU here is doing sampling, scheduling, and tokenization, not HPC compute). Most visibly, `nps` targets `["nps1", "nps4"]` — two settings with opposite tradeoffs (NPS1 interleaves memory across the socket for bandwidth, NPS4 favors locality). Accepting both means this knob can essentially never FAIL. If there is no conclusion, recording it is more honest than presenting it as a check. Same for the `0.90` boost ratio and the `8` MHz spread threshold — both are undocumented magic numbers, and on amd_pstate `cpuinfo_max_freq` *is* the boost ceiling, so the `peak > ceil_mhz` branch of the determinism inference is effectively unreachable and the "Overshooting the advertised ceiling" rationale in the docstring does not match runtime behavior.

**An ADMINISTRATOR account is minted before discovering the connection will not work.** The README's headline invocation is `sudo python3 scripts/platform_audit.py` with neither `--ca-cert` nor `--insecure`. Since BMCs almost universally ship self-signed certs, that command mints the account over KCS, opens the LAN channel, fails TLS verification, and revokes — an admin account exposed on the management network for a request that was never going to succeed. Suggest an unauthenticated `GET /redfish/v1` (the spec allows it) as a reachability/TLS probe before entering `TempBmcAccount`. Relatedly, `--insecure` plus the temp-account path sends freshly minted admin credentials to an unverified peer; one stderr line may be too quiet for that combination.

**The README does not mention that the default path creates a BMC account.** "Checks the EPYC tuning knobs" gives no hint that the tool will create and revoke a temporary ADMINISTRATOR account on the BMC. Many sites' security policies prohibit this outright, so it needs to be in the README, not only in the commit message and the class docstring.

**`report.py`'s socket count is more fragile than the other two copies of the same logic.** This PR now has three implementations of the same sysfs read:

```python
sockets = len(
    {p.read_text().strip() for p in cpu_root.glob("cpu*/topology/physical_package_id")}
)
```

`preflight.py` uses `{_read_sysfs(p) for p in ...} - {""}` and `platform_audit.py` uses `{read(p) for p in ...} - {""}`. Only the `report.py` copy leaves `read_text()` unguarded and skips the empty-string filter. An unreadable file there is caught by the outer `except Exception` and drops the **entire** platform block to `None` with only a `log.debug` — not one field degraded, all of them gone. Suggest one shared helper in `hyperloom.common` (`_read_sysfs` plus the socket/NUMA/NPS derivation) used by all three. Also, `platform: null` currently cannot distinguish "probe failed" from "not on Linux sysfs", which matters precisely when someone is trying to explain a delta after the fact.

**`_platform_fingerprint()` adds real I/O and a subprocess spawn to a dozen-plus report unit tests.** It hangs unconditionally off `_build_summary_dict()`, which is exercised by `test_report.py`, `test_roofline_comparison_pipeline.py`, `test_baseline_failure_summary.py`, `test_multimodal_gate.py`, and `test_report_helpers_coverage_unit.py`. On a Linux CI runner `/sys/devices/system/cpu/smt/active` exists, so every one of those calls really reads sysfs and — via `detect_gfx_arch(os.environ)` with the default `probe=True` — spawns `rocminfo` with a 3s timeout. Nothing monkeypatches it. This makes those tests hardware-dependent and slower.

**792 lines of security-sensitive code with no tests, most of it trivially testable.** `epyc_generation`, `resolve`, `verdict`, `effective`, `exit_code`, and `bmc_users` are all pure functions needing no hardware. One assertion would have caught the determinism issue above. I understand `scripts/` sits outside `[tool.coverage.run].source` and `testpaths`, so this matches repo convention — but given that this script creates admin accounts on a BMC and its output is consumed as a configuration verdict, a pure-function test module seems worth the exception. `_platform_fingerprint` and `_check_platform_tuning` are also untested and would only need `_read_sysfs` monkeypatched.

**Minor**

- `KNOBS["core_performance_boost"]["target"]` includes `"auto"`, which is dead: `verdict()` returns `UNKNOWN` for `v == "auto"` before consulting targets. Verified: `verdict('core_performance_boost', 'Auto')` -> `UNKNOWN`.
- `os_layer`'s `nps_effective` does not guard `nodes == 0` the way `preflight` does, so a container without `/sys/devices/system/node` yields `NPS0`, which then FAILs. Verified: `verdict('nps', 'NPS0')` -> `FAIL`.
- `final.md` renders `None NUMA nodes / None sockets` when those fields are `None`.
- With `ipmitool` absent from PATH, `bmc_ip()` returns `""` and the report says `"BMC not discoverable (no KCS/LAN)"`, reporting a missing tool as an unreachable BMC.
- `NAME_BLOCKLIST`'s bare `sev` will match any attribute name containing that substring; a word boundary would be safer.
- The platform block cannot distinguish MI300X from MI325X, since both are `gfx942`. If that distinction matters for provenance, `/sys/class/drm/card*/device/product_name` or a rocm-smi model string would resolve it.

---

## Suggestion: split the BMC layer into its own PR

The gfx fix, the preflight check, and the report fingerprint are small, low-risk, and clearly justified (with the corrected motivation above); they can go in once the silent-failure issues are addressed.

For `platform_audit.py`, I would split it:

- **The OS layer** (sysfs + MSR + measured frequency, covering boost, SMT, NPS, governor, determinism) needs no credentials, never touches the BMC, and carries no security risk. It is also what operators will reach for day to day. This can land now.
- **The Redfish/BMC layer** — minting a temporary ADMINISTRATOR account on the service processor to reach three additional knobs — should be its own PR with a security review. Its risk class is nowhere near the rest of this change; I found two paths above that leave an account behind while the exit code reports success. It would also help to quantify what APBDIS and DF C-states are actually worth for this workload, since that is the benefit being weighed against the risk.

Either way, the target profile needs a cited basis, NPS and SMT need either a justification or demotion to record-only, and the determinism verdict bug needs fixing before this is used to make configuration decisions.

xiaofei-zheng:
Follow-up to my review above — three items I initially set aside. Adding them for completeness, since the first one turns out to be the same theme as the blocking issues rather than a separate nitpick.

### The account-minting path is entirely unchecked

`2b5eba3`'s message says "Each step is now checked", but that only holds for the revoke path. Every step that *grants* access goes through the failure-swallowing `sudo()`:

```python
sudo(["ipmitool", "user", "set", "name", str(slot), self.user])
if not set_bmc_password(slot, pw):
    return self
self.slot = slot
sudo(["ipmitool", "user", "priv", str(slot), "4", "1"])
sudo(["ipmitool", "user", "enable", str(slot)])
sudo(["ipmitool", "channel", "setaccess", "1", str(slot),
      "callin=on", "ipmi=on", "link=on", "privilege=4"])
```

`set_bmc_password` is the only checked step. If `user enable` or `channel setaccess` fails silently, the script still hands the credentials to Redfish, gets a 401, and reports `"{host} (temp account made; {err})"` — which points the operator at an authentication or OEM-compatibility problem when the actual cause is that the account was never enabled, or never granted channel access. Same class as items 2 and 3 in the review above: a failure that produces a plausible-looking wrong answer instead of saying "I don't know."

Suggest `sudo_run()` for these four with the errors collected, and surfacing them through `bmc_note` so the failure names itself instead of being rediscovered as an HTTP status.

### `set_bmc_password` parses ipmitool's success string

```python
return ok and "successful" in (out + err).lower()
```

This depends on ipmitool's exact wording, which varies across versions and is not stable under a non-English locale. I left it out of the main review because it fails safe in both directions — a false negative aborts the mint in `__enter__`, and adds a `REVOKE_FAILURES` entry in `__exit__`, so it errs toward exit 3 rather than toward leaving credentials behind. Still worth a comment recording that dependency; the return code is probably the better primary signal, with the string match as corroboration rather than the sole criterion.

### `epyc_generation` returns "unknown" for non-numeric model numbers

`re.search(r"EPYC\s+(\d{4})", model)` does not match parts like `EPYC 9V84` (cloud-specific SKUs). Verified: returns `unknown`.

This is a correct degradation rather than a bug — it declines to answer instead of guessing a generation — so it only matters if such SKUs are in scope for this audit. If they are, note that the generation-keyed boost-ceiling expectation is simply unavailable on those hosts, which feeds into the `core_performance_boost` verdict.

ZhengGong-amd:
Checked out the branch and ran the changed code locally: the pure functions in `platform_audit.py`, plus `_platform_fingerprint()` and `_check_platform_tuning()` against this host's sysfs.

@xiaofei-zheng's review covers most of what I found. I independently reproduced two of the blocking items and won't restate them:

- `verdict('determinism', 'Performance (or Power at a uniform bin)')` → `PASS`. Confirmed.
- `account_enabled()` answering `False` on a read that failed, while being the trust anchor for the whole revoke. Confirmed.

Below is what I'd add. The first item is the one I'd want resolved before merge; the rest are smaller.

---

## The gfx fix removes the wrong source without wiring up the right one

Dropping `PYTORCH_ROCM_ARCH` is correct, and the escalation in `e27eb1d` from "screen by value shape" to "exclude outright" is the right call — a single-valued `PYTORCH_ROCM_ARCH=gfx942` is wrong in exactly the same way while looking plausible. But after the removal, `_GFX_ENVS` is down to `HYPERLOOM_GFX_ARCH` and `GFX_ARCH`, and **neither is written anywhere in this repository**. Grepping the whole tree (all file types, including shell and YAML) turns up exactly two readers — `common/provenance.py` and `kernel/tools/bypass_trace_analysis.py` — and no writers. Neither variable appears in `.env.template` or `docs/reference/environment-variables.md` either.

So in practice `detect_gfx_arch()` now rests entirely on `rocminfo` being on `PATH`. It lives in `/opt/rocm/bin`, and neither `install.sh` nor `install_baremetal.sh` adds that to `PATH`. The bare-metal install is also precisely the path that sets `PYTORCH_ROCM_ARCH` today, so on those nodes `gfx_arch` goes from "a possibly wrong value" to `null`.

`null` beats a wrong value, so this is still net positive. But the right answer is already in the tree:

```python
# src/hyperloom/inference_optimizer/gpu_types.py
_AMD_GPU_DISPATCH_IDENTITIES: dict[str, tuple[str, int]] = {
    "mi300x": ("gfx942", 304),
    "mi308x": ("gfx942", 304),
    "mi325x": ("gfx942", 304),
    "mi355x": ("gfx950", 256),
}
```

`--gpu-type` is locked for the session and is already the KB's hardware dimension. And in `build_provenance()`, `gfx_arch` is the only hardware field that does not accept an `args` override — the line directly above it does:

```python
"gpu_type": (_arg_first(args, "gpu_type") or _env_first(env, "GPU_TYPE")),
"gfx_arch": detect_gfx_arch(env, probe=probe),
```

Suggest resolution order: `HYPERLOOM_GFX_ARCH` → table lookup from `args.gpu_type` → `rocminfo`. That also gives the `probe=False` hermetic path a real answer instead of only ever `None`. `gpu_types.py` imports nothing but `os`, so moving that table into `hyperloom.common` doesn't invert the layering.

Also worth noting that this makes the change unfalsifiable in the failure case it fixes: there is no test asserting that a node with no gfx env and no `rocminfo` reports `null` rather than a stale value, and no operator-visible signal when that happens.

## `--quick` can never exit 0

`os_layer(quick=True)` sets `determinism_inferred = None`, so that knob resolves to `unknown` → `UNKNOWN` → `exit_code()` returns 2, unconditionally. Given the README frames this script for CI use, a flag whose whole purpose is to make the run fast guarantees a non-zero exit. Either `--quick` should exclude the measurement-dependent knobs from the verdict set, or the README should say it is not usable as a gate.

## The `platform` block is all-or-nothing, and its absence is invisible

Beyond the socket-count fragility already flagged: the `except Exception` in `_platform_fingerprint()` wraps the entire body, so one unreadable `physical_package_id` doesn't degrade one field — it drops CPU, SMT, NUMA, governor, boost, kernel, GPU, and stack together, leaving `platform: null` and a single `log.debug`. At default log level that is invisible.

The comment calls it a "cosmetic probe", which sits awkwardly against this PR's own argument that the data is needed to explain a delta after the fact. If it is cosmetic, the `WARN`-only framing is right and the log level is fine; if it is provenance, a failed probe should be at `warning` and should degrade per-field. I'd argue for the latter, consistent with how `detect_stack_fingerprint` handles a missing component (`"unknown"` per key rather than dropping the dict).

Related: `platform: null` cannot distinguish "not on Linux sysfs" from "the probe threw", which matters exactly when someone is reading an archived `final.json` months later.

## `__exit__` doesn't cover SIGTERM

The docstring accounts for SIGKILL and reboot, but nothing is registered for SIGTERM or SIGINT — no `signal.signal`, no `atexit`. SIGTERM is what a CI `timeout`, a job scheduler, or an orchestration layer sends, and it is far more common than SIGKILL. Since the consequence is a live ADMINISTRATOR account on the BMC with a known password, and the fix is a handler that re-raises after running the context manager, this seems worth covering.

## Smaller things

- **`write_minimal_final_json()` has no `platform` block.** The crash-safe fallback is arguably when a platform record is most useful, since nobody was there to watch the run. The `model_gate.py` paths go through `_build_summary_dict` and are fine.
- **`report.py`'s `_read_sysfs` is used to read `/proc/cpuinfo`.** Minor, but if the three copies of this helper get consolidated into `hyperloom.common` as suggested, the name is worth fixing on the way.
- **Bandit doesn't see the new script.** `bandit (medium+, production code)` passes because it scans `src/hyperloom`, not `scripts/`. Running it manually gives 6 findings on `platform_audit.py` (1 medium, which is a false positive on the `"0.0.0.0"` literal; the rest low). Nothing alarming, but for a file that mints BMC credentials it's worth having the scanner actually cover it.

## Documentation

This is the strongest part of the PR — the `detect_gfx_arch` docstring explains the failure, why the first attempt was insufficient, and why the variable is excluded outright, and the inline note on `_GFX_ENVS` pointing at `framework/targeted_build.py` should stop someone adding it back. Three small drifts:

- `docs/reference/environment-variables.md:282` describes `PYTORCH_ROCM_ARCH` as "Detected... Set automatically from the session `--gpu-type`". Accurate for its role as a compile target, but a reader can easily carry that over to arch detection — which is the exact misreading this PR exists to prevent. Half a sentence noting it does not participate in provenance detection would close the loop.
- README lists six knobs; `KNOBS` has seven (`power_profile` is missing).
- README's "Exits non-zero when a knob is off target" understates the contract: 2 means unresolved and 3 means a BMC account could not be revoked. Those need very different handling in CI, and 3 in particular should not scroll past as a generic failure.

---

## Summary

The gfx fix is real and the direction is right, but I'd call it half-finished: the wrong source is gone, the correct one that already exists in the tree isn't connected, and detection now hangs on a binary that isn't guaranteed to be on `PATH`.

The platform record I'm in favour of — `multi_node_session` marking the gap rather than papering over it is exactly the right instinct. The remaining issues are the three duplicated sysfs implementations and the all-or-nothing `except`.

On `platform_audit.py` I agree with the split @xiaofei-zheng proposed. The OS layer needs no credentials and carries no security risk; the Redfish layer — creating an ADMINISTRATOR account on the service processor — is in a different risk class from everything else here, and there are currently paths that leave the account behind while the exit code reports success. That half deserves its own PR and its own review.

rpoornac:
Thank you both — this was a thorough review and I verified every claim before
changing anything. **All of them hold.** Two are worse than described, and the
request for a cited basis was the most productive of them: chasing it down
removed two targets and put a citation and an explicit caveat behind each of the
three that remain.

I've taken @xiaofei-zheng's recommendation to split. Two branches are pushed:

- `fix/rpoornac/platform-tuning-os-layer` — gfx fix, preflight check, report
  record, and an OS-only auditor. No credentials, no security surface.
- `feat/rpoornac/platform-audit-bmc-layer` — the Redfish layer, on top of the
  first, for its own review.

Commit messages have been rebuilt, since two of them argued something untrue.

---

## Blocking

### 1. `verdict()` substring match on determinism — confirmed, and narrower than it looks

```
verdict('determinism', 'Performance (or Power at a uniform bin)')  ->  PASS
verdict('determinism', 'performance')                             ->  FAIL
```

The bare value fails correctly. It is specifically the OS-inferred path — the
one that runs when there is no BMC, i.e. the common case — that produced the
false PASS. The check silently approved the single configuration it exists to
catch, and only on the path where it is the sole line of defence.

Fixed as suggested: `infer_determinism()` returns a normalized `"power"` /
`"performance"` / `None`, with the caveat moved to a separate `note` field used
only for rendering. Matching is now exact after normalization — no target needed
substring semantics, so nothing was lost by removing it.

The target itself is unchanged and still Power, now cited to AMD 58011 §4.2.2;
see the citation section for why, and for the inference caveat that a FAIL on
this knob now prints.

I also dropped the overshoot comparison against `cpuinfo_max_freq`: under
`amd_pstate` that file *is* the boost ceiling, so the branch could never fire.

### 2. `__enter__` can skip revocation entirely — confirmed

`sudo(["ipmitool", "user", "set", "name", ...])` ran before `self.slot` was
assigned, so a password failure returned with `slot=None`, `__exit__` returned
immediately, `REVOKE_FAILURES` stayed empty, and the exit code reported success
— after the account name had already been written, and on the crash-recovery
path with an enabled ADMINISTRATOR account left live.

The in-code comment claiming the slot was claimed early was accurate only for
the steps *after* the password. Fixed by assigning `self.slot` as soon as the
slot is resolved, before any mutation. Revoking a slot that was never enabled is
idempotent and free; the reverse is not.

A sentinel found already enabled is now *recorded* as a credential exposure
rather than only printed, so it reaches the exit code — it is the fingerprint of
a previous run that could not clean up.

### 3. `account_enabled()` false on read failure — confirmed

It sat on the failure-swallowing `sudo()`, so a BMC that errored, timed out, or
formatted its output differently answered "not enabled" — a false confirmation
at the exact point the design calls its trust anchor. Replaced by
`account_status()` returning `"enabled"` / `"disabled"` / `None`, where `None`
means "could not read" and is appended to the credential-failure list.

While there, the four mint-path calls are checked too. My commit message claimed
"each step is now checked", which was only ever true of the revoke path. A silent
failure there produced a 401 from Redfish that pointed the operator at
authentication or OEM compatibility when the real cause was an account that was
never enabled.

### 4. The motivation claims do not hold — confirmed, and I had it wrong

You're right on both counts, and this is the correction I'm most glad to have.

Baseline and trials run on the same node, so BIOS state applies to both sides
and cancels out of the delta. "A delta credited to a code change could equally
have been SMT" does not describe a failure mode of a single session, and I
should not have written it.

I verified your alternatives rather than just accepting them. `--compare-against-gpu`
does pull an InferenceX reference into `target_analysis/target_baseline.json`,
which the report renders as "External baseline" — so cross-machine comparison is
real, just not where I claimed. The strongest argument is the one you gave: on a
de-tuned node the delta stays fair but the *optimum* is not, the optimizer
searches around a CPU-side bottleneck that does not exist on a correct machine,
and the winning config is then filed in the recipe KB for other people.

`bcf4116` is also wrong about the KB. `kb_hardware_slug()` takes `gpu_type`,
which comes from the CLI argument and never from `detect_gfx_arch()`. The blast
radius was the `gfx_arch` field of the provenance record. Both messages have been
rewritten, and the corrected reasoning is now in the code docstrings, where it
will be read.

---

## The gfx fix was half-finished — confirmed (@ZhengGong-amd)

This is the most useful finding in the review, because the change was net
positive and still wrong to stop where it did.

Confirmed by grep: `HYPERLOOM_GFX_ARCH` and `GFX_ARCH` have no writer anywhere
in the tree — only readers and tests. And no install script puts `/opt/rocm/bin`
on `PATH`. So detection rested entirely on a binary that isn't there, on exactly
the bare-metal nodes that set `PYTORCH_ROCM_ARCH` — turning a wrong value into
no value.

Adopted your suggested order: `HYPERLOOM_GFX_ARCH` → `--gpu-type` via the board
table → `rocminfo`. The table moved to `hyperloom.common.gpu_identity` and
`gpu_types.py` re-exports it, so there's one table rather than two. As you noted,
this also gives the `probe=False` hermetic path a real answer instead of only
ever `None`, and it means report generation no longer spawns a subprocess at all.

Worth noting for the record: the table is many-to-one — MI300X, MI308X and
MI325X are all `gfx942` — so the arch identifies the ISA, not the board, and
`gpu_type` remains the authority for anything that must tell them apart.

---

## Non-blocking, all addressed

| Finding | Resolution |
|---|---|
| Target profile has no cited basis | Every judged knob now cites AMD 58011; SMT and NPS dropped to record-only because ch. 5 varies them by workload |
| SMT target contradicts the PR's own conclusion | SMT is now recorded, not judged |
| NPS accepts both values, can never FAIL | Also recorded, not judged — the tradeoff is real and unsettled for this workload |
| No operator-visible signal when gfx resolves to null | Preflight now warns and names the three ways to fix it |
| `--quick` guarantees a non-zero exit | Measurement-dependent knobs report `SKIPPED`, excluded from the exit code; `--quick` now exits 0 on a good node |
| Three duplicated sysfs implementations | One `hyperloom.common.platform_probe`, used by preflight and report |
| Platform block is all-or-nothing and invisible | Per-field degradation, `warning` not `debug`, and a `status` field so "no sysfs" and "probe threw" are distinguishable |
| `_platform_fingerprint` adds I/O to unit tests | Resolves from `gpu_type` with `probe=False`; 141 report tests now run in 0.9s with no subprocess |
| No tests on security-sensitive code | 39 tests across both scripts, covering verdicts, determinism inference, exit-code ranking, EPYC parsing, core sampling, and the account lifecycle |
| Account minted before the connection test | Unauthenticated `GET /redfish/v1` settles reachability and TLS before anything is minted |
| README doesn't mention account creation | It does now, prominently |
| SIGTERM leaves the account behind | `SIGTERM`/`SIGINT` raise so the context manager unwinds; SIGKILL remains uncoverable, which is why a stale sentinel is treated as evidence |
| `write_minimal_final_json` has no platform block | Added — a run that died unattended is when it's most useful |
| `_read_sysfs` used for `/proc/cpuinfo` | Renamed `read_kernel_file` during consolidation |
| Bandit doesn't scan `scripts/` | CI now scans it; the `0.0.0.0` medium is a false positive (it rejects ipmitool's unconfigured-LAN sentinel, not a bind address) and is annotated |
| `"auto"` in the boost target is dead | Removed |
| `nps_effective` doesn't guard `nodes == 0` | Guarded in the shared helper, with a test — NPS0 is a value no BIOS can hold and read downstream as a real misconfiguration |
| `final.md` renders `None` | Unresolved fields render `?` |
| ipmitool absence reported as BMC unreachable | Distinguished |
| Doc drifts (env-variables, knob count, exit codes) | All three fixed; the README now states that `2` means unresolved and `3` means a credential problem |

### One correction to the `sev` suggestion

`\bsev\b` over-corrects. BIOS attributes are CamelCase, so `SevControl` has no
trailing word boundary and would have stopped being blocked — the opposite bug.
And under `re.IGNORECASE` a bare `[a-z]` lookahead also matches `C`, which
defeated my first attempt. The pattern is now `\bsev(?!(?-i:[a-z]))`, tested in
both directions: `SevControl` and `SEV_SNP_Support` blocked, `SeverityLevel`
allowed. My own test caught this, which is a fair argument for the tests you
asked for.

---

## The cited basis — and why it removed a target

Source: AMD, *BIOS & Workload Tuning Guide for AMD EPYC 9004 Series Processors*,
[pub. 58011 rev 1.0](https://docs.amd.com/v/u/en-US/58011-epyc-9004-tg-bios-and-workload),
ch. 5 "Workload-Specific BIOS Settings". It covers 9004 (Genoa); the knobs and
their semantics carry forward to 9005 (Turin), which is what the fleet runs.

Reading it closely argues against part of what I built, so I'd rather lead with
that than bury it.

**Chapter 5 does not publish one performance profile.** It publishes a column per
workload — CPU-intensive, Java throughput, Java latency, virtualization,
database, HPC/Telco — and most rows differ between them. That is a stronger form
of the point you both made: it is not just that my targets were uncited, it is
that AMD declines to give a single answer for several of them.

SMT and NPS are therefore recorded rather than judged, which also answers the
"can never fail" objection directly: chapter 5's own SMT and NUMA rows differ per
workload, so there is no single answer to assert.

**Determinism stays targeted at Power**, and it is worth explaining why, because
I briefly removed it and was wrong to. §4.2.2:

> **Performance** — "Uniform performance across identically configured systems in
> a datacenter."
> **Power** — "Maximum performance of any individual system by leveraging the
> capabilities of a given CPU to the maximum, resulting in a varying performance
> range across the datacenter or larger deployments."

My first reaction was that Power "makes systems differ" and so works against the
cross-machine comparability this PR argues for. That conflates two separable
things. Hyperloom's job is to find the best a given platform can do, and §4.2.2
is explicit that Power is how a platform delivers it; Performance determinism
buys fleet uniformity by leaving headroom unused on the better parts, which is
the opposite of what an optimizer is for. The cost of Power is paid by *recording*
the setting, not by avoiding it — every report already carries the mode each run
was in, so a system-to-system delta stays explicable. Constraining the setting to
buy comparability would have been paying for it twice, the second time in
throughput.

The judged set is therefore:

- **Core Performance Boost** — Enable/Auto by default, and no chapter 5 profile
  disables it (§4.1.3).
- **cpufreq governor** — an OS setting and therefore outside 58011. I've said so
  in the code rather than borrowing the guide's authority for it; the basis is
  that a ramping governor distorts latency percentiles, and it is the knob most
  often wrong on a freshly imaged node.
- **Determinism control** — Power, per §4.2.2, for the reason above.

One caveat now stated in the tool rather than left implicit: the OS layer infers
determinism from per-core frequency spread, and a uniformly binned part running
Power determinism produces the same low spread as a part running Performance
determinism. That verdict is marked `inferred`, and a FAIL prints a line pointing
at BIOS (or the BMC auditor) rather than inviting someone to change a BIOS
setting on the strength of a heuristic. Readings between the two thresholds
report "too close to call".

Since determinism needs load generation, `--quick` reports it `SKIPPED` rather
than UNKNOWN — which is what keeps a fast run usable as a gate.

The BMC layer's three knobs fared better — all are either the documented default
or the value chapter 5 selects for its latency-sensitive columns, and each cites
§4.2.1, §4.4.3 or §4.4.4 inline.

The spread thresholds remain heuristics, and are labelled as such, with an
ambiguous band that reports "too close to call" rather than picking a side.

## Operator-visible signal for null gfx

Fair point that the failure case was unfalsifiable. Preflight now warns when no
source resolves, and names all three fixes:

```
Preflight: WARNING — GPU architecture could not be resolved; provenance will
record gfx_arch as null, so an archived report will not say which ISA produced
its numbers. Pass --gpu-type (mi300x/mi308x/mi325x/mi355x), set
HYPERLOOM_GFX_ARCH, or put rocminfo (/opt/rocm/bin) on PATH.
```

Warn-only, and deliberately before the run: null provenance is cheap to fix
beforehand and impossible to fix afterwards.

---

## Closing this PR

I'm closing #1097 in favour of the two split PRs rather than force-pushing over
it, so your inline comments stay attached to the code they refer to. The branch
`fix/rpoornac/gfx-arch-and-platform-capture` stays in place for the same reason —
please don't delete it, or the line-level comments above lose their anchors.

- OS layer, gfx fix, preflight, report record → #https://github.com/AMD-AGI/Hyperloom/pull/1155
- Redfish/BMC layer → #https://github.com/AMD-AGI/Hyperloom/pull/1156, stacked on the first (1155)

## Verification

- Clean worktree: **11 failed, 11194 passed** — the same 11 pre-existing
  environment-dependent failures as the merge base, unchanged. The 5 extra I saw
  locally were artifacts of my own working clone (stale `egg-info`, a local
  `.env`), not regressions.
- `ruff` clean across all changed files.
- The OS auditor runs correctly on an MI355X node: `--quick` exits 0, the full
  run measures 5044 MHz peak / 33.3 MHz spread and passes all three judged knobs.

PR #1101, opened 2026-08-05
feat(trace): emit Explore spend telemetry and report failed LLM calls

luochen-amd:
## Summary

Two independent pieces of work, both closing gaps in what Langfuse can see.

### 1. Explore spend telemetry

- persist completed EXPLORE wall-clock segments across macro cycles and include the live segment while EXPLORE is active
- publish KB hit status, Explore elapsed seconds, total session elapsed seconds, and Explore ratio through the existing Langfuse `session_status` mirror
- keep pre-telemetry resumes explicitly unknown and minute-bucket volatile clocks so frequent state saves do not flood Langfuse

### 2. Failed LLM calls are now reported

A failed call never reached the ledger at all. The row is built from `BackendTurnResult.metadata`, which only exists once a turn has succeeded, so a turn whose model call never returned left no trace anywhere. Langfuse consequently showed **zero errors across 30,803 generations in 24h** while calls were demonstrably failing (`litellm.BadRequestError: AnthropicException` in the task logs), which made the LLM error rate unobservable — an absent row reads as "never happened", not "failed".

**The ledger.** `llm_calls.jsonl` gains a terminal `status` plus `error_type` / `error_message`, guarded by a closed vocabulary the way `component` already is. `LLMCallRecord.for_failure()` covers the no-result path; token counters stay `None` because nothing was measured, and the message is capped since litellm wraps the upstream payload verbatim. The emitter maps `status` onto the Langfuse observation `level` / `status_message`.

**What counts as a failure.** `BackendError` is far too broad to drive an error rate — it also covers a missing `--review` path, an unreadable `emit.json`, an unsupported runtime phase, an absent SDK, none of which touched a provider. A new `LLMCallFailed` marker is raised only around actual provider calls (Codex chat + responses, the Claude stream, both critic transports) and only that is recorded. It subclasses `BackendError`, so retry and error-streak accounting are untouched by the marking.

**Who writes the row.** Wiring only the reactor would have missed most of it, because the other producers never let a failure reach the Coordinator: the critic self-writes its rows, the specialist turn loop catches and breaks, and the scorer folds per-model exceptions into an `errors` map via `gather(return_exceptions=True)`. Each now writes at its own call site, where the model, task and phase are still known — after the `gather` fold, "which model failed" is all that survives. And exactly one writer per failure: a backend exposing `set_trace_context` already owns its trace rows, so the Coordinator's failure path honours that hook instead of adding a second, poorer row for the same failure.

Three consequences that were not obvious and are worth a reviewer's attention:

- **failures skip the pair buffer.** They have no conversation half, so waiting for one would hold the failure until session end — or let the status-blind `pair_key` marry it to a neighbouring successful call.
- **the spend rollup filters them out.** `_fold_call_into_bucket` increments `calls`, so counting failures would inflate call counts and skew per-decision attribution. `calls` and token totals are unchanged by this PR; failure visibility is Langfuse's job, not the rollup's.
- **`_start_obs` now degrades one kwarg rung at a time.** An SDK predating `level` / `status_message` must lose those two fields rather than the whole observation.

**One behaviour change beyond telemetry.** A Claude provider failure now raises `LLMCallFailed` instead of propagating the raw SDK exception, so it lands on the backend-error streak rather than `crash_count`. That is the more accurate bucket for an upstream fault and it matches what Codex already did, but it is a real change in failure handling, not just tracing. Cancellation stays a bare `BaseException` and is re-raised untouched; `NoIntentEmitted` is raised outside that block, so a call that succeeded with unusable output remains distinct from one that failed.

**Deliberately not covered: the kernel path.** It is a post-hoc scrape of an out-of-process child that prints `FORGE_LLM_USAGE` only on success, so a missing marker cannot be told apart from a failed call, and the attempt dict carries only business outcomes (`improved`, `best_ms`). Synthesizing an LLM error from those is the same conflation this change removes. Closing it needs the child (GEAK / KernelForge `forge_submit`) to emit a failure marker of its own; the reasoning is commented at the call site.

Backward compatible: rows written before `status` existed carry no such key and are read as successes, so backfilling an old session does not turn it red.

### Drive-by

`explore_elapsed_accum_s` was added to `policy.CORE_STATE_FIELDS` by (1) but not to the robustness role's vendored copy, leaving both lock-step contract tests (`test_role_contract` and `test_agent_roles_and_policy`) red on this branch since `a026ff06f`. Synced, mirroring the upstream entry's position in the phase-state-machine group.

## Test plan

- [x] `test_llm_trace_unit` + `test_langfuse_emitter` + `test_claude_backend` + `test_codex_backend` + `test_trace_utils` + `test_trace_edge_coverage` + `test_role_contract` — 184 passed, 1 skipped (no `claude_agent_sdk` locally)
- [x] `test_proposal_scorer` — 19 passed (5 deselected; they import the Coordinator, see below)
- [x] New coverage: failure row shape / message truncation / default status / closed-vocabulary rejection; `LLMCallFailed` still catchable as `BackendError`; Codex chat + responses raising the marker; Claude marking a non-timeout stream error and *not* marking cancellation; scorer writing an error row through the `gather` fold; immediate emit at `level=ERROR`; legacy rows staying `DEFAULT`; both `_start_obs` degradation rungs
- [x] Regression for the double-write: a self-tracing backend's provider failure yields **exactly one** row, and it is the backend's richer one (real review model)
- [x] End-to-end by hand: write one success + one failure row, confirm the level mapping, confirm the spend rollup keeps only the success
- [x] `ruff` / `pylint` / `bandit` / CodeQL clean
- [x] Code CI green — 24/24 (all 8 test shards, both coverage jobs, CodeQL, lint, docs, packaging)

Not runnable locally: anything importing the Coordinator, because `recipe_kb/local_store.py` imports `fcntl` (Linux-only). That covers the three Coordinator tests added here — marker discrimination, per-turn row count, and the exactly-one-row regression — plus `test_critic_agent_backend`. All are covered by the CI shards.

## Follow-up

Production Langfuse traffic is still dominated by long-running sessions launched from older checkouts, so the new `status` field is not yet visible there. Worth re-checking once tasks start from this code — the query is `level=ERROR` on `/api/public/observations`, which should stop returning an unbroken run of `DEFAULT`.

## Downstream

Cortex already consumes the Explore telemetry fields in its runtime-status pipeline; older sessions continue to render unknown values. The `llm_calls.jsonl` schema change is additive, and the breakdown collectors treat a missing `status` as success.

### DIFF
--- a/src/hyperloom/agents/robustness/role/envelope.py
+++ b/src/hyperloom/agents/robustness/role/envelope.py
@@ -137,6 +137,7 @@ CORE_STATE_FIELDS: frozenset[str] = frozenset(
         "phase_started_unix",
         "phase_history",
         "phase_budget_pct",
+        "explore_elapsed_accum_s",
         # Cyclic phase-machine state; locked so an LLM update_state cannot forge
         # macro-cycle / convergence / per-cycle budget state.
         "macro_cycle",
--- a/src/hyperloom/inference_optimizer/breakdown/collectors/decision.py
+++ b/src/hyperloom/inference_optimizer/breakdown/collectors/decision.py
@@ -44,6 +44,15 @@ _TOKEN_CACHE_CREATE_KEY = "cache_creation_input_tokens"
 _TOKEN_CACHE_READ_KEY = "cache_read_input_tokens"
 
 
+# Terminal-status key + its success value, re-declared for the same reason as the
+# token keys above. A ``status="error"`` row records a call that never returned,
+# so it carries no tokens and must stay out of every spend / call-count rollup.
+_STATUS_KEY = "status"
+
+
+_STATUS_OK = "ok"
+
+
 _TOKEN_KEYS_ALL: tuple[str, ...] = (
     _TOKEN_IN_KEY,
     _TOKEN_OUT_KEY,
@@ -105,12 +114,18 @@ def _load_llm_calls(
     session_dir: Path,
     warnings: list[str],
 ) -> list[dict[str, Any]]:
-    """Read every LLM-call row from the trace ledger and ext shards.
+    """Read every *successful* LLM-call row from the trace ledger and ext shards.
 
     Merges ``reports/trace/llm_calls.jsonl`` with every
     ``reports/trace/ext/*.jsonl`` shard written by out-of-process children.
     Best-effort: missing files / dirs yield ``[]``; malformed lines are
     skipped by :func:`_load_jsonl_safe`.
+
+    Rows whose ``status`` is not ``ok`` are dropped: they describe a call that
+    never returned, so counting them would inflate ``calls`` and skew the
+    per-decision spend attribution. Rows predating the ``status`` field have no
+    such key and are kept. Failure visibility is Langfuse's job (the emitter
+    maps them to ``level=ERROR``), not the spend rollup's.
 
     Args:
         session_dir (Path): Absolute session root.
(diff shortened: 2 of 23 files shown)


ZhengGong-amd (approved):
Problem addressed , approved

PR #1102, opened 2026-08-05
fix(kernel): support runtime-JIT deployment for installed AITER sources

jiaqiang-dot-liu:
## Summary

Installed AITER kernel artifacts could pass Forge micro-validation but never reach E2E. HyperLoom classified `aiter` / `aiter_meta` native sources as compiled, found no eager rebuild command for wheel layouts, interpreted the skipped rebuild as a failure, and immediately restored the original source. In the Mixtral session that motivated this change, k003 and k006 both produced correct, faster micro artifacts and then failed integration with `apply_failed: no rebuild command`.

This PR makes installed AITER deployment explicitly runtime-JIT: apply the complete source snapshot, invalidate the exact cache used by that runtime, start a fresh serving process, and let E2E remain the final KEEP/REVERT gate.

### 1. Target-pinned runtime-JIT deployment

- Recognizes split wheel layouts where Python/JIT code lives under `aiter` and device/codegen sources live under sibling `aiter_meta`.
- Pins `jit/build` to the selected target's install root instead of rediscovering AITER through the orchestrator process's import path or `VLLM_VENV_ROOT`.
- Supports native sources and Python codegen under installed AITER, including compiled headers and JIT sources outside `csrc`.
- Keeps editable `/sgl-workspace/aiter` Python-source behavior unchanged; full `setup.py develop` is not triggered for ordinary codegen edits.
- Records deferred runtime-JIT state in the apply manifest and propagates it through `apply_and_bench` and E2E integration.

### 2. Complete, bounded snapshot deployment

Forge patches can legitimately span AITER plus their vLLM/SGLang wrappers. `site-packages` remains the coordinate system for those relative paths, but it is no longer the write authorization boundary.

- Snapshot writes are limited to resolved `aiter`, `aiter_meta`, `vllm`, and `sglang` package roots.
- Explicit `repo_root=site-packages` cannot authorize writes to unrelated packages such as `torch` or `pip`.
- Write, delete, rename, post-verify, symlink, and multi-node fan-out paths use the same deploy-root containment checks.
- Unknown layouts retain fail-fast behavior and require an explicit valid mapping; they no longer fall back to `target.parent` and create nested paths.
- Discovery-only source-root configuration cannot expand pod write/delete permissions to `/` or arbitrary shallow paths.

### 3. Transactional source and JIT state

Single-node and multi-node integration now treat source files and AITER caches as one transaction.

- Uses unique JIT backup paths so an orphan from a crashed attempt cannot permanently block later optimization.
- Invalidates and backs up `aiter/jit/build` independently on every RayJob/Infera inference pod.
- Preserves every per-host, per-file backup for multi-file snapshots instead of overwriting earlier records with the final file.
- Rolls back successful pods when another pod fails apply.
- Restores source and baseline JIT state together; a missing baseline backup is detected before candidate cache deletion.
- Surfaces partial restore details, including already-restored targets and JIT restore failures, as structured results.
- Final KEEP promotes the applied state to the new baseline and finalizes that transaction, deleting its local and pod-side source/JIT backups. REVERT continues to restore them.

### 4. Safety and compatibility

- Destructive JIT paths require both framework-root containment and the exact `aiter/jit/build` package shape.
- Backup paths remain confined to the configured kernel backup root.
- `apply_snapshot` keeps its legacy optional `deploy_roots=None` calling contract; HyperLoom's integration path always supplies bounded roots.
- New pod CLI fields (`--jit-build-dir`, `--records-json`) have backward-compatible defaults for direct callers and older test fixtures.
- Existing manifest key readers retain compatibility with manifests created before this change.

## Effect on the motivating failure

The old behavior stopped after copying the candidate source:

```text
compiled=true + rebuild_command=[]
→ rebuild.status=skipped
→ apply_failed
→ source and JIT state restored
→ no E2E measurement
```

The new behavior is:

```text
validated Forge snapshot
→ bounded source apply
→ target-pinned JIT invalidation on every serving pod
→ deferred runtime JIT on fresh server startup
→ E2E KEEP / REVERT
```

This PR does not claim an E2E gain for k003 or k006; it restores the missing deployment/measurement path so E2E can make that decision.

## Test plan

- Installed `aiter_meta` native source apply without importable AITER or `VLLM_VENV_ROOT`.
- Import-path/target-path mismatch: only the target runtime cache is invalidated.
- Python codegen, non-`csrc` C++ sources, and CK headers trigger the correct runtime-JIT behavior.
- Editable AITER Python sources remain source-only.
- Cross-package AITER + vLLM snapshot succeeds; unrelated package and escaping symlink writes fail before mutation.
- Cross-process JIT restore, missing backup behavior, unique orphan-safe backups, and final KEEP cleanup.
- RayJob/Infera per-pod JIT invalidation, partial apply rollback, multi-file backup retention, structured invalid JSON, and partial restore reporting.
- Legacy direct `Namespace` callers and optional `apply_snapshot` arguments remain compatible.

Commands run on the final branch:

```text
pytest src/hyperloom/inference_optimizer/tests/test_kernel_patch_jit_transaction.py \
       src/hyperloom/inference_optimizer/tests/test_kernel_node_ops.py \
       src/hyperloom/inference_optimizer/tests/test_multi_node_scripts.py \
       src/hyperloom/agents/kernel/tests/test_apply_kernel_patch_isolated_aiter.py \
       src/hyperloom/inference_optimizer/tests/test_kernel_integrate_and_report.py
# 160 passed

pytest src/hyperloom/inference_optimizer/tests/test_common_utils.py \
       src/hyperloom/inference_optimizer/tests/test_coverage_gap_units.py \
       src/hyperloom/inference_optimizer/tests/test_external_multi_node.py \
       src/hyperloom/inference_optimizer/tests/test_integrate_payload_defaults.py \
       src/hyperloom/agents/kernel/tests/test_kernel_optimization_verification.py
# 274 passed
```

Python compilation, IDE lint diagnostics, and `git diff --check` also pass.

## Linked issues

N/A

## Breaking changes

No. New manifest fields and pod CLI arguments are additive, legacy readers/callers remain supported, and the public `apply_snapshot` fallback remains available.

### DIFF
--- a/src/hyperloom/agents/kernel/tests/test_apply_and_bench.py
+++ b/src/hyperloom/agents/kernel/tests/test_apply_and_bench.py
@@ -215,3 +215,78 @@ def test_gpu_vram_used_none_without_gpu_scope(monkeypatch):
     assert _ab._gpu_vram_used_mb("") is None
     assert _ab._gpu_vram_used_mb(None) is None
     assert _ab._gpu_vram_used_mb("  ,  ") is None
+
+
+def test_apply_and_bench_accepts_and_reports_deferred_rebuild(
+    tmp_path,
+    monkeypatch,
+):
+    patch_file = tmp_path / "kernel.py"
+    target = tmp_path / "target.py"
+    patch_file.write_text("VALUE = 2\n", encoding="utf-8")
+    target.write_text("VALUE = 1\n", encoding="utf-8")
+    manifest = tmp_path / "manifest.json"
+    measurements = iter(
+        (
+            {
+                "status": "ok",
+                "median": 100.0,
+                "tput_spread": {"p25": 99.0, "p75": 101.0},
+                "tpot_spread_ms": {"median": 1.0},
+                "p99_tpot_spread_ms": {"median": 2.0},
+                "vram_used_mb": 1000,
+                "reps": [100.0],
+            },
+            {
+                "status": "ok",
+                "median": 110.0,
+                "tput_spread": {"p25": 109.0, "p75": 111.0},
+                "tpot_spread_ms": {"median": 0.9},
+                "p99_tpot_spread_ms": {"median": 1.8},
+                "vram_used_mb": 1000,
+                "reps": [110.0],
+            },
+        )
+    )
+    monkeypatch.setattr(ab, "_find_benchmark_serving", lambda: tmp_path)
+    monkeypatch.setattr(
+        ab,
+        "_serve_and_bench",
+        lambda *args, **kwargs: next(measurements),
+    )
+    monkeypatch.setattr(
+        ab,
+        "apply_kernel_patch",
+        lambda **kwargs: {
+            "status": "ok",
+            "manifest_path": str(manifest),
+            "rebuild": {
+                "status": "deferred",
+                "mode": "runtime_jit",
+            },
+        },
+    )
+    monkeypatch.setattr(
+        ab,
+        "revert_kernel_patch",
+        lambda manifest_path: {"status": "ok"},
+    )
+    monkeypatch.setattr(
+        ab,
+        "_engagement_proof",
+        lambda *args, **kwargs: {"engaged": True},
+    )
+
+    result = ab.apply_and_bench(
+        patch_path=str(patch_file),
+        target_file=str(target),
+        backup_root=str(tmp_path / "backups"),
+        model="model",
+        out_dir=str(tmp_path / "out"),
+    )
+
+    assert result["status"] == "ok"
+    assert result["applied"][0]["rebuild"] == {
+        "status": "deferred",
+        "mode": "runtime_jit",
+    }
--- a/src/hyperloom/agents/kernel/tests/test_apply_kernel_patch_isolated_aiter.py
+++ b/src/hyperloom/agents/kernel/tests/test_apply_kernel_patch_isolated_aiter.py
@@ -58,7 +58,11 @@ def test_detect_strategy_isolated_aiter_csrc_compiled_no_rebuild_command(akp, tm
     venv_root, aiter_pkg = _make_isolated_aiter(tmp_path)
     monkeypatch.setenv("VLLM_VENV_ROOT", str(venv_root))
     site = aiter_pkg.parent
-    akp._CACHED_KN_TARGET_ROOTS = (str(site) + "/",)
+    monkeypatch.setattr(
+        akp,
+        "_CACHED_KNOWN_TARGET_ROOTS",
+        (str(site) + "/",),
+    )
 
     target = site / "aiter_meta" / "csrc" / "kernels" / "foo.cu"
     strat = akp._detect_strategy(target, allow_unknown_target=False)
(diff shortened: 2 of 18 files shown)


chaojhou (approved):
Approving. Reviewed all 9 commits.

**Verified fixes**

- `assert_aiter_jit_build_allowed` now calls `assert_target_path_allowed` before the shape check, so the `shutil.rmtree` in `restore_aiter_jit_build` is root-contained and not only shape-validated.
- `resolve_patch_target_roots` no longer accepts arbitrary env roots. Requiring the resolved directory to be a known package name under `site-packages`/`dist-packages`, or one of the three `/sgl-workspace` editable roots, is a stronger construction than a denylist -- an env-supplied `/` can no longer widen the write boundary.
- Single-node JIT backups are unique now (`jit_build_{time_ns}`), so a crashed run no longer leaves a leftover that blocks every later apply.
- `6dba96c2` also fixes a real data-loss ordering bug: `restore_aiter_jit_build` used to `rmtree(src)` before checking the backup existed, so a missing backup destroyed the candidate and left nothing to restore. The existence check is now ahead of the removal.

`finalize_patch_records` is bounded correctly -- both the source unlink and the JIT `rmtree` go through `assert_backup_path_allowed`, and it only runs on `decision == "KEEP"`.

**Non-blocking comments**

1. `3e84466` dropped `_AITER_META_CSRC_ROOT` (the bare `/aiter_meta/csrc/`) from the pod-side roots, but `orchestrator/framework/paths.py` still lists it in `_STATIC_PATCH_FALLBACK_ROOTS`. The new `site-packages/aiter_meta/` entries cover the standard layout, so this is likely a correction rather than a regression -- but since split-package aiter is this PR's subject, worth confirming no image ships `aiter_meta` at the filesystem root.

2. The module docstring still says "Superset of `orchestrator.framework.paths._STATIC_PATCH_FALLBACK_ROOTS`". After dropping the `atom` and `/app/xDiT/` roots that is no longer true. The atom removal has no functional effect (atom is single-node-only -- `--nodes>=2` exits 2 per the IR-8 guard -- and this module only serves the multi-node pod paths), so this is just a stale comment, but it should not claim an invariant it no longer holds.

3. `resolve_patch_target_roots` discards a non-conforming env root with a bare `continue`. An operator with a custom layout gets no signal that their root was ignored. A warning on discard would save a confusing debugging session.

4. On finalization: deleting the backups at KEEP removes the ability to undo a KEEP that is later found wrong. That is defensible if the patch and manifest are retained so the change stays reconstructible -- worth stating explicitly in the manifest docs which artifact is the authoritative record after finalization.

**On CI**

The last valid e2e pass on this branch is `265cd5f8f` (~157 min). Every failure since is platform-side, not this PR: `bef0ac8c0` got `HTTP 500: create workload failed` at submit (15s, the code never ran), `489fbcac7` reported `backend reports workload gone/failed` after a 159-minute queue, and `3e8446627` queued 382 minutes. A fresh run is in flight. Given this PR changes the site-packages write boundary and adds an irreversible finalization step, I would like one real e2e pass on the final commit before merge.

PR #1103, opened 2026-08-05
fix(profile): keep the torch-profiler bounds a replacing candidate wiped

chennyiiis:
## Problem

Several CI tasks were killed by the cgroup OOM-killer over the last few days. The kernel `dmesg` records name the victims precisely: `VLLM::Worker_TP` at 130.2 and 136.8 GiB anon-rss, `VLLM::EngineCore` at 107.2 and 107.3 GiB, with page cache at 4.9 MiB at the moment of death — 99.7% anonymous memory. The faulting instruction is inside glibc `_int_malloc` writing a chunk header, with `RDX` decoding to 4064 B / 20448 B chunks on a heap already ~100 GiB deep. So this was millions of small allocations piling up, not one oversized buffer.

Every victim was in the **roofline** phase.

## Root cause

`materialize_config_with_envs` injects the torch-profiler capture window into `EXTRA_VLLM_ARGS`:

```python
profiler_args_parts = [
    f"--profiler-config.delay_iterations {delay_iters}",
    f"--profiler-config.max_iterations {max_iters}",
]
...
if "delay_iterations" not in existing_vllm_args:
    envs["EXTRA_VLLM_ARGS"] = f"{existing_vllm_args} {profiler_args}".strip()
```

The per-task merge runs ~100 lines **later**, and its own comment says it must not drop the profile path's flags — but the `replace_args` branch does exactly that:

```python
if server_args:
    # Merge into (not overwrite) the framework env so the profile path's
    # graph-capture flags aren't dropped.
    ...
    if replace_args:
        envs[framework_env] = server_args
```

`writeback` marks `current_best["args_mode"] = "replace"` automatically as soon as a KEEP carries `remove_args` or `unset_envs`. From that point on the candidate's flag string replaces `EXTRA_VLLM_ARGS` wholesale, taking the bounds *and* the `--profiler-config.ignore_frontend True` that `profile_vllm.yaml` supplies. `extra_envs` (applied last and unconditionally) is a second way in.

In vLLM, `max_iterations = 0` means no limit, and `warmup_iterations = 0` means no `torch.profiler.schedule` at all, so nothing ever calls `on_trace_ready`:

```
ProfilerConfig(profiler='torch', torch_profiler_with_stack=True,
  torch_profiler_record_shapes=True, torch_profiler_with_memory=True,
  ignore_frontend=False, delay_iterations=0, max_iterations=0, ...)
```

That is the config vLLM logged in the gemma run that was killed. The profiler recorded from `/start_profile` at 06:28:58 until the kernel killed the worker at 06:57:53 — 29 minutes.

## Evidence

Ablation on one MI355X, Llama-3.1-8B-Instruct, identical serve flags and identical load generator:

| Configuration | Anon growth | Peak | Extrapolated to 107 GiB |
|---|---|---|---|
| Profiler off, 900 s, 91,701 requests served | **0.19 MiB/s** | 6.12 GiB | never |
| Profiler on, all options off | 20.3 MiB/s | 22.05 GiB | ~85 min |
| `+ with_memory + record_shapes + with_flops` | 47.8 MiB/s | 18.15 GiB | 32 min |
| **+ `with_stack` (the production set)** | **60.1 MiB/s** | 21.63 GiB | **24 min** |

24 minutes predicted against 29 minutes observed. The growth is perfectly linear with no plateau, and `smaps` puts 9.2 of 15 GiB in a single `[heap]` region — matching the malloc stack trace from the kernel log. glibc arenas hold only 448 MiB, so this is the main `brk` heap, not arena fragmentation.

Ruled out as causes (each ablated separately, all peaked at 5.7–6.1 GiB with no growth): `--quantization fp8`, ngram speculative decoding, prefix caching, `ROCM_AITER_UNIFIED_ATTN`, `--kv-cache-dtype fp8_e4m3`, `--max-num-seqs 1024`.

The production sessions corroborate the mechanism directly — both flip from bounded to unbounded at the point their `current_best` became replacing, and never recover:

| Session | Roofline runs with bounds | Roofline runs without |
|---|---|---|
| Llama-3.1-8B | 08-04 20:33 → 08-05 02:10 | 08-05 05:51 onward (OOM at 10:01) |
| gemma-4-26B-A4B | 08-04 09:02 → 08-05 01:36 | 08-05 05:29 onward (OOM at 06:57) |

A DeepSeek-V4-Flash session whose `current_best` was a single additive flag (`--kv-cache-dtype fp8_ds_mla`, so `args_mode="append"`) kept `delay_iterations 6080 / max_iterations 128`, logged `Max profiling iterations reached. Stopping profiler...` 5 seconds after starting, and survived.

## Fix

Re-assert the profiler bounds after the `extra_server_args` / `extra_envs` merges, warn when they had to be restored, and state `ignore_frontend` alongside them — the AsyncLLM-side profiler tracks no iterations, so bounding only the worker leaves the frontend capturing the whole range (which is how the API-server process became the victim in one of the kills).

Candidate flags still win for everything else. The append path is unchanged except that it no longer depends on the YAML to carry `ignore_frontend`.

## Test plan

- [x] `test_materialize_profile_bounds_survive_a_replacing_candidate` — reproduces the OOM configuration (`args_mode="replace"`) and asserts the bounds, `ignore_frontend`, the TraceLens capture flags and the candidate's own flag all survive, plus the restore warning.
- [x] `test_materialize_profile_bounds_survive_an_extra_envs_override` — covers the second path in.
- [x] `test_materialize_profile_states_ignore_frontend_exactly_once` — vLLM warns on duplicate keys, so the flag must not be doubled when the YAML already sets it.
- [x] `test_materialize_profile_adds_ignore_frontend_when_yaml_omits_it`
- [x] Existing suites green: 262 passed across `test_profile_and_kernel_handlers.py`, `test_workload_envs.py`, `test_workload_envs_branches_unit.py`, `test_cli_workload_envs.py`. The one failure, `test_profile_yaml_has_torch_profiler_enabled`, is a pre-existing Windows-only `cp1252` decode of `profile_sglang.yaml`; no YAML is touched by this change.
- [x] `ruff check` and `ruff format --check` clean.

## Follow-ups (not in this PR)

- The profile path enables `with_stack`, `record_shapes` and `with_memory`, which together triple the per-event cost (20 → 60 MiB/s). Bounded capture makes that survivable, but trimming what TraceLens does not consume would give margin back.
- Independent of memory: with `--quantization fp8`, AITER logs `Warning: Latency not found for MI_M=16, MI_N=16, MI_K=128...` — 3.2 million lines / 450 MB of `server.log` in 16 minutes (zero lines without fp8).
- The cgroups have `memory.high` unset, so there is no soft-throttle stage before the OOM killer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,28 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
   suppresses a pointless re-run. The verdict is matched tolerantly and only honoured
   when the manifest reports no KEEP, so it can never discard a validated fusion.
 
+- **Roofline runs no longer launch an unbounded torch profiler**: the profile
+  path injects `--profiler-config.delay_iterations/max_iterations` into
+  `EXTRA_VLLM_ARGS`, but the per-task merge that follows applies the candidate's
+  `extra_server_args` and `extra_envs` afterwards. A candidate carrying
+  `args_mode="replace"` (which `writeback` sets automatically as soon as a KEEP
+  needs `remove_args`) overwrote the whole flag string and took the bounds with
+  it, along with the `--profiler-config.ignore_frontend True` the profile YAML
+  supplies. vLLM reads a missing `max_iterations` as "profile until
+  `stop_profile`", so the worker accumulated every profiler event in host
+  anonymous memory — measured at 60 MiB/s with the production option set — until
+  the cgroup OOM-killer took the engine or worker process out mid-roofline, at
+  107–137 GiB RSS. Because `args_mode` is sticky on `current_best`, one such KEEP
+  turned *every* later roofline in that session into an OOM candidate.
+
+  `materialize_config_with_envs` now re-asserts the bounds after the
+  `extra_server_args`/`extra_envs` merges and warns when it had to restore them,
+  and it states `ignore_frontend` alongside the bounds (the AsyncLLM-side
+  profiler tracks no iterations and would otherwise capture the entire
+  `start_profile`..`stop_profile` range). Candidate flags still win for
+  everything else, and the append path is unchanged apart from no longer relying
+  on the YAML to carry `ignore_frontend`.
+
 ### Removed
 
 - **Kernel-agent LLM role retired** (breaking): the `kernel_agent` role has been
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,27 +36,38 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
   suppresses a pointless re-run. The verdict is matched tolerantly and only honoured
   when the manifest reports no KEEP, so it can never discard a validated fusion.
 
-- **Roofline runs no longer launch an unbounded torch profiler**: the profile
-  path injects `--profiler-config.delay_iterations/max_iterations` into
-  `EXTRA_VLLM_ARGS`, but the per-task merge that follows applies the candidate's
-  `extra_server_args` and `extra_envs` afterwards. A candidate carrying
-  `args_mode="replace"` (which `writeback` sets automatically as soon as a KEEP
-  needs `remove_args`) overwrote the whole flag string and took the bounds with
-  it, along with the `--profiler-config.ignore_frontend True` the profile YAML
-  supplies. vLLM reads a missing `max_iterations` as "profile until
-  `stop_profile`", so the worker accumulated every profiler event in host
-  anonymous memory — measured at 60 MiB/s with the production option set — until
-  the cgroup OOM-killer took the engine or worker process out mid-roofline, at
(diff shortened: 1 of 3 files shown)


chennyiiis:
Raw evidence for the numbers above lives on the shared NFS at `/shared_nfs/chenyi/oom-roofline-profiler-20260805/` — `RCA.md` plus an `evidence/` dir with the per-second memory samples (`mem-*.csv`), the `smaps` / cgroup dumps taken at each threshold crossing, and the ablation scripts under `evidence/scripts/` if anyone wants to re-run the matrix. 443 KB total; the multi-hundred-MB `server.log` files were left off the volume since it is at 100%.

This PR's own e2e confirms the normal path end to end — the roofline config it materialised carries both flags, and vLLM logged:

```
INFO 08-05 16:07:06 [wrapper.py:30] GPU profiling will stop after 128 worker steps, or when stop_profile is received.
```

xiaofei-zheng (approved):

PR #1104, opened 2026-08-05
Doc clean up for crisp user experience

marco-grond:


chaojhou (approved):

chaojhou (approved):
Approving. The review points are addressed and the docs consolidation holds up.

**Verified**

- `docs/compatibility.rst` is a clean revert, so Atom and xDiT are back in the matrix and no longer contradict the `--framework` choices built from the registry.
- The Quickstart -> Install rename is consistent now across the TOC, `index.rst`, both page H1s, and both cross-references. The two remaining "Quickstart" mentions in `examples/README.md` are correct as-is: the README framing itself as the quickstart and pointing at the docs for detailed instructions is a clean split.
- The 135-line `examples/README.md` reduction loses nothing. Each deleted block is present in `docs/install/install.md`: the setup-environment section (including the `.env` single-source-of-truth and re-run guidance at :189-190), the full source-checkout section at :269, and the container image list at :332-334. Both new anchors resolve against real headings.
- Moving the run-mode guidance directly under the `HYPERLOOM_RUN_MODE` options is the right placement, and restoring "Docker is preferred for a validated, reproducible stack" removes the earlier contradiction.

**Two follow-ups (cosmetic, not blocking this merge)**

1. `docs/install/install.md:90` ? the admonition fence needs braces: `{note}` rather than plain `note`. Without them MyST reads it as a fenced code block whose language happens to be "note", so it renders as a monospace box instead of a callout. Every other admonition in the repo uses braces (`docs/how-to/optimize.md:16`, `docs/reference/authentication.md:196`, `docs/reference/operations.md:13` and `:84`, `docs/reference/session-breakdown.md:9`). The sphinx build passes either way, which is why CI did not flag it.

2. `examples/README.md:86` ? same block, but this file is rendered by GitHub, which does not support MyST admonitions at all, braces or not. Matching the existing blockquote style already used at lines 19-25 would render correctly, for example a `> [!NOTE]` block.

**One operational note**

Worth adding the `skip-e2e-test` label. `ci-e2e` has no `paths-ignore`, so this docs-only PR has now dispatched a GPU run twice (median ~85 min), while `lint` and `tests-coverage` were correctly skipped by their path filters.

PR #1111, opened 2026-08-06
fix(knowledge): preserve remote compatibility paths

meinali-566:
## Summary
- stop forwarding the removed `forge-gemm-tune --kb-read` gate while preserving candidate and library quality controls
- audit and safely degrade selected GBrain Recipe reads, and close the selected remote store during shutdown
- query both current and legacy native-KG entity slugs so punctuation-bearing historical nodes remain readable
- make Forge knowledge configuration process-owned so a child environment cannot seed the global cache

## Test plan
- [x] 219 relevant Recipe KB, KG, GBrain, local graph, and Forge environment tests
- [x] Ruff on all changed production and test files
- [x] `git diff --check`


jiaqiang-dot-liu (approved):
LGTM

PR #1113, opened 2026-08-06
feat(kernel): resolve kernel sources from installed frameworks (retire op_to_source.json)

Ahmedhasssan-aig:
## Summary

The kernel optimizer used to locate a hot kernel's source via a static file, `op_to_source.json`, which goes stale whenever a framework moves a file, renames a kernel, or bumps its version. This PR replaces it with a resolver that looks sources up **live from the installed framework trees**, so it self-heals across moves, renames, and version drift — no hand-maintained table.

## How kernel-op sources are found

- **Native GPU kernels (`.cu` / `.hip`):** the active finder demangles the kernel's device symbol and looks it up in a fresh index built from the installed `csrc/` folders (definition bodies only), verifies the match, and returns the source file/line. Composable Kernel (`ck` / `ck_tile`) template instantiations have no single editable `__global__` source, so they are classified **non-patchable** (from the symbol alone) and not sent for rewrite.
- **Triton kernels (`.py`):** resolved from the profiler-recorded `kernel_file`, then the exact `@triton.jit` line is pinned via an AST parse (no import of the kernel).

## Resolution tiers per route

- **TraceLens route:** active finder (deterministic) → trace-stack → name-grep → LLM fallback.
- **Bypass route (`_bypass_report`):** Triton `.py` (AST) → active finder (native, by device symbol) → repo-scan by demangled kernel name. All native lookups are keyed on the **device symbol**, not the CPU-op name, so device kernels with no correlated `cpu_op` still resolve. A finder `non_patchable` verdict is authoritative and is **not** overridden by the repo-scan tier.

## Behavior change (intentional, reviewable)

The retired `op_to_source.json` marked its `aiter_ck` entries `patchable: true` and routed them to the ck-fellow backend. Resolving from the device symbol alone cannot recover that per-entry ck-fellow ownership, so CK instantiations are now classified non-patchable and no longer reach the ck-fellow branch. This trades hand-maintained CK routing (which could not generalize across versions) for coverage that self-heals; restoring CK→ck-fellow via a symbol-derivable classifier is left as a separate follow-up. See the note in `source_resolver.py`.

## Robustness

- `tracelens_analysis.py` runs as a standalone subprocess against the *installed* hyperloom, which may predate this checkout. It now reads `kernel_source_contract` method-name constants via `getattr` and demotes a too-old contract module to "absent", so it degrades to an unwritten artifact instead of aborting at import.
- The native-index disk cache falls back to a **user-scoped, owner-only** subdir of the system temp root (or `$HYPERLOOM_KSI_CACHE_DIR`), so it cannot collide/shadow across users on a shared host.
- Logging added across the new stack (discovery, index build, ambiguity, cache/read errors).

## Files changed

| File | Change |
|---|---|
| `tools/source_resolver.py` | **New.** Core resolver: device symbol → editable source; CK non-patchable gate (+ mangled fallback). |
| `tools/kernel_source_index.py` | **New.** Builds/caches the live native-kernel (`__global__`) definition index from installed `csrc/`. |
| `tools/source_env.py` | **New.** Discovers installed frameworks/versions; recursive fingerprint invalidates the cache on nested source edits. |
| `tools/_bypass_source_resolver.py` | Bypass entry: native via finder, Triton `.py` via trace + AST line pinning; repo-scan fallback. |
| `tools/_bypass_report.py` | Wires candidates through the three-tier chain; preserves the finder's `non_patchable` verdict. |
| `tools/tracelens_analysis.py` | TraceLens route uses the finder; removed the JSON mapping tier; contract-module import hardened. |
| `tools/kernel_optimization.py` | Recognizes the new resolution methods for the device-symbol focus block. |
| `common/kernel_source_contract.py` | Method-name constants for the finder (`active_finder` / `symbol_index`). |
| `tools/data/op_to_source.json` | **Deleted.** The static mapping is gone. |
| `pyproject.toml` | Dropped the deleted JSON's `package-data` glob. |
| `.github/CODEOWNERS` | Repointed ownership from the deleted JSON to the new resolver modules. |
| `tests/test_source_resolver.py` | **New.** Resolver tests (namespace-boundary CK detection, non_patchable preservation). |
| `tests/test_kernel_index_macros.py` | **New.** Index-builder tests (definition-only indexing). |
| `tests/test_source_env.py` | **New.** Framework-discovery + fingerprint tests. |
| `tests/test_live_source_stack.py` | **New.** Gated end-to-end resolver stack test (`HYPERLOOM_LIVE_STACK=1`). |
| `tests/test_bypass_source_resolver.py` | Editability filter + AST Triton path tests. |
| `tests/test_bypass_report.py` | Candidate-resolution expectations + finder `non_patchable` regression tests. |
| `tests/test_source_resolution_artifact.py` | Contract-module degrade-not-abort regression tests. |
| `tests/test_tracelens_csv.py` | Updated for the finder tier. |


.gitignore
tsrikris, line 1: Verifying that this is needed on this PR

src/hyperloom/agents/kernel/tools/source_resolver.py
```diff
@@ -0,0 +1,414 @@
...
+import os
+import re
+import shutil
+import subprocess  # nosec B404 - invokes c++filt with a fixed, non-shell argv.
+import time
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any
+
+try:  # package import (TraceLens route / tests)
```
tsrikris, line 48: Confirm why 2 different import paths are needed

src/hyperloom/agents/kernel/tools/source_resolver.py
```diff
@@ -0,0 +1,414 @@
...
+    """Resolve a kernel to its editable source in the installed tree (timed).
+
+    Resolution is driven entirely by the device kernel symbol against the live
+    :mod:`kernel_source_index` -- no static source mapping is involved:
+
+    1. symbol-derived non-patchable gate (CK templates / prebuilt assembly);
+    2. symbol-first lookup: base name -> ranked index records -> verified source.
+
+    Args:
+        op_name: Launching op name (e.g. ``_C::silu_and_mul``); carried for
```
tsrikris: Recommed removing silu_and_mul type kernel names from comments

src/hyperloom/agents/kernel/tools/source_resolver.py
```diff
@@ -0,0 +1,414 @@
...
+    for tag, stats in report.items():
+        print(f"Latency[{tag}]: {json.dumps(stats)}")
+    return 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(_main())
+
+
+__all__ = [
```
tsrikris: Recommend moving this up top

src/hyperloom/agents/kernel/tools/source_env.py
```diff
@@ -0,0 +1,250 @@
...
+        parts += [_dir_signature(cr) for cr in fr.csrc_roots]
+    return hashlib.sha1("|".join(parts).encode("utf-8")).hexdigest()[:16]  # nosec B324 - cache key, not security.
+
+
+def version_tag(frameworks: dict[str, FrameworkRoot]) -> str:
+    """Human-readable version tag for logs/reports (e.g. ``aiter0.1.16_vllm0.26.0``)."""
+    return "_".join(f"{n}{frameworks[n].version or 'unknown'}" for n in sorted(frameworks)) or "none"
+
+
+__all__ = [
```
tsrikris: Recommend moving this up top

src/hyperloom/agents/kernel/tools/kernel_source_index.py
```diff
@@ -0,0 +1,240 @@
...
+        symbol_index=symbol_index,
+        build_ms=round((time.perf_counter() - started) * 1000.0, 2),
+        file_count=file_count,
+        symbol_count=len(symbol_index),
+    )
+
+
+# --- cache ------------------------------------------------------------------
+def _cache_path(fingerprint: str) -> Path:
+    """Cache file path (dir from ``$HYPERLOOM_KSI_CACHE_DIR`` or a temp subdir)."""
```
tsrikris: Confirming HYPERLOOM_KSI_CACHE_DIR is a general env variable not created for this feature. Making sure this just adds a directory there or in a common directory independent of the container

tsrikris (commented):
Minor clarifications

.gitignore
Ahmedhasssan-aig: Yes, not needed. Dropping it

src/hyperloom/agents/kernel/tools/source_resolver.py
Ahmedhasssan-aig: The try/except just uses whichever import works for how the file was loaded. Both ways are need in pipeline, for pick and choose.

src/hyperloom/agents/kernel/tools/source_resolver.py
Ahmedhasssan-aig: Done

src/hyperloom/agents/kernel/tools/source_resolver.py
Ahmedhasssan-aig: Done

src/hyperloom/agents/kernel/tools/source_env.py
Ahmedhasssan-aig: Done

src/hyperloom/agents/kernel/tools/kernel_source_index.py
Ahmedhasssan-aig: It's a single self‑contained cache dir, fine to leave unset. If we want it to be in a common directory, we can set this variable to it.

tsrikris:
@jiaqiang-dot-liu @zhanglei-amd can you review this PR? We will address merge conflicts along with your comments. Would appreciate a quick turnaround.

CC: @chaojhou, @xiaofei-zheng

xiaofei-zheng:
## Review: live source resolution (retiring `op_to_source.json`)

The direction is right — a hand-maintained table inevitably rots, and resolving against the installed tree is the correct fix. But I found four issues that make the new resolver either point at the wrong file or silently drop optimizable kernels. Everything below is reproduced against this branch; repro script is at the end.

---

### Blocking

#### 1. `"ck::"` is matched as an unbounded substring, so unrelated kernels are silently dropped

`source_resolver.py:64`

```python
_CK_SYMBOL_MARKERS = ("ck_tile", "ck::", "ck_tile::")
```

`_non_patchable_kind` tests `marker in hay` against the entire demangled string, so any namespace or template argument ending in `ck` matches:

| symbol | verdict |
|---|---|
| `void vllm::block::my_kernel(float*)` | `aiter_ck` (wrong) |
| `void unpack::helper_kernel(int*)` | `aiter_ck` (wrong) |
| `void flashck::fmha_kernel(void*)` | `aiter_ck` (wrong) |
| `void ck_tile::Gemm(float*)` | `aiter_ck` (correct) |

This gate runs before any lookup and returns immediately, so the kernel is dropped from optimization with no log line — it will look like the kernel simply wasn't hot enough. Suggest a boundary-anchored match, e.g. `re.search(r"(^|[^A-Za-z0-9_])ck(_tile)?::", demangled)`. Note also that `"ck_tile::"` is fully subsumed by `"ck_tile"`.

#### 2. The index cannot distinguish a declaration from a definition, and ranking prefers the declaration

`_iter_global_defs` (`kernel_source_index.py:72`) accepts any identifier followed by `(` after `__global__` and never checks for a body, while `_NATIVE_EXTS` includes `.h/.hpp/.cuh`. Meanwhile `_rank_records` (`source_resolver.py:256`) breaks ties on `-len(path)`, and a declaration in `csrc/include/ops.h` is almost always a shorter path than the definition in `csrc/kernels/attn/paged_attention.cu`.

Given both files, `resolve()` returns `csrc/include/ops.h:2` — a header with no kernel body, which is then what GEAK is asked to rewrite. `_verify_symbol` cannot catch this: it is a plain `base in text` substring check, and the name appears in the declaration too.

The same scanner also indexes commented-out code and string literals:

```python
ksi._iter_global_defs('// __global__ void dead_kernel(int*);\n'
                      'const char* s = "__global__ void str_kernel(int*)";')
# -> ['dead_kernel', 'str_kernel']
```

Suggest requiring the first non-space character after the closing `)` to be `{`, and ranking "is a definition in `.cu`/`.hip`" above path length.

#### 3. The cache fingerprint is blind to edits inside subdirectories

`source_env.py:224`

```python
return f"{path}:{int(st.st_mtime)}:{sum(1 for _ in os.scandir(path))}"
```

This is the csrc root's own mtime plus its immediate child count — not recursive. Editing `csrc/sub/k.cu` leaves the fingerprint unchanged, so a stale index is reused. Real csrc trees are entirely nested, and GEAK itself rewrites these files during a run, so this is the common case rather than an edge case. The docstring currently claims the opposite:

> Changes exactly when a different framework/version is installed or a native source dir is modified, so a cached index is reused iff still valid.

Suggest a recursive signature (max mtime + total file count), or a TTL on the cache.

#### 4. `non_patchable` is swallowed on the bypass route

`source_resolver.py:83`

```python
return (self.source_file, self.method if self.source_file else "unresolved")
```

A `non_patchable` result carries an empty `source_file`, so the tuple is always `("", "unresolved")` — the `non_patchable` method can never be observed through this API. Both `source_resolver.resolve_source` and `_bypass_source_resolver.resolve_source` document `("", "non_patchable")` as a possible return value.

Net effect: the bypass route never learns that a kernel is a CK template, `classify_patchability` cannot report the real reason and falls through to the generic heuristics. The deterministic route uses `resolve()` directly and is unaffected, so the two routes now disagree on the same kernel.

---

### Should fix

#### 5. The production path rebuilds/reloads the index on every resolve; the benchmark bypasses that cost

`resolve()` calls `kernel_source_index.load_or_build()` whenever no `index=` is passed (`source_resolver.py:303`). That re-runs `discover_frameworks()` (iterating site-packages, with an `_has_native` os.walk per candidate dir) and re-parses the entire cache JSON. `_bypass_report.build_candidates` and `_expand_op_fanout` each call it once per candidate, and there is no in-process memoization anywhere.

Measured on a machine with **none** of the frameworks installed, so this is a lower bound: `discover_frameworks()` 28–41 ms per call, `json.load` of an 8k-symbol cache 7.6 ms per call. Inside a real ROCm image with aiter/vLLM csrc trees the walk will be considerably more expensive.

Worth noting that `_main --bench` passes `index=index` explicitly, so the latency report this PR ships does not measure any of this. Suggest a module-level singleton keyed by fingerprint.

#### 6. Coverage regression for kernels that have no `op_name`

`_bypass_report.py:445` still gates the native lookup on `op_name`:

```python
if not source_file and op_name:
    source_file, method = resolve_source(op_name, framework=framework, device_kernel_name=kname)
```

but the finder ignores `op_name` entirely ("carried for reporting, not used for lookup") and only needs `kname`. Previously the `resolve_by_kernel_name(kname)` tier below caught these; that tier is removed in this PR. So a kernel that has a device symbol but no associated cpu_op is now unresolvable, even though the finder has everything it needs. Either gate on `kname`, or keep the repo-scan tier.

Related: dropping the repo-scan tier is an independent coverage change that isn't mentioned in the description. Worth calling out explicitly, ideally with a before/after resolution rate on a real trace.

#### 7. `source_line` is written but never read

`_bypass_report.py:521` stores the AST-pinned `@triton.jit` def line on the candidate, but nothing in the repo reads it back (`source_line` appears exactly once outside unrelated code). Pinning the exact jit line is listed as a headline feature of this PR, yet the value never reaches the prompt or GEAK. Either wire it into `kernel_optimization.build_prompt` or drop the field until it has a consumer.

#### 8. The riskiest change in the PR has no test coverage

The four new test files all cover the three new low-level modules, which is good. But `OpResolver` in `tracelens_analysis.py` — where 429 lines of dispatch/composite/name-anchoring logic were replaced — has no references anywhere under `tests/`, before or after this PR. It is also the only path that *overrides* an already-resolved `source_file` via `apply_to`, so combined with the substring-only `_verify_symbol` it carries the highest mis-routing risk of anything here.

#### 9. `test_live_source_stack.py` contributes zero pytest cases

It is named `test_*.py` and lives in the tests package, but contains only `check_*` helpers and `_run_all()` — no `test_` functions — and `_require_live()` raises `SystemExit` rather than calling `pytest.skip`. Running the six touched test files collects 88 cases, none of them from this 348-line file. Suggest moving it under `scripts/` or converting it into real cases with `@pytest.mark.skipif`. It also writes artifacts into `results/`, which are now untracked since the `.gitignore` hunk was dropped.

#### 10. Failures are silent everywhere, with no diagnostics

`_bypass_source_resolver.resolve_source` swallows `(ImportError, OSError, ValueError)`; `_load_cache`, `_save_cache`, `_scan_file`, `_verify_symbol` and `_cxxfilt_base` all swallow and return empty. Correctness now depends entirely on the runtime environment — which frameworks are installed, where csrc lives, whether `c++filt` is present. If the resolution rate drops in production there is nothing to debug with. At minimum, log at debug level on a resolve miss and on a cache invalidation.

---

### Naming and docs not synced with the deletion

- `.github/CODEOWNERS:57` still assigns owners to the deleted `tools/data/op_to_source.json`.
- `stamp_onto` still writes `op_to_source_status` / `_kind` / `_patchable` / `_reason` / `_matched_route`, and `_invocation_spec.py` plus `classify_patchability` still read them.
- `classify_patchability` returns user-visible skip reasons literally prefixed `"op_to_source: ..."`, naming a mechanism that no longer exists.
- The `source_resolver` module docstring and `_non_patchable_kind` both state that prebuilt assembly kernels are detected from the symbol, but there is no assembly marker in `_CK_SYMBOL_MARKERS` — only `aiter_ck` can ever be returned.
- `OpResolver.__init__` keeps `mapping` "so existing call sites need no change", but the only call site in the repo is now `OpResolver()`.
- `OpResolution`'s `composite` / `dispatch` / `fanout` / `leaf_resolutions` branches are unreachable now that the finder only produces `kind="single"`, and `kernel_kinds` / `prebuilt_binaries` are always `[""]` — yet the `aiter_asm` handling in `apply_to` and `classify_patchability` is still there. That is a sizeable block of dead code plus dead comments.

---

### Nits

- The default cache dir is `/tmp/hyperloom_ksi/` with a predictable filename. Its contents are symbol -> path records that get handed to GEAK for rewriting, and `_load_cache` only validates the `fingerprint` field against itself. On a shared host another user could pre-create that file. A user-private default would be safer.
- The three new `test_source_env.py` cases assert `str(p).endswith("atom/csrc")`, hardcoding a forward slash. Fine on the Linux CI; `Path` comparison would be more portable.
- The branch is 86 commits behind `main`. Given the 429-line deletion in `tracelens_analysis.py` and the 321-line rewrite of `_bypass_source_resolver.py`, worth merging `main` in before another pass.

---

<details>
<summary>Repro script for issues 1–4</summary>

```python
import tempfile, time
from pathlib import Path
from hyperloom.agents.kernel.tools import source_resolver as sr
from hyperloom.agents.kernel.tools import kernel_source_index as ksi
from hyperloom.agents.kernel.tools import source_env as se

# 1) unbounded "ck::" substring
for s in ["void vllm::block::my_kernel(float*)",
          "void unpack::helper_kernel(int*)",
          "void flashck::fmha_kernel(void*)"]:
    assert sr._non_patchable_kind(s) == "aiter_ck"   # all three misclassified

# 2) header declaration outranks the .cu definition
root = Path(tempfile.mkdtemp()); csrc = root / "csrc"
(csrc / "include").mkdir(parents=True); (csrc / "kernels" / "attn").mkdir(parents=True)
(csrc / "include" / "ops.h").write_text(
    "namespace vllm {\n__global__ void paged_attn_kernel(const float*, float*);\n}\n")
(csrc / "kernels" / "attn" / "paged_attention.cu").write_text(
    "namespace vllm {\n__global__ void paged_attn_kernel(const float* q, float* o) { o[0]=q[0]; }\n}\n")
fw = {"vllm": se.FrameworkRoot("vllm", root, "1.0", (csrc,))}
idx = ksi.build_index(fw)
r = sr.resolve("x::y", framework="vllm",
               device_kernel_name="void vllm::paged_attn_kernel(float const*, float*)", index=idx)
assert r.source_file.endswith("ops.h")               # picked the declaration

# 2b) comments and string literals are indexed as kernels
assert [n for n, _ in ksi._iter_global_defs(
    '// __global__ void dead_kernel(int*);\n'
    'const char* s = "__global__ void str_kernel(int*)";')] == ["dead_kernel", "str_kernel"]

# 3) fingerprint blind to a nested edit
d = Path(tempfile.mkdtemp()); (d / "sub").mkdir()
(d / "sub" / "k.cu").write_text("__global__ void a(){}")
fwx = {"x": se.FrameworkRoot("x", d, "1.0", (d,))}
f1 = se.fingerprint(fwx); time.sleep(1.1)
(d / "sub" / "k.cu").write_text("__global__ void a(){}\n__global__ void b(){}")
assert f1 == se.fingerprint(fwx)                     # stale cache stays valid

# 4) non_patchable collapses to "unresolved"
res = sr.resolve("ck::op", device_kernel_name="ck_tile::Gemm<half>(void*)", index=idx)
assert res.method == "non_patchable"
assert res.as_legacy_tuple() == ("", "unresolved")   # verdict lost
```

</details>

xiaofei-zheng:
Follow-up to my previous comment — that one was about implementation bugs, this one is about the design of the symbol -> source mapping itself. Four points, roughly in order of how much I think they matter.

---

### 1. The qualified name is available and then deliberately thrown away

`_base_from_demangled` (`source_resolver.py:168`) reduces the demangled symbol to its last `::` component, and `build_index` keys the index on that bare name (`kernel_source_index.py:150`). So `vllm::reduce_kernel` and `sglang::reduce_kernel` collapse into one index entry, and `_rank_records` has to guess between them using a framework hint plus "whichever path is shorter".

With two same-named kernels in two frameworks:

```
index key -> records (namespace is gone from the key):
  'reduce_kernel': 2 records
      sglang: sglang/csrc/red.cu
      vllm:   vllm/csrc/reduction/reduce_kernels.cu

asked for sglang::reduce_kernel            -> sglang/csrc/red.cu     correct=True   confidence=high
asked for vllm::reduce_kernel, no fw hint  -> sglang/csrc/red.cu     correct=False  confidence=high
```

The second case returns the wrong framework's kernel and still reports `confidence=high`. The framework hint is what saves the first case, but it is not always available: `_expand_op_fanout` only honors `vllm`/`sglang`, so anything else (aiter in particular) arrives as an empty hint and the decision falls through to path length alone.

This is the part I'd most like to see changed, because the information needed for an exact match is right there in the demangled string. Suggest indexing on the qualified name as well — parse the enclosing `namespace` blocks during the scan, key on `ns::name`, and keep the bare name as a secondary index. Then resolution can try the exact qualified match first and only fall back to the bare name (at reduced confidence) when the symbol has no namespace. That removes the guesswork for the collision case entirely instead of trying to rank your way out of it.

<details>
<summary>Repro</summary>

```python
import tempfile
from pathlib import Path
from hyperloom.agents.kernel.tools import source_resolver as sr
from hyperloom.agents.kernel.tools import kernel_source_index as ksi
from hyperloom.agents.kernel.tools import source_env as se

root = Path(tempfile.mkdtemp())
v, s = root / "vllm" / "csrc", root / "sglang" / "csrc"
(v / "reduction").mkdir(parents=True); s.mkdir(parents=True)
(v / "reduction" / "reduce_kernels.cu").write_text(
    "namespace vllm {\n__global__ void reduce_kernel(const float* a, float* o) { o[0]=a[0]; }\n}\n")
(s / "red.cu").write_text(
    "namespace sglang {\n__global__ void reduce_kernel(const float* a, float* o) { o[0]=a[1]; }\n}\n")

fw = {"vllm":   se.FrameworkRoot("vllm", root / "vllm", "0.26.0", (v,)),
      "sglang": se.FrameworkRoot("sglang", root / "sglang", "0.4.0", (s,))}
idx = ksi.build_index(fw)
assert list(idx.symbol_index) == ["reduce_kernel"]          # one key for two kernels

res = sr.resolve("x::y", framework="",
                 device_kernel_name="void vllm::reduce_kernel(float const*, float*)", index=idx)
assert "sglang" in res.source_file and res.confidence == "high"   # wrong file, high confidence
```

</details>

### 2. `confidence` is never read, and one of its three values isn't a confidence

`ResolveResult.confidence` (`source_resolver.py:79`) is the only occurrence of the string `confidence` anywhere under `agents/kernel/` — nothing consumes it. The three call sites construct it positionally:

```python
ResolveResult("", None, "", False, "non_patchable", "symbol", 0.0, reason=nonp_kind)   # :315
ResolveResult(path, ..., True, "symbol_index", "high", 0.0)                            # :329
ResolveResult("", None, "", False, "unresolved", "none", 0.0, reason="no live match")  # :332
```

The `non_patchable` case passes `"symbol"` as the confidence — that's a provenance label, not a confidence level, and it reads like a positional-argument slip. Seven positional args into a dataclass is what makes that easy to miss; keyword arguments would have caught it.

More importantly, the field is exactly the mechanism this design needs and isn't using. Given that ranking is heuristic by construction, a resolve that matched a single record should not carry the same confidence as one that picked a winner out of several by path length. Suggest populating it honestly (single exact qualified match = high, bare-name match with one record = medium, tiebreak among several = low) and having `apply_to` refuse to override an existing `source_file` below some threshold.

### 3. The recall boundary of the new index is unknown, and may be smaller than the table it replaces

The index only scans `csrc` / `sgl-kernel/csrc` / `kernels` for literal `__global__`. Two things fall outside that:

- pip-installed vLLM ships no `csrc` at all — `source_env`'s own module docstring says so, and `test_live_source_stack.py:127` asserts `any(fr.csrc_roots for fr in fw.values())` rather than asserting vLLM has one, which suggests this is already known.
- CK template instantiations have no literal `__global__`, though those are deliberately gated out as non-patchable, so that part is by design.

Net, depending on the image layout, the kernels this mechanism can actually resolve may be mostly aiter's hand-written `__global__` kernels. The deleted table had 2617 lines of known coverage; the new mechanism's coverage is currently unmeasured.

I think this is the single most useful thing to add before merging: run one real trace through both the old table and the new finder and post the resolution rate side by side (resolved / non-patchable / unresolved counts). If it's comparable or better, that settles most of the discussion in this PR. If it's materially worse, that's worth knowing now rather than after the hot-kernel list quietly gets shorter.

### 4. The failure mode changed character, not just frequency

Worth stating explicitly because it affects how much the verification step matters:

- Old: the table goes stale, the recorded path doesn't exist, the open fails loudly and immediately.
- New: the index matches a plausible-looking file. The path exists, opens fine, and is writable — so GEAK rewrites it and produces a diff that looks reasonable.

The second is much more expensive to notice and diagnose. That's not an argument for keeping the table; it's an argument that with a live resolver the verification step has to be strict, and right now `_verify_symbol` is a whole-file substring check that a forward declaration, a comment, or a call site all satisfy. Since the index already stores the line number, checking that the recorded line is actually that kernel's `__global__` definition head costs essentially nothing and closes most of this gap.

---

Points 1, 2 and 4 are all small changes (index on qualified name, populate confidence and gate the override on it, verify the stored line). Together they'd move the failure mode from "silently rewrites the wrong file" back to "says it couldn't find the kernel", which for this component is the far better place to fail.

jiaqiang-dot-liu:
kernel_source_index._iter_global_defs() currently treats comments, strings, and forward declarations as kernel definitions. Candidate ranking then prefers framework/path heuristics, while _verify_symbol() only checks whether the base name occurs anywhere in the file.

I reproduced a case where a header containing only a commented declaration was selected over the real .cu implementation and reported with confidence="high".

The indexer should only record actual function definitions, validate the recorded line, and return unresolved when multiple candidates cannot be uniquely disambiguated. Matching should use qualified symbols/signatures where available, rather than only the base name.

jiaqiang-dot-liu:
The cache fingerprint only uses the csrc root directory’s integer-second mtime and immediate child count. It does not detect modifications to existing files or changes inside nested directories.

I reproduced a source update where the fingerprint remained unchanged and the old index was loaded from cache.

Please use a recursive source manifest containing at least relative path, file size, and mtime_ns, and include an index schema version. A failed record verification should trigger one forced rebuild instead of silently returning unresolved or using stale data.

jiaqiang-dot-liu:
A successful active-finder lookup is currently treated as an authoritative patchability verdict. Finding a file containing a matching symbol does not prove that it is safe to modify.

Finder results must still pass through the common patchability policy for vendor kernels, collectives, dispatch wrappers, generated sources, and unsupported implementation types.

The responsibilities should be separated:

SourceResolver: locate the implementation source
PatchabilityPolicy: decide whether it can be modified and select the backend

jiaqiang-dot-liu:
Removing op_to_source.json also removes information that cannot be reconstructed from a base symbol alone:

aiter_ck versus aiter_asm
prebuilt_binary
patchability
single / dispatch / composite
composite source ownership and fanout
The previous mapping contained source-backed CK kernels that were patchable and intended for ck-fellow, but the new implementation rejects CK symbols by namespace. Conversely, prebuilt assembly is not reliably identified.

Please preserve these semantics through a structured framework manifest or framework-specific classifier. Composite fanout must also be retained, unless its removal is explicitly reviewed and accepted as a breaking behavior change.

jiaqiang-dot-liu:
Several valid runtime cases are no longer resolved:

The bypass route does not call the native resolver when op_name is empty, even though lookup is driven by device_kernel_name.
Standard SGLang checkout layouts are not fully discovered.
Importable editable/source checkouts outside site-packages are skipped.
Real profiler formats such as graph-launch wrappers, .kd suffixes, and anonymous namespaces are not normalized correctly.
Triton kernels without kernel_file lose the previous fallback.
Native lookup should be attempted whenever a device symbol is available, independent of CPU-op attribution. Framework discovery and symbol normalization should cover the actual vLLM/SGLang/AITER runtime layouts.

jiaqiang-dot-liu:
The current AST logic can:

Treat unrelated decorators such as @numba.jit as Triton
Return an ordinary host helper from a launcher func
Select the first substring match when multiple JIT functions exist
Reject every source under /tmp, including legitimate checkouts
Please verify that decorators resolve to Triton, restrict function matches to confirmed Triton kernel definitions, and require a unique match. Ambiguous results should remain unresolved instead of selecting the first candidate. Generated-source detection should use provenance and trusted source roots rather than a blanket /tmp rule.

jiaqiang-dot-liu:
The deterministic TraceLens route can terminate the entire analysis when framework discovery or index I/O fails, while the bypass route degrades more gracefully. Expected discovery/cache errors should produce an unresolved result with an audit reason rather than aborting the pipeline.

The index should also be shared across one analysis run. Production calls currently repeat framework discovery, fingerprint calculation, and JSON deserialization for each candidate.

Finally, the legacy tuple conversion must preserve non_patchable; it currently converts every empty-source result to unresolved, contrary to its documented contract.

These cases should be covered by regression tests and run in PR CI. The current checks do not include the relevant pytest or live vLLM/SGLang integration suites.

jiaqiang-dot-liu:
Removing the static op_to_source.json is the right direction, since a manually maintained mapping cannot generalize across framework versions, environments, or newly introduced kernels. Separately, I would like to explore whether TraceLens could own the end-to-end hot-kernel analysis—including source location and provenance, input shapes, dtypes, and related metadata—and emit the results through a versioned, structured JSON contract that Hyperloom can consume directly. This could significantly reduce duplicated parsing and heuristic source reconstruction on the Hyperloom side.

ZhengGong-amd (commented):
Reviewed the branch locally (all 93 resolver/bypass tests pass; full kernel-agent suite green). Direction looks right — the notes below are about wiring gaps, one coverage regression, and observability.

## Must fix

**1. `_bypass_report.py:445` gates native resolution on `op_name`, but the finder never uses it**

```python
if not source_file and op_name:
    source_file, method = resolve_source(op_name, framework=framework, device_kernel_name=kname)
```

`source_resolver.resolve()` documents `op_name` as "carried for reporting, not used for lookup" — only `kname` matters. The `and op_name` guard is left over from the dict era, where the third tier `resolve_by_kernel_name(kname)` (removed here) covered the gap. `_bypass_trace_reader` fills `op_name` from `_majority_op(nm)`, which returns `""` for device kernels with no correlated `cpu_op` (common under HIP graphs). Those now go straight to unresolved. Should be `if not source_file and kname:`.

**2. `_non_patchable_kind()` has no mangled-symbol fallback, so classification depends on whether `c++filt` is installed**

`base_symbol()` falls back to `_base_from_mangled()` when `c++filt` is absent; `_non_patchable_kind()` only calls `_cxxfilt_base()`. Checked against the deleted JSON: `_ZN2ck15kernel_moe_gemmI...` demangles to something containing `ck::` and is gated, but the raw mangled string only contains `2ck15` and slips through. Same kernel, different verdict depending on whether binutils is in the image.

**3. CK coverage regression is not called out**

All 56 `aiter_ck` records in the deleted JSON were `patchable: true`, and `forge_submit.py:359` `_resolve_fellow()` has a dedicated ck-fellow branch for them. 32 of those symbols carry a `ck_tile`/`ck::` marker and are now hard-gated to `non_rewritable`, so nothing sets `kernel_kind="aiter_ck"` on the TraceLens route anymore and the ck-fellow path is unreachable from resolution. This may be intentional (a CK template instantiation has no single editable `__global__`), but it is a coverage reduction relative to `main` and should be stated and quantified in the PR description — or CK should keep routing to the ck-fellow instead of being gated out.

**4. No logging anywhere in the new stack — every failure is silent**

The three new modules plus the rewritten `_bypass_source_resolver` contain zero log statements; the old `_bypass_source_resolver` had `log = logging.getLogger(__name__)`. Replacing a static table with environment discovery adds failure modes that all present identically ("no candidates this round"):

- `discover_frameworks()` returns `{}` (pip vLLM ships no `csrc`, or a layout `_CSRC_DIRS` does not cover) → 0-symbol index → every native kernel unresolved
- `c++filt` missing → `_cxxfilt_base` returns `""`
- cache dir/write failure → `except OSError: pass`
- `_bypass_source_resolver.resolve_source` swallows `ImportError, OSError, ValueError` → `("", "unresolved")`

Minimum: log `version_tag / symbol_count / file_count` once on index build, warn when discovery is empty, and debug-log the swallowed exception instead of dropping it.

**5. The index is rebuilt/reloaded per candidate**

`resolve()` takes `index=`, but neither production route passes it (only tests and `--bench`). Each candidate therefore runs `load_or_build()` → `discover_frameworks()` (walks every site-packages child) → `fingerprint()` (stats every csrc root) → full `json.load` of the cache. Neither `resolve()` nor `load_or_build()` memoizes, yet the `kernel_source_index` docstring claims "built once per container … later runs are ~free". Either `lru_cache` `load_or_build` or build once in `_expand_op_fanout` / `build_candidates` and thread it through.

## Should fix

- **Ambiguity is now resolved silently.** The old `_build_repo_kernel_index()` mapped a multi-path symbol to `""` and refused it rather than "routing a rewrite at an arbitrary first-seen file". `_rank_records()` now picks a winner by `(fw_match, arch_match, -len(path))`, and `_verify_symbol()` is a whole-file substring check that every duplicate passes. At minimum log when a symbol has >1 candidate record.
- **`non_patchable` is lost on the bypass route.** `_bypass_report.py:447` only adopts `method` when `source_file` is non-empty, so `("", "non_patchable")` degrades to `"unresolved"` — "known not rewritable" and "not found" become indistinguishable in the report.
- **Cache location.** `tempfile.gettempdir()/hyperloom_ksi` bypasses `session/paths.py` (the path authority per CLAUDE.md), and a fingerprint-named file in shared `/tmp` is trusted unconditionally via `SourceIndex(**data)`.
- **Cache invalidation misses in-place edits.** `_dir_signature()` uses only the csrc root's mtime + immediate child count. The kernel agent patches `.cu` files deep in that tree, which does not bump the root mtime, so cached line numbers can go stale.
- **Coverage gate.** The two `_main` CLIs (~75 lines) are not in `[tool.coverage.run].omit`, which already has a "Stand-alone kernel-agent CLI tools" section, and `fail_under = 90` is enforced.

## Cleanup

- `ResolveResult.confidence` has no reader anywhere.
- `source_line` is written at `_bypass_report.py:521` but has no consumer; meanwhile the native route computes `line` in the index and then discards it in `as_legacy_tuple()`.
- The latency machinery (`_LATENCY`, `_LatencyBucket`, `_record_latency`, `latency_report`, `reset_latency`) only serves `--bench` and one test, but `_record_latency` runs on every `resolve()` and appends to an unbounded list — a slow leak over a long session.
- Dead on the TraceLens route now that the finder always returns empty metadata: `OpResolution.kernel_kinds` / `prebuilt_binaries` / `primary_kernel_kind` / `primary_prebuilt_binary`, the curated-metadata block in `apply_to`, the `aiter_asm` branch in `classify_patchability`, and the composite fan-out in `leaf_resolutions()`.
- `"symbol_index"` in `_CURATED_RESOLUTION_METHODS` is a no-op: the bypass route sets `source_resolution_method` but never `op_to_source_patchable`, so neither branch in `classify_patchability` fires and it falls through to the heuristics anyway.
- `_CK_SYMBOL_MARKERS`: `"ck_tile::"` is subsumed by `"ck_tile"`.
- `kernel_source_index._NATIVE_EXTS` indexes `.hpp`, but `is_editable_source._NATIVE_SOURCE_EXTS` rejects it, so any `__global__` in a `.hpp` is indexed and then always skipped. Align the two.
- `_find_csrc(pkg_dir)` also probes `pkg_dir.parent` (site-packages itself), so `_has_native()` re-walks the same shared dirs once per package with no memoization.
- `ruff format --check` reformats 2 of the 3 new modules; mypy flags `source_resolver.py:329` `int(rec.get("line") or 0)` (records are `dict[str, object]`).

## Docs / comments

- **`.github/CODEOWNERS:57` still points at the deleted `tools/data/op_to_source.json`** — the only dangling reference left in the repo.
- `source_resolver.py` module docstring says Triton is handled by `_bypass_source_resolver.editable_trace_source`; the wiring now uses `resolve_triton_py`.
- "prebuilt assembly (`.co`)" detection is claimed in the module docstring, `_non_patchable_kind`'s docstring, `resolve()`'s docstring, and the PR description, but the implementation only matches CK markers and can only return `"aiter_ck"` or `""`.
- `OpResolution`'s docstring still describes `dispatch` / `composite` and fan-out semantics; the finder only produces `single`.
- `classify_patchability` still emits `f"op_to_source: {reason}"` and stamps `op_to_source_*` keys (read by `_invocation_spec.py:1004-1005`) while the comments now say "finder". Keeping the key names for compatibility is fine, but say so in the comment.
- `test_live_source_stack.py` docstring points at `results/resolver_benchmarks_vllm/run_live_stack_test.sh`, which is not in the PR. It also uses `check_*` / `_run_all()`, so pytest imports it but never runs it — it is a manual script sitting in `testpaths`.
- The PR description lists a `.gitignore` change that is not in the diff.

ZhengGong-amd (commented):
Re-reviewed after the force-push. Most of the previous round is addressed: the `kname` gate, the mangled-symbol CK fallback (plus boundary anchoring), logging across all four modules, the in-process index singleton, recursive `_dir_signature`, the `.hpp` mismatch, the capped latency samples, `CODEOWNERS`, and the docstrings. Restoring `resolve_by_kernel_name` as tier 3 also undoes the coverage loss. Suite is green (1645 passed), `ruff check` clean.

## New — please fix

`tracelens_analysis.py:162` and `:4300` hard-access attributes that this PR adds to `kernel_source_contract`:

```python
_ACTIVE_FINDER_METHOD = _KSC.METHOD_ACTIVE_FINDER if _KSC is not None else "active_finder"
authoritative = {_KSC.METHOD_ACTIVE_FINDER, _KSC.METHOD_SYMBOL_INDEX, _KSC.METHOD_CURATED}
```

The guard covers "module missing" but not "module older than the script". `tracelens_analysis.py` runs as a standalone subprocess and imports the *installed* hyperloom, which need not be the same tree as the script (cf. `runtime/source-mirrors/`). I hit this for real — `AttributeError: module 'hyperloom.common.kernel_source_contract' has no attribute 'METHOD_ACTIVE_FINDER'`, killing the script at import rather than degrading. CI will be green because there the checkout and the install are one tree. Use `getattr(_KSC, "METHOD_ACTIVE_FINDER", "active_finder")`.

## Still open

- **`non_patchable` is still dropped on the bypass route.** `as_legacy_tuple()` now preserves it ("so callers can tell 'known not rewritable' from 'not found'"), but `_bypass_report.py` still only adopts `method` when `source_file` is truthy — and tier 3 then runs anyway, so a repo-scan hit can override the finder's explicit non-patchable verdict.
- **`"symbol_index"` in `_CURATED_LIKE_METHODS` is still inert.** Only `OpResolution.stamp_onto` (`tracelens_analysis.py:276`) writes `op_to_source_patchable`, so neither branch fires for a bypass candidate.
- **`ResolveResult.confidence` still has no reader** (now populated with four distinct values).
- **CK gating tradeoff is still undocumented.** All 56 `aiter_ck` records in the deleted JSON were `patchable: true` and `forge_submit._resolve_fellow` has a ck-fellow branch that resolution can no longer reach.
- Cache still under `tempfile.gettempdir()`; the two `_main` CLIs still not in `[tool.coverage.run].omit`; `ruff format --check` still reformats `_bypass_source_resolver.py`; `test_live_source_stack.py` still points at a `run_live_stack_test.sh` that is not in the repo.
- **PR description is stale**: still lists `.gitignore` (not in the diff), still claims prebuilt-assembly flagging (not implemented), still says "two-tier" for `_bypass_report` (now three), and omits the restored repo-scan tier, `CODEOWNERS`, and `kernel_source_contract.py`.

Ahmedhasssan-aig:
@ZhengGong-amd I have addressed all of your comments. Please merge the PR whenever you have time. I'd prefer not to delay it, as doing so will require additional rebasing and conflict resolution.

ZhengGong-amd (commented):
Everything from the previous rounds is addressed — thanks. Suite green (1652 passed), `ruff check` clean. Two small leftovers:

**1. The `_KSC` degrade is itself silent.** The crash is fixed, but both `if _KSC is not None:` guards around `write_source_resolution_artifact` (`tracelens_analysis.py:4220` and `:7432`) have no `else` and no log. On a version skew `kernel_source_resolution.json` simply is not written and the operator gets no signal — the same class of silent degrade as the first round. A single `log.warning` covers it.

**2. The `hasattr` allowlist is hand-maintained.** It is exactly complete today, but every read site already uses `getattr` with a default, so the list is a redundant second gate that will silently drift the moment someone uses a tenth attribute and forgets to add it — and the crash returns the same way. Either drop the list and rely on `getattr`, or add a test asserting the list matches actual usage.

ZhengGong-amd (approved):
Approved

Issue #1114, opened 2026-08-06
Diffusion support is xDiT-only: no SGLang multimodal_gen route, and Wan/FLUX.2 denoisers missing from the diffusion FLOP model


sushildubey171:
### Target release

Unsure

### Optimization domain

Inference

### Problem / use case

Hyperloom's diffusion support currently assumes **xDiT**. In `src/hyperloom/inference_optimizer/framework_registry.py`, `sglang` / `vllm` / `atom` are registered as `kind=serving` (throughput `tok/s`, GSM8K accuracy gate) and only `xdit` is `kind=scriptable`.

But SGLang also serves **diffusion** models through its `multimodal_gen` path — that is how we run Wan2.2-T2V-A14B and FLUX in production on gfx950. There is no way to point Hyperloom at such a workload today:

- the `sglang` route applies the LLM benchmark client (`benchmark_serving.py`) plus a GSM8K numeric eval, neither of which describes a text-to-video workload;
- the `xdit` route is a different serving stack, so it does not measure the thing we ship.

**Second gap — the analytic FLOP model is image-only.** `src/hyperloom/agents/kernel/tools/diffusion_flops.py::_CLASS_FAMILY` routes 13 `_class_name` values, all 2D image denoisers. Missing:

| model | denoiser `_class_name` | status |
|---|---|---|
| Wan2.2-T2V-A14B | `WanTransformer3DModel` | not routed |
| FLUX.2-dev | `Flux2Transformer2DModel` | not routed (only FLUX.1's `FluxTransformer2DModel` is) |

For Wan this is more than a table entry: the module has no notion of **latent frames** or **temporal VAE compression** — every constant is `text_tokens` / `vae_spatial`, and the docstring scopes it to "xDiT text-to-**image** models". So the workload-level analytic compute ceiling (`estimate_image_flops` → `diffusion_roofline.py`) cannot be produced for a video model at all.

### Proposed solution

**1. An SGLang diffusion route.** Either a distinct `sglang-diffusion` framework entry, or a modality flag on the existing `sglang` entry, carrying scriptable semantics: a video/image throughput unit, and a quality gate that is not a single reference PNG (the xDiT gate is LPIPS/SSIM/MSE against one image; video needs at minimum a per-frame or distributional equivalent).

[...]

PR #1115, opened 2026-08-06
feat(kernel): route FLUX.2 and count cross-attention Q/O projections

sushildubey171:
**Description: what and why**

Narrowed from the original scope at review request: the Wan 3D/video frame axis is split out to #1140 (draft), since nothing in the tree can currently point at a Wan workload. What remains is immediately usable on the `xdit` path.

### 1. FLUX.2 routing

`Flux2Transformer2DModel` had no `_CLASS_FAMILY` entry, so FLUX.2-dev fell through to the safetensors-header fallback instead of reading its own config. It keeps FLUX.1's dual + single stream structure, so it reuses the `flux` formula:

| | FLUX.1-dev | FLUX.2-dev |
|---|---|---|
| double + single layers | 19 + 38 | 8 + 48 |
| hidden | 3072 | 6144 (48 heads × 128) |
| `mlp_ratio` | absent → 4×hidden | **3.0** → 18432 |
| FFN | GELU (2 matmuls) | **SwiGLU (3 matmuls)** |
| forwards/step | 1 (guidance-distilled) | 1 (guidance-distilled) |

### 2. FLUX.2 SwiGLU

FLUX.2 uses a SwiGLU FFN but exposes no gating flag, so the `use_gated_mlp` heuristic counted it as a 2-matrix FFN. On the shipped config that is **224.0 TF vs a correct 282.5 TF** per forward — the 2-matrix count is 20.7% low (equivalently, correcting it raises the forward 1.261×). Classes in that position are now listed in `_GATED_FFN_CLASSES`. FLUX.1 and every other family are unaffected.

### 3. Cross-attention Q and O

The block counted only K and V. The query stream also needs Q and O (`q_tokens × h → h` each), charged to the query token count rather than the much shorter text length, so the omission grows with resolution — 7.93 TFLOP on a 75600-query block, **4.86%** of that layer. The module docstring only licenses dropping softmax and element-wise work.

The four-projection form is factored into `_cross_attention_block_flops`.

**Scope:** this fixes the **DiT/Sana-style** cross-attention path. `_sana_layer` is its only caller and is fixed here. `_unet_forward_flops` builds its cross-attention inline against a per-level channel count (`cout`) rather than the model hidden size, so **UNet still omits Q/O**. That is pre-existing on `main` and is left for a separate change, which would need the helper parameterized on `cout`.

**Behaviour change on a shipped model:** Sana's per-layer FLOPs rise **10.73%** at 1024² (4.7887e10 → 5.3025e10), because its 300 text tokens are close to its 256 image tokens (`vae_spatial=32`). The ceiling moves the honest way, but any Sana session's `ideal_ms` and `analytic_within_pct` shift with it.

The FLUX.2 test config is now the shipped config verbatim, including `mlp_ratio=3.0` — the previous one omitted it and silently validated the 4.0 default. Tests assert the resolved intermediate (18432), that gating is detected, the 282.5 TF forward, the 1.261× SwiGLU delta, that FLUX.1 stays ungated at 4×hidden, and the cross-attention decomposition (Q/O scale with queries, K/V with text length).

```
pytest src/hyperloom/agents/kernel/tests   ->  1618 passed, 29 skipped
ruff check                                 ->  clean
```

I did not run the full suite to completion locally (it exceeds ~10 min in my environment), so I ran the kernel-agent tree plus the modules that consume these call paths.

**Breaking changes: no**

FLUX.1 and every non-cross-attending family produce identical output. The two intended numeric changes are FLUX.2 (previously unrouted) and Sana (+10.73%/layer), both stated above.

Related: #1149 (`analytic_ceiling` has no `num_gpus`).

### DIFF
--- a/src/hyperloom/agents/kernel/tests/test_diffusion_flops.py
+++ b/src/hyperloom/agents/kernel/tests/test_diffusion_flops.py
@@ -47,6 +47,7 @@ def test_family_routing_for_each_class(tmp_path):
         "ZImageTransformer2DModel": "single",
         "SanaTransformer2DModel": "sana",
         "UNet2DConditionModel": "unet",
+        "Flux2Transformer2DModel": "flux",
     }
     for cls, fam in cases.items():
         d = _write_denoiser(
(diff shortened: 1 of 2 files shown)


lishuoshuo-amd:
The maths here is solid — we checked it independently. Reading the shipped config back out gives 13.63 B parameters for a model named A14B, and 720p × 81 frames resolves to 21 latent frames and 75600 tokens, matching the `(1, 75600, 5, 128)` operand shape in the trace. The causal-VAE off-by-one and the `(t, h, w)` patch read are both right.

Two problems and a scope question.

1. **`--frames` does not reach the call sites that write the report.** `tracelens_analysis.py` and `diffusion_roofline.py` call `analytic_ceiling` without `num_frames`, so a session can only ever use the table default of 81. On the production path that turns a currently-absent ceiling into `ideal_ms = 247779` against a true 39455 — 6.28x — which then drives `analytic_within_pct`.
2. **`forward_flops(g, h, w)` silently undercounts Wan by 73x**, because `frames` defaults to 1 on a public function. `frames: int | None = None`, falling back to `g.default_frames`, matches what `estimate_image_flops` already does.

Separately, cross-attention's Q and O projections are not counted — 4.86% of a Wan layer. `_sana_layer` has the same omission, so this is inherited rather than introduced, but the module docstring says only softmax and element-wise ops are dropped.

On scope: we are not planning the SGLang `multimodal_gen` route, so there is no way to point Hyperloom at a Wan workload — a repository-wide search finds no reference to Wan outside this PR. `Flux2Transformer2DModel` is immediately usable via `--model` on the `xdit` path. Would you be willing to narrow this to the FLUX.2 entry plus the cross-attention fix? The frame-axis work is good and worth revisiting if that route is ever built, but as it stands it serves a workload the tree cannot run.

One more thing worth knowing: `analytic_ceiling` has no `num_gpus`, so the number it produces is single-GPU. Video models are essentially always multi-GPU sequence-parallel, so on Ulysses-8 the ceiling is off by another 8x.

sushildubey171:
Narrowed as you asked, and thanks for checking the maths independently — the 4.86% figure reproduces exactly on my side too, which was a useful signal that we were looking at the same term.

**Scope.** This PR is now `Flux2Transformer2DModel` plus the cross-attention fix. The Wan frame-axis work is split out to a follow-up (draft, linked to #1114) rather than dropped, so it is ready if an SGLang `multimodal_gen` route ever exists — but you are right that shipping it now would add a code path the tree cannot exercise, and I would rather not carry untestable code in a merged PR. All four defects you found are fixed on that branch regardless, so the draft is not a parking lot for known-broken code.

**Cross-attention Q/O.** Fixed, and factored into `_cross_attention_block_flops` so the four-projection form has one definition. `_sana_layer` is the only current caller and carried the same omission, so it is fixed by construction rather than separately. Charging Q/O to the query token count rather than the text length is what makes the term grow with resolution — 7.93 TFLOP on a 75600-query block, 4.86% of the layer.

**FLUX.2.** Without the entry it fell through to the safetensors-header fallback instead of reading its own config. Now: 8 double + 48 single, h=6144, guidance-distilled so `cfg_batch=1`, 4096 image tokens at 1024².

Two of your points now apply only to the follow-up, but both are real and I have fixed them there rather than leaving them for the next round:

- `forward_flops(g, h, w)` defaulting `frames=1` — a public function silently undercounting by 73× is the wrong default. It now takes `frames: int | None = None` and falls back to `g.default_frames`, matching `estimate_image_flops`.
- `--frames` not reaching `tracelens_analysis.py` / `diffusion_roofline.py` — threaded through, since a ceiling that can only ever use the table default is worse than no ceiling.

**On `num_gpus`.** This one I think is worth raising beyond the follow-up: `analytic_ceiling` produces a single-GPU number today, and it is already consumed on the `xdit` path where `TP`/`XDIT_ULYSSES_DEGREE` defaults to 8 in `baseline_xdit.yaml`. So the 8× is not specific to video — a FLUX session at `TP=8` compares a whole-model FLOP count against one GPU's peak, and `analytic_within_pct` inherits it. I have deliberately not changed that here because it would alter numbers on a path you are actively using, and the right shape (divide the ceiling, or report per-GPU explicitly) is your call. Happy to open it as a separate issue with the arithmetic written out if that is useful.

`pytest src/hyperloom/agents/kernel/tests` → 1616 passed, 29 skipped. `ruff check` clean. Rebased onto current `main`.

lishuoshuo-amd:
Narrowing was the right call, and factoring `_cross_attention_block_flops` is the right shape. The latest version still has a few issues we should fix before merging.

**Blocking: FLUX.2 SwiGLU is not detected.** The current logic treats FFN gating as `family in ("sana",) or cfg.get("use_gated_mlp")`, but `Flux2Transformer2DModel` uses SwiGLU and does not expose `use_gated_mlp` in its config. That means FLUX.2 is still counted as a 2-matrix FFN instead of a 3-matrix SwiGLU FFN. On the shipped-shaped config this is **224.0 TF current vs. 282.5 TF expected**, a **20.7% underestimate**.

**Blocking: the FLUX.2 test config does not match the shipped config.** The test omits `mlp_ratio`, so it defaults to `4.0` and validates about **263.0 TF**. The shipped config uses `mlp_ratio=3.0`, which gives **224.0 TF** with the current ungated path, or **282.5 TF** once SwiGLU is counted correctly. The test should use the shipped-shaped config so it protects the production path.

**Blocking: the description / commit message overstates the cross-attention scope.** It says `_sana_layer` is the sole cross-attention consumer, but `_unet_forward_flops` also calls `_cross_attention_flops`:

```
total += _cross_attention_flops(s_tokens, txt, cout)
total += txt * _linear_flops(1.0, ctx, cout) * 2  # cross kv proj
```

Please update the wording to say this PR fixes the DiT/Sana-style path, while UNet remains separate unless you also parameterize the helper for UNet's per-level `cout`.

**Non-blocking / separate issue: UNet has the same Q/O omission.** Measured on an SDXL-shaped UNet at 1024², the missing Q/O is 8.05e10 against the 1.45e10 currently counted — 5.5x the cross-attention term and **8.31% of the whole forward**. This is pre-existing on `main`, so I would not block this PR on fixing UNet, but the PR text should not imply UNet is covered.

**Also worth calling out: Sana's numbers move.** Charging Q/O to the query count raises Sana's per-layer FLOPs by 10.73% at 1024² (4.7887e10 -> 5.3025e10, since `text_tokens=300` is close to its 256 image tokens). The ceiling moves the honest way, but it is a behaviour change on a shipped model and any Sana session's `ideal_ms` and `analytic_within_pct` shift with it.

On `num_gpus`: you're right, and it is not video-specific. `baseline_xdit.yaml:30` is `TP: 8  # = XDIT_ULYSSES_DEGREE`, so every shipped xDiT session already compares a whole-model FLOP count against one GPU's peak. Please open that issue — a separate change is right given it moves numbers on a path in active use.

sushildubey171:
All three blocking items fixed, and your numbers reproduce exactly on my side — 224.0 / 282.5 / 263.0 TF, and Sana 4.7887e10 → 5.3025e10.

**SwiGLU.** You're right that `mlp_ratio=3.0` was the tell — a 3.0 ratio only makes sense against three matrices. Classes that use SwiGLU without exposing a flag now live in `_GATED_FFN_CLASSES`, rather than widening the `use_gated_mlp` heuristic. FLUX.2 goes 224.0 → 282.5 TF/forward; FLUX.1 and everything else are untouched, with a test pinning that.

**Test config.** Now the shipped config verbatim, `mlp_ratio=3.0` included. Good catch that the old one validated the 4.0 default — it was testing a model that doesn't exist. Added assertions on the resolved intermediate (18432), that gating is detected, the 282.5 TF forward, and the 1.261× SwiGLU delta.

**Cross-attention scope.** Corrected in both the commit message and the description. `_unet_forward_flops` does build its cross-attention inline, and against a per-level `cout` rather than the model hidden size, so it is genuinely a different shape — not something the helper covers as written. The text now says this fixes the DiT/Sana path and that UNet still omits Q/O.

I left UNet alone as you suggested. Worth noting for whoever picks it up: parameterizing `_cross_attention_block_flops` on the projection width would cover it, since UNet's per-level `cout` is the only thing that differs.

**Sana.** Called out in the description as a behaviour change on a shipped model, with the per-layer delta and the reason (300 text tokens against 256 image tokens at `vae_spatial=32`).

**num_gpus.** Opened as #1149 with the arithmetic, including the point that a flat `/num_gpus` is an approximation under Ulysses — each rank holds all tokens but a subset of heads for attention, while the projections shard by token.

1618 passed, ruff clean.

xiaofei-zheng (approved):

PR #1116, opened 2026-08-06
fix(kernel): flag impossible Kineto durations, stop using annotation count as denoise steps

sushildubey171:
**Description: what and why**

Two independent defects in the bypass route, both surfaced while analysing a real Wan2.2 (SGLang, 720p/81f, Ulysses-8) trace of 55423 kernels.

### 1. Corrupt Kineto durations were trusted verbatim

Work dispatched to one hardware stream executes serially, so within a `(pid, tid)` row an event's end cannot run past the next event's start. On this trace **7 of 55423 events (0.013%)** violated that by **16251 ms** in total — two "elementwise kernels" at ~3.4 s and three 9.7 MB `Memcpy DtoH` at ~3.15 s.

Hot-kernel ranking is by *summed* duration, so those 7 events became **62% of reported GPU time** and displaced the true #2 and #3 bottlenecks. The report's headline was that exposed device memcpy dominated the workload. It does not. On that trace an Arbor/GEAK campaign would have spent its whole budget chasing a bottleneck that does not exist:

| | reported (raw) | after repairing the 7 durations |
|---|---|---|
| Exposed memcpy | **61.90%** | **0.46%** |
| 1 | SDPA 41.20% | SDPA 59.60% |
| 2 | `aten::copy_` 16.20% *(artifact)* | `aten::addmm` 18.68% |
| 3 | `aten::add` 15.97% *(artifact)* | `record_param_comms` 7.24% |

`_bypass_trace_reader` now checks each device stream pairwise, and the tool raises `bypass_impossible_kernel_durations` naming the stream, the total overrun and the worst offender.

Two design points worth flagging for review:

- **Pairwise, not an aggregate ratio.** I first implemented sum-vs-span and my own test caught it being too weak: a corrupt event also extends its stream's span, dragging the ratio back toward 1.0 and hiding exactly the case worth catching. `test_stream_overlap_detects_impossible_duration` asserts the aggregate ratio stays under 1.02 on a trace the pairwise check flags.
- **Per-stream, not pooled.** Pooling all device events would make summed duration exceed wall span on any genuinely concurrent multi-stream trace — a false positive. `test_stream_overlap_not_triggered_by_concurrent_streams` covers it.

Known limitation, documented in the docstring rather than papered over: a corrupt duration on a stream's *final* event is not detectable this way, because nothing follows it to contradict the value. There is a test pinning that behaviour.

Durations are only reported on, never modified — consistent with the existing "durations stay unclipped" contract in `_finalize`.

### 2. `num_denoise_steps` fell back to the annotation-window count

With no steady window, `inferred_denoise_steps` fell through to `attribution.annotation_window_count`, which counts `user_annotation` ranges — not denoise steps. On this trace that reported **21835 denoise steps for 8 actual steps**, a 2729× error on any per-step figure.

It is worse than the reported number suggests: `resolve_perstep_divisor` *prefers* the inferred value, so the bogus count also won as the diffusion per-step divisor even when `--num-denoise-steps` was passed explicitly — while the result still reported the requested value. Reported and used disagreed.

`_denoise_steps.count_profiler_steps()` already returns the correct answer and is used by the TraceLens route. Bypass now uses it as the fallback and no longer consults the annotation count. The mismatch warning text is corrected — it claimed "the trace-inferred per-step window is used for kernel shares", which described the divisor but not accurately.

**Linked issue(s)**

None open for these; they were found while working on #1114 but are independent of it.

10 new tests. Reader: clean trace stays silent, corrupt duration detected (with the aggregate-ratio counter-assertion), concurrent streams not flagged, worst-stream selection, rounding/short-stream/degenerate-input guards, and the documented final-event limitation. Tool: warning emitted and absent appropriately, denoise steps taken from `ProfilerStep#N` and explicitly *not* from a much larger annotation count, and explicit `--num-denoise-steps` still recorded plus flagged on mismatch.

```
pytest src/hyperloom/agents/kernel/tests            ->  1152 passed, 23 skipped
pytest <bypass_backend, bypass_report, diffusion_roofline>  ->  97 passed
ruff check                                          ->  clean
```

`ruff format --check` still reports the two files as unformatted, but those sites pre-exist on `main` (verified by re-running the check on a stashed tree) and are part of the formatting backlog CONTRIBUTING refers to; I reformatted only my own added lines rather than churning unrelated code. I did not run the full suite to completion locally — it exceeds ~10 min in my environment — so I ran the kernel-agent tree plus the modules that consume these call paths.

**End-to-end validation on the original trace**

```
raw trace:      7 event(s) overrun, 16251.2375 ms total   [warning raised]
                num_denoise_steps = 8   (was 21835)
repaired copy:  no warning                                [no false positive]
                num_denoise_steps = 8
```

The guard's 7 events / 16251.2375 ms independently reproduce a separate clip-based repair of the same trace (`clipped: 7`, `removed_ms: 16251.23746811719`) to the microsecond, from a different algorithm.

**Breaking changes: no**

Purely additive: one new optional `timeline.stream_overlap` key (present only when a stream is inconsistent) and one new health-warning code. No signature changes. The corrected denoise-step source changes `num_denoise_steps` only where it was previously the annotation count, which was never a meaningful value.

### DIFF
--- a/src/hyperloom/agents/kernel/tests/test_bypass_reader.py
+++ b/src/hyperloom/agents/kernel/tests/test_bypass_reader.py
@@ -743,3 +743,152 @@ def test_steady_state_with_zero_annotations_falls_back(tmp_path):
     assert out["aggregation_scope"] == "full_trace"
     assert out.get("steady_window_status")
     assert len(out["kernels"]) == 2
+
+
+# ---- device-stream duration sanity ---------------------------------------
+def _stream_trace(path: Path, events: list[dict]) -> Path:
+    path.write_bytes(json.dumps({"traceEvents": events}).encode("utf-8"))
+    return path
+
+
+def _gpu(name: str, ts: float, dur: float, pid: int = 1, tid: int = 1) -> dict:
+    return {"cat": "kernel", "ph": "X", "name": name, "ts": ts, "dur": dur, "pid": pid, "tid": tid, "args": {}}
+
+
+def test_stream_overlap_absent_on_consistent_trace(tmp_path):
+    """Durations that fit between their neighbours must not be flagged."""
+    tf = _write_trace(tmp_path / "ok.trace.json")
+    out = reader.analyze_trace(tf, top_k=0)
+    assert "stream_overlap" not in (out.get("timeline") or {})
+
+
+def test_stream_overlap_detects_impossible_duration(tmp_path):
+    """An event overrunning its successor on a serial stream is corrupt.
+
+    The corrupt event also extends the stream's span, so a sum-vs-span ratio
+    would read ~1.0 here; the pairwise check still catches it.
+    """
+    events = [
+        _gpu("k_a", 1000, 100),
+        _gpu("corrupt_kernel", 1100, 20_000),  # ends 21100, next starts 1200
+        _gpu("k_b", 1200, 100),
+        _gpu("k_c", 1300, 100),
+    ]
+    tf = _stream_trace(tmp_path / "bad.trace.json", events)
+    out = reader.analyze_trace(tf, top_k=0)
+    so = (out.get("timeline") or {}).get("stream_overlap")
+    assert so, "corrupt duration should be reported"
+    assert (so["pid"], so["tid"]) == (1, 1)
+    assert so["worst_event"] == "corrupt_kernel"
+    assert so["overlapping_events"] == 1
+    # Overruns k_b's start (1200) by 19900us.
+    assert so["worst_event_excess_ms"] == 19.9
+
+    # A naive sum-vs-span ratio would have looked innocent.
+    total_dur = sum(e["dur"] for e in events)
+    span = max(e["ts"] + e["dur"] for e in events) - min(e["ts"] for e in events)
+    assert total_dur / span < 1.02
+
+
+def test_stream_overlap_not_triggered_by_concurrent_streams(tmp_path):
+    """Two genuinely concurrent streams must not be mistaken for corruption.
+
+    Pooling every device event would make the summed duration twice the wall
+    span here, which is exactly the false positive the per-stream check avoids.
+    """
+    events = []
+    for i in range(5):
+        events.append(_gpu(f"s1_{i}", 1000 + i * 100, 100, pid=1, tid=1))
+        events.append(_gpu(f"s2_{i}", 1000 + i * 100, 100, pid=1, tid=2))
+    tf = _stream_trace(tmp_path / "concurrent.trace.json", events)
+    out = reader.analyze_trace(tf, top_k=0)
+    assert "stream_overlap" not in (out.get("timeline") or {})
+
+
+def test_stream_overlap_reports_worst_stream():
+    # Overruns must clear the absolute floor (1 ms) to be reported at all.
+    mild = [(0.0, 5_000.0, "mild"), (2_000.0, 4_000.0, "b"), (10_000.0, 10_000.0, "c")]
+    severe = [(0.0, 30_000.0, "severe"), (1_000.0, 2_000.0, "b"), (10_000.0, 10_000.0, "c")]
+    worst = reader._stream_overlap_health({(1, 1): list(mild), (1, 2): list(severe)})
+    assert worst["tid"] == 2
+    assert worst["worst_event"] == "severe"
+
+
+def test_stream_overlap_ignores_rounding_and_short_streams():
+    # Sub-microsecond overrun is timestamp rounding, not corruption.
+    tiny = [(0.0, 100.5, "a"), (100.0, 200.0, "b"), (1000.0, 1000.0, "c")]
+    assert reader._stream_overlap_health({(1, 1): tiny}) == {}
+    # Real but immaterial overlap stays below the reporting share.
+    small = [(0.0, 110.0, "a"), (100.0, 200.0, "b"), (100_000.0, 100_000.0, "c")]
+    assert reader._stream_overlap_health({(1, 1): small}) == {}
+    # Degenerate inputs must not divide by zero.
+    assert reader._stream_overlap_health({(1, 1): [(0.0, 10.0, "only")]}) == {}
+    assert reader._stream_overlap_health({(1, 1): [(5.0, 5.0, "a"), (5.0, 5.0, "b")]}) == {}
+    assert reader._stream_overlap_health({}) == {}
+
+
+def test_stream_overlap_final_event_is_not_detectable():
+    """Documented limitation: nothing follows the last event to contradict it."""
+    evs = [(0.0, 100.0, "a"), (100.0, 99_999.0, "corrupt_last")]
+    assert reader._stream_overlap_health({(1, 1): evs}) == {}
+
+
+# ---- review round 2: threshold denominator, absolute floor, severity ------
+def _spread(n: int, dur: float, stride: float, prefix: str = "k") -> list[tuple[float, float, str]]:
+    return [(i * stride, i * stride + dur, f"{prefix}{i}") for i in range(n)]
+
+
+def test_threshold_uses_summed_device_time_not_span():
+    """A lone corrupt duration must not hide behind the span it inflates.
+
+    1000 kernels spread over 60 s with one duration written as 2 s: the corrupt
+    event is 95% of the summed device time but only ~3% of the wall span, so a
+    span-denominator threshold reported nothing -- the same blind spot the
+    pairwise detector exists to avoid, moved into the threshold.
+    """
+    evs = _spread(1000, 100.0, 60_000_000 / 1000)
+    evs[500] = (evs[500][0], evs[500][0] + 2_000_000.0, "corrupt_2s")
+    device_us = sum(e - s for s, e, _ in evs)
+    span_us = max(e for _s, e, _ in evs) - evs[0][0]
+
+    out = reader._stream_overlap_health({(1, 1): list(evs)})
+    assert out, "corrupt duration dominating summed device time must be reported"
+    assert out["worst_event"] == "corrupt_2s"
+    assert out["severity"] == "warning"
+    # The share is meaningful against summed device time, negligible against span.
+    assert out["excess_share"] > 0.9
+    assert (out["excess_ms"] * 1000.0) / span_us < 0.05
+    assert abs(out["device_ms"] - device_us / 1000.0) < 1e-3
+
+
+def test_dense_jitter_is_not_escalated():
+    """Many sub-millisecond overruns are profiler jitter, not corruption.
+
+    Accumulates well past the 5% share while the worst single overrun is 4 us;
+    telling the user to recapture the trace here is a false alarm.
+    """
+    evs = _spread(20_000, 10.0, 6.0)  # each overruns the next by 4 us
+    assert reader._stream_overlap_health({(1, 1): list(evs)}) == {}
+
+
+def test_identical_timestamps_are_not_overlap():
+    """Events sharing a ts cannot contradict each other."""
+    same = [(0.0, 50_000.0, "a"), (0.0, 50_000.0, "b"), (0.0, 50_000.0, "c"), (100_000.0, 100_000.0, "d")]
+    assert reader._stream_overlap_health({(1, 1): same}) == {}
+
+
+def test_severity_grades_with_share_of_device_time():
+    """Marginal distortion is info; a materially wrong ranking is a warning."""
+    # ~10% of summed device time -> info.
+    mild = _spread(40, 10_000.0, 10_000.0)
+    mild[0] = (0.0, 10_000.0 + 40_000.0, "mild_overrun")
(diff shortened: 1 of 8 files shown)


lishuoshuo-amd:
This is the one we'd most like to land — it is framework- and model-agnostic, and it improves the `--framework sglang` bypass route you're already using. The core insight holds: work on one hardware stream is serial, so an event overrunning its successor is real evidence of a corrupt duration, and pairwise beats an aggregate ratio.

Three things need fixing first; each has a reproducer.

1. **The report threshold uses the polluted span as its denominator**, so it misses the case this was built for. `excess_us < 0.05 * span_us`, and the corrupt event inflates `span_us` too — the same blind spot the docstring argues against, moved from the detector into the threshold. On 1000 kernels spread over 60s with one duration written as 2s (95.2% of summed GPU time), `_stream_overlap_health` returns `{}`. Suggest comparing against the stream's summed device time, plus an absolute floor.
2. **No absolute floor means healthy traces are flagged** at the highest severity. A stream of short, dense kernels with microsecond timestamp collisions accumulates 5% while the worst single overrun is ~4µs, and the user is told the hot-kernel table is unreliable and to recapture. Suggest skipping adjacent events with identical `ts` (they cannot contradict each other) and grading the wording by severity.
3. **Directory input reads a different file than the reader does**, which nullifies the second fix. `count_profiler_steps(args.trace_input)` uses `sorted(p.glob("*.json"))[0]`, so a stray `config.json` wins over `trace.json` and the count comes back 0 where the reader resolved 8 steps. Suggest `count_profiler_steps(analyze.get("trace_file") or args.trace_input)`.

Two more that are not code fixes.

**Reported vs used divisor** — `resolve_perstep_divisor` prefers the inferred value while the result reports the requested one, and the new `test_explicit_denoise_steps_still_recorded` pins that disagreement. We would rather an explicit `--num-denoise-steps` win, since Hyperloom cannot know what an operator's `prof.step()` brackets.

**`(pid, tid)` as the stream key** — on ROCm, HIP streams beyond `GPU_MAX_HW_QUEUES` share a hardware queue; if Kineto records the queue rather than the stream, genuinely concurrent kernels would be read as corrupt. Could you confirm against a real Wan2.2 trace? We have no way to falsify it here.

Still only one conflict against `main`, in `test_bypass_trace_analysis.py`, and it is an end-of-file adjacency — both sides kept.

sushildubey171:
All three reproduced. Rebased onto current `main` and squashed; the conflict was the end-of-file adjacency you spotted, both sides kept.

**1. Threshold denominator.** Fixed — the share is now taken against the stream's summed device time rather than its span, since a corrupt event inflates its own span. Your reproducer lands at 95.2% against summed time and ~3% against span; `test_threshold_uses_summed_device_time_not_span` asserts both so the old form can't return unnoticed.

**2. Absolute floor.** Added, and it's the discriminator rather than the share: a stream now needs at least one overrun ≥ 1 ms before anything is reported. Your dense-jitter case accumulates 40% share with a 4 µs worst overrun and is silent. Adjacent events sharing a `ts` are skipped. Severity is graded — ≥25% of summed device time keeps the recapture wording, below that it drops to `info`.

**3. Directory input.** Fixed as suggested. `test_denoise_steps_read_the_file_the_reader_analyzed` builds a directory with a stray `config.json` sorting ahead of the real trace and asserts 4, not 0.

**4. Reported vs used divisor.** Agreed — an explicit `--num-denoise-steps` now wins and the reported value matches the divisor used. The test pinning the old disagreement is replaced.

**5. `(pid, tid)` as the stream key.** Measured on the real Wan2.2 ROCm trace (gfx950, Ulysses-8, 55423 device events, 2 rows):

| stream | events | overlaps | <1 µs | 1 µs–1 ms | 1–100 ms | >100 ms |
|---|---|---|---|---|---|---|
| pid=2 tid=3 | 54730 | 17 | 10 | 2 | **0** | 5 |
| pid=2 tid=50 | 693 | **0** | 0 | 0 | 0 | 0 |

Bimodal, with nothing in the 1–100 ms band. If the key were a hardware queue carrying concurrent HIP streams, overlaps would cluster at kernel-duration scale; instead that band is empty and the second row has none. It's one trace on one device, so not conclusive — but a false positive would have to clear both the 1 ms floor and the 5% share, and concurrent kernels overlap by roughly their own duration, not by seconds.

Real-trace behaviour is unchanged: 7 events, 16251.2375 ms, now reported as 51.6% of the 31474.1757 ms on that stream. 1627 passed, ruff clean.

lishuoshuo-amd:
All three reproduce as fixed, thanks — and thank you especially for measuring `(pid, tid)` on the real trace. The empty 1–100 ms band is a good argument: concurrent kernels would overlap by roughly their own duration, so a cluster there is what a mis-keyed stream would look like, and there isn't one. Combined with the 1 ms floor and the share gate, that's enough for us.

Verified here: the share is taken against summed device time, the dense-jitter trace is silent because the floor is the discriminator rather than the share, identical timestamps are skipped, severity grades at 25%, and `count_profiler_steps` reads `analyze["trace_file"]`. 1624 passed; the three failures all reproduce on pristine `main`, and the `ruff format` debt on those four files pre-dates this PR — no action needed on either.

One item left, and it's about how the divisor fix was made rather than what it does.

Reversing the arguments gets the right answer for bypass, but `resolve_perstep_divisor` is shared, and TraceLens still calls it in the documented order:

```
bypass:    resolve_perstep_divisor(requested_denoise_steps, inferred_denoise_steps)
tracelens: resolve_perstep_divisor(count_profiler_steps(...), num_denoise_steps)
```

The helper returns its first positive argument, so with `inferred=3, requested=9` bypass now divides by 9 and TraceLens by 3. `request_handlers` picks between the two routes for the same workload based on `HYPERLOOM_TRACE_ANALYSIS_ROUTE`, so the per-step figures in `diffusion_roofline.json` now depend on an environment variable rather than on the workload.

There's a readability trap alongside it: the parameters are named `inferred_steps, requested_steps` and the docstring says it "prefers the steps ACTUALLY IN the analyzed window", so the bypass call now passes `requested` into a parameter named `inferred_steps`. The next person to touch this will read the name, not the call order.

Could you change the helper rather than the call order — invert the documented preference, rename the parameters to match, and update the TraceLens call site in the same commit? If TraceLens should keep preferring the inferred count, then make both explicit (`prefer="requested"` / `prefer="inferred"`) so the difference is a decision rather than an argument position. Either shape is fine; what we'd like is for the precedence to be stated at each call site.

Nothing else blocking.

sushildubey171:
Good catch — reversing the arguments made bypass right and left the helper shared between two routes with opposite behaviour. Fixed in the helper.

`resolve_perstep_divisor` now takes `(requested_steps, inferred_steps)`, prefers the requested count, and its docstring says so. Both call sites pass by keyword, so the precedence reads at the point of use rather than depending on argument position:

```python
resolve_perstep_divisor(requested_steps=..., inferred_steps=...)
```

TraceLens is updated in the same commit, so with inferred=3 and requested=9 both routes now divide by 9 and the per-step figure no longer depends on `HYPERLOOM_TRACE_ANALYSIS_ROUTE`. I went with one precedence rather than a `prefer=` kwarg because your reasoning — Hyperloom cannot know what an operator's `prof.step()` brackets — applies to both routes equally, so a per-route difference would have been a distinction without a cause.

Helper tests are rewritten for the new precedence and keyword form, plus one asserting both routes resolve the same divisor across `(requested, inferred)` combinations.

1628 passed, ruff clean.

lishuoshuo-amd:
Moving the precedence into the helper is the shape we wanted, and the reasoning for one precedence over a `prefer=` kwarg is right: the argument applies to both routes equally.

Verified here. `resolve_perstep_divisor(requested_steps, inferred_steps)` prefers the requested count and the function docstring says so. Both production call sites — `bypass_trace_analysis.py:860` and `tracelens_analysis.py:6636` — pass by keyword, and those are the only two in the repo. Driving the helper with the arguments each call site actually binds: `requested=9, inferred=3` gives 9 on both routes, `requested` absent gives 3 on both, `requested=0, inferred=0` gives `None` on both. The five earlier fixes are untouched — `734cfbe5d` does not modify `_bypass_trace_reader.py` at all. 1628 passed.

One thing worth recording that argues for this change beyond what you claimed: the bypass summary field has always been `"num_denoise_steps": requested_denoise_steps or inferred_denoise_steps` (`bypass_trace_analysis.py:897`), so on `main` the reported value and the divisor already disagreed whenever both were positive. This does not just align the two routes, it makes each route internally consistent.

**The module docstring now contradicts the function.** Eighteen lines above the new docstring, `_denoise_steps.py:1-8` still says the divisor "MUST be the number of denoise steps ACTUALLY IN the analyzed data ... NOT the requested full sampler schedule". That is a substantive correctness claim, and it is the opposite of what the helper does. The trap we asked you to remove moved from the parameter names to the top of the file. `diffusion_roofline.py:135`, `:221` and `:270` still describe the parameter as "Denoise steps in the profiled window".

**`test_both_routes_resolve_the_same_divisor` cannot fail.** Both variables call the helper with identical keyword arguments, so the assertion is that the function is deterministic; it says nothing about the call sites. Two mutations confirm it: reverting `tracelens_analysis.py:6636` to the old argument order — which reintroduces exactly the route divergence this PR fixes — leaves the suite green, and swapping the arguments at `bypass_trace_analysis.py:860` does too. Neither call site has coverage. `test_explicit_denoise_steps_wins_and_is_flagged` looks like it covers this but asserts `result["num_denoise_steps"]`, which comes from the summary field at `:897` and never reaches the helper, so it survives both mutations as well. Asserting the argument binding at each call site, or comparing the divisor that lands in `diffusion_roofline.json` from both routes, would hold.

**The two CLIs disagree on the `--num-denoise-steps` default.** `bypass_trace_analysis.py:446` hardcodes `default=0` while `tracelens_analysis.py:6803` reads `default=int(os.environ.get('HYPERLOOM_NUM_DENOISE_STEPS', '0') or 0)`. With `HYPERLOOM_NUM_DENOISE_STEPS=9`, the flag not passed and `inferred=3`, bypass divides by 3 and TraceLens by 9 — the route dependence we set out to remove, through a different door. `request_handlers.py:4128-4136` forwards the flag to both routes without distinguishing them, so the window is narrow, but this one is new in this PR: on `main` the variable cannot affect the divisor because inferred always won. A matching env fallback on the bypass default closes it.

Not blocking, for the record: when `requested` is absent the two routes still fall back to different quantities — bypass uses `steady_window["step_count"]` (`bypass_trace_analysis.py:702`), TraceLens uses the deduplicated `ProfilerStep#N` count. Those can differ on the same trace, so route independence holds only while `requested > 0`. Pre-existing, and this PR strictly improves it.

sushildubey171:
All three fixed, and the mutation testing was the useful part of this — I reproduced both mutations before changing anything and confirmed the suite stayed green, so the coverage gap was real.

**Module docstring.** Rewritten. You're right that the trap moved rather than went away: the file header still made the opposite correctness claim eighteen lines above the function. Also fixed the three `num_denoise_steps` parameter descriptions and the CLI help in `diffusion_roofline.py`. The docstring now records your non-blocking point too — the fallbacks still differ (bypass uses `steady_window["step_count"]`, TraceLens the deduplicated `ProfilerStep#N` count), so route independence holds only while `requested > 0`.

**Call-site coverage.** Replaced. The old test called the helper twice with identical arguments, so it asserted determinism and nothing else. The new ones record what each route *binds* to each parameter, driving the real paths — `main()` for bypass, `write_reports()` for TraceLens — against a trace with three ProfilerStep markers so inferred (3) is distinguishable from requested (9). Verified against your two mutations plus a third:

| mutation | result |
|---|---|
| swap arguments at `bypass_trace_analysis.py:860` | bypass binding test fails |
| revert `tracelens_analysis.py:6636` to inferred-first | TraceLens binding test fails |
| revert the bypass env default to `0` | CLI default test fails |

Each fails exactly the test that covers it, and nothing else.

**CLI defaults.** Bypass now reads `HYPERLOOM_NUM_DENOISE_STEPS` like the TraceLens CLI. You're right that this one is new in this PR — on `main` the variable could not reach the divisor because the inferred count always won, so the previous commit opened it. Test covers the env default, the explicit flag winning over it, and the unset case.

Thanks also for the `:897` observation — the summary field and the divisor already disagreed on `main` whenever both were positive. That is a better argument for the change than the one I made, and I have folded it into the commit message.

1630 passed, ruff clean. `_bypass_trace_reader.py` is untouched by this round, so the five earlier fixes are unchanged.

sushildubey171:
@xiaofei-zheng , @lishuoshuo-amd please review it, thanks!

tsrikris:
@lishuoshuo-amd heads up that this entire path of work will be made redundant when native TraceLens support is brought up. We leave it up to you to decide how to handle the implication and burden of unverified/incorrect implementations.

CC: @chaojhou, @zhenygu, @devalshahamd, @kyle-hoffmeyer

xiaofei-zheng (approved):

PR #1117, opened 2026-08-06
fix(orchestrator): gate diffusion behaviour on registry kind, not the xdit name

### LINKED ISSUE #1114: Diffusion support is xDiT-only: no SGLang multimodal_gen route, and Wan/FLUX.2 denoisers missing from the diffusion FLOP model

sushildubey171:
**Description: what and why**

`hunyuan_image3` is registered `kind=scriptable` with `throughput_unit="img/s"` and ships `baseline_hunyuan_image3.yaml` / `profile_hunyuan_image3.yaml` — but several call sites decide "is this a diffusion workload?" by comparing the framework name to `"xdit"`. Those sites silently treat it as an LLM workload.

Before this change a `hunyuan_image3` session:

| | behaviour | should be |
|---|---|---|
| `SharedState._roofline_throughput_unit` | `tok/s` | `img/s` (its own registry entry) |
| bypass analysis report unit | `tok/s` | `img/s` |
| bypass diffusion-roofline sidecar | not emitted | emitted |
| bypass steady-state default | off | on |
| steady-window repeat threshold | LLM threshold | diffusion threshold (2) |
| `compute_roofline_breakdown_from_state` | LLM decode roofline | diffusion ceiling |

All of these now consult `framework_registry` (`is_scriptable` / `throughput_unit`), which is already the canonical predicate in 26 other places. `executors/profile.py:632` is the pattern I followed — it calls `is_scriptable()` for the server-less branch and only *then* narrows to `xdit` for the xfuser-specific check, with a comment noting hunyuan_image3 must not trigger it.

The kernel tools keep working standalone: they try the registry and fall back to a name check, mirroring the existing `_is_scriptable_framework` helper in `tracelens_analysis.py`.

**Deliberately not changed.** I went through every `"xdit"` literal in non-test code; most are correct and are left alone: the `xfuser` pip name and xDiT git URL, `XditAdapter` registration, per-framework config paths, the PATH guard for the `xdit` console script + `hipcc`, the XDIT_* env blacklist, the xDiT default grid seeds, and the xfuser baked-profiler verification. `cli/model_gate.py` already calls `is_scriptable()` with the literal only as an import-failure fallback.

The two health-warning codes keep their `bypass_xdit_*` spelling for wire stability even though the condition is now generic; only their messages change from "xDiT" to "diffusion". Renaming those codes felt like a separate call for you to make.

**Linked issue(s)**

Addresses item 3 of #1114 (see [my correction comment](https://github.com/AMD-AGI/Hyperloom/issues/1114#issuecomment-5207776396) there — my original "15 hardcoded comparisons" was a raw grep count and overstated it; this PR covers the subset that is genuinely wrong). Does not close #1114, whose main ask is an SGLang diffusion route.

This is also the prerequisite that issue names: with the kind checks registry-driven, adding a new scriptable framework gets closer to the "single-table edit" `framework_registry`'s own docstring promises. I deliberately did **not** add an `sglang-diffusion` entry here — without a matching Magpie wrapper script (external repo) it would be dead config, and #1114 is asking you how you'd want that shaped.

No test in the tree referenced `hunyuan_image3` before this change, which is why the divergence went unnoticed. The new tests assert against **every registry entry** rather than a hardcoded list, so a future framework cannot regress the same way:

- scriptable detection and throughput unit match `FRAMEWORKS[*].kind` / `.throughput_unit` for all entries, in both the tool and the reader
- the standalone fallback path (registry import forced to fail) still classifies xdit and hunyuan_image3 correctly
- `compute_roofline_breakdown_from_state` routes xdit **and** hunyuan_image3 to the diffusion ceiling, and sglang/vllm/atom to the decode path
- unknown/empty framework still resolves to `tok/s` + decode

```
pytest src/hyperloom/agents/kernel/tests                     ->  1146 passed, 23 skipped
pytest <roofline_ceiling, framework_adapters, diffusion_roofline,
        xdit_integration_units, backend_gating>              ->   264 passed
ruff check                                                   ->  clean
ruff format --check                                          ->  identical before/after (no new debt)
```

I did not run the full suite to completion locally — it exceeds ~10 min in my environment — so I ran the kernel-agent tree plus the modules consuming these call paths.

**Breaking changes: no**

Serving frameworks and unknown/empty names are unchanged: the registry falls back to `sglang`, so they still resolve to `tok/s` and the decode roofline. The only behaviour that changes is for `kind=scriptable` frameworks other than `xdit` — i.e. `hunyuan_image3` — which now gets the diffusion treatment its registry entry already declared. Warning code strings are unchanged.

### DIFF
--- a/src/hyperloom/agents/kernel/tests/test_bypass_reader.py
+++ b/src/hyperloom/agents/kernel/tests/test_bypass_reader.py
@@ -620,3 +620,14 @@ def test_steady_state_with_zero_annotations_falls_back(tmp_path):
     assert out["aggregation_scope"] == "full_trace"
     assert out.get("steady_window_status")
     assert len(out["kernels"]) == 2
+
+
+def test_reader_scriptable_detection_is_registry_driven():
+    """The steady-window threshold must apply to every scriptable framework."""
+    from hyperloom.inference_optimizer import framework_registry as fr
+
+    for name, spec in fr.FRAMEWORKS.items():
+        assert reader._is_scriptable_framework(name) is (spec.kind == fr.SCRIPTABLE), name
+    assert reader._is_scriptable_framework("hunyuan_image3") is True
+    assert reader._is_scriptable_framework("sglang") is False
+    assert reader._is_scriptable_framework(None) is False
--- a/src/hyperloom/agents/kernel/tests/test_bypass_trace_analysis.py
+++ b/src/hyperloom/agents/kernel/tests/test_bypass_trace_analysis.py
@@ -896,3 +896,48 @@ def test_single_graph_launch_low_busy_not_under_recorded(tmp_path):
     cov = bta._reader.analyze_trace(str(trace), top_k=0)["graph_coverage"]
     assert cov["graph_launch_count"] == 1
     assert cov["graph_under_recorded"] is False
+
+
+# ---- scriptable-framework detection is registry-driven -------------------
+def test_scriptable_detection_covers_every_registry_entry():
+    """Not just xdit: any kind=scriptable framework must be detected.
+
+    hunyuan_image3 ships baseline/profile configs and is registered
+    kind=scriptable, so gating on the literal "xdit" silently treated it as an
+    LLM workload.
+    """
+    from hyperloom.inference_optimizer import framework_registry as fr
+
+    for name, spec in fr.FRAMEWORKS.items():
+        assert bta._is_scriptable_framework(name) is (spec.kind == fr.SCRIPTABLE), name
+
+
+def test_scriptable_throughput_unit_matches_registry():
+    from hyperloom.inference_optimizer import framework_registry as fr
+
+    for name, spec in fr.FRAMEWORKS.items():
+        assert bta._scriptable_throughput_unit(name) == spec.throughput_unit, name
+    assert bta._scriptable_throughput_unit("hunyuan_image3") == "img/s"
+    # Unknown / empty keeps the previous tok/s default.
+    for unknown in ("", None, "not-a-framework"):
+        assert bta._scriptable_throughput_unit(unknown) == "tok/s"
+        assert bta._is_scriptable_framework(unknown) is False
+
+
+def test_scriptable_helpers_fall_back_without_the_package(monkeypatch):
+    """Standalone use (no importable inference_optimizer) still works."""
+    import builtins
+
+    real_import = builtins.__import__
+
+    def _no_registry(name, *a, **k):
+        if "framework_registry" in name:
+            raise ImportError("simulated standalone run")
+        return real_import(name, *a, **k)
+
+    monkeypatch.setattr(builtins, "__import__", _no_registry)
+    assert bta._is_scriptable_framework("xdit") is True
+    assert bta._is_scriptable_framework("hunyuan_image3") is True
+    assert bta._is_scriptable_framework("sglang") is False
+    assert bta._scriptable_throughput_unit("hunyuan_image3") == "img/s"
+    assert bta._scriptable_throughput_unit("sglang") == "tok/s"
--- a/src/hyperloom/agents/kernel/tools/_bypass_trace_reader.py
+++ b/src/hyperloom/agents/kernel/tools/_bypass_trace_reader.py
@@ -273,6 +273,28 @@ def stream_events(fileobj, bufsize: int = 8 * 1024 * 1024) -> Iterator[dict]:
             pos = 0
 
 
+def _is_scriptable_framework(framework: str | None) -> bool:
+    """Return whether ``framework`` is a server-less scriptable workload.
+
+    Prefers the canonical ``framework_registry`` so every ``kind=scriptable``
+    entry is covered rather than ``xdit`` alone; falls back to a name check so
+    the reader stays usable when run standalone (outside an importable
+    ``inference_optimizer`` package).
+
+    Args:
+        framework: Framework name (matched case-insensitively).
+
+    Returns:
+        bool: ``True`` for scriptable frameworks.
+    """
+    try:
+        from hyperloom.inference_optimizer.framework_registry import is_scriptable
+
+        return is_scriptable(framework)
+    except Exception:  # noqa: BLE001 - standalone use must not hard-fail
+        return str(framework or "").strip().lower() in {"xdit", "hunyuan_image3"}
+
+
 def _union_ms(intervals: list[tuple[float, float]]) -> float:
     """Return the union length (in ms) of ``[start_us, end_us)`` intervals."""
     if not intervals:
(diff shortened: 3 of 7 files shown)


lishuoshuo-amd:
Thanks for this — the direction was right, and #1136 adopted it: the predicate now lives in the registry as `FrameworkSpec.has_denoiser_config` rather than being matched against a name.

Closing this one, because the case it was written for is gone and part of it would now regress:

- `hunyuan_image3` was retired in #1136. The registry is `sglang`/`vllm`/`atom`/`xdit`/`custom`, and `is_supported("hunyuan_image3")` is `False`.
- Two of the five call sites are already fixed there: the bypass report reads `throughput_unit()` from the registry, and both trace routes gate the `diffusion_roofline.json` sidecar on `is_scriptable` instead of the `xdit` literal.
- `roofline_ceiling.py:1480` keeps `== "xdit"` deliberately. `custom` is scriptable but its model arrives from the operator at launch, so no denoiser geometry can be resolved and a guessed one is worse than none — we measured a HunyuanVideo checkpoint being labelled `FluxTransformer2DModel` and costed at one frame. Routing `custom` into the diffusion branch there also loses the `tok/s` ceiling a `custom`-wrapped LLM workload legitimately has.
- The two remaining literals (`_should_enable_steady`, the repeat threshold) are judgement calls for `custom` rather than defects, since we cannot know whether an operator's workload has homogeneous steps.

This has also grown to three conflicting files against `main`. If you'd like the last two sites made registry-driven, a small follow-up on top of current `main` would be easier to review than a rebase of this one.

sushildubey171:
Closing, as you suggested — #1136 is a better version of this than the PR was.

`FrameworkSpec.has_denoiser_config` is the right split. I had conflated "server-less" with "has resolvable denoiser geometry", and `custom` is exactly the case that breaks the conflation: scriptable, but with a model that only exists at launch. The HunyuanVideo-checkpoint-labelled-`FluxTransformer2DModel`-and-costed-at-one-frame example is a convincing argument that a guessed geometry is worse than none, and I hadn't considered that routing `custom` into the diffusion branch also throws away a legitimate `tok/s` ceiling for a `custom`-wrapped LLM.

Agreed on `roofline_ceiling.py:1480` keeping the literal, and on the two remaining sites (`_should_enable_steady`, the repeat threshold) being judgement calls for `custom` rather than defects — an operator's workload having homogeneous steps is not something the tree can know.

I'd already narrowed my own claim here after reading each site (comment on #1114): the "15 hardcoded comparisons" was a raw grep count and most of those literals are correct — the `xfuser` pip name, the git URL, `XditAdapter`, the config path maps, the PATH guard, the XDIT_* blacklist, the default grid seeds. Sorry for the noise that generated.

Not opening the follow-up for the last two sites for now, since you've called them judgement calls rather than bugs; happy to if you'd rather they were registry-driven.

PR #1121, opened 2026-08-07
fix(kernel): honor graph under-recording in the TraceLens idle gate

BaoYunkai:
## Summary

The high-idle gate suppresses the hot-kernel candidate list (skipping the forge kernel lane entirely) whenever GPU idle % exceeds the threshold (default 80%). The **bypass** trace route already skips that gate when the trace is a cuda/HIP-graph **under-recorded** capture — under continuous graph replay the profiler activity buffer overflows and only ~1 of N replays is recorded, so idle% is inflated (`bypass_graph_under_recorded`). The **TraceLens** route (deterministic + LLM-orchestrator) applied the idle gate with **no such guard**.

Result: a graph-mode workload that is actually compute/GEMM-bound (rocm-smi shows full saturation) gets reported as ~87% idle from the roofline capture, every hot kernel is suppressed, and forge-loop is wrongly skipped as "host/scheduling-bound".

Observed on Qwen3-8B sglang (`.../safe-opt-claw-top-models-forge-qwen-sgla-b0c92b15-a1/.../20260806T082051Z`): `busy_of_wall=0.093` roofline artifact → `report_source=skipped:high_gpu_idle_pct`, while the session's own scouts flagged the 87% idle as a "cuda-graph capture artifact" and the real workload as GEMM-bound.

## Changes

- `tracelens_analysis.py`:
  - `_graph_coverage_from_raw_trace()` — reuse the tested `_bypass_trace_reader.analyze_trace(..., emit_launches=True)` graph-coverage detection; fail-open to `{}`.
  - `_evaluate_idle_gate_with_graph_guard()` — when idle exceeds the threshold, first check graph under-recording; if under-recorded, skip suppression and return the `bypass_graph_under_recorded` warning instead of `high_gpu_idle_pct`.
  - Wire both TraceLens idle-gate sites (deterministic + LLM-orchestrator) through the guard.
- Aligns the TraceLens route with the semantics the bypass route already enforces; no new mechanism.

## Test plan

- New unit tests in `test_tracelens_csv.py`: under-recorded → skip suppression; not-under-recorded → plain gate applies; below-threshold → no probe; bad trace path → never raises.
- `pytest test_tracelens_csv.py test_idle_gate.py test_bypass_trace_analysis.py` — 273 passed, no regressions.

### DIFF
--- a/src/hyperloom/agents/kernel/tests/test_bypass_trace_analysis.py
+++ b/src/hyperloom/agents/kernel/tests/test_bypass_trace_analysis.py
@@ -940,3 +940,63 @@ def test_fully_recorded_idle_graph_still_suppressed_by_idle_gate(tmp_path, capsy
     assert "high_gpu_idle_pct" in codes
     assert "bypass_graph_under_recorded" not in codes
     assert not result["hot_kernels"]
+
+
+def test_finalize_graph_coverage_is_whole_trace_scoped_under_steady_window():
+    """Regression: recorded-launch coverage must be computed over the FULL event
+    stream, not the steady window. A fully-recorded 4-launch trace whose steady
+    window happens to clip to a single replay must NOT read as 1/4 under-recorded
+    (the denominator graph_launch_count is whole-trace, so the numerator must be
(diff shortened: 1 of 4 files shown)


jiaqiang-dot-liu (approved):
LGTM

PR #1126, opened 2026-08-07
Add ATOM framework Docker integration script

AMD-melliott:
## Context

atom is already a first-class registered framework in `framework_registry.py`, but there's no scripted path to actually stand it up in Docker mode the way there is for vLLM/SGLang. `hyperloom-custom-advanced`'s Docker section only lists vLLM/SGLang images, and Hyperloom's framework auto-detection doesn't probe for ATOM at all.

This script fills that gap for initial ATOM testing: provisions a container from rocm/atom:latest, optionally swaps in a local ATOM/aiter checkout via editable install (a personal fork branch, or a bleeding-edge main checkout newer than what's baked into the base image). This bleeding-edge-checkout capability mattered in practice: it's what let us validate a real upstream ATOM fix before it had even landed on main.

Validated end-to-end: `--framework atom` optimize path against DeepSeek-V4-Flash-0731 on 8x MI300X, container provisioning through TraceLens/kernel tracing under real 8-GPU load, no framework-side changes needed.

Not claiming full ATOM parity with vLLM/SGLang in the skills yet (that gap,  and others found along the way, is in a separate hackathon feedback submission).

### DIFF
--- a/scripts/atom_fork_docker_setup.sh
+++ b/scripts/atom_fork_docker_setup.sh
@@ -1,0 +1,149 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: 2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+# Bring up a Docker container for running Hyperloom's `--framework atom`
+# path against a local ATOM (and optionally aiter) fork checkout, instead of
+# whatever ATOM commit happens to be baked into the base image.
+#
+# Background: `atom` is a registered Hyperloom serving framework
+# (hyperloom/inference_optimizer/framework_registry.py) but Hyperloom does
+# not build/install it the way it does sglang/vllm -- `--install-framework`
+# only auto-installs those two (install_baremetal.sh's
+# resolve_installed_framework() has no atom probe), so ATOM must already be
+# importable in the target environment. This script provisions that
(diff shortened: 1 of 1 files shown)


lishuoshuo-amd:
Thanks for capturing this validation path. I can see the debugging value here, especially for replaying the specific ATOM/aiter fork environment used during the investigation.

That said, I don't think this should be merged as a maintained project script. It feels more like a snapshot of a personal debugging workflow than a reusable Hyperloom entry point, since it depends on a specific local checkout layout and carries environment-specific workarounds.

My recommendation is to close this PR rather than merge the standalone script. We do plan to support ATOM in the project flow, and a better follow-up would be to cover ATOM through the examples/docs path so users have a standard way to run Hyperloom with ATOM.

AMD-melliott:
@lishuoshuo-amd Thanks for the review! I'll close this PR. It was submitted as-is under the Hackathon feedback deadline and didn't have time to give it more polish before submitting. I'll plan to follow up through the examples/docs path once I have bandwidth to give ATOM support more testing.

One thing worth keeping on record: the bleeding-edge checkout capability (swapping in a local ATOM/aiter checkout via editable install) validated a real upstream ATOM fix before it had landed on `main`. That was genuinely useful during the investigation, even if the delivery mechanism here isn't the right long-term shape

PR #1129, opened 2026-08-08
refactor(llm): route every LLM call through an agent runtime or the single-shot entry point

xiaofei-zheng:
## Why

Hyperloom must never make bare, ad-hoc LLM API calls. Every interaction should
go through one of two sanctioned paths, so that one set of credentials can
serve every deployment shape a customer gives us:

* **Anthropic-only** (Claude / DeepSeek-style gateway),
* **OpenAI-only** (an OpenAI-protocol gateway, no Anthropic side),
* **both sides configured** (the default: critic on codex, the rest on claude).

Before this change the codebase had scattered provider access: five modules
built their own `AsyncOpenAI` client, four hand-rolled the Anthropic
`/v1/messages` protocol over `httpx` with `anthropic-version` hardcoded in each
one, and the TraceLens skill ran a 287-line hand-written tool-calling loop
instead of an agent SDK. The practical consequence was that an OpenAI-only
deployment silently fell back to an unauthenticated Claude backend, and an
Anthropic-only deployment hard-failed or silently degraded at several
single-shot call sites.

## How

Two sanctioned paths, and a guard that keeps them the only ones:

* **Agentic (multi-turn) work runs on an agent runtime.** The forge fellow now
  selects KernelForge's codex provider when the deployment is OpenAI-only, and
  the TraceLens skill runs on the Codex Agent SDK instead of a hand-written
  tool loop.
* **Single-shot inference is funnelled into `hyperloom.common.llm_config`,**
  which becomes the only module allowed to construct a provider client or speak
  a raw completion API. It resolves credentials for both protocols and, when
  only the Anthropic side is configured, derives the OpenAI-protocol endpoint
  from it so the single-shot layer works in an Anthropic-only deployment too.
* **An AST architecture guard** parses every first-party module and fails the
  build on a new client construction, bare completion call, or hand-rolled
  provider HTTP request outside those owners.

## What changed

Provider routing for the agent runtimes:

* `fix(kernel)`: run the forge fellow on codex when the deployment is OpenAI-only.
* `refactor(kernel)`: run TraceLens analysis on the Codex Agent SDK. This also
  removes a hardcoded `max_completion_tokens=2000` that was silently truncating
  6.2k-token reports mid-JSON.
* `fix(preflight)` / `fix(install)`: provision the codex runtime alongside the
  claude one, so a pod with only OpenAI credentials actually has the SDK.

The single-shot entry point:

* `refactor(llm)`: `llm_config` owns OpenAI client construction, and the
  Responses API goes through it too.
* `feat(llm)`: reach the OpenAI protocol from an Anthropic-only deployment.
* `refactor(llm)`: route the Anthropic transport through one shared contract,
  deleting `common/llm.py` and the four copies of `anthropic-version`.
* `refactor(llm)`: migrate the remaining call sites (codex and critic role
  backends, proposal scorer, framework ranker, audit, breakdown reporter, RCA
  engines) onto the contract.
* `refactor(llm)`: keep the breakdown reporter on a non-streaming request --
  routing it through `llm_config` had moved it onto the streaming helper, which
  changed the wire shape.

The guard:

* `test(llm)`: the ratchet pinned 18 pre-existing violations so the migration
  could land in stages; all of them are now migrated and the map is empty, so
  any new ad-hoc client fails the build.

## Test plan

- [x] Full suite: 12263 passed, 27 skipped. The 2 failures
      (`test_render_surfaces_source_dispatchability_and_task_groups`,
      `test_aiter_build_e2e_real_rocm`) reproduce on a pristine `main`
      worktree and are unrelated to this change.
- [x] Architecture guard passes with an empty violation map.
- [x] Live forge + codex run on MI355X in an OpenAI-only deployment: 1.0962x
      speedup over 4 iterations, validation SNR 103.46 dB, all checks passed.
- [ ] Live smoke run of the TraceLens Codex path against a real OpenAI-only
      gateway (mock-tested so far).


xiaofei-zheng:
Coordinated dependency: [KernelForge #109](https://github.com/AMD-AGI/KernelForge/pull/109). Final OpenAI-only and Anthropic-only e2e validation is in progress. Do not merge this PR until the dependency and both e2e runs pass.

haishuok0525 (approved):

PR #1131, opened 2026-08-09
Give the Codex Coordinator the transport and vocabulary its role already had

xiaofei-zheng:
## What this is for

An 8-hour OpenAI-only run finished green — `stop_reason=global_converged`,
`crash_count=0` — and delivered **0.00%**. The Anthropic-only run of the same
commit delivered **+4.50%**. Nothing in the logs announced a failure, because
each defect degraded silently. This branch is the Hyperloom half of closing that
gap; the KernelForge half is AMD-AGI/KernelForge#112.

## What was actually wrong

Every defect had the same shape: **where the Claude path had a mechanism, the
Codex path got a hand-written subset that nobody validated, and the prompts kept
describing the Claude mechanism.**

* **Intent vocabulary.** `roles/codex.py` hardcoded 5 intent types in prose while
  the orchestration role allows 10. `request` — the only legal route to the
  kernel agent — was missing, and `review_verdict`, which the role may not emit,
  was present. 256 turns produced 0 `request` intents and 44 denied
  `propose_action`. The Coordinator was also handed the *Critic's* output
  instructions verbatim, because `codex.py` was written when Codex served only
  the Critic.
* **Conversation continuity.** `CodexBackend` had no `conversational` attribute,
  so `push_full` was permanently true: 256 full pushes, 5,053,654 prompt chars,
  0 DELTA. The log line that would have revealed it is itself gated on
  `conversational`, so the degradation left no trace.
* **Seed gate.** A phase re-scope replaces the thread inside `run()`, after the
  Coordinator has already chosen DELTA — so a thin delta landed in a thread
  holding none of the state it omitted, and `run()` marked it seeded again,
  making every later tick of the phase a delta too.
* **Empty envelope.** An empty `intents` list satisfied every check, so a turn
  that decided nothing counted as a success and reset the backend's error streak.
* **Checkpoint turn.** Dropping the schema does not reach the `OUTPUT FORMAT`
  block, which is fixed in the thread's developer instructions for the thread's
  life, so a compaction was answered with an intent envelope, read as
  degenerate, and skipped — leaving the conversation growing.

## How they were fixed

| Area | Change |
|---|---|
| Vocabulary | The Coordinator's intent enum is derived from its role programmatically, the way the Claude side already does it, and the role gets its own output instructions |
| Transport | `CodexBackend` declares `conversational`, so SEED/DELTA and checkpoint compaction both work; the prompt describes the transport the Coordinator actually has |
| Seed gate | `needs_seed_for(system_prompt)` answers for the turn about to run, and the Coordinator loads the system prompt before composing |
| Envelope | An empty `intents` list is rejected rather than counted as a decided turn |
| Checkpoint | The schema-less turn carries an explicit per-turn override of the standing output format |
| Kernel gate | The summary names the gate that actually skipped an unattempted kernel |
| lm-eval probe | The probe anchors on the `export` itself, not on a right-hand side upstream rewrote — a whole-line literal had silently stopped matching in every InferenceX root |

## Verification

Measured across full OpenAI-only runs of this branch:

| | before | after |
|---|---|---|
| Coordinator reaching the kernel agent | 0 `request` intents in 256 turns, 44 denied `propose_action` | `integrate_handler` called, 0 denials of that class |
| Conversation | 256 full pushes, 5,053,654 chars, 0 DELTA | 7 SEED / 9 DELTA, re-SEED at all 5 phase seams |
| End-to-end gain | 0.00% | **+8.1%**, and **+5.52%** in the later kernel-focused run that also exercised the full kernel-rewrite chain |

Suite: **12815 passed**. Two failures are environmental and on paths this branch
does not touch: the documented `/tmp/pytest-of-root` flake in
`test_bench_patch_no_accuracy`, which passes in isolation, and
`test_aiter_build_e2e_real_rocm`, whose freshly created venv has no ROCm torch.

## Notes for review

* `7ada073ce` shipped with its tests deferred because the run it blocked was
  waiting; all three of its fixes are covered here.
* Two coverage gaps remain, both pre-existing: nothing exercises multiple
  intents in one turn, and nothing exercises backend-level usage accounting.
* Also open and not blocking: Codex has no context-pull surface, so the DELTA
  note is conditional on tools being mounted; `PolicyGate.allowed_tools_for_agent`
  still returns the Claude tool list regardless of provider; the Codex web-search
  env knobs were removed rather than remapped, because the SDK's
  `tools.web_search` takes an object whose shape could not be verified without a
  live call.

## A nondeterministic assertion this branch had introduced (71f6378)

The first CI run on this PR failed `test (py3.11 shard 1/4)`, and
`coverage (Python 3.11)` failed only as a downstream gate on that shard —
coverage itself was 90.08% and the `fail_under` step never ran.

```
FAILED test_codex_session.py::test_codex_session_starts_a_new_thread_after_a_reset
assert 1 == 2
 + where 1 = len({140546794365392})
 + where {140546794365392} = set([140546794365392, 140546794365392])
```

The assertion above it, `thread_start_calls == 2`, passed: the session did open
a second thread. Only the test's way of telling threads apart was unsound. It
recorded `id(self)`, and a reset frees the first thread before its replacement
is allocated, so CPython hands out the same heap address and the two collapse
into one set element. The fake now carries a serial issued from the shared
record, which counts what the assertions mean to count.

Introduced by `7ec5e4977` on this branch, latent because it passes whenever the
allocator does not reuse the address. Run 20 times consecutively after the fix,
passing every time; the CI shard and both coverage jobs are green.


ZhengGong-amd (approved):
Approved

PR #1132, opened 2026-08-10
feat(kernel): add collective forge optimization lane

zhanglei-amd:
## Problem

Multi-GPU collectives are visible in every trace and reachable by no backend.

Two independent gaps close the lane before it opens:

1. TraceLens attributes collectives as `nccl_summary` rows, which are a separate output from the per-category operations that feed the candidate pool. When every per-category operation is unresolvable, the pool finalizes empty and the session has nothing to optimize — even though the trace named a collective holding 27.8% of GPU time.
2. `harness_generator` emits single-device rigs (`device="cuda"`, no process group). A collective measured on one rank produces no inter-GPU traffic at all, so any "speedup" it reports is meaningless.

## Approach

A dedicated lane, from candidate discovery through E2E integration.

| Concern | Where |
|---|---|
| Get collectives into the candidate pool | `_nccl_summary_candidates.py`, wired into `tracelens_analysis.py` |
| Measure them honestly | `collective_driver_generator.py` |
| Run a campaign without corrupting the repo | `forge_collective.py` |
| Budget, gate and integrate | `request_handlers.py`, `phases/kernel.py`, `loop/writeback.py` |

### Candidate discovery

`nccl_summary` rows are injected into the enriched candidate pool with their device symbol demangled and matched back to a source file, so a collective becomes a first-class candidate rather than a footnote in the analysis.

### Measurement

Each candidate gets a generated `driver.py` and `program.md`. The rig self-launches under torchrun, seeds rank-distinct inputs (a collective that silently drops ranks fails parity instead of passing trivially), gates on SNR >= 30 dB against `torch.distributed`, reduces latency with MAX across ranks (the slowest rank bounds a collective), and times a replayed CUDA graph rather than Python dispatch.

`run_candidate` is deliberately left unimplemented. The trace gives the launch site and the shapes, but not the callee's parameter order; guessing produces a driver that fails deep inside the loop where it is expensive to diagnose.

### Workspace safety

An editable aiter install ignores worktree copies, so the campaign has to edit in place. That is made recoverable rather than avoided:

- a durable restore journal records the clean baseline before the first edit, so a hard crash leaves the repository restorable on the next run
- the validated published best is salvaged from the `forge_experiments` sidecar when the wrapper times out before forge-loop writes its result
- export happens before restore, so a timeout does not discard a validated KEEP

### Budget

Campaign and wall-clock budgets derive from the session's remaining minutes, reserving 45 minutes for the phases that follow and holding back a finalize window so export and restore cannot be killed mid-flight. A session that cannot fit preparation plus KernelForge's one-hour minimum campaign skips the lane instead of starting one it cannot finish.

## Validation

Ran end-to-end on 8x MI355X against MiniMax-M3-MXFP8 (session `20260809T094633Z`, 12h budget):

| Stage | Result |
|---|---|
| Candidate | `cross_device_reduce_2stage<bf16,8>` resolved to `custom_all_reduce.cuh:483`, 27.8% GPU time |
| Iteration 1 | KEEP, mean case speedup 2.64%, SNR 47.69 dB |
| Iteration 2 | REVERT, 0.96% — below incumbent, correctly rejected |
| Final micro | 10.83% mean case speedup |
| E2E | 2333.78 -> 2342.87 tok/s/GPU (+0.39%), below the KEEP threshold -> NEEDS_REVIEW, reverted |

The E2E verdict is the point: a 10.83% microbenchmark win did not survive the integrate gate, the patch was reverted, and the aiter checkout came back to its baseline commit. The lane reaches a verdict and cleans up after itself.

## How the lane is reached

The lane ships opt-in. `_on_enter_kernel` returns after `_run_geak_kernel_phase()`
whenever GEAK owns the phase, and the default kernel backend order is `("geak",)`,
so a default session never reaches the collective entry. It runs when the order
excludes GEAK (`KERNEL_OPT_BACKEND_ORDER=forge`) or when
`HYPERLOOM_COLLECTIVE_ONLY=1` forces the GEAK branch off. Its own gate then keys
on `TP > 1` and the exposed-communication share, not on the backend order.

## Not addressed

E2E throughput. This makes collectives optimizable and honestly measured; whether a given candidate converts to E2E gain is a separate question, and the first real run says this one did not.

Two pre-existing issues surfaced during that run. Attempt accounting is now fixed here: collective campaigns are persisted as capped history on `SharedState` and rendered into the kernel-optimization summary, so a rejected or failed campaign is counted instead of vanishing. The JIT build-cache restore refusal is fixed separately in #1135, since it is a revert-path bug in `apply_kernel_patch` rather than a collective concern.

## Changes outside the collective lane

`_server_patcher.py` updates the TraceLens sglang patch-detection sentinels. The
v1.0.1 patch set split `scheduler_profiler_mixin.py` into
`scheduler_components/profiler_manager.py` and dropped `http_server.patch`, so
the pre-v1.0.1 sentinels could never be satisfied: every apply reported success
and was then rolled back by the post-apply check. The four replacements cover the
annotation pipeline end to end (scheduler callback, profiler_manager toggle,
io_struct request fields, step-span aggregates).

`render.py` clears the CodeQL cyclic-import alert raised on this diff by taking
`_format_last_kernel_opt` from `_kernel_decisions` instead of `request_handlers`.

`coordinator_helpers.py` defers its `actions.executors` import into the one
function that uses it. `actions.executors` re-enters `coordinator_helpers`
through `session_breakdown`, so any importer reaching `coordinator_helpers`
first -- `phases.kernel`, which both new lane test modules import -- failed on a
partially initialised module. That made those two files uncollectable in
isolation, which is exactly how a CI shard runs them.

`_optimization_budget_minutes` gained a `KERNEL_OPT_BACKEND_BUDGET_MIN` override
and `_run_optimization_single` now always passes `--budget-minutes` rather than
only when the payload carried one. This changes the budget every GEAK/OOB
`kernel_opt` session receives, not just collective ones.

`integrate_handler` now reports `status="failed"` when a revert did not
complete, where it previously still reported `ok`. Any caller reading the
integrate return contract sees the stricter value.

`_maybe_revert_kernel_patch` no longer requires `apply_result["status"] == "ok"`
before attempting a revert, so a partially applied patch is now cleaned up
instead of being left in place.

## Test plan

- [x] `pytest src/hyperloom/agents/kernel/tests/` (1653 passed, 28 skipped)
- [x] `pytest src/hyperloom/inference_optimizer/tests/` (9765 passed; 2 pre-existing failures reproduce on main: `test_baremetal_aiter_install_preserves_system_triton_and_rechecks_alignment`, `test_aiter_build_e2e_real_rocm`)
- [x] Full 8x MI355X session: candidate discovery -> Forge campaign -> KEEP -> E2E integrate -> revert -> clean repository
- [x] Both new lane test modules collect and pass standalone (`test_collective_lane.py`, `test_e2e_source_and_collective.py`), the condition a CI shard reproduces
- [x] Source-resolution tests build their framework trees under `tmp_path` instead of assuming `/sgl-workspace`, so they no longer depend on the host layout


ZhengGong-amd (approved):
aPPROVED

PR #1140, opened 2026-08-10
feat(kernel): add the Wan 3D/video frame axis to the diffusion FLOP model

sushildubey171:
**Draft — not proposed for merge.** Split out of #1115 at review request, and parked here rather than dropped.

**Description: what and why**

Adds the 3D/video frame axis to `diffusion_flops.py` and routes `WanTransformer3DModel`. It is a draft because, as noted on #1115, there is currently no way to point Hyperloom at a Wan workload — no SGLang `multimodal_gen` route exists and the xDiT wrapper is image-shaped. Merging a code path the tree cannot exercise is worse than parking one, so this sits here until that changes (see #1114).

It is **not** a parking lot for known-broken code: the defects raised on #1115 are fixed here.

**The model.** The Wan VAE is causal, so `latent_frames = (frames - 1) // vae_temporal + 1` and the denoiser sees `latent_frames * lat_h * lat_w` tokens in one full-attention context. Attention is quadratic in that count, so frames dominate the ceiling. Wan gets its own family rather than reusing `single` because it cross-attends to text instead of concatenating it, so the O(n²) term is over video tokens alone. It reuses `_cross_attention_block_flops` from #1115, so it counts all four projections from the start.

**Review fixes carried here**

- `forward_flops(g, h, w)` defaulted `frames` to a literal `1` — a public entry point silently undercounting a video denoiser by its latent-frame count. Now `frames: int | None = None` falling back to `g.default_frames`, matching `estimate_image_flops`. `test_forward_flops_frames_defaults_to_the_family_not_one` pins it.
- `--frames` never reached the report writers. `num_frames` is threaded through `analytic_ceiling` into both `diffusion_roofline.py` and `tracelens_analysis.py`, each of which gains a `--frames` flag. Without this a session could only ever use the table default, which is what produced the 6.28× you measured.

**Config-reading fixes that fall out of the Wan config**

| | before | after |
|---|---|---|
| `patch_size` `[1,2,2]` | `ps[0]` read the temporal patch as spatial (1), inflating tokens 4× | reads `(t, h, w)` on the right axes |
| `ffn_dim` | unrecognised → `4*hidden` = 20480 | 13824, its real width |
| frames | no concept | `vae_temporal` / `patch_t` / `default_frames` |

`vae_temporal == 1` marks an image model, so every existing family is untouched and byte-identical.

**Validation**

Wan2.2-T2V-A14B at 720p × 81 frames resolves to 21 latent frames and **75600 tokens** — exactly the attention operand shape in a real Wan2.2 trace, `(1, 75600, 5, 128)` under Ulysses-8. That single number pins the temporal compression, the causal off-by-one and the spatial patch simultaneously.

```
WanTransformer3DModel  wan  h=5120 L=40  f=81(21lat) tok=75600+512  steps=8 x2
```

```
pytest src/hyperloom/agents/kernel/tests   ->  1624 passed, 29 skipped
ruff check                                 ->  clean
```

**Deliberately not addressed: `num_gpus`.** You flagged that `analytic_ceiling` is single-GPU, so Ulysses-8 is off another 8×. That is real but not Wan-specific — it already affects xDiT sessions, whose `baseline_xdit.yaml` defaults `TP` to 8, so a FLUX run at TP=8 compares a whole-model FLOP count against one GPU's peak and `analytic_within_pct` inherits it. Fixing it would move numbers on a path you are actively using, and the right shape (divide the ceiling vs. report per-GPU explicitly) is your call, so I have left it alone here.

Stacked on #1115 — the FLUX.2 commit is its parent, so review that one first.

### DIFF
--- a/src/hyperloom/agents/kernel/tests/test_diffusion_flops.py
+++ b/src/hyperloom/agents/kernel/tests/test_diffusion_flops.py
@@ -47,6 +47,7 @@ def test_family_routing_for_each_class(tmp_path):
         "ZImageTransformer2DModel": "single",
         "SanaTransformer2DModel": "sana",
         "UNet2DConditionModel": "unet",
+        "Flux2Transformer2DModel": "flux",
     }
     for cls, fam in cases.items():
         d = _write_denoiser(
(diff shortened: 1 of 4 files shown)


lishuoshuo-amd:
Thanks for splitting this out and keeping it as a draft.
For now, we are going to close this PR rather than continue reviewing Wan/video-specific FLOP support directly. Hyperloom already has the `custom` framework path, and we would like to first validate the SGLang diffusion / Wan workload through `custom` before adding a dedicated framework or model-specific FLOP path.
Could you please try your real SGLang diffusion / Wan workload with `custom` first and share what is missing or awkward? Based on those concrete gaps, we can decide whether we should improve `custom`, add a dedicated SGLang diffusion framework, or revisit Wan-specific analytic support later.
This is not a rejection of Wan support itself. We just do not want to merge a model-specific path before we have an executable workload path in Hyperloom and clear evidence that `custom` is insufficient.

Issue #1146, opened 2026-08-11
Hyperloom exceeds time budget


marco-grond:
### Target release

v0.2

### Optimization domain

Inference

### Problem / use case

Hyperloom exceeded the specified time budget (spent 3h55 when time budget was set to 3h) on Qwen3-8B on MI300X

### Proposed solution

Enforce the specified time limit to ensure it is not exceeded

zoroyihan7:
## Root cause: the budget is cooperative, not a hard deadline

3h55 against a 3h budget is 55 minutes of overshoot, which matches the order of magnitude of one or two unbounded accuracy tails (#1144, 30 minutes each) plus the close/finalize tail. I suspect #1144 accounts for most of it. That said, the budget mechanism itself cannot currently stop an overshoot either. Five distinct gaps (line references against `main` @ `7e67c49fd`):

### Two clocks that disagree

Enforcement uses a monotonic deadline computed on entry to `Coordinator.run()`:

```
src/hyperloom/orchestrator/loop/coordinator.py:1570
deadline = time.monotonic() + effective_minutes * 60.0
```

Reporting uses persisted wall-clock `start_ts` (`state/shared_state.py:3114-3152`), and the breakdown computes `elapsed_minutes` against *the moment the breakdown is generated* rather than the session end (`inference_optimizer/breakdown/collectors/sessions.py`). A clean resume preserves the original `start_ts` while `run()` issues a fresh full deadline. The reported 3h55 most likely comes from `session_breakdown.json`, not from `final.json` — which has no start/end/elapsed fields at all.

### The deadline is only checked after a whole tick, and in-flight work is drained rather than cancelled

`coordinator.py:1686` evaluates the deadline only after the full tick body completes. There is no `asyncio.timeout` around the tick and no global `wait_for` around `Coordinator.run()`. SIGINT/SIGTERM only set `_stop` (`coordinator.py:1589`).

The dispatcher is explicit about draining: `loop/dispatcher.py:122-135` documents "stop spawning NEW ... variants — drain in-flight, then return", and the pump loop at `:181-212` waits on `FIRST_COMPLETED` until `inflight` is empty without ever cancelling. Entering closing cancels only *queued* tasks.

### The session deadline never reaches most executors

Grepping `session_deadline_sec` shows only four consumers: `state/shared_state.py`, `actions/executors/_grid_runner.py`, `explore.py`, `sweep.py`. Notably absent:

[...]

PR #1155, opened 2026-08-11
fix(provenance): record host platform state and resolve gfx arch correctly

rpoornac:
Supersedes #1097, which this splits in two. The Redfish/BMC half is #1156 and carries a different risk class; nothing in this PR touches credentials.

## What this fixes

**A wrong GPU architecture in every provenance record.** `detect_gfx_arch()`
read `PYTORCH_ROCM_ARCH`, which names the archs a wheel is *compiled* for, not
the installed device. On MI355X nodes running the standard ROCm image it holds a
multi-arch list, so detection returned its first entry — `gfx90a`, an MI200 part
two generations off — and, because an env hit short-circuits the probe,
suppressed the `rocminfo` call that would have answered correctly.

Screening by value shape is not enough: a single-valued
`PYTORCH_ROCM_ARCH=gfx942`, which vendor images do set, is wrong the same way
while looking plausible. It is excluded outright. Resolution is now
`HYPERLOOM_GFX_ARCH` → `--gpu-type` via the board table → `rocminfo`, because
excluding the variable alone would have left detection resting on a binary that
neither install script puts on `PATH`, turning a wrong value into no value on
exactly the bare-metal nodes that set it.

**No record of how the host was tuned.** Nothing captured SMT, NUMA layout,
governor or boost state, and nothing warned when they were wrong.

Within one session this does not corrupt an A/B: baseline and trials run on the
same node, so host tuning applies to both sides and cancels out of the delta.
What it does not cancel out of is everything the session exports — the absolute
throughput, and the configuration the optimizer selects. On a de-tuned node the
search is conducted around a CPU-side bottleneck that does not exist on a correct
one, and the winning config is then filed in the recipe KB for other people. It
also makes the two genuinely cross-machine comparisons uninterpretable: an
InferenceX reference pulled in by `--compare-against-gpu`, and an archived
`final.json` re-read months or a reimage later.

## What it adds

- `hyperloom.common.platform_probe` — one shared probe replacing three drifted
  copies of the same sysfs reads. Every field degrades independently, and
  "no host CPU sysfs" is distinguishable from "the probe broke".
- A preflight warning when the governor or boost will skew the run.
- A `platform` block in `final.json`/`final.md`, including the crash-safe path.
- `scripts/platform_audit.py` — a host auditor that judges Core Performance
  Boost, the cpufreq governor and determinism control, and records SMT and NPS
  without a verdict. Targets cite AMD's BIOS & Workload Tuning Guide for EPYC
  9004 (pub. 58011). Reads only `/sys`, `/proc` and — as root — the HWCR MSR.

## Review notes

Every finding from @xiaofei-zheng and @ZhengGong-amd on #1097 was reproduced
before being fixed; the point-by-point response is on that thread. Highlights:

- The determinism verdict returned PASS for the OS-inferred `Performance`
  string, because substring matching found "power" inside the explanatory prose
  — the check approved the one configuration it existed to catch. Matching is now
  exact, the value is normalized, and there is a regression test.
- SMT and NPS dropped to record-only: chapter 5 of 58011 varies both by workload,
  so there was no single answer to assert. Determinism keeps its Power target
  (§4.2.2, maximum performance of an individual system); the cost — that
  platforms then differ — is paid by recording the setting, which is what keeps a
  system-to-system delta explicable.
- Report generation no longer spawns `rocminfo`; 141 report tests run in 0.9s.
- Two commit messages on #1097 argued something untrue (that a same-machine A/B
  could misattribute a delta, and that the gfx value reached the recipe KB). Both
  are corrected in the rebuilt history.

## Testing

- Clean worktree: 11 failed, 11194 passed — the same 11 pre-existing
  environment-dependent failures as the merge base, unchanged.
- 32 new tests: the shared probe (fake `/sys` tree, so hardware-independent),
  gfx resolution order, and the auditor's pure functions.
- `ruff` clean. Auditor verified on an MI355X node: `--quick` and the full run
  both exit 0.


zoroyihan7:
Independent confirmation of the `detect_gfx_arch()` half, on a different part from the one in the description.

A 3-hour session whose artifacts came in on #1146 exported this `manifest.json`:

```json
{
  "gpu_type": "mi300x",
  "gfx_arch": "gfx90a"
}
```

MI300X is `gfx942`. `gfx90a` is MI200/MI250 — the same two-generation miss the PR describes on MI355X, so this is not specific to that board. And the record already contains the correct answer one line above: `gpu_type` was resolved to `mi300x` while `gfx_arch` was reading `PYTORCH_ROCM_ARCH` and taking its first entry, which makes the `--gpu-type` → board-table step in the new resolution order (`HYPERLOOM_GFX_ARCH` → `--gpu-type` → `rocminfo`) sufficient on its own for this node, with no `rocminfo` on `PATH` required.

Two things worth noting for reviewers weighing the risk:

- **It reaches the exported artifact.** This is the session manifest — the record that leaves the machine and gets compared against other sessions. Anything keyed on `gfx_arch` for KB lookup or cross-run comparison has been matching MI300X runs against MI200 entries.
- **It is silent.** The whole session log contains no mention of `gfx`, `PYTORCH_ROCM_ARCH` or `rocminfo` — no warning, no probe, nothing to notice. A wrong value that never announces itself is the case the "screening by value shape is not enough" argument is really about, and this run is a concrete instance of it.

No objection from me on the resolution order; the internal consistency of this manifest is a good argument for putting `--gpu-type` ahead of `rocminfo` rather than after it.

ZhengGong-amd:
## Should fix before merge

### 1. `scripts/tests/` is not collected in CI

`testpaths = ["src/hyperloom/**/tests"]` and `pytest_ci_args` passes no path. Verified:
default collection is 13083 tests, **0** of them from `scripts/tests`; run explicitly, all
44 pass. The 23 new auditor tests are dead. Add `scripts/tests` to `testpaths`, or move the
tested logic into the package. (Same fix is needed for #1156.)

### 2. `_check_gfx_arch_resolvable` runs before `--gpu-type` is resolved, so it cries wolf

`_preflight(args)` is called at `cli/__init__.py:1579`; the probe-vs-hint resolution that
sets `args.gpu_type` and `os.environ["GPU_TYPE"]` runs at `cli/__init__.py:1903-1943`,
after it. So the check sees the raw hint and an unset `GPU_TYPE`, leaving only `rocminfo`.
But `_autodetect_gpu_type()` falls back to `torch.cuda...gcnArchName`, which needs no
`PATH` — so on a bare-metal node without `/opt/rocm/bin` on `PATH` the warning fires while
provenance will in fact resolve `gfx_arch` seconds later. That is the exact host class the
check targets, and it is the failure mode `_check_platform_tuning`'s own comment warns
about ("train readers to ignore the check entirely"). Move the call after GPU resolution.

### 3. `--help` epilog contradicts the code

> Recorded-only knobs (determinism, SMT, NPS) never affect the exit code.

`determinism` is in `CHECKED` and does affect the exit code — a real run exited 2 because
of it. Leftover from when determinism was RECORD-only.

### 4. The `gpu.count` comment says the opposite of what the code does

> PCI devices bound to amdgpu — the count the run actually had, independent of any
> `*_VISIBLE_DEVICES` masking.

`/sys/bus/pci/drivers/amdgpu` gives the **host** device count, not the run's. Verified
`"count": 8` on a node regardless of masking. The field feeds cross-machine interpretation,
so the comment matters: "host device count, before any `*_VISIBLE_DEVICES` masking".

### 5. `os_layer()`'s "Never raises" is not true

`cpu_identity()` guards the NUMA `os.listdir` but not the socket one. On a host without
`/sys/devices/system/cpu` it raises `FileNotFoundError` (reproduced). Note this is the same
"one copy missed a guard" drift that `platform_probe.py`'s own docstring complains about.

### 6. `platform_probe.py`'s module docstring describes history that is not in `main`

> Three call sites grew their own copy of this logic (CLI preflight, the run report, and the
> standalone auditor) and they drifted -- one dropped the empty-string filter, one omitted
> the zero guard

Grepping `main` for `smt/active|physical_package_id|scaling_governor|system/node` returns
**zero hits**: all three call sites are new in this PR, so those drifted copies only ever
existed in #1097's iterations. Same for `nodes_per_socket`'s "NPS0 ... previously read
downstream as a real misconfiguration" and `infer_determinism`'s "a previous version
returned the prose ..." — 8 `previously` / `an earlier` / `used to` instances across the
two PRs. Keep the constraint, drop the account of an unmerged version (commit messages are
the right home). The PR body's "one shared probe replacing three drifted copies" needs the
same correction, since `platform_audit.py` deliberately adds a fourth copy.

Not a length complaint: prose ratio is 28–34% in the new files versus 37% in
`provenance.py` and 32% in `report.py`, i.e. in line with the house style. The issue is
only that the prose points at a history a reader cannot verify.

## Minor

### 7. `gpu_types.py` deduplication is half-done

`_AMD_GPU_DISPATCH_IDENTITIES` is now a re-export, but `_AMD_GPU_TYPES` still hardcodes the
same four boards, `_GFX_TO_RUNNER` is a partial inverse, and `parser.py`'s `choices` is a
fourth copy. `_AMD_GPU_TYPES = frozenset(AMD_GPU_DISPATCH_IDENTITIES)` closes the loop the
new docstring promises.

### 8. Precedence wording conflicts with the existing convention

`_resolve_gpu_type` says the probe wins "to catch wrong-host typos that corrupt KB";
`detect_gfx_arch`'s new docstring says `gpu_type` "is a stronger source than a probe". No
real conflict — `args.gpu_type` is already probe-corrected before provenance is built — but
say so in the docstring. Also add the `GPU_TYPE` env fallback to the documented resolution
order; the code has `gpu_type or _env_first(env, "GPU_TYPE")` but the numbered list does
not mention it.

### 9. `_crash_safe_platform`: double fallback plus a private cross-layer import

It lazily imports the orchestrator's private `_platform_fingerprint`, whose own docstring
promises it always returns a dict and which is already catch-all internally. Worse,
`report.py` imports `MessageBus`/`SqliteConnection` at module level, so the crash-safe path
pulls a heavy import chain in after a crash. Moving `_platform_fingerprint` into
`hyperloom.common.platform_probe` (where the shared probe already lives) and importing it
publicly from both callers removes the private dependency and one `try/except`.

### 10. A measurement no verdict uses

`measure_peak_mhz()` + `per_core_spread()` generate ~10s of load, but `peak_mhz` and
`boost_ceiling_mhz` are only printed. The CPB `basis` text still says "the effect is
directly measurable here as achieved MHz", implying a peak-vs-ceiling comparison that
`infer_determinism`'s docstring says was removed. Either restore the check or reword to
"recorded only". Related: `_sleepy` exists only to drive a sleep loop, and
`[f for f in (core_freq_mhz(target) for _ in _sleepy(4, 0.4)) if f]` is harder to read than
the direct form.

### 11. The determinism heuristic is not stable enough to gate on

Same EPYC 9575F, five consecutive runs: spread `7.9` MHz (UNKNOWN, exit **2**), then
`21.0`, `18.1`, `20.9`, `15.1` (power, exit **0**). `DETERMINISM_SPREAD_MHZ = 8.0` sits
inside the observed jitter, so the exit code flips on an unchanged host. Take a median over
several rounds, or make determinism RECORD-only — it is inferred anyway, and a FAIL already
tells the operator to confirm in BIOS.

### 12. `_format_md` comment vs. code

> Say why it is missing rather than omitting the line

The code does omit the line when `status != "ok"` and there is no `reason` (for summaries
predating the `platform` field). Say that, or emit a "not recorded" line.

### 13. README entry

The `[58011]:` link definition sits mid-list and splits the "Key entry points" list in two
(`- Documentation source: docs/` becomes a new list). Move it to the end of the section.
The bullet itself is 18 lines against one-line neighbours — keep ~2 lines plus a link and
move the reasoning to `docs/`.

---

Items 1–6 are what I'd want before merge; items 7–13 are fine as follow-ups, though items 9
and 11 are worth an issue rather than being dropped.

Parts I'd keep as they are: per-field degradation in `platform_probe`;
`probe_cpu_platform()` returning `None` only for "no host CPU sysfs"; `_platform_fingerprint`
always carrying `status` and logging at `warning` rather than `debug`. The premise also
checks out on a real node — `PYTORCH_ROCM_ARCH` there is `gfx90a;gfx942;gfx950;gfx1100;...`
while `rocminfo` reports `gfx942`, so the current `detect_gfx_arch` does return `gfx90a`
and does suppress the probe.

rpoornac:
All thirteen are fixed and pushed, and the description above is updated to match. Every
one reproduced as you described.

A few worth calling out rather than leaving to the diff:

- **1** was larger than it looked — it was this PR's entire test coverage evaporating in
  silence. `scripts/tests` is now in `testpaths`; those 27 tests had been collected 0
  times while reporting green. Collecting them made `platform_audit.py` a measured file
  for the first time and put the coverage gate under 90%, so `scripts/*` is now omitted
  alongside the CLI drivers and operator tools already there. #1156 inherits both.
- **11**: I took the RECORD-only option rather than the median. Your five runs settle it —
  an 8.0 MHz threshold sitting inside 7.9–21.0 MHz of jitter measures the noise, not the
  machine. The Power target stays documented and the value stays recorded, which is what
  keeps a system-to-system delta explicable; it just no longer moves the exit code. That
  is also what makes **3** true rather than merely consistent.
- **6**: I swept the rest of the diff rather than only the files you quoted, and found two
  more — including the NPS0 sentence, which lives in `test_platform_probe.py` rather than
  in the probe. Both now state the constraint. The body's "three drifted copies" is
  corrected too, since `platform_audit.py` deliberately keeps a standalone fourth. #1156
  had three more, fixed there.
- **13**: down to 7 lines with the link moved, but the reasoning went into the script
  rather than `docs/` — it is the kind of thing you want in front of you while reading the
  check, not on a separate page. I applied the same cut to #1156's entry.
- **9**: worth more than it looked. The crash path went from 365 modules (~95 ms) to 167
  (~26 ms), and a subprocess test now asserts the orchestrator stays out of `sys.modules`.

Full suite with CI's marker filter: **13,018 passed, 34 skipped, 12 failed** in 3m22s. All
12 failures reproduce on a clean `origin/main` worktree.

ZhengGong-amd:
Re-checked at `401bf25`. All 13 items are addressed, and several went further than asked — `platform_fingerprint` moved into `common` with a single net and `multi_node=None` on the crash path; determinism demoted to RECORD-only; the board list now derives in `parser.py` and `gpu_probe.py` too. Verified: `scripts/tests` collects (52 tests, 13143 total), `os_layer()` survives a host with no `/sys/devices/system/cpu`, three full audit runs exit 0, retrospective narrative down from 8 instances to 1 (and that one is about `main`). `ruff`: 14 findings, all pre-existing, none in new files. Test failures: 11, all in `test_preflight_auth_override.py`, identical on `main`.

Three new items:

1. **`framework-agent/scripts/env.sh` should not land.** 7 lines, generated ("Do not hand-edit"), hardcodes `export PATH="/workdir/hyperloom/venv/bin:..."`, resurrects a top-level dir that is not on `main`, and was explicitly untracked by an earlier commit (`chore(framework-agent): untrack auto-generated scripts/env.sh`). Came in with the last commit. `.gitignore` has `kernel-agent/env.sh` but no framework-agent equivalent, which is how it got back in.

2. `test_crash_safe_platform_does_not_drag_in_the_orchestrator` fails outside an editable install of the same checkout — the subprocess gets no `PYTHONPATH`, so it imports whichever `hyperloom` is installed. Its sibling 50 lines up already sets `env["PYTHONPATH"]`; adding `_REPO_ROOT / "src"` the same way fixes it (verified: 3 passed).

3. Nit: `preflight.py:1007` still spells `(mi300x/mi308x/mi325x/mi355x)` by hand — the last copy now that `choices` derives from the table.

Drop `env.sh` and this is good to merge from my side; items 2 and 3 are optional.

rpoornac:
Thanks @ZhengGong-amd!!

All three addressed in `c5ec8c72`:

1. **`env.sh`.** Dropped: `git add -A` in the item-6 commit swept up an untracked generated file. Untracked again, working copy left alone, and `framework-agent/scripts/env.sh` added to `.gitignore` beside the `kernel-agent` entry, since that missing line is the whole reason it could come back. `framework-agent/` is no longer in the diff against `main`.

2. **The subprocess probe.** Fixed as you described, with both children now sharing a `_child_env()` that puts this worktree's `src` first. I took it one step further: the probe prints where its `hyperloom` was imported from and the test asserts that path sits under this worktree. Otherwise the test passes for the wrong reason on exactly the machines where it currently passes — a green result that says nothing about the sources in the diff. Verified from a foreign cwd with a hostile `PYTHONPATH`.

3. **`preflight.py:1007`.** Derived from `AMD_GPU_DISPATCH_IDENTITIES`, with a test asserting the printed list matches. Worth doing rather than leaving as a nit: it prints only on hosts where nothing resolved the arch, so a stale list is least likely to be noticed precisely where an operator is relying on it to say what to pass. That is the fifth copy and, as far as I can find, the last.

Full suite after the change: **13,044 passed, 34 skipped, 12 failed** — byte-identical failure set to the run before it, all reproducing on `main`. Your 11 in `test_preflight_auth_override.py` don't fire in my container, so that looks environmental on one side or the other; none of them touch this diff.

ZhengGong-amd (approved):
Approved

PR #1156, opened 2026-08-11
feat(scripts): audit BIOS-only EPYC knobs over Redfish

rpoornac:
Split out of #1097 at the reviewers' request. Stacked on #1155; the only commit of its own is the BMC layer.

## Why this is a separate PR

Everything in PR A reads `/sys` and `/proc`. This does not. With no `--bmc-user`
on file it **mints a temporary ADMINISTRATOR account on the BMC** over the in-band
KCS channel, uses it for a handful of HTTPS GETs, then revokes it and verifies the
revocation.

Root on the host already holds full BMC authority through KCS, so this is not a
privilege escalation. But it is a different risk class from reading sysfs, many
sites prohibit creating service-processor accounts outright, and the failure mode
— a live admin account on the management network — is not one that should ride
along in a PR about provenance records. It wants its own review.

Pass `--bmc-user` with an existing read-only account to avoid account creation
entirely.

## What it covers

The three knobs the OS cannot see on kernels without `amd_hsmp`: the platform
High Performance profile, APBDIS, and DF C-states. Unlike the OS-layer knobs, all
three are supported by AMD 58011 — each is either the documented default or the
value chapter 5 selects for its latency-sensitive columns, and each cites §4.2.1,
§4.4.3 or §4.4.4 inline.

## Account lifecycle

Every path that can leave an account behind is prevented or reported. Three of
these were bugs @xiaofei-zheng and @ZhengGong-amd found on #1097, and two of them
reported success while leaving an account enabled:

- **The slot is claimed before any mutation.** It was previously recorded only
  after the password step, so a password failure returned with `slot=None`,
  `__exit__` returned immediately, and nothing was recorded — while the account
  name had already been written and, on the crash-recovery path, an enabled
  ADMINISTRATOR account was left live with a clean exit code.
- **`account_status` distinguishes "read it, disabled" from "could not read it".**
  Built on a failure-swallowing runner it answered "not enabled" for any BMC that
  errored or timed out — a false confirmation at the exact point the design calls
  its trust anchor.
- **Every access-granting step is checked.** A silent failure produced a 401 from
  Redfish that pointed the operator at authentication or OEM compatibility when
  the real cause was an account that was never enabled.
- **SIGTERM and SIGINT unwind the context manager.** SIGKILL and a power cut
  remain uncoverable, which is why a sentinel found already enabled is recorded as
  a credential exposure rather than merely printed — it is the fingerprint of a
  previous run that could not clean up.
- **Reachability and TLS are settled before anything is minted**, with an
  unauthenticated GET of the service root. The documented invocation previously
  created an account, opened the LAN channel, then failed certificate
  verification.

Exit `3` means an account was left enabled or could not be confirmed revoked, and
outranks every knob verdict. It should page someone rather than scroll past as a
generic failure.

## Other

TLS verification is on by default; most BMCs ship self-signed certificates, so
expect `--ca-cert` or an explicit `--insecure`. The password for `--bmc-user`
comes from `$BMC_PASSWORD` or a prompt, never the command line, where the process
table would expose it to every local user.

Bandit now scans `scripts/` — the one file in the tree that mints credentials was
the one the scanner never saw.

## Testing

17 tests drive the lifecycle against a fake `ipmitool`, including the two paths
that previously reported success while leaving an account enabled. Nothing in the
test suite touches a real BMC. Not yet exercised end-to-end against live
Supermicro/Dell hardware — that is the main thing I would want a reviewer with
BMC access to try.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -102,7 +102,11 @@ jobs:
         id: bandit_scan
         continue-on-error: true
         run: |
-          bandit -r src/hyperloom -x "*/tests/*" -q --severity-level medium
+          # scripts/ is included deliberately: platform_audit_bmc.py mints BMC
+          # credentials and runs privileged subprocesses, so it is exactly the
+          # code a security scanner should see, even though it ships outside the
+          # package.
+          bandit -r src/hyperloom scripts -x "*/tests/*" -q --severity-level medium
 
       - name: Advisory notice when Bandit failed
         if: always() && steps.bandit_scan.outcome == 'failure'
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -105,7 +105,10 @@ jobs:
           # scripts/ is included deliberately: platform_audit_bmc.py mints BMC
           # credentials and runs privileged subprocesses, so it is exactly the
           # code a security scanner should see, even though it ships outside the
-          # package.
+          # package. This widens the full-tree invocation only -- pre-commit's
+          # Bandit hook is passed explicit filenames and already covered
+          # scripts/ -- and the step stays advisory (continue-on-error), so it
+          # reports rather than gates.
           bandit -r src/hyperloom scripts -x "*/tests/*" -q --severity-level medium
 
       - name: Advisory notice when Bandit failed
--- a/README.md
+++ b/README.md
@@ -84,13 +84,14 @@ feedback on how to improve Hyperloom by completing the
   root — the HWCR MSR; no credentials, nothing written. Exit `0` on target, `1` a
   knob is wrong, `2` unresolved, which CI should treat as missing coverage rather
   than as a failure. The BIOS-only knobs are not reachable this way; see below.
-- BIOS audit over the BMC: `sudo python3 scripts/platform_audit_bmc.py` — covers the
-  three knobs the OS cannot see (High Performance profile, APBDIS, DF C-states),
-  targeted per [58011][58011] §4.2.1, §4.4.3 and §4.4.4. **Mints a temporary
-  ADMINISTRATOR account on the BMC** unless `--bmc-user` names an existing one, so
-  running it is a deliberate choice; exit `3` means such an account was left enabled
-  or could not be confirmed revoked, and should page someone. The script's docstring
-  has the account lifecycle and the rest of the exit codes.
+- BIOS audit over the BMC: `sudo python3 scripts/platform_audit_bmc.py --bmc-user <ro>`
+  — covers the three knobs the OS cannot see (High Performance profile, APBDIS, DF
+  C-states), targeted per [58011][58011] §4.2.1, §4.4.3 and §4.4.4. Without
+  `--bmc-user` it refuses to run unless `--allow-account-creation` is passed, because
+  that path **mints a temporary ADMINISTRATOR account on the BMC**; exit `3` means
+  such an account was left enabled or could not be confirmed revoked, and should page
+  someone. The script's docstring has the account lifecycle and the rest of the exit
+  codes.
 - Documentation source: `docs/`
 
 [58011]: https://docs.amd.com/v/u/en-US/58011-epyc-9004-tg-bios-and-workload
--- a/README.md
+++ b/README.md
@@ -85,18 +85,12 @@ feedback on how to improve Hyperloom by completing the
   knob is wrong, `2` unresolved, which CI should treat as missing coverage rather
   than as a failure. The BIOS-only knobs are not reachable this way; see below.
 - BIOS audit over the BMC: `sudo python3 scripts/platform_audit_bmc.py` — covers the
-  three BIOS-only knobs (High Performance profile, APBDIS, DF C-states), each
-  targeted per [58011][58011] §4.2.1, §4.4.3 and §4.4.4 respectively.
-  **This creates a temporary ADMINISTRATOR account on the BMC** unless you pass
-  `--bmc-user` with an existing account: with no credentials on file it mints a
-  sentinel account over the in-band KCS channel, uses it for a few HTTPS GETs,
-  then revokes it and verifies the revocation. Root on the host already holds
-  full BMC authority through KCS, so this is not an escalation — but many sites
-  prohibit creating service-processor accounts, so it must be a deliberate
-  choice. TLS verification is on by default; most BMCs ship a self-signed
-  certificate, so expect to pass `--ca-cert` or an explicit `--insecure`. Exit
-  `3` means an account was left enabled or its state could not be confirmed, and
-  should page someone rather than scroll past as a generic failure.
+  three knobs the OS cannot see (High Performance profile, APBDIS, DF C-states),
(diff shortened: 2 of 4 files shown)


ZhengGong-amd:
## Blocker

### 1. Exit code is computed inside the `with`, so `__exit__` failures never reach it

`main()`:

```python
with TempBmcAccount() as acct:
    if acct.slot is None or not acct.password:
        print(f"Could not obtain BMC access: {acct.note or 'no slot'}", file=sys.stderr)
        return exit_code([])        # evaluated BEFORE __exit__ fills CREDENTIAL_FAILURES
```

Reproduced:

| case | output | exit |
|---|---|---|
| no reusable BMC slot | `Could not obtain BMC access: no reusable BMC user slot` | **0** |
| password step rejected | full `*** SECURITY: could not verify revocation ... may remain enabled with ADMINISTRATOR privilege` banner | **0** |

The tool reports "every checked knob is on target" on the path most likely to leave a
privileged account behind, which defeats the documented `exit 3` contract.

Fix: store `rows`/`err`, leave the `with`, then compute and return the exit code once at
the end. Also make "could not obtain BMC access" `EXIT_UNKNOWN`, not `EXIT_OK`.

## Major

### 2. SIGTERM does not produce exit 3 either

`install_signal_handlers` raises `KeyboardInterrupt`; `__exit__` runs and records
`could not confirm final account state`, then the exception leaves `main()` — traceback,
**exit 130**. That path is the entire reason the handler exists. Catch `KeyboardInterrupt`
in `main()` and `return exit_code(...)`.

### 3. A failed `set account name` does not abort the mint

A password failure returns early; a name failure only appends to `mint_errors`. Verified
that after a failed name step the code still runs `user priv 3 4`, `user enable 3` and
`setaccess ... callin=on` — leaving an enabled ADMINISTRATOR account on a slot whose name
is not controlled, after which Redfish auth as `hlaudit` 401s. That is the same
misdirection this PR fixes for the other steps. Abort on a failed name step, like the
password step does.

### 4. Account creation has no opt-in gate

Running with no arguments mints an ADMINISTRATOR account. README says it "must be a
deliberate choice", but nothing in the code enforces that. Add
`--allow-account-creation` (or a tty confirmation) and require `--bmc-user` otherwise.

## Minor

### 5. `scripts/tests/` is not collected in CI

`testpaths = ["src/hyperloom/**/tests"]` and `pytest_ci_args` passes no path, so all 17
tests — including the credential-lifecycle regressions — never run. Add `scripts/tests`
to `testpaths`, or move the tested logic into the package.

### 6. `ipmitool_present()` is skipped when `--bmc-host` is passed

Minting still needs `ipmitool`; with `--bmc-host` and no `--bmc-user` the check is
bypassed and the run lands on the exit-0 path above. Gate the check on "an account will be
minted", not on "the BMC address must be discovered".

### 7. Single-element `steps` loop

`__enter__` loops over a one-item list (`set account name`); leftover from a multi-step
version. Inline it.

### 8. `CREDENTIAL_FAILURES` is a module-level global

Tests have to `clear()` it. Attaching failures to the `TempBmcAccount` instance and
aggregating in `main()` is cleaner, and removes the scope problem behind item 1.

## Docs / comments

### 9. `resolve()` comment cites a knob this file does not have

`# Prefer the shortest name: "SMTControl" over "SMTControlPolicyOverride"` — SMT is not in
`BIOS_KNOBS` (it is RECORD-only in the OS-layer tool). Use `ApbDis` /
`ApbDisPolicyOverride`, which is what the test actually exercises.

### 10. Soften the Bandit claim

The `lint.yml` step is `continue-on-error: true` and Bandit already exits 1 on
pre-existing `src/hyperloom` findings (6 medium + 1 high: `_vendor/kb_store_client.py`
B310, `dispatcher.py` B608, `explore.py` B324, `kernel.py` B108). The new script itself
adds **0** medium+ findings, so it introduces no noise — but the change only widens an
advisory scan, it does not gate. Also note that pre-commit's Bandit hook receives explicit
filenames and therefore already covered `scripts/`; this only changes the CI full-tree
invocation.

### 11. Trim the README entry

13 lines stacked on #1155's 18 lines blow up the one-line-per-item "Key entry points"
list, and the `[58011]` link definition sits mid-list (it splits the list in two). Keep
~2 lines plus a link, move the risk write-up to `docs/`, and move the link definition to
the end of the section.

---

The PR states it has not been exercised end-to-end against live Supermicro/Dell hardware.
Given items 1–4 all sit on the credential-lifecycle surface, a real-hardware run should
land before this goes into `main`.

Parts I'd keep as they are: `account_status` returning `None` for "unreadable" instead of
`"disabled"`; `set_bmc_password` returning no text, plus the test guarding that contract;
revoking channel access before `user disable`; the `privilege=15` → `1` fallback.

rpoornac:
All eleven are fixed and pushed, and the description above is updated to match. 

Item 1 earned the blocker label. Reproduced before touching anything, driving the pre-fix
script with a fake ipmitool whose password step fails:

*** SECURITY: could not verify revocation of BMC account 'hlaudit' in slot 3 on this node.
*** It may remain enabled with ADMINISTRATOR privilege and be reachable over the LAN channel.
***   - randomize password: ipmitool rejected the new password
OLD (pre-fix) exit: 0
NEW exit: 3

A few worth calling out rather than leaving to the diff:

1 and 8 turned out to be one item. exit_code() now takes the credential failures as
an argument, so a caller cannot reach a verdict before revocation was attempted — the
old signature is what allowed the bug to be written at all. "Could not obtain BMC access"
is EXIT_UNKNOWN, as you suggested.

4: I went with the flag rather than a tty confirmation, so an operator at a terminal
and a CI job are refused identically; a prompt would have made the non-interactive path
the permissive one. The test asserts zero ipmitool invocations on that path, since a
gate that mints first and asks later is not a gate.

10: corrected in the body and in the workflow comment. Your reading is right on all
three counts — advisory step, pre-commit already covered scripts/ via explicit
filenames, and this widens only the CI full-tree invocation.

Unfortunately, I don't have access to all OEM platforms to validate, we can try when we get systems access/ 

Tests are at 25, up from 17, and running in CI for the first time now that #1155 puts
scripts/tests in testpaths.

ZhengGong-amd:
Re-checked at `6c06435`. All 11 items addressed, no new findings. Re-ran the scenarios I had reproduced:

| scenario | before | now |
|---|---|---|
| default invocation, no flags | mints an ADMIN account | refuses, exit 2 |
| no reusable slot | 0 | 2 |
| password rejected (SECURITY banner printed) | 0 | 3 |
| revocation unconfirmable | 0 | 3 |
| SIGTERM mid-audit, revocation unconfirmable | 130 + traceback | 3 |
| SIGTERM mid-audit, revocation clean | — | 2 |
| clean run, three knobs resolved | — | 0 |
| `set account name` fails | granted priv, enabled, opened channel | aborts; revocation still runs |

`exit_code(rows, credential_failures)` is a better fix than the one I suggested: passing the failures in means a caller structurally cannot compute the code before `__exit__` has run, rather than relying on the `return` staying outside the block.

LGTM from my side, pending the real-hardware run the PR already flags as outstanding.

PR #1157, opened 2026-08-11
vLLM version specific patch application; use new profiler configs

### LINKED ISSUE #929: Support latest vLLM and SGLang
The latest official releases are vLLM v0.26.0 and SGLang v0.5.17.

TraceLens currently supports:

vLLM: v0.14.0 – v0.24.0
SGLang: 0.5.9, 0.5.11 – 0.5.16
Please update inference-analysis support to cover the latest versions (missing vLLM v0.25.0 / v0.26.0 and SGLang v0.5.17).

devalshahamd:
- **Description:**

-  Hyperloom applies TraceLens patches to vLLM at profile runtime. vLLM upstreamed those profiler config fields in **v0.26**, so patching is not required for newer versions. 
- We are adding patches for vLLM and SGLang latest versions (vLLM till v25.0, and SGLang till v0.5.17). The profiler config names have been modified.  for example, in SGLang: `roofline_annotations` → `detailed_annotations`, and the env var `SGLANG_ENABLE_CUDA_GRAPH_CAPTURE_TRACE` → `SGLANG_GRAPH_BATCH_CAPTURE`.

  **What changed.**

  - **Version gate (`_server_patcher.py`).** `ensure_vllm_patched_for_tracelens()` is split into a capability check and a patch path. For vLLM ≥ 0.26 it skips patching entirely and just verifies `capture_torch_profiler` / `detailed_trace_annotation` are present in `vllm/config/profiler.py`; below 0.26 it takes the existing patch path unchanged. An operator pin (`HYPERLOOM_VLLM_PATCH_EXACT_VERSIONS`) still forces patching on any version. If a 0.26+ build somehow lacks the fields, it fails soft — the flags are dropped rather than booting a server that rejects them.
  - **Supporting refactor.** Extracted `_discover_vllm_install()` (version + install-root probe, including the isolated-venv fallback) so the capability check and `_discover_vllm_plan()` share one probe instead of duplicating it; `_discover_vllm_plan()` takes an optional `install=` to reuse it. Extracted `_markers_present()` and reused it in `_is_patched()` — behavior identical, just no longer duplicated.
  - **vLLM flag (`_workload_envs.py`).** Emit `--profiler-config.capture_torch_profiler True` instead of a computed `capture_traces` path. vLLM now owns the sidecar location; trace discovery was already directory-name based (`capture_traces/` under the workspace or its parent), so extraction is unaffected.
  - **SGLang rename.** `detailed_annotations` replaces `roofline_annotations` at every load-bearing site — the request-body setter, the three `extra_sentinels` tuples used to detect an already-patched install, and the two multi-node marker lists in `apply_tracelens_patch_multinode.py`. These are patch-detection markers, not cosmetic: a stale name makes a fully patched install look unpatched. Also updated the SGLang env var and the now-stale log/warning strings in `profile.py` and `_inferencex_patcher.py`, plus the `profile_vllm.yaml` / `profile_sglang.yaml` headers.
  - **TraceLens pin.** Bumped in both places that carry it (`agents/kernel/scripts/install.sh`, `agents/kernel/tools/tracelens_analysis.py`) to the revision whose patches use the new names. Verified no other copy of the SHA exists in the repo.

  **Effect.** Profiling works across all current vLLM and SGLang releases, and because the config now ships upstream, **future vLLM releases are supported with no additional patch work** — no new `config_vllm_v<version>.patch` per release.

- **Linked issue(s):**  https://github.com/AMD-AGI/TraceLens/issues/929

- **Tests:**

  - **Added** `test_server_patcher_vllm_native_gate.py` (9 cases): 0.26 and a 0.27 rc/local build skip patching entirely (asserts the patch path is never entered), a 0.26 build missing the fields fails soft, pre-0.26 still patches *and* reuses the probe rather than repeating it, an operator pin forces patching on a native version, undiscoverable vLLM fails soft, plus boundary checks on `_vllm_profiler_is_native` and `_markers_present`.
  - **Updated** flag/key assertions in `test_profile_and_kernel_handlers.py`, `test_workload_envs_golden_lock.py` (golden `PROFILE_EXTRA_BODY`), `test_inferencex_patcher.py`.
  - **Repurposed** `test_materialize_profile_capture_dir_wins_over_a_stale_yaml_path` → `..._capture_flag_wins_over_a_stale_yaml_value`. The old premise (a stale capture *directory*) no longer exists, but the last-wins guarantee still matters, so it now asserts an injected `True` beats a YAML-pinned `capture_torch_profiler False`.

- **Breaking changes:** **Yes — coordinated with TraceLens; the two must land together.**

  - Requires the new TraceLens pin. An older TraceLens checkout ships patches using the old names, which this code no longer recognizes as applied.
  - Any operator override still setting `SGLANG_ENABLE_CUDA_GRAPH_CAPTURE_TRACE` is now ignored; use `SGLANG_GRAPH_BATCH_CAPTURE`.
  - Custom YAMLs or `extra_envs` passing `roofline_annotations` or `--profiler-config.capture_torch_profiler_dir` will no longer take effect (the latter is rejected by vLLM 0.26+).
  - No change for vLLM ≤ 0.25: same patch path, same behavior.


src/hyperloom/inference_optimizer/assets/configs/profile_sglang.yaml
```diff
@@ -51,7 +51,7 @@ benchmark:
     # --- Profiler: shape + callstack annotations (#126) ---
     SGLANG_PROFILE_WITH_STACK: "true"
     SGLANG_PROFILE_RECORD_SHAPE: "true"
-    SGLANG_ENABLE_CUDA_GRAPH_CAPTURE_TRACE: "1"
```
tsrikris, line 54: Just confirming this name change is intended!

tsrikris (commented):
LGTM!

src/hyperloom/inference_optimizer/assets/configs/profile_sglang.yaml
devalshahamd: Yes this is intended. We were using SGLANG_ENABLE_CUDA_GRAPH_CAPTURE_TRACE earlier, but then upstream added a bit different functionality under SGLANG_ENABLE_CUDA_GRAPH_CAPTURE_TRACE, so we renamed it

zoroyihan7:
Field evidence for the version gate, from a 3-hour session whose artifacts came in on #1146. It ran vLLM **0.26.0** — the exact version this PR stops patching — and it cost the run 81 minutes.

The patch attempt and its consequences, straight from `optimize.log`:

```
01:28:59 _server_patcher [WARNING] `git apply --check -p1` AND fuzzy `patch -p1 --dry-run` both
         failed for vllm (version 0.26.0, patch config_vllm_v0.24.0.patch), and it is not already
         applied (reverse check failed); fail-soft skip
01:28:59 _workload_envs [WARNING] TraceLens runtime patch unavailable for framework=vllm; profile
         will omit annotation-only flags and roofline analysis may be degraded.
02:28:45 profile [WARNING] trace structure check: [1] capture_traces/ subdirectory missing —
         graph capture didn't fire. Verify EXTRA_VLLM_ARGS / EXTRA_SGLANG_ARGS include the
         TraceLens flag and the server-side patch landed.
02:28:45 profile [WARNING] trace structure check: 1 issue(s) detected — TraceLens downstream
         analysis may be degraded.
```

Three things this confirms:

1. **The 0.24 patch does not apply to 0.26 by any route.** Strict and fuzzy both fail, and the reverse check rules out "already applied" — so the fallback that is supposed to cover minor upstream drift does not stretch two minor versions. The version gate is the right shape; fuzz was never going to close this.
2. **Fail-soft is not free.** The skip is silent at the point of decision and only surfaces an hour later as a missing `capture_traces/`. The profile plus roofline arm ran 01:28:59 → 02:50:25 — 81 minutes, **45% of the session's 3-hour budget** — for a degraded analysis that the system had enough information to predict at second zero.
3. **On 0.26 the whole cost was avoidable**, which is this PR's premise: the fields are upstream, so the correct behaviour was to verify and proceed at full fidelity rather than fail-soft into a degraded one.

A suggestion that follows from (2), for this PR or a follow-up: when the capability check fails on a 0.26+ build (the `fails soft — the flags are dropped` path in the description), it would help to return that as a *structured* degradation rather than only a log line. #1146 is adding budget-driven decisions about whether an 80-minute roofline is worth starting, and "the annotations will be missing" is exactly the input that decision needs. Right now the two warnings above are the only trace, and nothing downstream reads them.

Happy to test this branch against 0.26.0 if a run slot is available — that combination is now documented to fail on `main`, so it is a clean before/after.

tsrikris (approved):

Issue #1144, opened 2026-08-11
Accuracy-gate EOS handling broken


marco-grond:
### Target release

v0.2

### Optimization domain

Inference

### Problem / use case

The accuracy-gate EOS patch is broken on the current InferenceX layout (eos_string=</s> is wrong for Qwen) → a few reasoning samples run to the full 12,288-token budget, adding a ~30-min tail per accuracy gate that eats the time budget.

### Proposed solution

Correct EOS detection/termination per model family; validate in preflight.

zoroyihan7:
## Correction: the anchor half of this is already fixed on main

My first read of this was against a branch based on the 2026-08-09 main, where the eval-probe patch anchored on an exact whole line that no longer exists in the pinned InferenceX checkout, so the patch silently no-opped and the accuracy eval ran with no early-exit protection. That was real and it did break the reported run.

It is **already fixed on main**. Commit `17ff2fa2a` (2026-08-10 17:54 +0800) replaced the line-literal with a prefix regex:

```
src/hyperloom/orchestrator/actions/executors/_inferencex_patcher.py:225-228
_EVAL_PROBE_ANCHOR_RE = re.compile(
    r'^(?:[ \t]*)export PYTHONPATH="\$\{patch_dir\}.*$',
    re.MULTILINE,
)
```

I verified this against `benchmarks/benchmark_lib.sh` at the pinned commit `a4bb43af` — it now matches `    export PYTHONPATH="${patch_dir}${PYTHONPATH:+:${PYTHONPATH}}"` correctly.

Worth flagging how this was missed: that fix shipped inside a PR titled "Give the Codex Coordinator the transport and vocabulary its role already had", with no issue link, roughly 15 hours before this issue was filed. From the issue tracker there was no way to tell it had been fixed.

## The issue is not resolved — the anchor was never the actual cause

The reported symptom is "**a few** reasoning samples run to the full 12,288-token budget". Even with the patch now applying correctly, nothing bounds that. The probe cannot help here, and forcing it to would make things worse.

[...]

PR #1162, opened 2026-08-12
fix(eval): bound the candidate arm on the same terms as the baseline, and let the fingerprint say so

zoroyihan7:
Follow-up to #1150 (issue #1144). Two gaps that #1150 left open, both about the
same thing: the accuracy gate is **differential**, so it only means something
when both arms generated under the same bounds.

## The asymmetry

`#1150` put per-request generation bounds and model-derived terminators into the
shim appended to lm-eval's `lm_eval_sitecustomize.py`, and made the **baseline**
arm verify the install and hard-fail (`eval_probe_unpatchable`) when the target
file is present but unpatchable.

The grid arm (explore / sweep / integrate_patch) did neither:

- it called `ensure_eval_probe_patched(...)` and **discarded the return value**;
- it only called it at all **when `$INFERENCEX_PATH` was set**, while the
  baseline arm falls back to env discovery (`$MAGPIE_PATH/InferenceX`).

So a session whose checkout is only reachable through `MAGPIE_PATH` benched
candidates with **no bounds** against a **bounded** baseline, and said nothing.
For a healthy model the 4096-token bound is a no-op, so this stays invisible —
right up until the pathological never-EOS case #1144 exists to catch, where the
unbounded candidate also burns its whole timeout.

The grid arm now asserts the install on the same terms as the baseline, and
preserves the baseline's distinction rather than inventing a stricter one:

| condition | verdict |
|---|---|
| target present, unpatchable, round runs eval | fail variant, `error_class=eval_probe_unpatchable` |
| target absent entirely | warn (unrecognized layout, not a broken contract) |
| round runs no eval (`RUN_EVAL` falsey) | proceed (no eval contract to keep symmetric) |

`keep_going_on_failure` is honoured: the install is `flock`-serialized, so a
transient loss can succeed on the next variant.

The failure is labelled rather than left to the generic path — nothing launches,
so that path would find no `benchmark_*` directory and file an eval-contract gap
as `no_benchmark_workspace`.

## The fingerprint

The three bounds knobs (`HYPERLOOM_EVAL_MAX_TOKENS`, `HYPERLOOM_EVAL_DERIVE_STOP`,
`HYPERLOOM_EVAL_STOP_STRINGS`) decide where each individual answer is truncated,
but did not participate in the eval-contract fingerprint. Changing one left the
contract looking identical to a run scored under different rules, which is
exactly what the enablement re-run path compares fingerprints to prevent. They
join the probe knobs there, for the same stated reason.

Verified they stay *out* of the digest where they should: a tunable server arg
still produces an identical fingerprint, so drift detection is not defeated by
the very thing a candidate is allowed to change.

## Drive-by

`AGENTX_PREFLIGHT_RETURNCODE` was never imported into `_grid_runner`, so the
AgentX preflight failure path three lines from this change raised `NameError`
and took the **whole action** down, instead of returning the structured rc its
own comment promises ("so the grid records a failed benchmark instead of
crashing"). One-line import. Caught by `ruff` F821, pre-existing on `main`.

## Also moved

`_materialized_run_eval_disabled` → `_workload_envs.materialized_run_eval_disabled`,
because both arms now need it and neither can host it for the other
(`_workload_envs` imports `_grid_runner` at module level; `baseline` imports
both). Now fails closed on an unreadable config: reading "eval disabled" from a
config that would not parse would switch the bounds guard off precisely when
least is known about the run.

## Test plan

- [x] New `test_eval_bounds_arm_symmetry.py`, 17 tests.
- [x] Verified the tests are not vacuous: reverting only `_accuracy_gate.py`
      makes all three fingerprint cases fail with **identical digests on both
      sides** (`5cf9d42ad05693df`); reverting only `_grid_runner.py` fails all
      five grid cases, including the behavioural one (installer never called
      when `$INFERENCEX_PATH` is unset).
- [x] 3761 passed across all 104 test files that import any touched module.
- [x] `ruff check` clean on every touched file (it also had the two pre-existing
      errors before this branch; one of them is the `NameError` fixed above).

### Pre-existing failure, not from this branch

`test_custom_framework.py::test_an_unparseable_pin_does_not_take_the_run_down`
fails when it runs after `test_framework_rewrite_specialist.py`. Reproduces
identically on unmodified source with just that file pair, and is normally
hidden because `pytest-randomly` shuffles the order — I only saw it by passing
`-p no:randomly`. Cross-test pollution, unrelated to this change; flagging so it
is not read as a regression here.

### DIFF
--- a/src/hyperloom/inference_optimizer/tests/test_eval_bounds_arm_symmetry.py
+++ b/src/hyperloom/inference_optimizer/tests/test_eval_bounds_arm_symmetry.py
@@ -26,8 +26,10 @@ from unittest.mock import patch
 import pytest
 import yaml
 
+from hyperloom.orchestrator.actions.executors import _accuracy_gate
 from hyperloom.orchestrator.actions.executors._accuracy_gate import (
     eval_contract_fingerprint,
+    materialized_run_eval_disabled,
 )
 from hyperloom.orchestrator.actions.executors._grid_runner import (
     GridVariant,
@@ -36,9 +38,6 @@ from hyperloom.orchestrator.actions.executors._grid_runner import (
 )
 from hyperloom.orchestrator.actions.executors._subprocess_kill import (
     EVAL_PROBE_UNPATCHABLE_RETURNCODE,
-)
-from hyperloom.orchestrator.actions.executors._workload_envs import (
-    materialized_run_eval_disabled,
 )
 
 _PATCHER = "hyperloom.orchestrator.actions.executors._grid_runner"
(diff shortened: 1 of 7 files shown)


chaojhou (approved):
Approving. Verified the load-bearing claims against `main` rather than taking them on trust:

- **Fingerprint change is compatibility-safe.** `_extract_eval_contract_fields` skips absent keys (`if v is not None`) instead of defaulting them, so adding the three bounds knobs only changes the digest for configs that actually set them. Those knobs arrived with #1150 on Aug 12, so the window of persisted fingerprints that could shift is about a day. Drift detection is not weakened either — the PR's own check that a tunable server arg still yields an identical digest holds.
- **The `AGENTX_PREFLIGHT_RETURNCODE` NameError is real.** `_grid_runner.py:1006` on `main` references it while the constant lives in `_subprocess_kill.py:242` with no import, so the AgentX preflight failure path took the whole action down instead of returning the structured rc its own comment promises. Good catch on a one-line fix.
- **The arm asymmetry is the right framing.** The gate is differential, so an unbounded candidate measured against a bounded baseline is only invisible while the model terminates normally — precisely not the case #1144 exists to catch. Preserving the baseline's present-but-unpatchable vs absent distinction rather than inventing a stricter one is the correct call.
- **`materialized_run_eval_disabled` fails in the safe direction:** an unreadable config reads as eval-enabled, keeping the bounds guard on when least is known about the run.

Also worth noting for others: the mutation check (reverting only `_accuracy_gate.py` yields identical digests on both sides of the three fingerprint cases) is what makes the new suite credible rather than just green, and flagging the unrelated `test_custom_framework.py` ordering failure up front saves the next reader a false regression hunt.

PR #1163, opened 2026-08-12
[CLEAN] Retire the unreachable kernel-harness stack; fix benchmark workspace selection

ZhengGong-amd:
Three problems surfaced while tracing the kernel path:

1. A large stack of code — harness generation, Hyperloom-authored Forge drivers,
   several trace artifacts — had **no reader and no reachable caller**. Some of it
   was worse than dead: it wrote credentials and unredacted LLM transcripts to
   disk.
2. Benchmark result collection picked its workspace with
   `sorted(glob("benchmark_*"))[-1]`, so a **stale directory from a prior attempt
   could yield `status="succeeded"` for a run that crashed**.
3. Three actions — `vendor_kernel_config`, `operator_tuning`,
   `deep_kernel_analysis` — carried a full catalogue surface (action yaml, phase
   allowlist, prompt entry, verdict class, runs-workspace slot) but **no executor
   and no request handler**, so every request for them was answered with
   `unknown_kernel_kind`. They were authored in `0ab1c748` for the `kernel_agent`
   LLM role; `ba2a5be2` (#1095) retired that role without removing the surfaces
   it left behind. A 16h vLLM run on Mixtral-8x7B spent seven KERNEL_AGENT turns
   proposing them and collecting hard rejections.

Problem 3 was possible because the catalogue was split across four surfaces —
`actions/_meta/*.yaml`, the phase allowlist, the `action_surfaces` sets, and the
prompt — so a name could fall out of one and survive in the rest.

## Correctness fixes

- **Reject stale benchmark workspaces.** `snapshot_workspaces` /
  `select_run_workspace` (new, in `benchmark_result.py`) replace the three inline
  lexicographic-last globs in `baseline.py` and `_grid_runner.py`. Only workspaces
  that did not exist before the subprocess started are eligible, so a crash can no
  longer report success off a leftover directory.
- **Unique output dir per benchmark round.** `session_paths.unique_runs_dir`
  probes `-2`, `-3`, … via `mkdir(exist_ok=False)` (atomic under concurrency),
  replacing four deterministic names. `task_id` / `idempotency_key` are untouched,
  preserving task-registry dedup.
- **`parallel_e2e_runner` no longer spawns a rejected flag** — it passed the
  removed `--test-harness-path`; the value now goes through `--benchmark-file`.
- **Drop the hardcoded `/workspace/` rescue default.** Salvage now runs only when
  `$INFERENCE_OPTIMIZER_RESCUE_PATHS` is set explicitly (RULE F2).

## Removals — kernel harness

| Removed | Why it was unreachable |
|---|---|
| `--test-command` / `--test-harness-path` contract (end to end) | only reachable under `KERNEL_OPT_BACKEND_ORDER=forge`; the whole-pipeline GEAK phase never calls `kernel_optimization.py` |
| `harness_generator.py`, `skills/unittest/` | sole consumer of the above |
| Forge autogen drivers (`_AUTOGEN_*`, `_build_driver_adapter`, …) | KernelForge's `preflight_task` rewrote every Hyperloom-authored driver anyway; now always delegates to the task-preparer |
| `agent_transcript.jsonl` + the Codex item-normalization chain | no reader; both writers persisted raw tool input/results with **no redactor** |
| `mn_input_params_*.json` | wrote `--extra-env` credentials **in cleartext**; its redaction matched top-level keys only, and `INFERENCE_OPTIMIZER_EXTRA_ENV` is a container key |
| `orchestration_turns.jsonl` | full LLM messages past a redactor that inspects values, never keys |
| `semantic_audit.json` (work_dir copy), `_gc_old_profile_traces`, `apply_and_bench.py`, `experiments/` | no reader / kept alive by tests alone / zero callers |

Docs (`SKILL.md`, `references/kernel.md`, `references/benchmark.md`,
`kernel-execution-path.md`) are corrected to match — they credited
`apply_and_bench` as the IR-4 enforcer via two functions that do not exist.

## Removals — action surface

| Removed | Why it was unreachable |
|---|---|
| `vendor_kernel_config`, `operator_tuning`, `deep_kernel_analysis` and every reference | no executor, no `KERNEL_REQUEST_HANDLERS` kind; only reachable failure was `unknown_kernel_kind` |
| `actions/registry.py` + the 20 `_meta/*.yaml` (1337 lines) | the catalogue has no external supplier, so the yaml layer bought a schema validator, a file scanner and package-data entries without buying configurability |
| `preferred_backend`, `preferred_model`, `max_turns` | documented as operational, read by nothing outside the registry — `d6d137eb` refreshed `preferred_model` across 20 files on the assumption it took effect |
| `params_schema` blocks in 10 yamls | `ActionMetadata` has no such field; never parsed |
| `_noop_prep` and the kernel-owned no-op executors | PolicyGate already denies a delegate for those actions; the stub only stood ready to report an unexecuted action as `succeeded` |
| `run_fusion` from the dispatch table | `KernelPhase` awaits `run_fusion_handler` directly, so no request ever carried that kind |
| `KERNEL_OPT_BACKENDS` env | no production reader; two test files still set it and asserted an outcome that no longer depended on it |
| `GRID_INJECTABLE_ACTIONS`, `_EVIDENCE_PRODUCER_ACTIONS`, `VALID_FAMILIES`, `VALID_PIPELINE_PHASES`, `VALID_VERDICT_CLASSES` | defined, imported, re-exported — never read |
| 16 of 18 `request_handlers` re-exports, 3 `SharedState` forwarding shims | no caller through the module namespace and none in-file |
| `test_action_registry.py`, `test_kernel_request_handlers_helpers_coverage_unit.py`, `test_stall_branches_unit.py` | covered the deleted loader, or held byte-identical copies of a neighbour's tests |

Action metadata now lives in `ACTION_CATALOGUE`
(`protocol/action_surfaces.py`) as frozen records, modelling only the fields
production code reads. `KERNEL_ACTION_REQUEST_KINDS` replaces the if/elif chain
in `_format_emit_hint` whose `else` branch was what advertised the dead actions'
names as request kinds.

## Breaking changes

- Sessions recorded under the old build may name the three removed actions in
  `state.json` / `coordinator.db`. They are **not resumable** and no migration is
  provided.
- Editing an action yaml no longer changes anything, because there is no yaml.
- `_geak_enabled` no longer falls back to the persisted
  `shared_state.kernel_optimizer`, so `KERNEL_OPT_BACKEND_ORDER` is the single
  source of truth for the kernel backend on a resume as well. The state field
  itself is unchanged and still feeds the session breakdown.

## Deliberately not fixed

Recorded in `gate-bypass.issue.md` with anchors, kept out of this diff so the
decision to fix them stays separate from the decision to delete dead code:

- **The KERNEL_AGENT phase gate is bypassed for every real request kind.**
  `gate.py:1150` matches a request *kind* against a set of action *names*; the
  two intersect only at `integrate`. `run_optimization`, `run_gemm_tuning` and
  `trace_analyze` never reach `_validate_phase_action`. Before this PR the gate
  fired only for the three actions that had no implementation.
- `geak_prompt_patcher.py`, `robustness_pulse.py` and
  `apply_compatibility_filter` look dead to a call graph while sitting on a
  supported path — deleting them would be a product decision.
- `parallel_e2e_runner.py` has no in-repo caller but is a standalone CLI listed
  as a subprocess driver, so grep cannot rule out external invocation.
- The remaining 16 `SharedState` shims: collapsing them was measured at 45
  production and ~205 test call sites, each of which would import the private
  `_kernel_decisions` and thread `state` by hand. That widens the import surface
  onto a private module to save 170 lines of facade.

## Accepted trade-offs

- The Codex path is now silent per item during a turn (Claude still logs assistant
  text). `analysis.md` plus the status sidecar remain the evidence of a run.
- Langfuse `session_start` still ships the container keys — left alone
  deliberately; that instance is self-hosted.
- `diffusion_flops.py` is kept. A static audit called it test-only, but
  `tracelens_analysis.py:6660` and `diffusion_roofline.py:585` reach it through a
  bare `import diffusion_flops` under the `tools/` sys.path convention, which a
  `from x import y` grep misses.

## Testing

New regression coverage for the workspace bug: crash with stale-only workspace,
silent exit with stale-only workspace, a stale name that would previously sort
last, and a successful run coexisting with a stale peer — plus unit tests for both
new helpers and one for the task-preparer placeholder path.

The action-surface half is verified by differencing failure sets rather than
counting passes: `inference_optimizer/tests` (9883 tests) was run on this branch
and on the base commit in a separate worktree, and the two sets are identical at
22 pre-existing, environment-dependent failures. That comparison is what caught
the one real regression in this PR — the inline fast-action whitelist resolved to
the empty set once the catalogue became a plain dict, because
`_inline_action_whitelist` duck-typed the registry through `reg.names()`. It is
fixed here, and `action_registry` is now a `Coordinator` class attribute so the
missing-registry branches that hid it are gone rather than re-guarded.

`pyproject.toml` coverage-omit and package-data entries updated for the deleted
modules.


xiaofei-zheng (approved):

chaojhou (commented):
Three things I would want resolved before this merges. None of them is large, but the first hides a possible silent behaviour change and the second reintroduces the pattern this PR exists to remove.

## 1. `test_harness_path` is now always `None` — dead, or an undocumented breaking change?

`_invocation_spec.py:962-963` still writes this key into the invocation spec's `tests` dict:

```python
"test_harness_path": _absolute_path(
    candidate.get("test_harness_path"),
    base_dir=repo_root,
),
```

It is the only occurrence left in the tree, and the CLI flag that populated `candidate["test_harness_path"]` was removed by this PR — so the value is now unconditionally absent.

Both readings need a different action, and right now it is ambiguous which applies:

- **KernelForge does not read `tests.test_harness_path`** → this is exactly the write-with-no-reader pattern the PR is removing everywhere else; delete it.
- **KernelForge does read it** → the field silently went from populated to always-`None`, which is a behaviour change across a repo boundary and belongs in *Breaking changes*.

I cannot settle this from inside this repo, which is why I am asking rather than asserting it is dead.

## 2. `select_run_workspace` still selects lexicographically while its docstring says "newest"

```python
fresh = [p for p in root.glob("benchmark_*") if p.is_dir() and p.resolve() not in known_before]
return max(fresh, default=None)
```

`max()` over `Path` compares by string, so this is lexicographic-last — the same ordering whose use in `sorted(glob("benchmark_*"))[-1]` is the bug this PR is fixing. The docstring promises "the newest workspace this run created".

The `known_before` filter is what actually fixes the reported defect, so this is not a live bug as long as a round produces exactly one fresh workspace. But shipping the mismatch means the next reader will believe the selection is time-ordered when it is not, and the failure mode returns the moment a round creates two.

Either make it explicit:

```python
return max(fresh, key=lambda p: p.stat().st_mtime, default=None)
```

or keep `max()` and change the docstring to say lexicographically-last, with a note that a round is expected to produce a single fresh workspace. I would prefer the former, since the whole point of the change is to stop inferring recency from a name.

## 3. `gate-bypass.issue.md` should be a tracked issue, not a file at the repo root

The position is the smaller half of this. The substance is that this PR changes the KERNEL_AGENT phase gate from *almost* dead to *entirely* dead:

`gate.py` matches a request **kind** against a set of action **names** (`gated_kind in KERNEL_AGENT_OWNED_ACTIONS`), and the two intersect only at `integrate`. Per this PR's own analysis, the gate previously fired only for the three actions that had no implementation — the three this PR removes. So after this change no reachable request kind reaches `_validate_phase_action` at all.

That is worth a tracked issue with a severity and an owner. A 6.3 KB `*.issue.md` at the repo root will rot in place, and it is also not a location the repo uses for anything else. Please file it and drop the file from the diff, referencing the issue number from the PR description instead.

---

Two notes that are not asks.

**The `None` path is handled well.** Since `select_run_workspace` can now return `None` where the old code always got *something*, I checked all four call sites for an unhandled case that could stall a run. Both main sites fall back to the parent directory for artifact harvesting and then classify immediately as a failure — `baseline.py` persists captured stderr/stdout precisely because no `server.log` exists in that case, and `_grid_runner.py` builds `no_ws_error_summary` from the server-log death excerpt or the redacted output tail. Neither waits, retries, or passes silently. That is more thorough than the fix strictly needed.

**`SKILL.md:284-285` is a false positive, in case another reviewer flags it.** It still names `vendor_kernel_config` / `operator_tuning` / `deep_kernel_analysis`, but as a removal note ("they never had an implementation"), not as an action list the model can propose from. Correct to keep.

PR #1164, opened 2026-08-12
fix(knowledge): stop handing the forge child a knowledge mode it cannot satisfy

BaoYunkai:
## Problem

`KnowledgeConfig.apply_to_child_env` passed Hyperloom's own `KNOWLEDGE_STORE_MODE=remote` down to the KernelForge child while, three lines later, deleting `GBRAIN_BASE_URL` and `GBRAIN_TOKEN` from that same child environment. GBrain is the only remote backend the child knows: no KernelForge source reads `KB_STORE_URL` or `KB_STORE_TOKEN` at all.

The child therefore failed closed on startup, every time:

```
=== forge-loop (cli) stdout ===
Error: KNOWLEDGE_STORE_MODE=remote requires GBRAIN_BASE_URL and GBRAIN_TOKEN
=== forge-loop exception ===
forge-loop exited rc=1
```

This is deterministic, not a missing-credential accident in one deployment: any run with `KNOWLEDGE_STORE_MODE=remote` loses 100% of its kernel work.

## Production evidence

Classifying every `forge_loop.log` under the shared forge workspaces:

| day | forge found a gain | startup killed by this bug | total attempts |
|---|---|---|---|
| Aug 1 | 19 | 0 | 40 |
| Aug 5 | 8 | 0 | 17 |
| Aug 9 | 20 | 0 | 54 |
| Aug 10 | 15 | 95 | 128 |
| Aug 11 | 0 | 199 | 209 |
| Aug 12 | 0 | 7 | 25 |

The first affected attempt is `20260810T122220Z`; 301 attempts died this way, and per-session `integrated` kernel counts went to 0 across 197 sessions.

## Fix

A KernelForge child now always launches against its local store, and is handed no credential it has no client for. Hyperloom keeps owning KB Store publication and still collects the child's experience through its result document (`KernelExperienceBridge.collect_result`), so no data path is lost.

## Tests

- `test_a_remote_parent_never_tells_a_child_a_mode_it_cannot_satisfy` reproduces the regression: it fails on `main` with `- local + remote`.
- `test_a_child_is_handed_no_credential_it_has_no_reader_for` pins that KB Store secrets stop at the process boundary.
- Three tests that had encoded the broken contract are updated, including `test_forge_knowledge_env.py::test_a_remote_run_still_launches_the_child_against_its_local_store`.
- `258 passed` across the KB/knowledge-plane and forge-child env suites.

## Test plan

- [x] `pytest src/hyperloom/inference_optimizer/tests/test_knowledge_plane_phase1.py src/hyperloom/agents/kernel/tests/test_forge_knowledge_env.py`
- [x] Regression sweep over `test_cli_kb_unit.py`, `test_cli_local_kb_root.py`, `test_close_phase_sequencer.py`, `test_coordinator_kb_writes.py`, `test_local_graph_store.py`, `test_local_kb_requirements.py`, `test_remote_recipe_v2.py`
- [ ] One `KNOWLEDGE_STORE_MODE=remote` session reaches a forge-loop iteration instead of rc=1

### DIFF
--- a/src/hyperloom/agents/kernel/tests/test_forge_knowledge_env.py
+++ b/src/hyperloom/agents/kernel/tests/test_forge_knowledge_env.py
@@ -90,9 +90,15 @@ def test_the_card_is_resolved_from_the_candidate_when_the_environment_is_silent(
     assert forge_submit._resolve_gpu_type({"arch": "gfx942"}) == ""
 
 
-def test_remote_child_env_forwards_kb_store_alone_when_gbrain_is_absent(
+def test_a_remote_run_still_launches_the_child_against_its_local_store(
     monkeypatch: pytest.MonkeyPatch,
 ) -> None:
+    """A ``remote`` run must not hand the child a mode it cannot satisfy.
+
+    KernelForge resolves remote knowledge through GBrain only, and this env
+    withholds GBrain, so a child told ``remote`` exits rc=1 on startup instead
+    of optimizing anything.
+    """
     _avoid_unrelated_fellow_setup(monkeypatch)
     monkeypatch.setenv("KNOWLEDGE_STORE_MODE", "remote")
     monkeypatch.setenv("KNOWLEDGE_LOCAL_ROOT", "unchanged/root")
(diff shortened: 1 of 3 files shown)


BaoYunkai:
Closing: the premise no longer holds. KernelForge grew a KB Store experience backend, so a child told `remote` with `KB_STORE_URL`/`KB_STORE_TOKEN` and no GBrain credentials now starts normally.

Upstream fix: `92c4976` "fix(kb): let KB Store-only children start" and `d42a8d1` "fix(kb): let each remote backend activate on its own credentials", merged to KernelForge main as PR #107 on 2026-08-12. Remote mode there now requires credentials for *at least one* backend rather than GBrain specifically.

Verified against that KernelForge revision, feeding it the exact child environment `apply_to_child_env` produces on `main`:

```
OK    remote + KB_STORE only        -> mode=remote
RAISE remote + no credentials       -> requires credentials for at least one backend
```

So this PR is not only unnecessary, it would be harmful: forcing the child to `local` silently disables the remote experience sharing KernelForge just built.

The production failure this PR was written from was real -- 301 forge-loop attempts exited rc=1 between Aug 10 12:22 UTC and Aug 12, with zero integrated kernels across 197 sessions -- but it was fixed from the KernelForge side, and by the better of the two available fixes. My analysis had searched a stale KernelForge checkout that predated PR #107.

What remains worth doing, and what I got wrong here, is that nothing in CI makes the two repositories agree: reading one side's source was not enough to establish the contract. A follow-up should feed the `apply_to_child_env` output to the *installed* `kernel_agents` validator so drift in either direction fails a check instead of a session. Filed separately.

PR #1165, opened 2026-08-12
fix(kernel): let a rejected advisory forge-loop option cost the option, not the attempt

BaoYunkai:
## Problem

`_run_loop_via_cli` builds one argv and launches `python -m kernel_agents.cli forge-loop`. click rejects an unknown option *before* running anything, so an installed KernelForge that predates a single option this launcher passes turns the entire attempt into a loss:

```
Usage: python -m kernel_agents.cli forge-loop [OPTIONS]
Error: No such option '--shapes-json'. (Did you mean one of: '--agent-options-json', '--result-json'?)
forge-loop exited rc=2
```

`--shapes-json` is the option that hit production. KernelForge itself prints `--shapes-json is deprecated and ignored by evaluation; it is retained only for schema-v1 checkpoint compatibility` — so the attempt was lost over an option that contributes nothing to the search. Hyperloom and KernelForge are separate repos wired together through `$FORGE_PATH`, so this skew is a normal operating condition, not an anomaly.

## Production evidence

Classifying every `forge_loop.log` under the shared forge workspaces:

| day | attempts killed by a rejected option | total attempts |
|---|---|---|
| Aug 11 | 4 | 209 |
| Aug 12 | 18 | 25 |

On Aug 12 this was the dominant failure mode. Earlier the same class showed up as `Invalid value for '--max-hours'` (26 attempts, Jul 25-28).

## Fix

When a launch fails with an unknown-option rejection, the launcher drops that option and relaunches — at most once per droppable option, so a skew cannot loop. Only advisory options are droppable, each for a stated reason:

- `--shapes-json` — deprecated upstream and ignored by evaluation
- `--e2e-pct` — input to an end-to-end projection, not to the search
- `--framework` — forge-loop infers it from the kernel path when omitted

Load-bearing options stay loud: retrying without `--kernel` would optimize something other than the candidate. The drop is logged as a warning and written into `forge_loop.log`, so a version skew stays diagnosable instead of silently changing the run.

## Tests

- `test_a_rejected_compat_option_costs_the_option_not_the_attempt` reproduces the regression: it fails on `main` with `assert 1 == 2` (no relaunch) and pins that the dropped option's value goes with it.
- `test_a_rejected_load_bearing_option_still_fails_the_attempt` pins that `--kernel` is not silently dropped.
- `132 passed, 2 skipped` across the forge CLI, codex-provider, export/restore, knowledge-env and driver-fallback suites.

## Test plan

- [x] `pytest src/hyperloom/agents/kernel/tests/test_forge_long_horizon_cli.py`
- [x] Regression sweep over `test_forge_codex_provider.py`, `test_forge_export_restore.py`, `test_forge_knowledge_env.py`, `test_forge_submit_driver_fallback.py`
- [ ] A session against a KernelForge without `--shapes-json` reaches a forge-loop iteration

### DIFF
--- a/src/hyperloom/agents/kernel/tests/test_forge_long_horizon_cli.py
+++ b/src/hyperloom/agents/kernel/tests/test_forge_long_horizon_cli.py
@@ -3485,3 +3485,124 @@ def test_submit_falls_back_to_forge_loop_on_an_incompatible_producer(tmp_path, m
     assert verdict["eligible"] is False
     assert verdict["reason"] == "capability_artifact_schema_unsupported"
     assert result["best_ms"] == 1.0
+
+
+def _forge_loop_result_stdout() -> str:
+    payload = {
+        "baseline_ms": 2.0,
+        "best_ms": 1.0,
+        "mean_case_speedup": 2.0,
+        "search_start_mean_case_speedup": 1.0,
+        "total_improved": True,
+        "incremental_improved": True,
+    }
+    return f"__FORGE_RESULT__{json.dumps(payload)}__FORGE_RESULT__"
+
+
+def _run_loop_with_popen(tmp_path, monkeypatch, fake_popen):
+    workspace = tmp_path / "worktree"
+    workspace.mkdir()
+    kernel = workspace / "kernel.py"
+    driver = workspace / "driver.py"
+    kernel.write_text("pass\n")
+    driver.write_text("pass\n")
+    experiments = tmp_path / "attempt" / "forge_experiments"
+    experiments.mkdir(parents=True)
+
+    monkeypatch.setattr(forge_submit, "_ensure_forge_on_path", lambda: "/forge/src")
+    monkeypatch.setattr(forge_submit, "_apply_fellow_env", lambda _env: None)
+    monkeypatch.setattr(forge_submit.subprocess, "Popen", fake_popen)
+
+    return forge_submit._run_loop_via_cli(
+        worktree_kernel=str(kernel),
+        driver=str(driver),
+        workspace=str(workspace),
+        shapes={"primary": {"M": 128}},
+        snr_threshold=30.0,
+        max_iters=8,
+        max_hours=1.0,
+        branch="forge/session/kernel",
+        gpu_target="gfx950",
+        gpu_type="mi355x",
+        fellow="triton-fellow",
+        program_md_file="",
+        invocation_spec_file="",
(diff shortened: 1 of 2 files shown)


BaoYunkai:
Re-checked against the relocated KernelForge (`AMD-BRAIN-Internal/KernelForge`, main at `935826a`). This is no longer a latent risk -- it is currently fatal.

Probing all 26 options this launcher passes, one at a time, against two KernelForge trees. An unknown option makes click exit 2 during parsing, so "No such option" is a reliable existence probe, and unlike `--help` it also sees hidden options:

| option | KernelForge main | an older installed KernelForge |
|---|---|---|
| `--shapes-json` | **missing** | ok |
| `--e2e-pct` | **missing** | ok |
| `--gpu-type` | ok | **missing** |
| other 23 | ok | ok |

So on `main` today, every forge-loop launch exits rc=2 before running anything, and each attempt is lost -- the same failure that killed 22 attempts on Aug 11-12. Both missing options are already in this PR's droppable set, so the recovery covers it.

Note the skew runs both ways: the older tree is missing `--gpu-type`, which this PR deliberately does not drop. Dropping it silently would let the child file its experience under no card at all, which Hyperloom's own tests call out as accumulating nothing while looking healthy. That direction wants a CI contract check rather than a runtime guess; filed separately.

Added `test_two_rejected_options_in_a_row_are_both_dropped` for the case observed above: click names only the first unknown option, so recovery has to survive being told about them one at a time. `93 passed, 2 skipped`.

BaoYunkai:
Closing in favour of #1175 and #1176, which address the same failure without the runtime guesswork.

This PR recovered from a rejected option by dropping it and relaunching. That kept an attempt alive, but it treated a fixed, knowable fact -- KernelForge no longer has `--shapes-json` or `--e2e-pct` -- as something to discover at runtime, three subprocess launches at a time, behind a hand-written allowlist of what is safe to drop.

#1175 removes the two dead options outright, and fixes the quieter breakage that came with them: `case_coverage` was derived from `--shapes-json`, so no current forge-loop reports it, and checkpoint recovery vetoed on that silence -- discarding every salvageable best from a timed-out campaign.

#1176 replaces the runtime tolerance with a check: introspect the installed forge-loop's click command, including hidden options, and compare it against the argv the launcher actually builds. That catches drift in both directions before a session pays for it -- including `--gpu-type`, which this PR deliberately would not have dropped and therefore could not have helped with.

PR #1170, opened 2026-08-13
fix(kernel): stop sending forge-loop the options it never declared

jiaqiang-dot-liu:

  Hyperloom's forge backend passed two options to `python -m kernel_agents.cli forge-loop` that KernelForge does not declare: `--shapes-json` and `--e2e-pct`. `forge-loop` is a click command without `ignore_unknown_options`, so it rejects them during argument parsing:

  ```
  $ python -m kernel_agents.cli forge-loop --workspace /tmp --shapes-json '{}'
  Error: No such option '--shapes-json'. (Did you mean one of: '--agent-options-json', '--result-json'?)   # exit 2

  $ python -m kernel_agents.cli forge-loop --workspace /tmp --e2e-pct 1.0
  Error: No such option '--e2e-pct'.                                                                        # exit 2
  ```

  `--shapes-json` was unconditional, so **every** forge-loop attempt died at exit 2 before doing any work; Hyperloom only saw `forge-loop exited rc=2`. `--e2e-pct` was a second latent failure hidden behind the first.

  Both were removed upstream by KernelForge `6bb0fba`, which is worth naming: `--shapes-json` had been a `hidden=True`, deprecated-and-ignored option, so it was tolerated until that commit deleted it outright. The same commit also dropped `case_coverage` from the checkpoint payload — the other half of the fallout, addressed on main in the meantime (see below).

  No information is lost by removing the options. KernelForge receives the shape suite through `--invocation-spec-file`, and the spec's `tests.driver_contract.case_selectors` is built from the *same* `task_group_shape_cases()` call that produced the `--shapes-json` payload — plus per-case `arguments`/`outputs`/`raw_arg_spec`, which the option never carried. For candidates without a task group, the raw per-tensor dimensions still travel in `invocation.arguments`.

  Removed along with the options, since they existed only to feed them:
  - `_forge_e2e_pct()` and the invalid-GPU-share warning in `submit()`
  - the `shapes` and `e2e_pct` parameters on `_run_loop_via_cli()`

  **Kept:** the `shapes` dict in `submit()`. It feeds `evaluate_rewrite_route()` for the FlyDSL routing decision and `_validated_forge_checkpoint()` for checkpoint case coverage.

  `forge-rewrite-by-flydsl` is untouched: it genuinely declares `--shapes-json`, and all 21 options Hyperloom sends it are valid.

- **Review findings addressed**

  **1. The checkpoint recovery veto — mostly landed on main, one thing added here.** Review pointed out that `6bb0fba` also removed `case_coverage` from both KernelForge checkpoint write paths, so Hyperloom's gate compared against a permanently absent key and dropped every checkpoint for any candidate with a grouped case set or a derivable named dimension. Reproduced against a real git repo and a checkpoint in KernelForge `main`'s exact shape:

  ```
  KernelForge main   (no case_coverage)           -> DROPPED
  pre-6bb0fba        (case_coverage written back) -> ACCEPTED
  ```

  While this branch was in review, `97cf53219` (#1179) landed the same fix on main by the same reasoning — absence and an empty list both carry no information, so only a reported coverage that disagrees is evidence. **Its one-line condition is what this branch merged and kept**, and the duplicate test this branch had written was dropped in favour of upstream's `test_a_checkpoint_that_reports_no_coverage_is_still_recoverable`.

  What this PR adds on top is the missing log: upstream still returns `None` silently when a reported coverage genuinely disagrees. Losing a best the producer already validated and committed is too expensive an outcome to leave unattributable, so the veto now warns with both the expected and the reported coverage before discarding.

  Worth recording for whoever picks this up next: restoring a *meaningful* coverage check would need KernelForge to put `case_times` (its current vocabulary for "which CASE_IDs were measured", already used by its own preflight) into the KEEP checkpoint — today only WARM_START carries it. That is a cross-repo change and deliberately not here.

  **2. A non-zero exit reported only its return code.** `forge-loop exited rc=2` made a producer that refused its own argv indistinguishable from one that crashed while measuring; the click error went to the forge log, which nobody opens when the only thing reaching the orchestrator is a number. This is precisely why the drift stayed invisible. Both child launchers now carry the reason into the raised error via `_forge_failure_tail()`, which prefers the `Error:`/`Usage:` line the CLI prints over the tail of the transcript, and skips `__FORGE_RESULT__` lines because one of those is a whole JSON document. The rewrite launcher had the identical gap and is fixed too. The capability probe in `_flydsl_rewrite.py` already did this correctly and is the precedent the helper follows.

  **3. The driver-fallback test lost half its coverage.** The original assertion read `shapes` off the stubbed `_run_loop_via_cli`, checking both that the grouped selectors resolve correctly and that `submit()` passes them on. With `shapes` off the argv, the test now spies on the real `_validated_forge_checkpoint` — the consumer that still reads them — so both halves are covered again without stubbing its verdict.

  **4. The argv guard's rationale was about to expire.** Its comment justified itself by `forge-loop` aborting on an undeclared option. KernelForge is adding tolerance for unknown options, so that will stop being true; the comment now states the durable reason instead — a tolerating producer drops the value silently and a strict one aborts, so either way it never reaches the loop and the argv must not imply otherwise.

  Not done here, on purpose: a cross-repo argv contract test in CI. The KernelForge tolerance change covers the same failure mode more directly (unknown options are dropped, named on stderr, and reported as `ignored_cli_options`), so it is worth re-evaluating whether a second line of defence is needed once that lands rather than building two overlapping ones.

  None. Overlaps #1179, which landed the checkpoint-coverage half independently; merged and reconciled here.

  Before merging `origin/main`: `python -m pytest src/hyperloom/agents/kernel/tests -q` -> **1673 passed, 7 skipped, 12 subtests passed**. `ruff check` clean apart from one pre-existing unused import in `forge_gemm_tuning.py`, a file this PR does not touch (confirmed by re-running with these changes stashed). The post-merge run is left to CI.

  Added:
  - A `caplog` assertion on the coverage-mismatch case, so a dropped checkpoint is never inferred from a return value alone.
  - `_forge_failure_tail` unit tests: usage line outranks the transcript, fallback to the last lines, result blobs skipped, empty input, length bound.
  - `test_nonzero_exit_reports_the_child_reason_not_only_the_code`, driving a stubbed child that rejects its argv.

  Updated:
  - Dropped the `shapes=` argument at 7 `_run_loop_via_cli` call sites and the 3 `_forge_e2e_pct` unit tests.
  - `test_generated_argv_...`: replaced the `--shapes-json` expectation with a guard asserting `--shapes-json` / `--e2e-pct` / `--kernel-kind` are absent.
  - `test_forge_submit_driver_fallback.py`: the spy described above.

  Cross-repo verification, re-run after the merge: captured the argv `_run_loop_via_cli` actually builds and fed it to the real `forge-loop` click command from a current KernelForge checkout. It parses cleanly — 19 options on the default provider branch, 24 with the OpenAI-only pin (`--agent-backend`, `--agent-fallback-provider`, `--model`).

  Merge note: the only conflict was the coverage condition described above. Nothing else on main needed mirroring — its `forge_submit.py` changes retire the embedded kernel-harness driver templates (`--shape`/`--mode`/`--warmup` on Hyperloom's own generated drivers) and leave the forge-loop argv untouched.

  No. This restores an argv the producer rejected outright, and makes the failure that hid it legible.


chaojhou (approved):
Approving. I checked the three claims this change rests on rather than taking them on trust, and all three hold.

**"No information is lost" is accurate.** This was the one worth verifying, because if the shape suite did not actually reach KernelForge the change would be silently dropping the workload-driven GEMM shape selection from #1099 rather than fixing a crash. The two paths are genuinely the same source: `_invocation_spec.py:661` iterates `task_group_shape_cases(candidate)` and builds `case_selectors` at `:725`, and that is the same function `forge_submit.py:4375` called to build the removed `--shapes-json` payload. The spec additionally carries per-case `arguments` / `outputs` / `raw_arg_spec`, which the option never did.

**The severity is as described.** `--shapes-json` sat in the unconditional segment of the `cmd` list, between `--workspace` and `--snr-threshold`, so every forge-loop invocation aborted in argument parsing at exit 2 and the whole path was unrunnable.

**Removing `--e2e-pct` is functionally a no-op.** Since KernelForge never declared the option it never received the value, so deleting `_forge_e2e_pct()` cannot cost it any input. Worth stating explicitly because "we deleted the GPU-share projection input" reads more alarming than it is.

**Keeping the `shapes` dict in `submit()` is justified.** Both remaining consumers exist: `evaluate_rewrite_route` at 4395 for the FlyDSL routing decision and `_validated_forge_checkpoint` at 4629 for checkpoint case coverage. And no unused import is left behind — `math` still has five call sites (the SNR `log10` computations and the `mean_case_speedup` finiteness check), so removing `_forge_e2e_pct` does not orphan it.

Moving the grouped-selector assertion in `test_forge_submit_driver_fallback.py` off the stubbed loop kwarg and onto `_shapes_from_candidate()` is the right call — it keeps the selector-resolution coverage that would otherwise have been deleted along with the argument.

---

Two notes for the planned KernelForge-side follow-up, since that is where the durable fix lives.

**Prefer "ignore and report" over "ignore silently".** Accepting undeclared options without aborting is the right instinct, but a silent ignore would have made this specific bug much harder to find, not easier. The exit 2 is what surfaced it — `forge-loop exited rc=2` on every attempt. Had KernelForge silently ignored `--shapes-json`, Hyperloom would have believed it was passing workload shapes while Forge quietly tuned against defaults, with no signal at all and a real quality cost. If the unrecognized option names are logged *and* echoed back in the result JSON (`unrecognized_options: [...]`), Hyperloom can detect interface drift on the next run instead of someone noticing degraded tuning weeks later. Crash becomes recoverable without giving up observability.

**Consider exporting the declared-option list so the cross-repo check can be automated.** The manual verification described here — feeding the argv `_run_loop_via_cli` builds to the real `forge-loop` click command — is exactly the right check, but it is one-off. `test_cli_invocation_pins_the_forge_loop_contract` can only pin expectations on this side, and the new negative guard covers the three specific options that broke rather than a fourth undeclared one added later. KernelForge is not currently reachable from CI (`kernel_agents` is not a dependency), so if the follow-up exposes the option set in a machine-readable form, Hyperloom could assert its generated argv parses against the pinned CLI. That addresses the interface-drift problem at its root, whereas tolerating unknown options only removes the symptom.

PR #1172, opened 2026-08-13
feat(kb): replay agent overlays as one validated warm set

meinali-566:
## Summary
- keep Explore, Framework, and Kernel knowledge on one seven-tuple `inference:` Recipe and one `optimized_throughput` champion
- publish one final `value.config` snapshot from durable `current_best`; Explore/Framework owner sections contain ordered patch/artifact refs only
- use CLOSE builders as the sole Kernel GEMM/Fusion/Rewrite projection, including source, patch, tuned-table, and rewrite experience artifacts
- consume the KB Store Hyperloom Recipe View and PostgreSQL-backed identity search for exact and degraded donor selection
- retain local RecipeKB as an isolated compatibility backend; remote T0 performs identity/search only and CLOSE owns publication

## Warm start and replay
- search exact identity first, then same architecture class, same GPU ISA/topology, compatible bf16/fp16, and nearest non-newer framework version
- load final config and global patch timeline through `RecipeReplayKB`; load prior Kernel columns through the read-only `KernelAgentKB`
- apply Recipe overlays and Kernel changes once, then grade the complete set with one E2E correctness/throughput benchmark
- promote the already-validated Recipe checkout as the session-authoritative InferenceX path; patches are not re-applied to a second canonical target
- KEEP preserves the physical Kernel changes for subsequent optimization; drift, quality failure, or apply failure invokes one shared Recipe+Kernel rollback implementation
- rollback failures retain durable recovery state and stop the session with an explicit reportable reason; there is no Config/Env-only salvage retry

## Publication
- durable post-save outbox stages only accepted Explore/Framework patch sets
- remote CLOSE uses staged owner patches plus replayed prior overlays, one final config snapshot, and CLOSE-only Kernel builders
- pure warm replay does not republish; a later new KEEP writes the validated cumulative result under the original target seven-tuple
- old remote/native records are projected by the KB Store View as history-only when they are not replayable

## Removed duplication
- removed `prs_tested`, Config/Env-only fallback replay, owner config snapshots/conflict merge, duplicate canonical patch persistence, duplicate Writeback rollback loops, incremental Kernel staging, dead SDK methods, and redundant KB tests
- deliberately do not create a separate Hyperloom `kernel:` record; Kernel knowledge remains tied to the inference workload's E2E result

## Scope
- single-node text unified diffs
- no binary patch replay, setup-command replay, Agent conflict resolver, or multi-node patch fan-out

## Test plan
- [x] latest-main combined Kernel/KB regression suite: 748 passed
- [x] consolidated KB architecture suite: 532 passed
- [x] CI failure regression suite: 71 passed
- [x] Ruff on changed production/tests
- [x] `git diff --check`
- [x] Gitleaks, docs, packaging, REUSE, Bandit, Pylint passed on rebased head


chaojhou:
Reviewed at head `0cbb26c3d520` across donor selection, replay/rollback, the benchmark path, and publication. A lot of the hard parts are right — the Recipe-overlay apply is genuinely atomic, the crash-window ordering is correct, `KernelAgentKB` really is read-only, and the write identity is the recipient's own seven-tuple so a degraded donor's record cannot be poisoned. I have a longer list, but these three are the ones I would hold the merge for. Everything below is verified against the code at that head.

## 1. Precision-bound configuration crosses the bf16/fp16 relaxation verbatim

`_precision_is_compatible` is symmetric by construction and nothing downstream re-checks:

```964:970:src/hyperloom/orchestrator/knowledge/recipe_kb_t0.py
def _precision_is_compatible(target: str, candidate: str) -> bool:
    """Allow exact precision plus the first bf16/fp16 compatibility pair."""
    target_value = str(target or "").strip().lower()
    candidate_value = str(candidate or "").strip().lower()
    if candidate_value == target_value:
        return bool(target_value)
    return {target_value, candidate_value} == {"bf16", "fp16"}
```

The replayed payload is `value.config` read verbatim and merged last-wins into the server command line. I read `sanitize_publish_server_args` (`remote_recipe/sanitize.py:166-202`) in full: it drops credential-shaped options and absolute-path operands, and that is all — there is no allowlist of option names anywhere in the file, and the strings `--dtype`, `--quantization` and `--kv-cache-dtype` do not appear in it. All three are first-class recognized single-value flags (`_grid_server_args.py:418-422`). So an fp16 donor carrying `--dtype float16` silently forces fp16 on a bf16 recipient. The env side has the same shape: `AITER_CONFIG_GEMM_BF16` is a *relative* path under an allowed prefix, so `sanitize_publish_env_mapping` keeps it (only absolute paths are dropped, `sanitize.py:157-158`) and an fp16 recipient is pointed at a bf16 tuned-GEMM table.

What makes this the first item rather than the most severe one: the mechanism that is supposed to catch it cannot, by construction. A wrong-dtype run posts entirely believable throughput, so the single E2E benchmark that grades the warm set will pass it. This is a silent wrong measurement, not a crash.

There is no precondition in the code — the tier is live for any donor pair differing only in precision.

## 2. One unresolvable patch path permanently poisons the outbox and costs the session its entire publication

A row whose patch member does not resolve is retained forever, with nothing to clear or age it out:

```402:404:src/hyperloom/orchestrator/loop/writeback.py
            if row.get("missing_patch_sources"):
                retained.append(row)
                continue
```

and a non-empty outbox aborts the whole record build, not just that owner's section:

```1011:1015:src/hyperloom/orchestrator/knowledge/remote_recipe/values.py
    if pending_sections:
        raise RemoteRecipeValidationError(
            "required section staging is incomplete: "
            f"{[row.get('id') for row in pending_sections if isinstance(row, Mapping)]!r}"
        )
```

The exception propagates out of `write_final_remote_recipe` into `finalize_recipe_and_journal`, which records an audit error and publishes nothing. The final config snapshot, every other accepted KEEP, and all three Kernel columns are lost because of one missing `.patch` file.

The realistic trigger is not a bug in the patch producer. `_drain_agent_keep_outbox` re-resolves the stored absolute paths on resume, so a crash followed by attempt-directory GC is sufficient to turn a good session into a total publication failure. This is worse than before the PR because the independent `kernel:` record — whose deleted docstring named this exact purpose, "so kernel optimizations land even when the recipe write itself failed" — is removed in the same change, so there is no longer a second sink, no degrade-and-warn path, and no operator override.

Section-level fail-closed is defensible. Record-level almost certainly is not what was intended.

## 3. The shared kernel rollback restores one file of a multi-file patch and reports success

Three things compose. The target resolver returns at the *first* `+++` header:

```409:415:src/hyperloom/orchestrator/phases/prelude.py
        for line in text.splitlines():
            if line.startswith("+++ "):
                candidate = line[4:].split("\t", 1)[0].strip()
                if candidate.startswith("b/"):
                    candidate = candidate[2:]
                if candidate and candidate != "/dev/null":
                    return candidate
```

`_snapshot_warm_kernel_target` then backs up only that one resolved path (`prelude.py:698-718`, single `target`, single `NNNN.bin`). And the revert short-circuits to the snapshot path whenever snapshots exist:

```677:679:src/hyperloom/orchestrator/phases/prelude.py
        if snapshots:
            return PreludePhase._restore_warm_kernel_snapshots(snapshots)
        from ..kernel.request_handlers import _maybe_revert_kernel_patch
```

So `_maybe_revert_kernel_patch` — the manifest-aware restore that knows `source_backups` for every file plus the artifact copies — is unreachable, while the apply side (`apply_snapshot`) writes every descriptor in the diff. A patch touching N files rolls back 1 and returns `{"ok": True}`. Because that reads as success, `warm_replay_pending` is cleared, no stop reason is set, and the session continues on a partially patched framework tree.

Two stated guarantees break together here: the rollback is "one shared implementation" only at the entry point, and "rollback failures retain durable recovery state and stop the session" never fires because this is not classified as a failure. It applies to every rollback path — drift, quality failure, dispatch denial, enqueue failure, resume — and it is also the recovery mechanism for the fact that the PolicyGate warm-replay target check runs at dispatch (`sub_agent_runner.py:214`), after PRELUDE has already written the files (`prelude.py:938`; task created at `:1519`). The gate detects an under-covered multi-file patch; this rollback is what is supposed to undo it.

**One question decides the severity, and it is yours to answer:** do warm kernel patches ever span multiple files? `apply_snapshot` writing every descriptor, and rewrite/fusion records carrying a plural `source_paths`, both suggest the shape is supported — but I have no multi-file sample to confirm it. If yes, this is the highest-priority item in the PR. If fusion/rewrite patches are always single-file in practice, it is a latent defect rather than a live one and can be scheduled.

Either way the fix is small: drive the revert off the apply manifest, or snapshot every path the diff declares rather than only the first.

xiaofei-zheng (approved):

PR #1173, opened 2026-08-13
fix(ci): stop an ordinary PR comment from cancelling the e2e run

zoroyihan7:
## Problem

`ci-e2e` listens on `issue_comment` so `/retest` can re-run it without a new
commit, and its concurrency group was keyed on the PR number with
`cancel-in-progress: true`:

```yaml
group: ci-e2e-${{ github.event.pull_request.number || github.event.issue.number || github.ref }}
```

GitHub resolves `concurrency` when a run is **created** — before the `resolve`
job's `if` can decline the work. So **any** comment on a PR created a run that
joined that PR's group, cancelled the in-flight run, and only then skipped
itself.

The cost is a whole single-GPU e2e run, and the evidence is on this repo:

| | |
|---|---|
| 05:36:52Z | PR #1161 e2e starts (run `31670922249`, `pull_request`) |
| 05:39:25Z | a review reply is posted on #1161 |
| 05:39:28Z | run `31671071877` created (`issue_comment`, branch `main`) → **`skipped`** |
| 05:39:42Z | run `31670922249` → **`cancelled`**, 49s in, workload deleted |

The skipped run never ran a step; taking the group was its entire effect. Recent
history holds four more `issue_comment` runs of this shape, so the failure is not
a one-off — commenting on a PR while its e2e is in flight has been silently
killing it, and the run reports red as `e2e fail`, which reads as a test failure.

## Fix

Key a comment that *cannot* start a run to a group of its own, so it cancels
nothing. Everything that could preempt a run before still does:

| trigger | group | preempts |
|---|---|---|
| `pull_request` (new commit / label) | `ci-e2e-<pr>` | yes |
| `issue_comment` with `/retest` | `ci-e2e-<pr>` | yes — intended |
| `issue_comment`, anything else | `ci-e2e-noop-comment-<id>` | no |
| `workflow_dispatch` | `ci-e2e-<ref>` | yes |

The `/retest` predicate is spelled exactly as in `resolve.if`, so the comments
that can start a run are precisely the comments that can cancel one. That
equality is what the fix rests on, so it is pinned by a test rather than a
comment.

## Test plan

- [x] `test_ci_e2e_concurrency_guard.py` — 4 tests: `cancel-in-progress` still
      set; a non-retest comment is keyed per comment id; the group and
      `resolve.if` agree on the retest predicate; all three triggers still
      contribute a key
- [x] Reverting the group to its previous value fails exactly the two tests that
      describe the bug (`test_a_comment_that_cannot_start_a_run_cannot_cancel_one`,
      `test_the_group_and_the_gate_agree_on_what_a_retest_is`) and leaves the
      other two passing
- [x] `test_packaging_lint.py` still green (new test module stays out of the wheel)
- [ ] Behaviour itself needs a real run to observe: after merge, a comment on a PR
      with e2e in flight should leave the run alive, and `/retest` should still
      preempt it

## Note

Drift guard aside, the group expression cannot be unit-tested as an expression —
there is no local evaluator for GitHub's `&&`/`||` ternary. I walked the four
trigger cases by hand against the truthiness rules; the table above is that
walk-through.

Separately: `_find_repo_root()` is now duplicated verbatim in
`test_packaging_lint.py` and `common/tests/test_llm_client_architecture.py`. This
PR does not add a third copy (it looks for the workflow it guards, not the repo
root), but a shared checkout-discovery helper for tests looks worth extracting on
its own.

### DIFF
--- a/.github/workflows/ci-e2e.yml
+++ b/.github/workflows/ci-e2e.yml
@@ -31,8 +31,21 @@ on:
 
 # A newer commit / retest on the same PR cancels the in-flight run; the cleanup step
 # below cancels the workload so no GPU is leaked.
+#
+# A comment that is NOT a `/retest` command must stay OUT of that group. GitHub
+# evaluates `concurrency` when the run is created -- before the `resolve` job's `if`
+# below -- so an ordinary review comment on a PR would otherwise cancel that PR's
+# in-flight run and then skip itself, losing a multi-hour GPU run to a comment. Such
+# comments get a group of their own (keyed by comment id) so they cancel nothing. The
+# `/retest` predicate is kept identical to the one in `resolve.if`, so exactly the
+# comments that can start a run are the comments that can cancel one.
 concurrency:
-  group: ci-e2e-${{ github.event.pull_request.number || github.event.issue.number || github.ref }}
+  group: >-
+    ci-e2e-${{
+      (github.event_name == 'issue_comment' && !contains(github.event.comment.body, '/retest'))
+      && format('noop-comment-{0}', github.event.comment.id)
+      || (github.event.pull_request.number || github.event.issue.number || github.ref)
+    }}
   cancel-in-progress: true
 
 # The workflow token writes the live commit status, the sticky PR report comment, and
--- a/src/hyperloom/inference_optimizer/tests/test_ci_e2e_concurrency_guard.py
+++ b/src/hyperloom/inference_optimizer/tests/test_ci_e2e_concurrency_guard.py
@@ -1,0 +1,96 @@
+# SPDX-FileCopyrightText: 2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Guard for the ``ci-e2e`` concurrency group.
+
+The workflow answers three triggers (``pull_request`` / ``issue_comment`` /
+``workflow_dispatch``) and cancels in-progress runs that share its group. GitHub
+resolves ``concurrency`` when the run is *created*, before the ``resolve`` job's
+``if`` can decline the work, so a group keyed on the PR number alone lets any
+comment on a PR cancel that PR's in-flight multi-hour GPU run and then skip
+itself.
+
+The fix keys non-``/retest`` comments to a group of their own, which only holds
+while the group expression and ``resolve.if`` agree on what a retest comment is.
+These tests pin that agreement; there is no way to unit-test the cancellation
+itself short of running the workflow.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import pytest
+import yaml
+
+_RETEST_COMMAND = "/retest"
+
+
+def _find_workflow() -> Path | None:
+    """Locate ``ci-e2e.yml``; returns None when running from an installed wheel."""
+    for parent in Path(__file__).resolve().parents:
+        candidate = parent / ".github" / "workflows" / "ci-e2e.yml"
+        if candidate.is_file():
+            return candidate
+    return None
+
+
+_WORKFLOW = _find_workflow()
+
+pytestmark = pytest.mark.skipif(
+    _WORKFLOW is None,
+    reason="ci-e2e guard needs the source checkout (.github/workflows/)",
+)
+
+
+@pytest.fixture(scope="module")
+def workflow() -> dict:
+    assert _WORKFLOW is not None
+    return yaml.safe_load(_WORKFLOW.read_text(encoding="utf-8"))
+
+
+@pytest.fixture(scope="module")
+def concurrency_group(workflow: dict) -> str:
+    return " ".join(str(workflow["concurrency"]["group"]).split())
+
+
+@pytest.fixture(scope="module")
+def resolve_condition(workflow: dict) -> str:
+    return " ".join(str(workflow["jobs"]["resolve"]["if"]).split())
+
+
+def test_an_in_flight_run_is_still_preempted(workflow: dict) -> None:
+    """The point of the group: a newer commit must not queue behind the old run."""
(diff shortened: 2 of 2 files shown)


xiaofei-zheng (approved):

chaojhou:
The diagnosis is right, the expression is right, and the evidence holds up — I'd take this. One gap worth deciding on before merge, because the residual class it leaves is unusually likely to fire on this PR's own review thread.

## Verified

**The expression evaluates as the table claims.** GitHub's `&&` / `||` return an operand rather than a boolean, and `&&` binds tighter than `||`, so it parses as `((A) && format(...)) || (B)`. Walking it:

- `pull_request` — `A` is false at the first conjunct, `false && format(...)` short-circuits to `false`, `false || (B)` yields `pull_request.number`. Group `ci-e2e-<pr>`.
- `issue_comment` + retest — `!contains(...)` is false so `A` is false, same fall-through; `pull_request.number` is null on an `issue_comment` payload, so `B` resolves to `issue.number`. Same group as the PR's own runs, so it preempts. Intended.
- `issue_comment`, anything else — `A` is true, `true && format(...)` returns the formatted string, which is truthy, so `||` returns it. Group `ci-e2e-noop-comment-<id>`, unique per comment.
- `workflow_dispatch` — `A` false, both numbers null, `B` falls to `github.ref`.

`format(...)` can never return a falsy value here and PR numbers are never 0, so there is no input that collapses a branch.

**The evidence chain is real.** Run `31670922249` is `pull_request` on `fix/1143-proposal-schema`, `cancelled`; run `31671071877` is `issue_comment` on `main`, `skipped`; both updated at `05:39:47Z`, 14 seconds after the comment run was created. That is the cancellation, and it confirms the premise that `concurrency` resolves before `resolve.if` runs.

**It is more than four.** The last 100 `ci-e2e` runs hold **34** `issue_comment` / `skipped` runs against 27 `pull_request` / `cancelled`. Not every comment run overlapped an in-flight run, but 34 draws is the exposure this removes.

**The folded scalar is fine.** `>-` keeps the newlines rather than folding them, because the continuation lines are more-indented than `ci-e2e-${{` — parsing the file yields `'ci-e2e-${{\n  (github...\n}}'`. It works anyway: there is exactly one `${{` / `}}` pair and every newline falls inside it, so the expression parser consumes them and the group name comes out clean. The test's `" ".join(...split())` normalization makes the assertions immune to the folding either way, which is the right call.

## The gap: there are three retest predicates, not two

The PR rests on "exactly the comments that can start a run are the comments that can cancel one", and pins it with `test_the_group_and_the_gate_agree_on_what_a_retest_is`. That agreement is real but it covers the two predicates that do not decide anything:

- L45 — concurrency group: `contains(github.event.comment.body, '/retest')`
- L68 — `resolve.if`: `contains(github.event.comment.body, '/retest')`
- L147 — the `decide` step, which is what actually sets `run`: `grep -qiE '(^|[[:space:]])/retest(-failed)?([[:space:]]|$)'`

L147 is strictly narrower: it demands `/retest` as a standalone token. So a comment that mentions the command without it standing alone — backticked, or `/retesting` — still takes `ci-e2e-<pr>`, cancels the in-flight run, passes `resolve.if`, allocates the self-hosted runner, and only then gets rejected by the anchored grep, leaving `run=false` and no replacement run. That is the same bug this PR fixes, narrowed from "every comment" to "comments mentioning the command".

Narrow, but self-referential: writing about the retest command is exactly what a review thread on this PR does. I checked before posting this — #1173's own e2e is `completed`/`success` and nothing is in flight on `fix/ci-e2e-comment-cancels-run` — otherwise this comment would have killed it, since it contains the substring several times.

I did not find a historical instance: no completed `issue_comment` run in the last 100 has a non-skipped conclusion, and the two currently in flight are genuine retests with their `e2e` job running. So the residual is latent rather than observed.

**Suggestion.** Spell all three as a prefix match, which GitHub expressions can express and the shell can anchor:

- L45 / L68 → `startsWith(github.event.comment.body, '/retest')`
- L147 → anchor the regex to `^/retest(-failed)?([[:space:]]|$)`

`/retest-failed` still matches the prefix, and every accidental-mention case disappears. It does drop "please /retest" as a trigger, so the doc line at L10 would need to say the comment must begin with the command — which is the usual slash-command contract anyway. If you would rather not change user-facing behaviour in this PR, the alternative is to say in the workflow comment that the group is deliberately looser than the shell, so the next reader does not take the "identical predicate" claim at face value.

Either way, the drift test can then cover the predicate that actually decides — asserting the shell regex is `^`-anchored is a real agreement check, where today the test is blind to the one that matters.

## Same shape on the label path (latent, follow-up)

`pull_request` subscribes to `labeled` / `unlabeled` (L20) and `resolve.if` declines every label except `retest` / `skip-e2e-test` (L70-71) — so a human adding any unrelated label to a PR creates a run in `ci-e2e-<pr>`, cancels the in-flight run, and skips. Identical mechanism, identical cost.

Dormant today, and I checked why rather than assuming: no PR in the last 30 carries a label, `auto-label.yml` is scoped to `issues: [opened]` so it never touches PRs, and the "Consume retest label" step deletes the label with `GITHUB_TOKEN`, whose events do not create workflow runs — so the documented label path does not cancel the run it just started. Worth a follow-up rather than a blocker, and the same noop-group trick extends to it.

## Small things

Test placement matches the existing convention (`test_packaging_lint.py` is the sibling meta-test in the same directory), and CI confirms the wheel-contents and packaging-lint guards still pass with the new module. The `format('noop-comment-{0}', ...)` key is unique per comment, and since `issue_comment` is subscribed to `created` only, an edited comment cannot re-enter the group.

On the note about `_find_repo_root()` — agreed that a shared test-side checkout discovery helper is worth extracting, and agreed it should not happen here.

PR #1174, opened 2026-08-13
fix(kb): accumulate the kernel-agent KB per optimization, not per document

BaoYunkai:
## Problem

The kernel-agent record holds one column per backend (`gemm`/`fusion`/`rewrite`),
but it was published as an indivisible document:

- a single document score (the best kernel gain in the session) decided whether
  to write at all, and
- the write replaced the whole record.

So a session that only tuned GEMM, and scored higher than what was published,
erased a rewrite an earlier session had learned — even though nothing better
about that rewrite had been found:

| step | session | published record |
| --- | --- | --- |
| 1 | learns a rewrite, gain 20% | `rewrite=k1` |
| 2 | tunes only GEMM, gain 30% | `gemm=v1` — **`k1` gone** |

Giving the kernel agent its own record removed this all-or-nothing behaviour at
the recipe level and reintroduced it one level down. The KB therefore did not
accumulate: a later run could leave it holding less than an earlier one, and
PRELUDE's warm replay silently had less to restore over time.

The publish gate had the mirror-image flaw. Because it compared one score,
adding a *second* kernel that did not raise the best gain looked like no
improvement, and the new column was dropped.

## Fix

Reconcile against what is already published rather than replacing it.

- **Match per optimization, not per document.** Records are keyed within their
  column by the optimization they describe — GEMM by variant name or tuned
  table, fusion and rewrite by kernel — and the higher end-to-end gain wins each
  slot. A session contributes exactly what it improved and inherits the rest;
  different kernels occupy different slots.
- **Re-stage inherited artifacts.** The document is still replaced wholesale, so
  the files behind an inherited record are copied back into the upload set;
  otherwise its refs would dangle.
- **Gate on the merge, not on a score.** A write happens when the merge changed
  something. `write_record` performs that ungated publish and shares its
  implementation with `write_if_better`, which keeps its own score gate.
- Reading the published record is best-effort: an unreadable incumbent degrades
  to publishing this session's own columns rather than losing them.

### DIFF
--- a/src/hyperloom/inference_optimizer/tests/test_kernel_kb_columns.py
+++ b/src/hyperloom/inference_optimizer/tests/test_kernel_kb_columns.py
@@ -12,6 +12,7 @@ from hyperloom.orchestrator.knowledge.remote_recipe.values import (
     KERNEL_AGENT_METRIC,
     build_kernel_agent_knowledge,
     kernel_agent_canonical_id,
+    merge_kernel_columns,
 )
 from hyperloom.orchestrator.knowledge.kernel_kb_columns import stage_kernel_columns
 from hyperloom.orchestrator.knowledge.remote_recipe._vendor.kb_store_client import (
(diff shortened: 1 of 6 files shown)


BaoYunkai:
### Champion semantics, measured

Probed the service against a throwaway identity (`kernel:hyperloom-champion-probe-*`), no real workload touched:

| Promotion | Result | Champion after |
|---|---|---|
| session A, value 10.0 | accepted | A |
| session B, value 10.0 (equal, different session) | accepted | B |
| session C, value 5.0 (lower) | accepted | C |

`get_best_record()` follows the champion each time. So `POST /champion` does no value comparison at all: an equal *or lower* value is accepted, a 409 means a concurrent write, and keep-if-better is entirely the client's responsibility.

### Fixes

1. **Read failure destroyed the accumulated record.** The published record is the base of every merge and the write replaces the document wholesale, so treating an unreadable incumbent as absent published this run's columns alone. The read failure now fails the write, reported as `error: published_record_unreadable`. Corrected the docstring that claimed the opposite.

2. **Champion never taken over at an equal score.** The conflict retry only re-promoted on a strictly higher score, leaving the accumulated document off-champion for every identity written under the old per-run scheme. An accumulating record now re-promotes over an equal score; the competing recipe record keeps the strict comparison. `_FakeStore.set_champion` also records promotions the way the service does, instead of dropping them and reporting a champion that never moved.

3. **Ref displacement matched on basename.** A record naming two files that share a basename across directories had the second ref repointed at the first, so its bytes went unreferenced, were dropped from the upload, and would replay as the wrong kind of artifact. Matching is on the whole ref now.

Each fix has a regression test that reproduces the reported failure without it.

xiaofei-zheng (approved):

PR #1175, opened 2026-08-13
fix(kernel): stop sending forge-loop two options it no longer has

BaoYunkai:
## Problem

`_run_loop_via_cli` passes `--shapes-json` and `--e2e-pct` to `python -m kernel_agents.cli forge-loop`. KernelForge has neither any more:

- `--shapes-json` went with `33b7723` *"make drivers own full-suite evaluation"* ("Remove shape and workload coupling"), and `6bb0fba` removed the leftover shim.
- `--e2e-pct` went with `2158a15` *"harden forge-loop failure boundaries"* ("remove unreachable profiling surfaces and stale contracts").

click rejects an unknown option during parsing, before running anything, so each launch exits rc=2 and the attempt is lost. Probing all 26 options this launcher passes against the current KernelForge:

| option | KernelForge main | an older installed KernelForge |
|---|---|---|
| `--shapes-json` | **missing** | ok |
| `--e2e-pct` | **missing** | ok |
| `--gpu-type` | ok | **missing** |
| other 23 | ok | ok |

This is the failure that killed 22 attempts on Aug 11-12 and became the dominant one on Aug 12; against a current KernelForge it costs every attempt.

## Fix

Neither value fed anything but those two arguments, so the launcher no longer takes them. `e2e_pct` is still computed and still gates dispatch on the Hyperloom side; only the argv entry is gone.

A second, quieter breakage came with the same upstream change. `case_coverage` in a forge-loop checkpoint was derived from `--shapes-json`, so no current forge-loop reports it, and `_validated_forge_checkpoint` vetoed recovery whenever it disagreed with the expected coverage -- including when it was absent, which is now always. Every salvageable best from a timed-out campaign was discarded, and the run simply looked like it recovered nothing. Coverage now vetoes only when it is reported and disagrees; absent or empty carries no information.

## What is deliberately not here

`--gpu-type` shows the skew runs both ways: an older KernelForge does not have it, and this PR does not make Hyperloom stop sending it, because the child needs the card to address its experience. Nothing in CI makes the two repositories agree on this argv, which is the actual gap; a contract check is filed separately.

## Tests

- `test_the_invocation_omits_options_kernelforge_no_longer_has` -- fails on `main`, pins both options out of the argv.
- `test_a_checkpoint_that_reports_no_coverage_is_still_recoverable` -- fails on `main` with `assert None is not None`; covers absent, empty, and reported-but-disagreeing coverage.
- `test_cli_invocation_pins_the_forge_loop_contract` loses the `--shapes-json` expectation.
- One assertion in `test_grouped_multi_shape_task_requires_one_prepared_driver` went with the argument it inspected; that test still pins the prepared driver, which is what its name is about.

## Test plan

- [x] `pytest src/hyperloom/agents/kernel/tests` -- 1714 passed, 6 skipped
- [ ] A session against the relocated KernelForge reaches a forge-loop iteration instead of rc=2
- [ ] A timed-out campaign salvages its validated best again


BaoYunkai:
Closing in favour of #1170, which landed first and does the same removal more carefully: it re-points the assertion in `test_grouped_multi_shape_task_requires_one_prepared_driver` at `_shapes_from_candidate` instead of dropping it, and it also retires `_forge_e2e_pct` and the dispatch gate that read it.

One part of this PR is not in #1170, so it is now #1179: `case_coverage` was derived from `--shapes-json`, and `_validated_forge_checkpoint` vetoed recovery whenever the reported coverage disagreed with the expected one. No current forge-loop reports the field, and an absent field disagrees with everything, so every salvageable best from a timed-out campaign was discarded -- silently. Removing the option does not fix that; it makes the silence permanent.

PR #1176, opened 2026-08-13
feat(kernel): gate a launch on the forge-loop argv contract

BaoYunkai:
Stacked on #1170; review that one first.

## Why

Hyperloom and KernelForge are separate repositories, resolved against each other at runtime through `$FORGE_PATH`, and deployments track KernelForge closely -- during the Aug 11-12 failures the deployed build had already retired `--shapes-json` two days before that removal reached KernelForge main. So the standing risk is Hyperloom lagging behind the KernelForge it will actually run against.

click rejects an unknown option *while parsing*, before running anything, so that lag does not degrade an attempt, it costs the whole thing. Worse, the session does not fail: it finishes reporting success with zero kernel optimizations, which is how Aug 10-12 stayed unnoticed for three days across 301 attempts. rc=2 at least leaves a message; this leaves a healthy-looking run.

The existing guard, `test_forge_loop_cli_accepts_the_provider_flags`, greps `forge-loop --help` for three flags. That cannot see the options most likely to be retired: `--shapes-json` was `hidden=True`, so it never appeared in help text, before or after removal.

## The gate

`cli/preflight.py` runs on every launch and exits 2 on a hard failure. The new check sits next to the TraceLens hard-gate and follows it: skipped for `--no-kernel`, since a run that will not dispatch forge has nothing to break.

It resolves KernelForge through `_ensure_forge_on_path()` -- the same resolution a dispatch performs -- so it reads the build this session will use rather than whatever is newest upstream, and introspects the click command rather than `--help` so hidden options are visible. On a mismatch it names the options and refuses to start, turning "finishes healthy with no kernel work, three days unnoticed" into "fails in seconds, names the option".

## What is in the test suite, and why only that

`FORGE_LOOP_OPTIONS` is the gate's only account of what the launcher sends, because nothing in a running image can capture an argv. `test_the_declared_option_set_matches_the_argv_the_launcher_builds` pins that declaration against the argv actually built, with every conditional option satisfied. Without it a stale declaration would let the gate pass while the run still lost every attempt.

That test needs no KernelForge, so it holds wherever the suite runs. An earlier revision of this PR also compared against the locally installed KernelForge; that has been dropped. CI installs none, so it skipped there, and on a dev node it inspected a stale pip install and reported a mismatch no deployment would ever see. The gate covers that ground properly, in the only place that knows which build will run.

## Verified

| environment | outcome |
|---|---|
| KernelForge main on `$FORGE_PATH` | `forge_loop_contract_gaps()` returns `[]`; preflight prints contract OK |
| the stale pip-installed build on this node | returns `['--gpu-type']`; preflight exits 2 naming it |
| no KernelForge importable | returns `None`; preflight reports unchecked and does not block |

Suites: `src/hyperloom/agents/kernel/tests` plus the gate tests -- 1675 passed, 5 skipped. The same two files pass with no KernelForge (5 passed) and against KernelForge main (5 passed).

## Follow-up worth deciding

`main` carries no `required_status_checks` rule (only `deletion`, `non_fast_forward`, and one required review), so no check gates a merge today. Catching this class at PR time rather than at launch would mean comparing the argv against KernelForge main in CI, which needs a credential for `AMD-BRAIN-Internal`; the workflows have no secret for it.

## Test plan

- [x] `pytest src/hyperloom/agents/kernel/tests src/hyperloom/inference_optimizer/tests/test_preflight_forge_contract.py`
- [x] Same two contract files with KernelForge absent, stale, and at main
- [ ] A launch against a KernelForge missing an option refuses to start
- [ ] Decide whether CI should compare against KernelForge main

### DIFF
--- a/src/hyperloom/agents/kernel/tests/test_forge_cli_contract.py
+++ b/src/hyperloom/agents/kernel/tests/test_forge_cli_contract.py
@@ -1,0 +1,143 @@
+# SPDX-FileCopyrightText: 2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""The forge-loop argv contract, checked against the installed KernelForge.
+
+Hyperloom and KernelForge are separate repositories, wired together at runtime
+through ``$FORGE_PATH``, so nothing else makes them agree on this argv. click
+rejects an unknown option while parsing, before running anything, which is why a
+single retired or renamed option costs an entire attempt instead of degrading
+it: 22 attempts were lost to ``--shapes-json`` alone.
+
+Reading ``--help`` is not enough. Hidden options are absent from it, so a check
+built on help text cannot see the ones most likely to be retired -- ``--help``
+listed neither ``--shapes-json`` nor its removal.
+"""
+
+from __future__ import annotations
+
+import sys
+import time
+from pathlib import Path
+
+import pytest
+
+_BACKENDS_DIR = Path(__file__).resolve().parent.parent / "tools" / "backends"
+sys.path.insert(0, str(_BACKENDS_DIR))
+import forge_submit  # noqa: E402
+
+
+def _installed_forge_loop_options() -> set[str]:
+    """Every option the installed forge-loop accepts, hidden ones included."""
+
+    cli = pytest.importorskip(
+        "kernel_agents.cli",
+        reason="KernelForge is not installed, so its argv cannot be checked",
+    )
+    command = cli.main.commands.get("forge-loop")
+    if command is None:
+        pytest.fail("the installed kernel_agents CLI has no forge-loop command")
+    supported: set[str] = set()
+    for param in command.params:
+        supported.update(param.opts)
+        supported.update(param.secondary_opts)
+    return supported
+
+
+def _maximal_launcher_argv(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> list[str]:
+    """Build the argv with every conditional option present.
+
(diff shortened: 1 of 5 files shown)


BaoYunkai:
Closing. Neither half reaches far enough to be worth carrying.

The CI half cannot compare against anything on its own: KernelForge is a private repository in another organisation, so the job needs a read credential added as a repository secret before it does more than report that it could not look. A guard that ships inert, waiting on someone else to arm it, is worse than no guard -- it reads as covered.

The launch half does work, and does gate: it resolves KernelForge exactly as a dispatch would and refuses to start when an option is missing. But it only speaks at launch, on the machine that already has the mismatched pair, which is not where a contract break should be caught.

Worth keeping from the investigation, since none of it is recorded elsewhere:

- `tools/preflight_optimizer.py` is documented as the outer launcher preflight, but nothing calls it -- it is operator-run only. `cli/preflight.py` is the one that runs on every launch.
- `main` carries no `required_status_checks` rule (only `deletion`, `non_fast_forward`, and one required review), so no check gates a merge today, for any PR.
- The existing `forge-loop --help` guard cannot see hidden options, which is why `--shapes-json` was invisible to it both before and after its removal.
- Probing all 26 options the launcher passed showed KernelForge main missing `--shapes-json` and `--e2e-pct`; #1170 removes both.

The two real fixes stand on their own: #1170 stops sending the retired options, and #1179 stops discarding a salvaged best over coverage nobody reports.

Issue #1141, opened 2026-08-11
Containerized execution is unsupported but mandatory


marco-grond:
### Target release

v0.2

### Optimization domain

Inference

### Problem / use case

The optimizer hard-enforces baremetal serving (run_mode=local in bypass_runner.py) yet its benchmark scripts assume the vLLM-ROCm container's environment, and no ROCm vLLM wheel exists on PyPI. Net effect: you must run Hyperloom inside the serving container, which is non-obvious and cost real investigation time.

### Proposed solution

We need to make containerized execution a first-class, documented mode with automatic environment detection.

zoroyihan7:
Triage notes from tracing this alongside #1142–#1146. Not taking the issue over (@lishuoshuo-amd owns it) — just recording what I found so the fix scope is clear.

## The enforcement is real

- `orchestrator/actions/executors/bypass_runner.py:1048-1050` hard-rejects anything other than `local`:
  > `bypass runner supports --run-mode local only, got {args.run_mode}`
- The default benchmark backend always passes `--run-mode local` to Magpie (`actions/executors/benchmark_backend.py:97` and `:179`; default set at `benchmark_backend.py:23-27`).
- Container-internal paths are hardcoded in places, e.g. `orchestrator/phases/kernel.py:44` → `/sgl-workspace/aiter/aiter/configs`.

So "you must be inside the serving container" is an accurate description of current behaviour.

## One correction to the issue title: the container path *is* documented

`docs/install/install.md:374-404` has a "Docker (source)" section with recommended images, a full `docker run` invocation, and the `docker exec -w "$REPO_ROOT" ...` usage pattern. So this is less "unsupported" and more "documented, but not discoverable and not enforced". Three concrete reasons it reads as unsupported:

1. It is titled "Docker (source)", sits at the end of the install doc, and is worded as *recommended* rather than *required*.
2. `install.md:128-143` ("Scenario B: Bare metal + Docker") states that setup writes `.env` only and does not start a container — container lifecycle is delegated to "the selected demo skill". That indirection makes the entry point hard to find.
3. **There is no detection.** This is the part that costs investigation time.

## The detection already exists — it is just only used for labelling

[...]

PR #1182, opened 2026-08-13
fix(preflight): gate optimize on an importable ROCm serving framework

lishuoshuo-amd:
Refs #1141.

## What the issue got right, and what it got wrong

The reporter's conclusion — run Hyperloom inside the serving container — was correct. It just should not have been something he reverse-engineered.

Two corrections to the issue title. Containerized execution is **supported and documented**, and `examples/README.md` states Docker is the *recommended* run mode, with bare metal called out as "for advanced users". And it is **not mandatory** once a ROCm framework is installed on the host: the bare-metal installer pulls vLLM from `wheels.vllm.ai` and SGLang from `pypi.amd.com/rocm-*`, so "no ROCm vLLM wheel exists on PyPI" is true but never a blocker.

What was actually missing is the check that tells you which situation you are in. Going through `/hyperloom-setup` would have stopped him at Phase 1 of `install_baremetal.sh`; calling `optimize` directly bypassed all of it, so the run proceeded and died somewhere unrelated-looking.

## `run_mode=local` is not the switch he thought it was

Worth recording since the naming is the whole trap. `local` means "do not start another container", not "bare metal". Mechanically: `local` runs one bash command in whatever environment the process is already in (`benchmarker.py`), while `run_mode=docker` assembles a full `docker run` and executes the same command inside a **newly started** container. With Hyperloom inside a serving container, `local` starts the server *in that container* — which is containerized execution.

The switch he wanted is `HYPERLOOM_RUN_MODE` (`baremetal` | `docker`), a separate axis. And `local` is not a shortcut: Magpie refuses `server_lifecycle` under any other mode ("Docker/Ray executions cannot reuse a server process across Magpie invocations"), and the grid search depends on persisting one server across rounds.

## The gate

`optimize` now fails up front when the selected serving framework is missing, naming the remedy. Extending an existing check rather than inventing one — `install_baremetal.sh` Phase 1 has always done this; it just never covered this entry point, which left the two disagreeing about what a usable host is.

**Importability alone is not enough.** The PyPI wheel is the CUDA build: it imports fine and dies at GPU init. The ROCm verdict is tri-state, so an unanswerable probe (absent torch, signal death, timeout) warns instead of reporting a wrong diagnosis. Deliberately not `adapters.verify_torch_is_rocm`, which collapses inconclusive into `False` and waits 30 minutes.

**Every candidate interpreter is considered**, not just the first that imports. `install_baremetal.sh` defaults vLLM to an isolated `$VLLM_VENV_ROOT` venv, so the ROCm build routinely lives there while a stray CUDA wheel sits in the benchmark interpreter — stopping at the first import rejected a working host. An inconclusive verdict outranks a refuted one, so the run is blocked only when every importable candidate is provably wrong.

**Container detection covers cgroup v2, podman and Kubernetes.** Under cgroup v2 a container's `/proc/1/cgroup` is often just `0::/` with no marker, and `/.dockerenv` only covers Docker. Getting this wrong is worse than not knowing: a user already inside a container would be told to set `HYPERLOOM_RUN_MODE=docker`, reproducing the exact confusion this PR exists to remove. Signals now include `/run/.containerenv`, `KUBERNETES_SERVICE_HOST`, on-disk image markers via `common.provenance`, and PID 1 sitting at the cgroup root. `HYPERLOOM_IMAGE` is deliberately **not** proof — the demo skills export it on the *host* to pick an image — and a test pins that.

**Only the evidence actually gathered is claimed.** For vLLM the platform is checked; for every other framework only `torch.version.hip`, so a CUDA SGLang beside a ROCm torch would pass. Upstream offers nothing better (`sglang`'s own `is_hip()` just re-reads `torch.version.hip`), and this repo agrees — there is no `verify_sglang_rocm`. So the wording names its evidence instead of overstating it.

**Probe stderr is surfaced**, bounded to a few truncated lines. It was captured and discarded, which is precisely the output needed to diagnose a broken ROCm stack.

**The remote case points at the supported path.** `BENCHMARK_BASE_URL` is offered first and the skip variable is presented as a last resort; the earlier wording steered users into disabling a check instead of configuring the exemption that already existed.

Exempt: a remote server, external multi-node, scriptable frameworks, and `HYPERLOOM_SKIP_FRAMEWORK_CHECK`.

## Guidance carries nothing that can perish

The first version pinned image tags — an eighth copy of versions already duplicated across seven docs and skills, and the only one in Python. Living in an error path, no test fails when they bump; the gate would confidently name an image that no longer exists. A doc pointer was worse: an installed distribution ships no `docs/` at all.

What remains is a command and an env var, both present in every install shape, and image choice already belongs to the demo skill. A guard test rejects any version, GPU-arch or repo doc path re-entering the message.

## Probe budget

Worst case was 3 interpreters × (60s + 180s) ≈ 12 minutes, uncomfortably close to the 30-minute timeout this PR criticises elsewhere. Now bounded to ~140s: the scan short-circuits after the first timeout, and each budget matches an existing precedent for the same work (`build_utils.probe_torch_abi` for the torch probe, the heaviest inline probe in `targeted_build.py` for vLLM). A timeout on importability now warns instead of exiting — the old code read it as "absent" and hard-failed slow machines.

## Verification

- Full suite green with the credential env cleaned: **13153 passed, 43 skipped, 0 failed**. Every failure previously called "pre-existing" was the developer shell's real `OPENAI_*`/`ANTHROPIC_*` values leaking into assertions.
- Every new test was confirmed to fail before its fix, on behaviour rather than a missing symbol.
- Container detection verified both ways on a cgroup v2 host: `False` there, and `True` inside a private cgroup+pid namespace built with `unshare` (no container runtime involved).
- `main` merged in; `ruff check` clean. `preflight.py` has three pre-existing `ruff format` hunks, all in untouched code — not reformatted.

## Known limit

On a cgroup v2 host whose PID 1 sits at the cgroup root (a systemd-less WSL, a minimal init), detection reports "container". `systemd-detect-virt --container` is not a fix — it calls WSL a container too. The bias is deliberate and documented: a false negative gives actively wrong advice, a false positive only omits the `HYPERLOOM_RUN_MODE` hint while still offering the install remedy.


ZhengGong-amd (approved):
Approved

PR #1185, opened 2026-08-13
fix: derive paths and env precedence from the caller instead of assuming root

### LINKED ISSUE #1142: Derive paths from the $HOME directory
### Target release

v0.2

### Optimization domain

Inference

### Problem / use case

Hardcoded /root/.claude (install), /root/.aiter probe in clean_stale_aiter_locks() (uncaught PermissionError), git "dubious ownership" on mounted repos, and .env silently overriding USER_DATA_PATH.

### Proposed solution

Derive paths from $HOME, catch permission errors, mark mounted repos safe, and define explicit env-var precedence.

### Expected impact

_No response_

### Alternatives considered

_No response_

### Additional context

_No response_

lishuoshuo-amd:
Closes #1142.

All four failures the issue names, plus the wider ones found while proving each fix. Every one of them stays invisible only while the process happens to run as root.

## 1. Claude credentials were written where nothing reads them

`install.sh` wrote `/root/.claude/config.json`, while every reader resolves `Path.home()`. As root those are the same file, which is why the divergence survived. Under any other uid: `/root` unwritable aborts the whole installer under `set -euo pipefail`; `/root` writable but a different uid writes credentials the runtime never reads, reports success, and never mentions `/root` in the failure.

This was drift, not intent — the comment above the block and both sibling log lines already said `~/.claude`.

`_home_dir()` mirrors `posixpath.expanduser` rather than substituting `${HOME}`, verified against CPython source and empirically:

| `HOME` | `Path.home()` |
| --- | --- |
| `/tmp/x` | `/tmp/x` |
| unset | passwd entry |
| empty | `/` |

A *present* `HOME` wins even when empty; only a wholly absent one falls back to passwd. Trailing slashes are stripped in full because `expanduser` uses `rstrip('/')`, so `HOME=/foo//` must yield `/foo` — a single `${HOME%/}` would not. Substituting `${HOME}` would diverge on both inputs and, under `set -u`, crash where `write_env_file` already referenced it bare.

The test is **differential**: it runs the shell helper and `Path.home()` against one controlled environment and demands byte-identical output, pinning the invariant rather than one sampled path.

**The same function also installed the Claude CLI globally under `/usr/local`.** `run` executes bare under `set -e`, so off root that `npm install -g` takes the installer down for the identical reason, three lines away. The fallback prefix is `$HOME/.local`, which is already where `write_env_file` probes for the binary.

## 2. The aiter lock sweep raised where it documents that it cannot

`clean_stale_aiter_locks` states "never raises (errors counted)" and guards all three sweep I/O sites — but the resolution running *before* the sweep did not. `Path.is_dir()` re-raises `EACCES` because pathlib ignores only `ENOENT`, `ENOTDIR`, `EBADF`, `ELOOP`, so the `/root/.aiter/build` fallback aborted the call outright for any non-root uid. Guarding the sweep looked sufficient precisely because the failure happens earlier.

Skipped trees now reach `stats` and a warning; an all-zero stats dict read as a clean sweep of a tree nobody looked at. The probe order it depends on is pinned too — `$HOME` first, `/root` last — which nothing tested, since the existing sweep test blanks the tuple.

## 3. `.env` inverted env precedence in every documented launch recipe

`install.sh` already snapshots and restores protected variables around its dotenv load. That protection lived in one script only: the documented recipes ran a bare `set -a` sourcing, which re-exports every assignment in the file, so a stale `USER_DATA_PATH` there silently redirected the run to another workspace. `SKILL.md` promised the opposite, and scoped the guarantee to `install.sh` and the CLI preflight — excluding the recipes it documents.

The recipes now re-export a snapshot of the caller's environment. **Empty assignments are filtered out**, matching `install.sh`'s non-empty-only rule: an exported-but-blank value is a gap for `.env` to fill, and restoring it would send `${USER_DATA_PATH:-/workspace/hyperloom}` to the pod-local default.

The credential-only recipe needed the same treatment: filtering the load to `ANTHROPIC_*`/`OPENAI_*` keeps path variables safe but still lets a stale token replace the caller's — and `install.sh` snapshots those same four vars for exactly that reason.

`eval`'s stderr is no longer suppressed. `eval` continues after a failed assignment and reports the *last* statement's status, so a `|| warning` fired only when the failure came last. Unsuppressed, a clean re-eval prints nothing and a failure names the variable and the reason. Measured under `dash` too: it aborts on a readonly assignment either way, so this turns a silent death into a diagnosed one.

No shared helper, deliberately: `examples/*/SKILL.md` install outside the package tree via `data_files`, and the package root differs between wheel and source layouts, so no single sourceable path resolves in both. The duplication is held by tests that execute the snippets **extracted from the markdown**.

The snapshot is wider than `install.sh`'s fixed list, and the comment now says so instead of claiming parity. Narrowing to that list would not help the case that actually bites — `USER_DATA_PATH` is on it.

## 4. A bind-mounted checkout was unusable for git

The documented container recipe mounts the repo at the same path, so the checkouts the optimizer patches are routinely owned by a different uid. git then refuses **every** operation, reads included, and nothing configured an exception.

Which verbs are refused was measured, not assumed: `rev-parse`, `status`, `tag`, `show`, `clean`, `ls-files`, `checkout`, `reset`, `commit`, `add`, `branch`, `worktree`, `update-ref`, `revert`, `fetch`. The `apply` family is **not** — it works the tree as plain files and never validates ownership, even under `--index` — so those sites are untouched.

Three details decide whether the exception works, each measured against a real refusal via git's own `GIT_TEST_ASSUME_DIFFERENT_OWNER` hook:

- It must name the repository **root**. Ownership resolves there, so naming the `-C` path is ignored when that path is a subdirectory — and callers pass a `framework_root` that need not be the checkout root. Hence the walk up for a `.git` entry, which is a **file** in linked worktrees.
- `-c` must precede the subcommand, or git treats it as the subcommand's own argument.
- Command config is protected config, so `-c` is honoured for `safe.directory` though repository config is not. This matches how the commit path already injects an identity for checkouts lacking `user.email`, and needs no environment rewriting — which avoids `GIT_CONFIG_COUNT`, whose unparseable values make git fail hard.

**"Every load-bearing git call goes through `executors/_git.py`" was wrong**, and that claim is retracted. Reaching git directly, with real consequences:

| Site | Consequence of a refusal |
| --- | --- |
| `request_handlers.py` | Falls back to the on-disk file, so a patch's base snapshot silently comes from a dirty worktree instead of HEAD |
| `targeted_build.py` | Reports "no AITER version tags found in the cloned repo" — pointing at the repo, not at permissions |
| `baseline.py` | `rev-parse` / `reset --hard` / `clean -fd` on the InferenceX mirror, located by `cwd=` rather than `-C` |
| `forge_submit.py` | ~30 call sites incl. `worktree`, `branch`, `checkout`, `update-ref`, `revert` — the kernel agent's main path |
| `isolation.py`, `apply_tracelens_patch_multinode.py` | Private `_run_git` wrappers, both `cwd=`-located |
| `_nogit_patch.py`, `phases/framework.py` | A real git tree reads as "not a git tree", so patching drops to the non-git channel |
| `forge_fusion.py` | Repo root and tracked-file status fall back to guessing a package root |

The helper moved to `common/git_safety.py` and grew `cwd=` support for the callers that never pass `-C`. `tools/` modules import it inside functions: module level is stdlib-only there so those scripts still run standalone on remote nodes, which their other `hyperloom.common` imports already observe.

Left alone: `provenance.py` and `session/manifest.py` take a short SHA for run provenance, documented best-effort, returning empty on failure. Adding a mechanism for a probe designed to degrade would be an overreaction — and taking them as representative is exactly what made this look harmless at first.

Also fixed while in `phases/framework.py`: its `except` returned `False` instead of `continue`, abandoning the remaining roots on the first spawn error. Unrelated to ownership, pre-existing.

## 5. `USER_DATA_PATH`'s default suited only a container

`/workspace` is a container convention: the ROCm images ship it writable, so the fixed `/workspace/hyperloom` default is right there and unusable anywhere else. On bare metal off root the directory does not exist and `/` is not writable, so the `mkdir -p` sites in `install.sh`, `local_setup.sh` and `preflight_kb.sh` aborted under `set -e` before anything was installed — which is what kept "non-root can run it" from actually holding.

Unset `USER_DATA_PATH` now resolves to `/workspace/hyperloom` only when `/workspace` is writable, else `<cwd>/session`. Container behaviour is unchanged, and the fallback is what the setup skill has been offering all along, settling the disagreement between the skill and every other site.

The resolver is duplicated three times because those modules cannot import one another — `tools/` scripts run standalone and the `fa` CLI must not depend on `inference_optimizer`. A test asserts all three agree under both branches, since nothing else would catch them drifting.

## Verification

- Full suite green with the credential env cleaned: **13153 passed, 43 skipped, 0 failed** in 3m20s (`-n 16`). Every failure previously called "pre-existing" was the developer shell's real `OPENAI_*`/`ANTHROPIC_*` values leaking into assertions.
- Every new test was confirmed to fail before its fix, on behaviour rather than a missing symbol: `mkdir: cannot create directory '/root': Permission denied`; `HOME: unbound variable`; `PermissionError ... '/root/.aiter/build'`; `assert '/from/dotenv' == '/from/caller'` on all five recipes; `- key-from-caller / + key-from-dotenv`; `fatal: detected dubious ownership`.
- Recipe snippets verified under both `bash` and `dash`.
- `main` merged in. The one conflict was in `forge_submit.py`'s in-place restore block, which `main` rewrote; upstream's logic was taken and this branch's git-routing re-applied on top, rather than keeping the older local version.
- `ruff check` clean. Several touched files are not `ruff format`-clean on `origin/main`; each complaint was compared against `git show origin/main:<path>` and none were reformatted.

## Known limit

`GIT_TEST_ASSUME_DIFFERENT_OWNER` forces the ownership check to fail but leaves files writable, so the git tests prove the refusal is lifted, not that a genuinely unwritable mount becomes writable. In the common shape — container root, host-owned files — root can write; if a mount really is unwritable, the failure moves from "dubious ownership" to an `EACCES` from the same call.

## Deliberately out of scope

- `agents/kernel/tools/apply_kernel_patch.py` uses `Path(os.environ.get("HOME", "~")).expanduser()`, which is **not** equivalent to `Path.home()`: on an empty `HOME` it yields `.`, a relative path. Same class of drift, but it drives path classification rather than credential placement.
- `assets/slurm/run_hyperloom.sbatch` exports `USER_DATA_PATH` before sourcing `proxy.env`, so a hand-added entry there could still override it. `proxy.env.template` carries none of the protected path variables today, so this is latent.


ZhengGong-amd (approved):
Approved

xiaofei-zheng (approved):

PR #1187, opened 2026-08-13
Unblock bare-metal setup on hosts that ship a preinstalled engine

chennyiiis:
## What

Setup plumbing only. Phase 1 preflight in `install_baremetal.sh` refused to pass
on a host whose image already ships a serving engine that Hyperloom itself
cannot install, and the rest of setup only ever understood sglang/vllm — so even
once preflight passed, the run finished misconfigured.

### Preflight

- **`FRAMEWORKS` is overridable and probes every registered serving framework.**
  It was hardcoded to `sglang,vllm`, so a host with neither installed failed
  preflight with "no serving framework importable". Phase 1 still passes as soon
  as one entry imports.
- **The torch/triton alignment check is advisory under `--install-framework none`.**
  It stays fatal whenever a framework layer is actually being built against the
  pin. A prebuilt image is allowed to carry its own triton.
- **`sgl_kernel` is only probed once SGLang is the engine in play.** Other hosts
  were being warned about a dependency they do not use.

### Framework resolution (the part that made a preflight pass useless)

`resolve_installed_framework()` still hardcoded sglang/vllm, so a host that
passed Phase 1 on a third engine ended with `install complete` and a broken
runtime: no `FRAMEWORK` in `.env`, any stale `FRAMEWORK` from a previous image
left in place, and a next-step prompt naming an engine that is not installed. It
warned but never escalated, so the operator walked into the wrong downstream
path.

- `resolve_installed_framework()` walks `$FRAMEWORKS` — the same list Phase 1
  probes — so the two cannot drift apart again. Order is preserved, so an
  existing sglang host still resolves to sglang.
- New `framework_probe_python()` is the single place that knows vLLM lives in
  its own venv under `FRAMEWORK_ENV=isolated`. Preflight, resolution and the
  profiler hotfix all route through it instead of repeating the rule three
  times.
- `write_runtime_dotenv()` **removes** `FRAMEWORK` when nothing is importable,
  instead of leaving a stale value behind.
- `print_next_steps()` names the engine actually detected instead of defaulting
  to sglang whenever `--install-framework` is none.

### Phase 3 profiler hotfix

`rocm_profiler_hotfix_compatible()` gated on sglang/vllm specifically and
skipped otherwise. The hotfix patches the ROCm profiler libraries that
`torch.profiler` loads, which is engine-agnostic, so it now gates on "some
serving framework is present". The ROCm-7.2 eligibility check above it is
unchanged.

Verified on an affected host: `torch 2.10.0+rocm7.2.4`, `hip 7.2.53211` — it
passes the version gate and was previously skipped only for want of
sglang/vllm, i.e. that host was silently losing the profiler fix.

No documentation, examples, or compatibility-matrix changes — deliberately kept
out of this PR.

## Test plan

Tests drive the real functions rather than matching source text: the installer
is sourced with its `main` invocation stripped and run against a fake host, and
the assertions read the resulting `.env` and the printed prompt.

- [x] `pytest test_setup_cli.py` → **53 passed** on Linux (ROCm MI355X host).
- [x] **Negative control**: run the same suite against the previous revision of
      this branch — 5 of the 6 new tests fail there and pass here. The sixth is
      `test_baremetal_framework_resolution_keeps_sglang_precedence`, which passes
      both ways by design (it is the no-regression guard).
- [x] `bash -n install_baremetal.sh` → syntax clean.
- [x] End-to-end on a host that ships only a preinstalled engine: setup
      completes, and a subsequent `optimize` run went through the full
      PRELUDE → EXPLORE → SWEEP → CLOSE lifecycle with `crash_count=0`, a
      validated throughput gain recorded, and the accuracy gate holding.

## Notes

Two Windows-only test failures in this file predate this branch (the bash
subprocess tests cannot resolve paths under PowerShell); they pass on Linux.

### DIFF
--- a/docs/compatibility.rst
+++ b/docs/compatibility.rst
@@ -151,9 +151,19 @@ accordingly.
      - MI355X
    * - ``rocm/hyperloom:vllm-v0.24.0-rocm7.2.0``
      - MI300X / MI325X / MI355X
+   * - ``rocm/atom:latest`` (``--framework atom``)
+     - MI300X / MI355X
 
 Browse all available tags at
-`hub.docker.com/r/rocm/hyperloom/tags <https://hub.docker.com/r/rocm/hyperloom/tags>`_.
+`hub.docker.com/r/rocm/hyperloom/tags <https://hub.docker.com/r/rocm/hyperloom/tags>`_,
+and the ATOM images at
+`hub.docker.com/r/rocm/atom/tags <https://hub.docker.com/r/rocm/atom/tags>`_.
+
+The ATOM image is published by the `ROCm/ATOM <https://github.com/ROCm/ATOM>`_
+project rather than under ``rocm/hyperloom``. It ships ATOM and AITER but not
+SGLang or vLLM, so bare-metal setup inside it reports those two as missing and
+passes on ``framework atom: OK``. Run setup there with
+``--install-framework none``; ATOM itself is not installable by Hyperloom setup.
 
 Bare-metal recommended environment
 -----------------------------------
--- a/docs/compatibility.rst
+++ b/docs/compatibility.rst
@@ -151,19 +151,9 @@ accordingly.
      - MI355X
    * - ``rocm/hyperloom:vllm-v0.24.0-rocm7.2.0``
      - MI300X / MI325X / MI355X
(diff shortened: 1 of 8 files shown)


ZhengGong-amd (approved):
Approved

Issue #1149, opened 2026-08-11
analytic_ceiling has no num_gpus: analytic_within_pct overstates attainment by the parallel degree on every xDiT session


sushildubey171:
### Target release

Unsure

### Optimization domain

Inference

### Problem / use case

`diffusion_flops.analytic_ceiling` divides a **whole-model** FLOP count by a **single GPU's** peak, so `ideal_ms` is a one-GPU number. Every shipped xDiT session is multi-GPU: `assets/configs/baseline_xdit.yaml:30` is

```yaml
    TP: 8                          # = XDIT_ULYSSES_DEGREE (sequence parallel degree)
```

so the ceiling is compared against a workload whose FLOPs are spread across 8 devices. `analytic_within_pct` inherits the error:

```python
report["analytic_within_pct"] = round(est["ideal_ms"] / (actual_us / 1e3) * 100.0, 2)
```

`ideal_ms` is a whole-model figure while `actual_us` comes from a single rank's trace, so at `TP=8` the reported attainment is roughly 8x too high. This is not specific to video — it applies to the FLUX sessions running on the `xdit` path today.

### Proposed solution

Give `analytic_ceiling` an explicit device count and use it:

```python
def analytic_ceiling(model_dir, *, gpu_type="mi355x", precision="bf16",
                     height=1024, width=1024, num_steps=None, cfg_batch=None,
                     num_gpus: int = 1):
    ...
    est["num_gpus"] = num_gpus
    est["ideal_ms"] = est["total_flops"] / (pk * 1e12 * num_gpus) * 1e3
```

with the call sites in `diffusion_roofline.py` and `tracelens_analysis.py` passing the session's `TP` / `XDIT_ULYSSES_DEGREE`.

The alternative is to leave `ideal_ms` whole-model and label it as such, dividing only where it meets a per-rank measurement. Either is defensible; what matters is that the two sides of `analytic_within_pct` end up on the same basis.

One thing worth deciding explicitly: sequence-parallel attention is not a clean 1/N split of the model's FLOPs. Under Ulysses each rank holds all tokens but only a subset of heads for the attention itself, while the linear projections shard by token. A flat `/num_gpus` is a good first approximation and far better than the current implicit 1, but it will not be exact for attention-dominated workloads.

### Expected impact

Correctness of a reported metric rather than throughput. Today `analytic_within_pct` overstates attainment by roughly the parallel degree on every multi-GPU diffusion session.

### Additional context

[...]

PR #1189, opened 2026-08-14
fix(xdit): analyse the raw capture, not a split fragment

luochen-amd:
## Why

Over the last three days, 48 of 56 xDiT sessions produced no roofline snapshot at
all, and the one model that did produce them produced garbage: an empty ceiling
and a throughput of 0.1 tok/s/GPU. The failing runs all raised the same error:

```
RuntimeError: Trace contains zero GPU kernel events (...); the upstream profile
run captured CPU-only activity. Re-run profile with the torch.profiler GPU
activities enabled (no LD_PRELOAD competing for ROCprofiler-SDK) ...
```

The profiler was not the problem. The GPU data was captured correctly every
time.

## Root cause

`discover_trace_inputs` put the raw capture and the splitter's fragments in the
same priority bucket, leaving alphabetical order to choose between them. From a
real failing FLUX.1-dev roofline directory:

| prio | file | size | gpu kernels |
|---|---|---|---|
| 1 | `decode_only_steady_state_..._rank_0.trace.json.gz` | 938 | **0** ΓåÉ probed |
| 1 | `execution_details.json` | 3965 | no traceEvents |
| 1 | `mixed_steady_state_..._rank_0.trace.json.gz` | 932 | 0 |
| 1 | `rank_0.trace.json.gz` | **910818** | **30196** ΓåÉ wanted |
| 1 | `rank_0.trace_annotation_iteration_{0,1,2}.json.gz` | ~890 | 0 |

`decode_only_...` beats `rank_0...` on the first letter. The CPU-only preflight
opens `trace_files[0]` and nothing else, so it read a 938-byte fragment, counted
zero kernels, and blamed the profiler for a capture with 30k kernel events in the
same directory.

**One selection bug, both symptoms:**

- Fragment holds no kernels ΓåÆ preflight fails ΓåÆ no roofline. 48 of 56 sessions.
- Fragment holds *some* kernels ΓåÆ preflight passes and the analysis then runs on
  the fragment. `ERNIE-Image-Turbo`'s fragment held 512 of its 19,635 kernels, so
  its roofline was computed from 2.6% of the trace: no ceiling, 0.1 tok/s/GPU.
  This is why ERNIE looked like the one model that worked.

## Why #1160 did not help

[#1160](https://github.com/AMD-AGI/Hyperloom/pull/1160) stopped xDiT traces being
split going forward, which was a real fix at a different layer. It is deployed ΓÇö
`stale_framework_overridden` and `--skip-split` are present in the source the
sessions stage from, `roofline.py` was updated two minutes after the merge, and
the merge commit is an ancestor of the `main` the sessions run. It could not help
here because the fragments already on disk still won the sort, and the selection
itself was never what it changed.

## What changed

- Splitter output sorts last, keyed on the containing `trace_split/` directory and
  the annotation-sidecar name rather than on guessing from phase names, so a new
  phase name cannot silently reintroduce this.
- The CPU-only preflight probes candidates in discovery order instead of only the
  first, bounded at 8 opens. A leading fragment can no longer condemn the capture
  behind it, and the error now names every file it probed.

## Test plan

- [x] 4 new tests reproduce the production layout (real filenames and sizes) and
      assert the raw capture leads, that the leading candidate is the one holding
      the kernels, and that the pre-existing merged-trace preference is unchanged
- [x] Verified the new tests fail with the fix reverted ΓÇö all 3 ordering
      assertions flip
- [x] Full file: 9 failed / 208 passed before, 9 failed / 212 passed after. The 9
      are pre-existing Windows path artifacts in unrelated `resolve_launcher` /
      `grep_for_keyword` tests
- [ ] Confirm on the next xDiT roofline run that `trace_gpu_kernel_events` logs a
      non-zero count and a ceiling reaches the dashboard

## Note for whoever picks this up

Sessions stage Hyperloom from `/shared_nfs/haiskong/Hyperloom` rather than a
pinned SHA ΓÇö the logs say `using existing Hyperloom source (no clone)` and carry
no source SHA. That checkout needs updating for this to reach running sessions,
and it also means there is no record in any session of which code it ran.

## Decision: size ordering applies to every capture, not only xDiT

`_trace_input_sort_key` is shared, so this changes rank selection beyond the
workload being fixed. For a multi-rank capture with no `merged-` trace -- where
`_preferred_main_trace_path` falls through and hands the directory to discovery --
the analysed rank moves from "rank_0, alphabetically" to "the largest rank".

That is deliberate. Alphabetical order carries no information about which rank
recorded a usable window, whereas the largest file is the one most likely to hold
a complete one, and the preflight already has to reject empty candidates either
way.

Two consequences worth stating plainly rather than discovering later:

- Which rank represents a run can now differ between two analyses of the same
  session, because rank sizes vary between captures. Previously it was always
  rank_0.
- This is a behaviour change for frameworks other than xDiT, even though nothing
  about them was broken. xDiT captures hold a single `rank_N.trace.json.gz`
  (measured: 276 of 276 directories), so the workload this PR fixes is not
  affected by it at all.

If per-rank stability matters more than picking the fullest rank for some
framework, the ordering should become conditional rather than shared -- but that
is a different change from this one, and nothing observed so far needs it.

### DIFF
--- a/src/hyperloom/agents/kernel/tests/test_tracelens_csv.py
+++ b/src/hyperloom/agents/kernel/tests/test_tracelens_csv.py
@@ -1945,6 +1945,85 @@ def test_annotation_sidecars_sort_after_a_capture_even_outside_trace_split(tmp_p
     assert traces[-1] == sidecar
 
 
+def test_fragments_flat_beside_the_capture_are_still_demoted(tmp_path):
+    """The demotion must not depend on the splitter nesting its output.
+
+    Production nests fragments under trace_split/ today, but keying only on the
+    directory would leave a flat layout exactly as broken as before: the phase
+    names sort ahead of rank files on the first letter.
+    """
+    trace_dir = tmp_path / "torch_trace"
+    trace_dir.mkdir()
+    raw = _rank_trace(trace_dir / "rank_0.trace.json.gz", kernels=40)
+    flat_fragment = _rank_trace(
+        trace_dir / "decode_only_steady_state_prefill_0_decode_1_bs1_conc1"
+                    "_rank_0.trace.json.gz", kernels=0)
+
+    _kind, traces = tla.discover_trace_inputs(trace_dir)
+
+    assert traces[0] == raw
+    assert traces[-1] == flat_fragment
+
+
+def test_an_eight_rank_flat_capture_does_not_exhaust_the_probe_budget(tmp_path):
+    """xDiT runs at TP=8, so a flat layout could present eight fragments first.
+
+    With the fragments still in the default bucket they would consume the whole
+    probe budget before any rank file was opened, and the run would fail with the
+    error this change exists to remove.
+    """
+    trace_dir = tmp_path / "torch_trace"
+    trace_dir.mkdir()
+    for rank in range(8):
+        _rank_trace(
+            trace_dir / f"decode_only_steady_state_x_rank_{rank}.trace.json.gz",
+            kernels=0)
+        _rank_trace(
+            trace_dir / f"mixed_steady_state_x_rank_{rank}.trace.json.gz",
+            kernels=0)
+    captures = [_rank_trace(trace_dir / f"rank_{r}.trace.json.gz", kernels=25)
+                for r in range(8)]
+
+    _kind, traces = tla.discover_trace_inputs(trace_dir)
+
+    leading = traces[:tla._KERNEL_PROBE_LIMIT]
+    assert any(p in captures for p in leading), (
+        "a real capture must be reachable inside the probe budget")
+    assert tla.count_gpu_kernel_events(traces[0]) > 0
+
+
+def test_a_non_trace_sidecar_does_not_lead_discovery(tmp_path):
+    """execution_details.json is swept in by the *.json glob but is not a trace.
+
+    It sorts ahead of rank_0.trace.json.gz alphabetically, so before size
+    ordering it was trace_files[0] in every healthy nested capture: one wasted
+    probe, and a promotion logged on every run, which made the log line
+    meaningless exactly when it should have meant something.
+    """
+    trace_dir = tmp_path / "torch_trace"
+    raw = _xdit_roofline_capture(trace_dir)
+    sidecar = trace_dir / "trace_split" / "execution_details.json"
+    assert sidecar.exists()
+
+    _kind, traces = tla.discover_trace_inputs(trace_dir)
+
+    assert traces[0] == raw
+    assert traces.index(sidecar) > 0
+
+
+def test_the_larger_trace_leads_within_a_bucket(tmp_path):
+    """Size is the part of the ordering that needs no name recognition."""
+    trace_dir = tmp_path / "torch_trace"
+    trace_dir.mkdir()
+    small = _rank_trace(trace_dir / "aaa_unknown_shape.trace.json.gz", kernels=1)
+    large = _rank_trace(trace_dir / "zzz_unknown_shape.trace.json.gz", kernels=400)
+    assert large.stat().st_size > small.stat().st_size
+
+    _kind, traces = tla.discover_trace_inputs(trace_dir)
+
+    assert traces[0] == large
+
+
 def test_merged_trace_still_wins_over_a_raw_rank_capture(tmp_path):
     """The pre-existing preference is unchanged: merged first when present."""
     trace_dir = tmp_path / "torch_trace"
@@ -2203,12 +2282,18 @@ def _drive_main_over_capture_dir(tmp_path, trace_dir):
     return captured
 
 
-def _rank_trace(path: Path, kernels: int) -> Path:
+def _rank_trace(path: Path, kernels: int, cpu_events: int = 0) -> Path:
+    """Write a rank trace with the given number of GPU kernels.
+
+    ``cpu_events`` pads with host-side events, which is what a CPU-only capture
+    actually looks like: a large file with no kernels in it.
+    """
+    events = [{"cat": "kernel", "name": "void real_kernel<...>", "dur": 5.0}
+              for _ in range(kernels)]
+    events += [{"cat": "cpu_op", "name": f"aten::some_host_op_{i}", "dur": 1.0}
+               for i in range(cpu_events)]
     with gzip.open(path, "wt") as fh:
-        json.dump({"traceEvents": [
-            {"cat": "kernel", "name": "void real_kernel<...>", "dur": 5.0}
-            for _ in range(kernels)
-        ]}, fh)
+        json.dump({"traceEvents": events}, fh)
     return path
 
 
@@ -2222,11 +2307,13 @@ def test_analysis_input_follows_the_candidate_that_passed_the_preflight(tmp_path
     """
     trace_dir = tmp_path / "torch_trace"
     trace_dir.mkdir()
-    empty = _rank_trace(trace_dir / "rank_0.trace.json.gz", kernels=0)
+    # A CPU-only rank is a big file with no kernels in it, so size ordering puts
+    # it first on merit. Ordering cannot help here; only the promotion can.
+    empty = _rank_trace(trace_dir / "rank_0.trace.json.gz",
+                        kernels=0, cpu_events=400)
     populated = _rank_trace(trace_dir / "rank_1.trace.json.gz", kernels=12)
+    assert empty.stat().st_size > populated.stat().st_size
 
-    # Discovery still leads with rank_0: the promotion, not the ordering, is what
-    # this test is about.
     _kind, traces = tla.discover_trace_inputs(trace_dir)
     assert traces[0] == empty
 
--- a/src/hyperloom/agents/kernel/tests/test_tracelens_csv.py
+++ b/src/hyperloom/agents/kernel/tests/test_tracelens_csv.py
@@ -2229,12 +2229,15 @@ def _find_splitter_cmd(captured):
     )
 
 
-def _drive_main_over_capture_dir(tmp_path, trace_dir):
+def _drive_main_over_capture_dir(tmp_path, trace_dir, extra_argv=None):
     """Drive tla.main() with a capture *directory* and capture subprocess argvs.
 
     A sibling of :func:`_drive_main_capturing_subprocess`, which always passes a
     single file. Multi-rank selection only shows up when discovery has more than
     one candidate to choose from.
+
+    ``extra_argv`` appends CLI flags, which is how the ``--skip-split`` route
+    that scriptable workloads actually take gets exercised.
     """
     import os as _os
     from unittest.mock import patch
@@ -2267,6 +2270,7 @@ def _drive_main_over_capture_dir(tmp_path, trace_dir):
         "--budget-minutes", "1",
         "--no-llm-orchestrator",
     ]
+    argv += list(extra_argv or [])
 
     env_backup = dict(_os.environ)
     try:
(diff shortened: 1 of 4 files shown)


luochen-amd:
## Verified against production, not just fixtures

Ran the old and the new selection over every xDiT roofline capture directory
written in the last three days on crusoe — 257 directories with trace files — and
compared what the CPU-only preflight would decide for each.

| | preflight passes | preflight fails |
|---|---|---|
| before | 21 | **236** |
| after | **257** | 0 |

Every one of the 236 failures was the same shape: the leading candidate was a
~900-byte `decode_only_steady_state_..._rank_0.trace.json.gz` fragment with no
kernel events, while `rank_0.trace.json.gz` sat in the parent directory with the
real capture in it.

`ERNIE-Image-Turbo` is among the rescued directories, which confirms the second
half of the diagnosis: it was passing the preflight on a fragment that happened
to hold 512 of its 19,635 kernel events, and its roofline was therefore computed
from 2.6% of its own trace. That is the empty ceiling and the 0.1 tok/s/GPU on
the dashboard, not a metric-units problem.

### One correction to my own measurement

A first pass reported 38 `HiDream-I1-Fast` directories as still failing. That was
an artefact of the verification script, not the fix: it capped its scan at 64 MB
of decompressed bytes, and HiDream's captures are 32 MB gzipped with the first
kernel marker at ~174 MB decompressed (FLUX's sit at ~5 MB). Re-scanned with no
ceiling, they carry kernel events like the rest. The shipped code parses the file
rather than scanning a prefix, so it was never subject to this.

Worth knowing for review: HiDream traces being ~35x larger than the other models'
means `count_gpu_kernel_events` parses ~175 MB for them. That was already true
before this change for whichever file sorted first; the ordering fix means the
first file opened is now the one that answers the question, so the probe loop
exits after one file in the normal case. `_KERNEL_PROBE_LIMIT = 8` bounds the
pathological case where a capture genuinely has no kernels.

chaojhou:
I can corroborate the diagnosis independently — I was reading `_trace_input_sort_key` earlier today for an unrelated triage, and the four branches on `main` really do drop both `decode_only_steady_state_..._rank_0.trace.json.gz` and `rank_0.trace.json.gz` into the same `return (1, name)` bucket. Neither starts with `merged-`, neither matches `TP-\d+-DECODE`, neither starts with `bs_`/`graph_capture`. So alphabetical order decides and `d` beats `r`, exactly as you describe. Worth noting *why* the existing demotion missed it: the priority-2 rule was clearly written to demote phase shards, but it only matches the vLLM/SGLang `177-TP-0-DECODE` naming, not the splitter's `decode_only_steady_state_...` naming. Keying on the directory instead of the phase name is the right correction, and the reasoning in the docstring is good.

The second commit is the part I'd call out as unusually good practice. Probing several candidates and then still handing `trace_files[0]` to the analysis would have been strictly worse than the bug it replaced — a silent wrong input instead of a loud wrong error — and you caught it, fixed it, and logged the promotion so a run that switched inputs is explicable afterwards. That is the failure mode this class of fix usually ships with.

One thing I would like resolved before merge, and it turns on a fact I cannot check from here.

## The ordering half only fires if the fragments really are under `trace_split/`

`_is_derived_trace` demotes on two conditions: a `trace_split` component in `path.parts`, or a `trace_annotation_iteration_\d+` name. Your evidence table is a flat listing with no directories, so I could not tell which layout production has. I ran your production filenames through both the old and the new key:

**Fragments nested under `trace_split/`** (the layout `_xdit_roofline_capture` builds): works as intended. `rank_0.trace.json.gz` leads after `execution_details.json`, the preflight promotes on probe 2, the fragments sort to priority 4.

**Fragments flat beside the capture**: the demotion never fires — `decode_only_steady_state_..._rank_0.trace.json.gz` has no `trace_split` component and does not match the sidecar regex, so it stays at priority 1 and still leads. Ordering is unchanged from `main`, and the whole fix then rests on the probe loop.

That second case is not hypothetical for this workload. xDiT ships at `TP: 8` (`profile_xdit.yaml:18`), so a capture has eight ranks. Simulating a flat 8-rank directory against the new key:

```
  candidates=25  probed=8/8
  verdict: FAILS: 'zero GPU kernel events' (limit exhausted)
      1. prio=1 decode_only_steady_state_x_rank_0.trace.json.gz   kernels=0
      2. prio=1 decode_only_steady_state_x_rank_1.trace.json.gz   kernels=0
      ...
      8. prio=1 decode_only_steady_state_x_rank_7.trace.json.gz   kernels=0
      9. prio=1 execution_details.json                            kernels=0
     10. prio=1 mixed_steady_state_x_rank_0.trace.json.gz         kernels=0
```

Eight `decode_only_*` fragments exhaust `_KERNEL_PROBE_LIMIT` before any `rank_N.trace.json.gz` is reached, and the run fails with the same "zero GPU kernel events" error this PR exists to remove.

This also inverts the comment above the constant:

```python
#: How many discovered files the CPU-only preflight will open before giving up.
#: Each probe decompresses a trace, so this is a cost ceiling rather than a
#: correctness one -- the raw capture sorts near the front, and a directory that
#: needs more than a handful of probes has a real capture problem.
_KERNEL_PROBE_LIMIT = 8
```

"The raw capture sorts near the front" is a consequence of the demotion predicate being complete, not an independent property. Where the predicate does not fire, 8 becomes a correctness ceiling.

Your 257-directory production run is strong evidence and I am not disputing it — but it measures the two halves together, so a result of 257/257 is consistent with either half doing the work. If the fragments are nested, everything above is moot and I would just suggest saying so in the PR body, because the flat table invites the misreading. If any layout has them flat, the options are to demote on the name shape as well as the directory (`^(decode_only|mixed|prefilldecode)_steady_state_.*_rank_\d+`), or to stop leaning on the probe limit — sorting by descending file size within a priority bucket would put the 910 KB capture ahead of the ~900-byte fragments regardless of naming or directory, and costs nothing.

## Two smaller things

**Probe cost on the failing path.** `count_gpu_kernel_events` calls `open_json`, which parses the whole file before counting, and `max_events` only bounds the counting loop. On HiDream captures — which by your own measurement decompress to ~175 MB — a directory that genuinely has no kernels now parses up to eight of those before raising, where it previously parsed one. Bounded and worth the correctness, but it is a real change in how long a failing roofline takes to fail, and the "exits after one file in the normal case" claim is optimistic even in the good layout: `execution_details.json` sorts before `rank_0.trace.json.gz`, so the normal case is two probes and the promotion log line fires on every run.

**`execution_details.json` is not a trace.** It is swept in by the `*.json` glob in `discover_trace_inputs`, has no `traceEvents`, and consequently becomes `trace_files[0]` in the nested layout. Excluding non-trace JSON at discovery would remove a wasted probe and stop the promotion path from triggering on every healthy run, which would make the log line mean something when it does appear.

## Note

Unrelated to this PR, but relevant to anyone reading the resulting rooflines: the analysis still runs on a single rank, so for a TP=8 xDiT session the per-kernel totals cover one rank's worth of work. That is the basis mismatch behind [#1149](https://github.com/AMD-AGI/Hyperloom/issues/1149), and it is independent of this fix — but the two compound on the same `analytic_within_pct` number, so a session that looked wrong for the reason you found here may still look wrong afterwards for that one.

On the staging note in your description: agreed that a checkout with no recorded SHA is the more serious operational problem, since it means no session can say which code it ran. Worth its own issue rather than a line in this PR.

luochen-amd:
Thanks — the flat-layout point was the right thing to press on, and the `_KERNEL_PROBE_LIMIT` comment was doing exactly what you said: asserting a property that only held because the predicate happened to be complete.

## The layout question, measured

Walked every xDiT roofline capture directory written in the last four days on crusoe:

| | dirs |
|---|---|
| fragments nested under `trace_split/` | **276** |
| fragments flat beside the capture | **0** |
| capture dirs holding exactly one `rank_N.trace.json.gz` | 276 |
| capture dirs holding eight | 0 |

So the flat 8-rank case is not what was breaking us — despite `TP: 8`, the capture lands as a single `rank_0.trace.json.gz` rather than one file per rank. Your simulation was right about what *would* happen; it just isn't the shape production writes today.

That does not make it moot, and I have taken both of your suggestions rather than only documenting the layout:

- Phase fragment names are now demoted wherever they sit (`^(decode_only|mixed|prefill_only|prefilldecode)\w*_steady_state\w*_rank_\d+`), not only under `trace_split/`.
- Within a bucket, largest file first. As you say it costs nothing, and it is the only part of the ordering that needs no name recognition — a fragment shape nobody has seen yet still sorts behind a 910 KB capture.

`_KERNEL_PROBE_LIMIT`'s comment now says what actually makes the ordering hold, and stops claiming to be a pure cost ceiling.

## One correction, on `execution_details.json`

It does not reach `trace_files[0]` on this branch, and the promotion does not fire on healthy runs. It lives *inside* `trace_split/`, so the directory rule from the first commit already demotes it to priority 4 — the flat listing in my table was misleading about that too, and I should have shown the tree.

The production re-run bears it out. Same capture, patched selection:

```
trace_gpu_kernel_events=30112 (probed=rank_0.trace.json.gz=30112)
```

One probe, first candidate, no promotion logged. So "the normal case is two probes and the promotion line fires on every run" doesn't happen — though it would have if the sidecar sat beside the capture, which is another thing size ordering now covers.

## The probe cost on the failing path

Agreed, and it is a real regression in how long a failure takes. A directory that genuinely has no kernels now parses up to eight candidates, and on HiDream captures that is ~175 MB each. Bounded, but not free. `count_gpu_kernel_events` parsing the whole file before counting is the underlying cost — `max_events` bounds the loop, not `open_json` — and a streaming presence check would make both this path and the healthy one much cheaper. I would rather not fold that into this PR; happy to open it separately if you agree it is worth doing.

## On #1149

Useful, thank you — and it explains something I could not account for. The single-rank basis means the ERNIE roofline will still be wrong after this lands, just wrong for your reason instead of mine. I had put its empty ceiling down entirely to the 2.6% fragment; if the analysis only ever covers one rank of eight, that is a second factor on the same number.

## On the staging note

Agreed, that belongs in its own issue rather than this PR body. I will file it.

chaojhou:
Re-reviewed at `b5cf757f9`. Both points are resolved, and you took my simulation and turned it into a regression test, which is the outcome I was hoping for rather than the one I expected.

Taking your correction first: you are right that `execution_details.json` never reached `trace_files[0]` on the previous head, because it sits inside `trace_split/` and the directory rule already demoted it. I read that off the flat table and inferred a layout that was not there. It changes nothing now — `test_a_non_trace_sidecar_does_not_lead_discovery` covers the case whether or not production nests it — but the inference was mine to get wrong and I did.

The layout measurement is the useful part of your reply, and the detail that matters most is one I would not have guessed: 276 of 276 directories hold exactly one `rank_N.trace.json.gz` despite `TP: 8`. That is why the flat 8-rank scenario was never what was breaking you, and it is worth having in the record — the capture shape does not follow the parallel degree, so reasoning about rank count from `TP` is a trap for the next person.

Checked the new code rather than taking the commit message for it.

**The name predicate holds on the real shapes.** I ran `_PHASE_FRAGMENT_RE` against the production filenames and a set of near-misses: all four fragment forms match, and `rank_0.trace.json.gz`, `merged-177.trace.json.gz`, `177-TP-0-DECODE.trace.json.gz`, the annotation sidecar and `execution_details.json` all correctly do not. `\w*` being able to match empty means `decode_only_steady_state_rank_0.trace.json.gz` with no middle segment also matches, which is what you want. The one shape it does not catch is a fragment ending `.json.gz` without the `.trace` infix — covered by the directory rule today, and by size ordering regardless, which is precisely the argument for having all three.

**Size ordering degrades correctly.** `_trace_file_size` returning 0 on `OSError` means a vanished or unreadable file sorts to the back of its bucket rather than the front, and keeping `name` as the third element means equal-size files still order deterministically. Both are the right way round; neither is obvious enough to be accidental.

**The bucket priorities are unaffected**, so `merged-` still wins outright and the `TP-\d+-DECODE` demotion is unchanged — `test_merged_trace_still_wins_over_a_raw_rank_capture` pins the first and I checked the second by hand.

`test_the_larger_trace_leads_within_a_bucket` is the test I would have asked for: naming the files `aaa_unknown_shape` and `zzz_unknown_shape` means it fails if size ordering is dropped and name ordering silently takes over, which is the regression that would otherwise reappear quietly.

## One thing to note rather than change

Size ordering reaches past xDiT. `_trace_input_sort_key` is shared, so for any multi-rank capture that has no `merged-` trace — a serving framework where `_preferred_main_trace_path` falls through and hands the directory to discovery — the analysed rank changes from "rank_0, alphabetically" to "the largest rank". I think that is the better choice on the merits, since the largest rank is the one most likely to carry a complete window. But it is a silent change in which rank represents the run, in shared code, and rank sizes vary between runs, so re-analysing the same session could land on a different rank than the first pass did. Worth a line in the PR body so it reads as a decision rather than a side effect. Your measurement says xDiT captures have a single rank file, so this is not about the workload you are fixing.

## On the follow-ups

Agreed on both, and agreed they belong elsewhere. The streaming presence check is now clearly the right thing to defer: with size ordering the healthy path is one probe, so the cost you would be optimising only lands on directories that are genuinely failing, and it is a change to `count_gpu_kernel_events` semantics that deserves its own tests rather than riding in here. Same for the staging-checkout issue — a session that cannot say which code it ran is a bigger problem than this PR, and burying it in this description would lose it.

No further comments from me. Once e2e lands this looks ready.

chaojhou (approved):

luochen-amd:
Thanks — this found real gaps, including two places where the change reintroduced the very thing it was written to remove. All of it is fixed in `2d66209ab`. Taking them in your order.

**(a) The same-source bug in `profile.py`.** Agreed this was the important one, and agreed on the framing: `_is_capture_trace` already establishes that a derived slice must never be promoted as the primary trace, so this is the same reason one directory along. `_trace_files_for_dir` now excludes `trace_split/` and returns largest-first. The size ordering matters as much as the exclusion — the fallback consumers you point at (`writeback.py:2705` and `roofline.py::_extract_trace_path`) take `trace_files[0]`, and alphabetical order is what made a 938-byte chunk beat a 910 KB capture. Three tests: the exclusion, the ordering, and the anchoring from (b).

You're right that this is latent rather than live today, but the failure mode is worth naming: a chunk reaching `--trace-input` as a single file takes the `("file", [path])` branch, and at that point every mitigation in this PR is bypassed. Better excluded at the source.

**(b) `_is_derived_trace` over absolute parts.** Correct, and worse than a cosmetic issue: with every candidate in bucket 4, ordering falls back to filename and the original bug reproduces exactly — from nothing but a directory name. Now bounded to the capture root, in both files. `test_capture_under_an_ancestor_named_trace_split_still_orders_correctly` puts the capture under `.../trace_split/run/torch_trace/` and names it so it also sorts last alphabetically, so it fails if either the anchoring or the ordering regresses.

**(c) Size within the bucket.** This landed in `b5cf757f9`, which I think predates your pass — `_trace_input_sort_key` returns `(priority, -size, name)`. I agree with the reasoning and it's why I kept the naming rules as a supplement rather than the mechanism, and why the new `profile.py` ordering is size-based too. Thanks for pointing at `max(trace_files, key=_safe_size)` as the in-repo precedent; that's a better argument for it than the one I'd written down.

**(d) Production takes the untested branch.** Fair, and this was the gap that mattered most in the test set: asserting on splitter argv covers a branch these sessions never enter. `test_skip_split_route_analyses_the_promoted_candidate` drives `--skip-split --analysis-route deterministic`, per `request_handlers`. The path is only observable by intercepting `_run_deterministic_tracelens_steps`, since that route runs in-process and never puts the trace on a subprocess argv — which is presumably why it had no coverage.

**(e) mypy.** Both errors traced to the dead `else None`, so the redundancy fix below removed them: 34 → 32 on the file locally, and no `Path | None` diagnostics remain. Left `ruff format` alone since both files fail it on base and reformatting would bury the diff; `ruff check` is clean.

**Redundancy: `if trace_files else None`.** Right on all three counts, and fixed as one line. `discover_trace_inputs` raises `FileNotFoundError` for an empty directory and returns `[trace_input]` for a file, so the list is never empty; the guard bought a `Path | None` that then had to be re-checked downstream. Left the outer `and trace_files` alone as you suggested.

**Probe cost on the failing path.** Agreed the honest note wasn't enough — the cost lands where the failure already is. Two things now bound it. Size ordering means a healthy capture answers on the first probe, so the happy path is unchanged; and the loop stops on a cumulative byte budget (512 MiB), always after at least one probe, recording `(stopped after N bytes probed)` so a truncated probe set is visible rather than indistinguishable from a complete one. I did not switch to a streaming count in this PR — that changes the shared `count_gpu_kernel_events` contract and deserves its own change.

**Silent failure 1: corrupt vs CPU-only.** This is the one I'm most glad you caught, because the promotion turned a pre-existing wart into an active misdirection, and in exactly the direction the PR exists to fix. `_count_kernels_if_readable` now separates them, and the message no longer asserts a cause it hasn't established. I kept `count_gpu_kernel_events` as the single place kernel events are recognised — the disambiguating parse runs only on a zero answer, which is the only ambiguous one, so nothing is parsed twice on the healthy path. (Keeping that seam also turned out to matter: bypassing it broke `test_deterministic_main_fails_before_high_idle_gate`, which patches it.)

**Silent failure 2: promotion not in any artifact.** You're quoting my own commit message back at me and the criticism holds — `append_log` does not make a run explicable downstream. Now emits a `trace_analysis_input_promoted` trace-health warning carrying `leading_candidate`, `analysed` and the full probe record, plus `artifacts["tracelens_analysis_input"]`. The probe record is what makes the warning diagnostic rather than decorative: it distinguishes `rank_0=0` from `rank_0=unreadable`. Two tests, one asserting it appears on promotion and one asserting it stays absent otherwise, since an informational warning that fires on every healthy run costs the Coordinator the signal.

**`discover_capture_folder` not following the promotion.** Fixed; it now receives the analysed trace. Indifferent same-directory, wrong across directories, as you say.

Full file is 223 passed / 9 failed locally, against 218 / 9 on base — same nine failures, all Windows-only path and launcher assertions, plus the five new tests. `profile.py`'s test module can't be collected on Windows (`fcntl`), so I verified those three against the real function directly and they'll run in CI.

PR #1190, opened 2026-08-14
fix(kernel): rewire fusion onto kernel-agents forge-fuse

ZhengGong-amd:
## Summary

KernelForge absorbed `forge_fusion` into `kernel_agents` and deleted the
`forge-fusion` console script. This points Hyperloom's fusion tool at the new
entry, removes the code that only existed because fusion was a separate package,
and fixes a pre-existing gap that was discarding a kept compile-pass claim.

**Merge order: the KernelForge PR lands first.** Branch
`feat/zgong/fusion-as-fellow` on `AMD-BRAIN-Internal/KernelForge`. Until it does,
`kernel-agents forge-fuse` does not exist and the availability probe here
correctly reports fusion unavailable.

Options, environment variables, the stdout sentinel, the `fusion_manifest.json`
schema and exit code 3 are all unchanged on the KernelForge side, so only the
invocation moved.

## The three commits

**`fix(kernel): point the fusion tool at kernel-agents forge-fuse`**

`python -m forge_fusion.cli run` becomes `python -m kernel_agents.cli forge-fuse`.

The availability probe mattered more than the invocation. Both of its prongs -- a
PATH lookup for a console script that no longer exists, and a `find_spec` on a
module that moved -- are now permanently false, and the failure is silent:
`run_fusion_handler` returns `"failed"`, which is not in the skip-set
`_fusion_required_before_kernel_opt` consults, so every KERNEL re-entry retries
and fails again without ever stopping the session. Fusion would simply never run.

The probe now tests `kernel_agents.fusion` rather than `kernel_agents`, so a
KernelForge predating the absorption is reported unavailable instead of failing
later inside the subprocess. The PATH prong is gone: the tool is invoked through
`sys.executable -m`, so PATH never decided anything.

**`refactor(kernel): drop the fusion wrapper's compatibility code`**

`--no-author` and `--no-validate` were never emitted -- the orchestrator's payload
builder does not set either key. `_git_toplevel` and `_package_root` re-derived,
on this side of the boundary, the export root KernelForge already reports; they
were a fallback for a missing `artifacts.repo_root`, and KernelForge sets that
field exactly when it sets a patch.

**`fix(kernel): stop discarding a kept compile-pass claim`** (pre-existing bug)

When the chain is already covered by a framework compile pass that shipped
switched off, KernelForge flips the default, runs a serving A/B and exports the
diff -- *instead of* running the authoring loop, so `compile_pass` is populated
while `fusion_loop` and `validation` are null. `_normalize_manifest` only read
the latter two, so a successful claim came back as `no_improvement` / `REVERT`
and its patch was dropped. That is the cheapest win available, one where the
framework already wrote the kernel.

The function now branches explicitly on which path ran. A claim reports no env
flags (the win is a flipped default that the patch carries), and its serving
tok/s ratio rides in `kernel_speedup` with `serving_speedup` and
`compile_pass_flag` alongside so the number's origin is legible.

## Test plan

- [x] `PYTHONPATH=src pytest src/hyperloom/agents/kernel/tests/ src/hyperloom/inference_optimizer/tests/test_kernel_request_handlers_units.py` -- 2159 passed
- [x] New cases cover a kept and a non-kept compile-pass manifest
- [x] `ruff check` clean on every touched file
- [ ] Five failures in that run (`TestForgeCollectiveCoverage` x2,
      `test_install_env_credentials` x2, `test_tracelens_csv` x1) reproduce
      unchanged on pristine `origin/main` and are unrelated to this change

## Note for deployment

`_incontainer.sh` lives in neither repository. If it still runs
`pip install <KernelForge>/src/forge_fusion`, that install now fails: the
directory and its sub-pyproject were removed. It should install the KernelForge
root instead, which is what `ensure_kernel_agents()` in `install.sh` already does.


chaojhou (commented):
Reviewed against `e4f250d`, cross-checked with KernelForge #9 (`c59ce43`).

## Verdict

**Blocked, and not yet verifiable end to end.** The invocation, the option names and the manifest contract line up with what `kernel-agents forge-fuse` actually accepts, and the compile-pass fix is a real bug fix. But KernelForge #9 currently fails before any authoring session on the default `--validate` path (two P0s reported there), so nothing on this side can be exercised against it yet. The findings below are the ones that live in this repo.

All 26 checks pass here. The `CONFLICTING` state is textual, not a code problem — see the rebase note at the end.

## Blocking

**1. A KEPT result with no usable artifact is reported as success, and then blocks its own retry.** `_normalize_manifest` sets `status="ok"` / `kept=True` purely from `fusion_loop.kept` or `compile_pass.kept` (`forge_fusion.py:255,318-321`), without checking that `artifacts.patch` names an existing file or that `repo_root` and `source_file` are non-empty. KernelForge's export is best-effort and can legitimately return an empty `FusionArtifacts`. The consequence chain is closed:

- `_integrate_fusion` returns early on missing patch/target (`phases/kernel.py:2646-2648`), so nothing is applied.
- `_fusion_required_before_kernel_opt` treats `last_fusion.status == "ok"` as done (`phases/kernel.py:2622-2623`), so fusion never runs again this session.

A validated fusion is silently dropped and cannot be retried. Suggest validating the artifact contract before claiming success and returning a retryable error class otherwise.

**2. A stale manifest can be read as this run's result.** `_normalize_manifest` is called regardless of the child's exit code (`forge_fusion.py:429-433`), and neither the wrapper nor the handler removes a previous `fusion_manifest.json` first. The workspace is a fixed path per task id (`request_handlers.py:4242`), so a rerun that dies before writing its manifest — an import error, a Click usage error, a crash — re-reads the previous run's KEEP and hands its patch to integrate. Suggest unlinking the manifest before launch plus a freshness check, and accepting only `rc=0`, or `rc=3` together with an `llm_unavailable` manifest from this run.

**3. The install path can leave an old KernelForge in place and report it ready.** `install.sh:857` short-circuits on `import kernel_agents.cli, openai_codex`, and the post-install verification at `install.sh:886` checks the same names. Neither mentions `kernel_agents.fusion`. On a machine that already has a pre-absorption KernelForge, the install is skipped, the check passes, and every fusion request then returns `forge_fusion_not_found` at runtime. This PR updates the comments in that function but not the probe. `test_install_kernel_agents_idempotent.py:128-133` currently pins the old behaviour, so it needs updating alongside.

**4. The default payload trips KernelForge's explicit refusal.** `fuse_all_confirmed` defaults to `True` (`request_handlers.py:4257`) and the wrapper appends `--fuse-all-confirmed` whenever it is truthy (`forge_fusion.py:144-147`). KernelForge raises a `UsageError` for that flag when the recipe set mixes a compile-pass candidate with authored fusions (`command.py:667-674`), and a usage error writes no manifest — which then lands on finding 2. Suggest either defaulting to `False`, or handling the mixed case explicitly (claim the compile pass first, author on a later round).

## Rebase note

The branch is 46 commits behind `main` and the simulated merge shows one real conflict, in `forge_fusion.py`: this PR deletes `_package_root` / `_git_toplevel` while `main` has since added `safe_directory_args` to them. Deleting them is the right call given KernelForge sets `artifacts.repo_root` whenever it sets a patch — but that is exactly the assumption finding 1 asks you to verify at runtime rather than trust.

After rebasing, please keep `main`'s two recent fixes in `request_handlers.py`: the `safe_directory_args` wrapping of the git invocations, and the trailing-newline normalisation before `git apply`.

## Correct as written

The compile-pass fix is a genuine bug fix, and branching on which path ran rather than inferring from a null `validation` is the right shape. Naming the serving tok/s ratio with `serving_speedup` / `compile_pass_flag` alongside `kernel_speedup` keeps the number's origin legible. The `llm_unavailable` handling is careful in a way that matters: guarding on `not kept`, matching the verdict tolerantly, and mapping the outage to a retryable shape rather than into `complete` — which would have satisfied the idempotency gate and skipped fusion for the rest of the session. The process-group teardown and the wrapper timeout grace are right in principle; they just cannot reach a grandchild that starts its own session (reported on the KernelForge side). Command construction is an argv list throughout, with no shell interpolation.

xiaofei-zheng (approved):

PR #1191, opened 2026-08-14
feat(kernel): route mori EP dispatch/combine to the KernelForge vendor playbook

zili-amd:
## Summary

Mori's EP `dispatch`/`combine` kernels are pip-installed compiled binaries with no rewritable source, so `classify_patchability()` used to reject them as `vendor_binary` before any tuning could happen ([KernelForge PR #88](https://github.com/AMD-AGI/KernelForge/pull/88)). This adds a **vendor operator playbook**: a registry that recognizes known vendor operators and routes them to a validated KernelForge task bundle instead of source rewriting.

**Implemented:**
- `vendor_operator_playbooks.json` + `_vendor_operator_playbooks.py` — registry + matcher, disambiguates `dispatch` vs `combine` by trailing method segment, routes to `examples/mori_ep_dispatch_combine/`.
- `tracelens_analysis.py` — `classify_patchability()` checks the registry before rejecting as `vendor_binary`; groups dispatch+combine into one aggregate GPU% target.
- `forge_submit.py` — new `_submit_vendor_playbook()` path with a session-scoped lock so dispatch and combine share **one** forge-loop session instead of running twice.
- Fixed a result-dict field gap that made `kernel_optimization.py` report `PARTIAL`/no-speedup even when forge-loop had validated a real speedup internally.

## Bugs found & fixed during review (4 rounds, 10 total, all with regression tests)

| # | Severity | Issue | Fix |
|---|---|---|---|
| 1 | High | Reused (combine) role lost its optimized artifact (path mismatch) | Aligned `cli_workspace`/`output_dir`; added a physical artifact copy as a safety net |
| 2 | High | Exception after claiming the group lock left it stuck all session | Wrapped post-claim work so failures always release the lock |
| 3 | Medium | `aggregate_gpu_pct` computed but never used for gating | Wired into all 3 gate call sites |
| 4 | Medium | Registry comment overstated what `kb_path` does | Corrected docs (metadata only) |
| 5 | Blocking | Vendor-playbook artifact could reach `apply_kernel_patch`'s legacy full-file-replace and overwrite the real installed operator | Blocked vendor-playbook results from ever reaching auto-integrate |
| 6 | Blocking | Failed results cached for the whole session, no retry | Added a 10-min TTL on failure caching (successes stay cached permanently) |
| 7 | High | A dead lock-holder (SIGKILL/OOM) blocked submissions for up to an hour | Claims now carry `{pid, timestamp, nonce}` and are safely stolen once stale |
| 8 | High | Dispatch and combine could double-count the same forge-loop speedup as two wins | Only the role that actually ran forge-loop is counted |
| 9 | High | `_candidate_haystack()` didn't check `trace_launcher_file`, so graph-captured kernels (the real production trace shape) never matched the registry at all | Added the field + 2 regression tests reproducing the exact real-trace shape |
| 10 | Blocking | (tech-lead review) `_build_hot_kernel_summaries()`'s key whitelist dropped `patch_strategy`/`vendor_playbook_*` fields, so `untried_hot_reusable_kernels()` silently degraded the aggregate/floor gate to bare `gpu_pct` on the real production path (`hot_kernels_top15`) — losing both the split-load pass-through *and* the playbook's own floor enforcement | Added the 4 fields to the whitelist; 2 new regression tests through the real `record_trace_analyze()` entry point (existing tests bypassed the projection and couldn't catch this) |

## Testing

- **Unit**: `test_vendor_operator_playbook_mori.py` 18/18 passing. Full `agents/kernel/tests/` + `inference_optimizer/tests/` suites checked before/after the latest fix: identical pre-existing, unrelated failure set both times (missing `pytest-asyncio` plugin) — no regressions.
- **Live e2e — isolated micro-workload** (MI300X, dispatch/combine-bottlenecked): correctly routed to the vendor playbook; forge-loop found a validated **1.25x** speedup (`combine_zero_copy=True`, tuned block/warp grid), correctness **PASS** against mori's own reference test-suite math, single-session dedup confirmed (combine reused dispatch's cached result in 0.1s). **This is the run that proves mori is actually re-tuned by KernelForge, with a real measured win, end to end.**
- **Live e2e — full production workload** (8× MI300X, DeepSeek-Coder-V2-Lite, TP1/EP8/DP8, `mori_high_throughput`): proves the *automatic* half of the pipeline off a real serving trace, with **zero manual intervention** — TraceLens classified dispatch+combine as `vendor_playbook` on its own (42.7% aggregate GPU%), the orchestrator transitioned `PRELUDE → KERNEL_AGENT`, `forge_submit` resolved the target and spawned a real forge-loop subprocess (fresh git worktree, aiter-cache, `claude-opus-5` implementer/supervisor), and the `combine` candidate correctly deduped onto the same session instead of spawning its own. forge-loop's baseline benchmark then hit a native abort *inside mori's own compiled shmem library* (`ShmemStates::CheckStatusValid()`, ×8 ranks) before a clean timing round could complete — so this run does **not** produce a second measured speedup, but the crash trace itself is further confirmation that mori's real operator construction path was reached (not mocked). Handled safely by the orchestrator (`REVERT`, `crash_count: 0`, no stuck lock) — see "Known follow-up" below.
  - **Net read**: detection → routing → dispatch → dedup is proven automatic on a real production trace; the actual *tuning* result (a measured speedup) is proven only on the isolated micro-workload above, not yet on a live production workload.
- **CI**: all checks green, including `ci-e2e` (a transient CI-infra disk-quota fault on the first run was unrelated to this PR's code — cleared on retest).

## Known follow-up (not blocking)

On the full production run, forge-loop's own baseline benchmark crashed with a native abort inside mori's shmem library (`ShmemStates::CheckStatusValid()`). Handled safely — `REVERT`, `crash_count: 0`, no stuck lock, no false-positive speedup — but this specific task bundle couldn't produce a *measured* speedup on that run.

Root cause is most likely resource exclusivity, not a missing rendezvous: mori's `shmem` is a GPU-initiated **symmetric-memory heap** shared across all ranks on a node, and `driver.py` already does its own complete standalone `dist.init_process_group` + `shmem_torch_process_group_init` — it doesn't depend on an external process group. The live vLLM server being optimized was still serving on the *same* 8 GPUs when forge-loop's driver tried to claim its own independent 8-rank mori heap on top of it; two competing mori-EP groups can't coexist on one 8-GPU box. A real fix means pausing/releasing the live server's GPU allocation around this class of kernel_opt attempt (or routing distributed collective kernels to a phase where the GPUs are guaranteed free) — an orchestrator-level coordination change, not a `driver.py` patch, and risky to rush (a botched pause/resume could hang a live production server, a worse outcome than today's safe no-op). Filed as follow-up rather than patched here; not a regression risk as-is since the pipeline already fails safe.

## Test plan

- [x] Unit tests: `pytest .../test_vendor_operator_playbook_mori.py -q` (18/18)
- [x] Full kernel-agent + inference_optimizer regression suites (no new failures vs. `main`)
- [x] Live e2e, isolated micro-workload (dispatch+combine routed, single session, real speedup)
- [x] Live e2e, full production workload (automatic routing, dedup, graceful failure handling)
- [x] CI `ci-e2e` smoke check green
- [ ] Follow-up: confirm the GPU-exclusivity hypothesis above, then either pause/release the live server's GPUs around distributed vendor-playbook attempts or route them to the Coordinator's collective lane, so routed EP kernels can produce a measured speedup on a live DP+EP serving workload

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -190,6 +190,11 @@ hyperloom = [
 "hyperloom.agents.kernel" = [
     "scripts/*.sh",
 ]
+# Vendor-operator-playbook registry (see tools/_vendor_operator_playbooks.py):
+# read at runtime by classify_patchability()/forge_submit.py, so it must ship.
+"hyperloom.agents.kernel.tools" = [
+    "vendor_operator_playbooks.json",
+]
 # ``agentx/deploy.py`` copies these into InferenceX ``benchmarks/`` at runtime
 # and raises FileNotFoundError when they are absent from the install.
 "hyperloom.inference_optimizer.assets.agentx" = [
--- a/src/hyperloom/agents/kernel/tests/test_vendor_operator_playbook_mori.py
+++ b/src/hyperloom/agents/kernel/tests/test_vendor_operator_playbook_mori.py
@@ -1,0 +1,465 @@
+###############################################################################
+# SPDX-FileCopyrightText: 2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+#
+# See LICENSE for license information.
+###############################################################################
+
+"""Vendor-operator-playbook routing for mori's EP dispatch/combine.
+
+Closes the gap KernelForge PR #88 left explicit: Hyperloom's own
+TraceLens-driven pipeline had no path to mori (a pip-installed compiled
+library, classified ``vendor_binary`` with no rewritable source) even though
+a validated KernelForge forge-loop task bundle exists for it. These tests
+pin, end to end within Hyperloom's own boundary:
+
+1. the registry matcher recognizes mori dispatch/combine by name (
+   ``_vendor_operator_playbooks``);
+2. ``classify_patchability`` + ``_finalize_candidates`` route both candidates
+   to ``reusable_native_kernel=True`` / ``patch_strategy="vendor_playbook"``
+   and sum their GPU share (dispatch+combine are one logical round trip);
+3. ``forge_submit.submit()`` recognizes ``patch_strategy="vendor_playbook"``,
+   copies the KernelForge ``examples/mori_ep_dispatch_combine/`` bundle into
+   the forge worktree, and invokes forge-loop with the bundle's own
+   driver/config/program.md, the mori KB env knob, and the six documented
+   tunable params as ``--target-functions``;
+4. dispatch and combine are invoked as **one** Forge session, not two: the
+   second candidate to reach ``submit()`` reuses the first's result instead
+   of launching a second forge-loop.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import sys
+from pathlib import Path
+
+import pytest
+
+_TOOLS_DIR = Path(__file__).resolve().parents[1] / "tools"
+if str(_TOOLS_DIR) not in sys.path:
+    sys.path.insert(0, str(_TOOLS_DIR))
+_BACKENDS_DIR = _TOOLS_DIR / "backends"
+if str(_BACKENDS_DIR) not in sys.path:
+    sys.path.insert(0, str(_BACKENDS_DIR))
+
+import tracelens_analysis as tla  # noqa: E402
+import forge_submit  # noqa: E402
+import kernel_optimization as ko  # noqa: E402
+from _vendor_operator_playbooks import (  # noqa: E402
+    match_vendor_operator_playbook,
+    _reset_vendor_operator_playbooks_cache,
+)
+
+_MORI_SITE_PACKAGES_FILE = "/opt/venv/lib/python3.12/site-packages/mori/ops/dispatch_combine.py"
+
+
+@pytest.fixture(autouse=True)
+def _fresh_registry_cache():
+    _reset_vendor_operator_playbooks_cache()
+    yield
+    _reset_vendor_operator_playbooks_cache()
+
+
+def _mori_dispatch_candidate(**overrides) -> dict:
+    candidate = {
+        "kernel_id": "k010",
+        "name": "mori::EpDispatchCombineOp::dispatch",
+        "operation": "dispatch",
+        "duration_us": 700.0,
+        "call_count": 10,
+        "source_file": _MORI_SITE_PACKAGES_FILE,
+        "source_type": "unknown",
+        "library": "mori",
+        "shapes": [],
+    }
+    candidate.update(overrides)
+    return candidate
+
+
+def _mori_combine_candidate(**overrides) -> dict:
+    candidate = {
+        "kernel_id": "k011",
+        "name": "mori::EpDispatchCombineOp::combine",
+        "operation": "combine",
+        "duration_us": 300.0,
+        "call_count": 10,
+        "source_file": _MORI_SITE_PACKAGES_FILE,
+        "source_type": "unknown",
+        "library": "mori",
+        "shapes": [],
+    }
+    candidate.update(overrides)
+    return candidate
+
+
+# --- 1. registry matcher -----------------------------------------------------
+
+
+def test_match_vendor_operator_playbook_matches_mori_dispatch_and_combine():
+    dispatch_match = match_vendor_operator_playbook(_mori_dispatch_candidate())
+    combine_match = match_vendor_operator_playbook(_mori_combine_candidate())
+
+    assert dispatch_match is not None
+    assert dispatch_match["id"] == "mori_ep_dispatch_combine"
+    assert dispatch_match["role"] == "dispatch"
+    assert combine_match is not None
+    assert combine_match["id"] == "mori_ep_dispatch_combine"
+    assert combine_match["role"] == "combine"
+
+
+def test_match_vendor_operator_playbook_ignores_unrelated_kernels():
+    gemm_candidate = {
+        "name": "aiter::gemm_a8w8",
+        "operation": "gemm",
+        "source_file": "/sgl-workspace/aiter/aiter/gemm_a8w8.py",
+        "library": "aiter",
+    }
+    assert match_vendor_operator_playbook(gemm_candidate) is None
+    # "mori" alone (no dispatch/combine marker) must not match either --
+    # the registry requires both an identity marker and a role marker.
+    mori_other = {"name": "mori::shmem::init", "library": "mori"}
+    assert match_vendor_operator_playbook(mori_other) is None
+
+
+# --- 2. classify_patchability + _finalize_candidates -------------------------
+
+
+def test_classify_patchability_routes_mori_dispatch_and_combine():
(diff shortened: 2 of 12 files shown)


chaojhou:
Reviewed the full diff. The design holds up and the nine listed findings are genuinely closed in the code — the deploy refusal is defense-in-depth in both places that matter (`_queue_kernel_keep()` for auto-enqueue, `_fill_integrate_defaults_from_state()` → `integrate_handler()` for an LLM-named `kernel_id`), the claim/steal protocol is sound (`O_CREAT|O_EXCL`, nonce-verified `os.replace`, failure-cache TTL and claim staleness aging out together), and `submit()`'s route check is safe because `kernel_optimization.py` loads candidates off `candidates_path` as full dicts, so `patch_strategy` really does survive to the branch.

One thing to fix before merge.

### The aggregate/floor gate is a no-op on the production path in `untried_hot_reusable_kernels()`

`SharedState._build_hot_kernel_summaries()` builds `summary_entry` as an explicit key whitelist (`shared_state.py`, ~3199-3224), and that is what becomes `last_trace_analyze["hot_kernels_top15"]`. None of `vendor_playbook_aggregate_gpu_pct`, `vendor_playbook_min_gpu_pct_floor`, `vendor_playbook_group_id` or `patch_strategy` is in it. `untried_hot_reusable_kernels()` reads `info.get("hot_kernels_top15") or info.get("hot_kernels")`, and in production `hot_kernels_top15` is always populated, so `effective_hot_kernel_gpu_pct()` / `effective_hot_kernel_min_gpu_pct()` degrade to bare `gpu_pct` / `min_gpu_pct` at that site.

Of the three gate call sites, the two in `_batch_kernel_candidates` are fine — that function reads the candidates artifact off `candidates_path`, which carries the full dicts. Only this one is affected, and it loses both directions of the intended behavior:

- the split-load case the change exists for (7% + 5% clearing a 10% floor together) is still dropped, so a mori group whose members are each under threshold never shows up as owed work; and
- the playbook's own `min_gpu_pct_floor: 10.0` is not enforced either, so with a loosened `HYPERLOOM_KERNEL_OPT_MIN_GPU_PCT` (a small fixture at 1.0, say) the group is reported untried and can burn a whole forge-loop session below its own floor. That is the unsafe direction.

The two new tests can't catch this because `_set_trace()` assigns `state.last_trace_analyze` directly and bypasses the projection. Fix is adding those keys to `summary_entry`, plus one regression test that goes through `record_trace_analyze()` instead of hand-building the dict.

Two small things while you're in that test: `test_untried_hot_kernels_vendor_playbook_group_gated_on_aggregate` only asserts `assert untried`, which passes even if just one member clears the gate — worth asserting the specific kernel ids. And its comment says the two rows "share source_file+name", but the names differ (`::dispatch` vs `::combine`), which is precisely what keeps both members in the list rather than collapsing them in the identity dedup.

Also worth letting `ci-e2e/run` land on 873a610 before merging.

zili-amd:
Fixed in 1a637b8 — thanks for catching this, good find.

Confirmed your read exactly: `_build_hot_kernel_summaries()`'s `summary_entry` whitelist never carried `patch_strategy`/`vendor_playbook_group_id`/`vendor_playbook_aggregate_gpu_pct`/`vendor_playbook_min_gpu_pct_floor`, so `untried_hot_reusable_kernels()` silently degraded to bare `gpu_pct`/`min_gpu_pct` on the real `hot_kernels_top15` production path (the two `_batch_kernel_candidates` call sites were unaffected, as you noted, since those read full dicts off `candidates_path`).

- Added the four fields to `summary_entry`.
- Added two new regression tests through the real `record_trace_analyze()` entry point (one per direction: split-load pass-through, and floor-still-enforced) — verified both fail with a `KeyError` without the fix and pass with it.
- Fixed `test_untried_hot_kernels_vendor_playbook_group_gated_on_aggregate`'s weak `assert untried` to assert the specific kernel-id set (turns out both k010 and k011 surface without task_groups metadata, since neither the group-key nor identity dedup collapses them — verified empirically before asserting).
- Fixed the comment claiming the two rows share `source_file+name` — they do not; the differing names (`::dispatch` vs `::combine`) are exactly what keeps them distinct.

Ran the full `agents/kernel/tests/` + `inference_optimizer/tests/` suites before/after: identical 1612 pre-existing, unrelated failures (missing `pytest-asyncio` plugin) both times — no regressions.

Also waiting on `ci-e2e/run` to land on the new head before merge, per your note.

chaojhou (approved):
Verified the fix in `1a637b871`. Both halves are there: `_build_hot_kernel_summaries()`'s `summary_entry` now carries `patch_strategy` / `vendor_playbook_group_id` / `vendor_playbook_aggregate_gpu_pct` / `vendor_playbook_min_gpu_pct_floor`, and the two new tests go through the real `record_trace_analyze()` entry point — one for the pass direction (7% + 5% clearing the 10% default as 12%) and one for the unsafe direction (the playbook's 10.0 floor still blocking under a loosened `min_gpu_pct=1.0`) — each asserting the projection actually carries the fields, which is exactly the assertion that fails without the change. The weak `assert untried` is now `assert set(untried) == {"k010", "k011"}` and the misleading comment about the rows sharing `(source_file, name)` is corrected.

Also checked the new keys don't disturb the surrounding logic: the `kernel_roofline` inclusion test just below only inspects its own key tuple (`bound_type`, `arithmetic_intensity`, ...), so a `None`-valued vendor field on an ordinary kernel doesn't change which rows land in `kernel_roofline_top15`, and `breakdown/collectors/kernels.py` re-projects explicit keys, so the extra keys are inert downstream.

On the Read the Docs failure — agreed, unrelated to this PR:

- the same commit passes the repo's own Sphinx build (Actions `Docs` run 32183961767, `head_sha 1a637b8719b1…`, `build (sphinx html)` green in 47s running `sphinx-build -b html docs docs/_build/html`);
- the fix commit touches only `shared_state.py` and one test file, and no docs file changed anywhere in the PR;
- RTD succeeded on the four earlier commits of this same branch (125d853b, 2bfba7c1, 873a6104), so neither the docs tree nor `.readthedocs.yaml` is broken here;
- RTD is failing intermittently on unrelated open PRs too (#1222, #1216) while a dozen others pass; and
- the failures return in 34-62s versus 78-101s for the successes, i.e. it is failing early (clone/env/install or a queue rejection), not at the end of a real documentation build.

A re-run of RTD build 4332364 should clear it; worth a glance at that build's "Installing" step if it recurs.

Approving. The four non-blocking items from my earlier pass (`recommend_backends()` still returning `[]` for a `vendor_binary`-typed playbook candidate, `vendor_playbook_independently_counted` having no production reader, both roles consuming a top-N slot, and the breadth of the `"mori"` substring marker now that `trace_launcher_file` is in the haystack) are fine to leave for follow-ups. The one worth filing alongside the GPU-exclusivity item is the deployability gap: with integrate refused, a vendor-playbook KEEP is measured but never shipped, so the tuned launch config has no route to a live server yet.

PR #1194, opened 2026-08-15
feat(breakdown): build the optimization view from what producers recorded

haishuok0525:
## Description: what and why

The optimization read model was rebuilt from `state.json`, which can only re-infer who owned a change from phase timestamps. This switches it to the operations, measurements and adoptions producers recorded at author time, keeping the `state.json` model as the fallback for pre-recorder sessions.

Three commits, each standing on its own:

**`project optimizations from author-time records`** — the new read model, plus the two accounting defects the old one carried:

- The gain chain started from zero, so throughput a session had already gained before its first adopted step was absorbed into that step. Steps now contribute percentage points relative to the baseline and the remainder is reported as `unattributed_gain_pct` instead of inflating a neighbour. On the gemma session this moves `kernel_agent` from 7.490365 to 7.116912 and accounts for the 0.373453pp an unadopted PRELUDE `integrate_patch` had produced.
- A measurement id was derived from its operation, which is stable per kernel, so re-measuring a kernel overwrote the readings an earlier adoption had been decided on. Ids now carry the occurrence the caller holds, adoptions pin the readings they cited, and evidence overwritten before this is reported as `adoption_pinned_stale`.

**`trace the recorder's write path`** — `HYPERLOOM_BREAKDOWN_TRACE=1` logs every write with its section, fragment, entity id, call site on both sides of the SDK, and the fields a merging write changed. Recording is best-effort and every call site swallows exceptions, so a doubted number previously had nothing behind it. Sits below DEBUG, switched on by that variable alone, and costs one level check when off.

**`name a measurement's occurrence by its run, not its values`** — an occurrence was derived from the readings it identified, so anything that moved a value moved the key. GEMM tuning reports its kernel-time ratio when it finishes and is recorded again once end-to-end validation has numbers to add beside it; that second pass moved the key the run's metrics shared and re-filed the untouched ratio under a new id. Breakdown 99751 reports two readings of `best_speedup=1.0577` where the tuner measured once, and the adoption pinned the invented one as its evidence.

An occurrence is now named by the run a reading came off — an attempt id, the benchmark report it was graded on, the stamp the producer put on the entry. Values remain a last resort for callers with no run to name, and only ever the metric's own reading, never a sibling's. Separately, `integration_id` was the default for `kernel_e2e` and is structurally absent on env-only integrates, which name no artifact and so never enter the kernel-patch queue that issues those ids; those are now named by the benchmark they were graded on, recorded relative to the session.

**`redact the write trace before it is emitted`** — the trace line is built from payload values and from the fragment name, and a fragment name comes from an entity id that can carry an error excerpt. Errors are where credentials surface, which is why the code that persists them already redacts; the trace read the same material and did not. It now goes through `redact_secret_values` once per line at the point the fields are joined, so ids and the fragment name are covered along with the values, behind the same level check as everything else.

## Linked issue(s)

None.

Added: a regression that amending a tuning record with e2e numbers consolidates onto the existing reading instead of splitting it, and one that two env-only integrates graded on different benchmarks stay apart. Both fail on the parent commit.

Full suite on this branch, rebased onto main: **21 failed, 13081 passed**. Clean `origin/main` in a separate worktree: **25 failed, 13050 passed**. Every failure on this branch also fails on main — **no new failures**. (The four extra on the baseline are worktree artifacts: they need the untracked `_bypass_repo_scan_fixture/`.)

```
python3 -m pytest -q -m "not critic_agent_e2e and not robustness_agent_e2e and not targeted_build_e2e"
```

Verified on production breakdown 100035 (`GLM-5.2-MXFP4_20260813T181834Z_5bef041e`), built at the head of this branch: the 99751 phantom duplicate is gone, `best_speedup` lands as one validated reading, the eight `conc_sweep` points stay eight, the env-only integrate is named by its benchmark report rather than a null, and gain closes exactly (validated 0.523409 = attributed 0.523409 + unattributed 0.0) with no `adoption_pinned_stale`.

That session exercised the action, sweep, `gemm_tuning` and env-only `kernel_e2e` paths. It did not reach the GEAK route, `micro_speedup` (both kernels died pre-dispatch) or the patch branch of `kernel_e2e`, which rest on unit tests here. Their fallback is the metric's own value, so an unnamed run on those paths degrades to today's behaviour rather than to something new.

## Two red checks, both looked into

**CodeQL** reports one high `py/clear-text-logging-sensitive-data` on the trace's log call. Its flow starts at a string literal named `secret` in `test_record_action_attempt_redacts_secrets_from_persisted_errors` — the test that exists to prove that value gets masked — and follows it through `record_action_attempt` into an entity id, a fragment name, and finally the trace line. The production path redacts at both ends now, and the two tests above cover it, but CodeQL does not model the project's redaction as a sanitizer, so the alert stands. It needs a dismissal rather than another code change.

**Read the Docs** fails while the in-repo `build (sphinx html)` job passes on the same commit. Building the docs locally with RTD's own pinned set (`docs/sphinx/requirements.txt`) succeeds: 21 warnings, all pre-existing myst cross-references in `install.md` and `upgrade.md`, none in the files this PR touches. The failure is on the RTD side.

## Breaking changes

No. The `state.json` model remains the fallback for sessions without recorder fragments, so existing archives project as before.


ZhengGong-amd (approved):
Approved

ZhengGong-amd (approved):
Approved

PR #1195, opened 2026-08-15
Bump the recommended vLLM stack to v0.27.1 (ROCm 7.2.3)

xiaofei-zheng:
## Summary

Moves the recommended vLLM stack from 0.24.0 to 0.27.1 and consolidates every container image reference onto the `rocm/hyperloom` repository.

- **Image**: `rocm/hyperloom:vllm-v0.24.0-rocm7.2.0` -> `rocm/hyperloom:vllm-v0.27.1-rocm7.2.3`
- **Bare metal**: `VLLM_VERSION` default `0.24.0` -> `0.27.1`; `VLLM_ROCM_VARIANT` stays `rocm723`
- **SGLang**: unchanged, since `sglang-v0.5.16-rocm7.2.0-mi300x` / `-mi350x` are still the newest published tags

The slurm example registry also moves off the upstream `vllm/vllm-openai-rocm:v0.24.0` image, so every image documented in the repo is now a `rocm/hyperloom` ref.

### Why the doc narrative changed

Before this PR the vLLM Docker image was ROCm 7.2.0 while bare metal resolved to 7.2.3, and `compatibility.rst` carried a paragraph explaining that mismatch. The new tag is itself 7.2.3, which lines up with the unchanged `rocm723` wheel variant, so the two setup modes now agree and that paragraph is rewritten to say each framework's bare-metal ROCm layer matches its image.

### No production code change

vLLM >= 0.26 ships `capture_torch_profiler` and `detailed_trace_annotation` upstream, so the native-profiler gate added in #1157 (`_VLLM_NATIVE_PROFILER_MIN_VERSION = (0, 26)`) already skips the TraceLens patch set for 0.27.1. The only non-doc edits are comments that had gone stale: `bypass_engine.py` and its test described `--profiler-config` as a "vllm 0.24" mechanism, and the kernel `install.sh` named a stripped `lmsysorg/sglang` v0.5.9 / rocm700 image as its example of a runtime lacking `git` / `patch` / `moreutils`. The former lose the version anchor because the interface is unchanged through 0.27.1; the latter now describes upstream stripped images generically, since naming a `rocm/hyperloom` tag there would wrongly imply our own images ship without those binaries.

## Verification

Checked against the real artifacts rather than assumed:

- `rocm/hyperloom:vllm-v0.27.1-rocm7.2.3` exists on Docker Hub (pushed 2026-08-15) and is the newest vLLM tag there; both SGLang tags exist and v0.5.16 is the newest SGLang.
- `vllm-0.27.1+rocm723-cp312-cp312-manylinux_2_34_x86_64.whl` exists under `wheels.vllm.ai/rocm/0.27.1/rocm723/`, so the raised bare-metal default resolves.
- Cross-checked v0.27.1's `vllm/config/profiler.py`: all seven fields this repo injects (`profiler`, `torch_profiler_dir`, `capture_torch_profiler`, `detailed_trace_annotation`, `ignore_frontend`, `delay_iterations`, `max_iterations`) are present, and `max_iterations=0` still means "no limit", so the `_profiler_bound_holds` guard keeps its meaning.
- That same file is what `_server_patcher`'s native path greps for its two sentinels, so `ensure_vllm_patched_for_tracelens` returns True on 0.27.1 and the two annotation flags are still injected. No silent profile degradation.

Note that this jumps three vLLM minor versions and one ROCm patch level at once: 0.27.1 is the first `rocm/hyperloom` image on ROCm 7.2.3, every earlier vLLM tag there is 7.2.0.

## Test plan

- [x] `test_baremetal_doc_version_consistency.py` passes. It locks `install_baremetal.sh` defaults to the `compatibility.rst` strings, which is the easiest thing to get wrong here.
- [x] `test_bypass_backend.py` profiler cases pass; the file's failure count is identical to `main` locally (the local box is Windows and lacks process-group support).
- [ ] CI on Linux.
- [ ] Optional follow-up on a GPU host: `docker run` the new image and confirm a profile round trip produces a torch trace plus the `capture_traces` sidecar.

### DIFF
--- a/docs/compatibility.rst
+++ b/docs/compatibility.rst
@@ -122,7 +122,7 @@ The following inference frameworks are supported:
      - 7.2.0
      - Default framework
    * - vLLM
-     - 7.2.0
+     - 7.2.3
      - Do not mix frameworks within one session
    * - Atom
      - 7.2.0
@@ -149,7 +149,7 @@ accordingly.
      - MI300X / MI325X
    * - ``rocm/hyperloom:sglang-v0.5.16-rocm7.2.0-mi350x``
      - MI355X
-   * - ``rocm/hyperloom:vllm-v0.24.0-rocm7.2.0``
+   * - ``rocm/hyperloom:vllm-v0.27.1-rocm7.2.3``
      - MI300X / MI325X / MI355X
 
 Browse all available tags at
@@ -170,7 +170,7 @@ Hyperloom does not install ROCm or torch itself.
      - Notes
    * - ROCm
      - 7.2.x
-     - Bare-metal patch level differs per framework: the vLLM stack uses ROCm 7.2.3 and the SGLang stack uses ROCm 7.2.0 (see the note below). ``docker`` mode uses the ``rocm/hyperloom`` images (ROCm 7.2.x) for both frameworks.
+     - The patch level differs per framework and is the same in both setup modes: the vLLM stack uses ROCm 7.2.3 and the SGLang stack uses ROCm 7.2.0 (see the note below).
    * - Python
      - 3.12
      - Required by the vLLM ROCm wheel.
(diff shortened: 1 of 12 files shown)


ZhengGong-amd (approved):
docs, approved

PR #1198, opened 2026-08-17
fix(loop): keep approved variants when a sibling is rejected

### LINKED ISSUE #1143: Scout→ critic schema mismatch
### Target release

v0.2

### Optimization domain

Inference

### Problem / use case

The critic rejected its own scouts' proposals for including predicted_gain_pct (specialist_quantitative_claim_violation). The only novel optimization ideas never reached benchmarking, and the run had no time to resubmit.

### Proposed solution

Enforce one shared proposal schema across scouts, critics, and executors; internally generated proposals must never fail on format.

### Expected impact

_No response_

### Alternatives considered

_No response_

### Additional context

_No response_

zoroyihan7:
Closes #1143. Follow-up to #1161.

#1161's agreed fix had two halves. (a) strip the forbidden fields at source —
landed. (b) pass `approved_variant_names` so a partial reject stops discarding
every variant — this PR. With both halves on the tree, the reported failure
cannot reject a format slip, and one genuine reject cannot sink siblings the
Critic still let through.

A mixed `verdict_map` used to collapse `reject` over `advise` and then
materialise the whole grid. One genuine reject therefore threw away every
sibling the Critic had approved or advised through. The filter on
`_materialize_approved_proposal` already existed; the live verdict path never
passed it.

### What changes

- After the per-entry hold, collapse the **proceedable** subset (`approve` /
  `advise`) first. The summary is that subset's collapse, and those names are
  the materialize filter.
- A map with no proceedable variant still collapses the whole map (`reject` /
  `needs_review`) and does not dispatch.
- An explore grid that the filter empties is skipped rather than enqueued empty.
- Single-verdict paths, all-reject maps, and all-advisory-hold maps are
  unchanged.

Not in this PR, and not required to close #1143: one automatic format-class
resubmit, and a shared pydantic proposal model (`reason` / `rationale` / `note`).
Those remain optional follow-ups if wanted later.

### Test plan

- [x] `test_a_genuine_reject_does_not_sink_advised_siblings`: advise + reject +
      approve materialises `{v_a, v_c}` under summary `approve`.
- [x] Mixed maps pass only proceedable names; advisory-only holds still keep
      the held variants; all-reject still rejects.
- [x] `collapse_verdict_map` unit: proceedable subset, advise surviving a
      sibling reject, no-proceedable stays reject.
- [x] Empty filter does not enqueue an explore task.
- [x] 247 passed across `test_critic_verdict_map.py`,
      `test_framework_agent_reauthor.py`, `test_coordinator_async_batch2_unit.py`.

### DIFF
--- a/src/hyperloom/inference_optimizer/tests/test_critic_verdict_map.py
+++ b/src/hyperloom/inference_optimizer/tests/test_critic_verdict_map.py
@@ -25,7 +25,9 @@ from hyperloom.inference_optimizer.protocol.intent import (
     validate_envelope,
 )
 from hyperloom.orchestrator.loop.coordinator_helpers import (
+    collapse_verdict_map,
     collapse_verdicts,
+    proceedable_variant_names,
     verdict_held_to_its_rule,
     verdict_map_entry_grounds,
     verdict_map_entry_held_to_its_rule,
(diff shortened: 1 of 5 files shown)


ZhengGong-amd (approved):
Merge after minor revision

xiaofei-zheng (approved):

PR #1199, opened 2026-08-17
fix(kb): harden Recipe finalization and warm replay

meinali-566:
## Summary
- make terminal Recipe publication idempotent across normal CLOSE and graceful teardown, persist the outcome before breakdown/Langfuse, and retry failed writes without changing stop-reason semantics
- preflight SGLang warm-replay config against the target workload and repair undersized last-wins `--context-length` values before any Patch/Kernel mutation
- use unified Patch headers as the sole Fusion/Rewrite target authority: do not persist `target_file`/`target_files`, dynamically resolve every target under one trusted framework root, snapshot all existing/new files, and fully roll back failed replay
- remove redundant Fusion source artifacts; unified Patch bytes plus benchmark evidence are sufficient for replay
- protect durable finalize fields with `CORE_STATE_FIELDS` so Agents cannot forge `written`/`skipped` state

## Production issues addressed
- graceful `time_exhausted` and other stop-check exits could miss the full CLOSE sequencer and depend on an unobservable one-shot fallback
- Recipe finalize occurred after session breakdown and Langfuse flush, so successful/skipped/error outcomes were absent from telemetry
- failed finalize could coexist with `close_sequence_done=true`, preventing a retry
- SGLang Recipes with `--context-length 6144` were replayed against `ISL=8192, OSL=1024`, causing `0/192` successful requests
- a Qwen Fusion Patch modified both `qwen3.py` and a new `qwen3_fused_ops.py`, while the old single-target metadata model caused repeated dispatch rejection

## Behavior
- publication eligibility remains unchanged: a Session needs a non-replay KEEP, finite positive throughput, a complete Recipe contract, and a score above the incumbent Champion
- `no_new_keep`, `not_better_than_champion`, and invalid throughput remain normal explicit skips
- no new stop reasons are introduced and stop reason is not a publication gate
- Patch destinations are derived only from safe unified-diff headers; target metadata is runtime-only and never written to the Recipe

## Rollout
- merge/deploy this PR before enabling [KB Store schema2 View migration](https://github.com/AMD-BRAIN-Internal/primus-cortex-internal/pull/13)
- 20 schema2 Explore-config migration sessions are already staged as non-Champion current-format candidates; promote them only after this preflight is deployed

## Test plan
- [x] 567 focused CLOSE/KB/replay/coordinator/policy tests
- [x] 11 smoke tests across Hyperloom and KB Store
- [x] Ruff, IDE lint, and `git diff --check`

## Follow-ups
- add equivalent workload-bound validation for vLLM/Atom (`--max-model-len`, batching limits)
- add external recovery for SIGKILL, force-delete, or node loss where in-process finally hooks cannot run


ZhengGong-amd (approved):
Approved

PR #1216, opened 2026-08-17
Close the accuracy-gate gaps that let a broken model be promoted

zengleixin-amd:
## Why

Breaking a model's numerics is itself a large throughput win. Across the retained session pool the accuracy-collapsed runs are a **median 22% faster** than the healthy ones in the same session (77 of 101 comparable sessions; up to **+392%** on GLM-5.2-MXFP4). An optimizer scored on throughput therefore selects *for* configurations that make the model emit garbage.

Warm replay was the worst gap: **45 of 241 promoted replays (19%)** were promoted while the model was broken ? the worst carrying a `+23.95%` gain on a config scoring **0.0000** on gsm8k against a 0.9014 baseline. The score had been measured; the promotion path never read it from the right place.

### Warm-replay accuracy gate (R1 + R2)

- **Every successful warm-replay now runs eval** on the warmup round (`baseline.py`). A recipe from another session is not a locally verified candidate, so the high-risk-only trigger used by explore does not apply here.
- **`--no-eval` still means no eval.** The forced warmup eval is skipped when the operator disabled eval for the session, matching what the baseline path on the same executor already did. Such a session promotes replays unjudged, which is the trade that flag already makes everywhere else.
- **Accuracy is harvested from the task directory that holds both rounds**, not from the measure round workspace. Verified against 852 recorded replays: scores sit in `warmup_round` 320 times, in `measure_round` **zero** times.
- **Judgment uses the existing absolute threshold** (`accuracy_passed(threshold=0.05)`) when a positive baseline accuracy exists. Degradation above 0.05 rejects promotion; within tolerance passes.
- **When no baseline reference exists** (`baseline_accuracy <= 0`, e.g. the baseline eval failed), a measured replay score is judged against the enablement absolute floor (`DEFAULT_ENABLEMENT_ACCURACY_FLOOR = 0.5`) so a collapsed replay cannot promote on throughput alone.
- **Failed measurements do not block the run.** If eval cannot produce a score, the reason is recorded in `eval_error` and promotion continues.
- **All outcomes are recorded** on `warm_replay_outcome` and flow into breakdown: `eval_ran`, `replay_accuracy`, `baseline_accuracy`, `eval_error`. `eval_ran` is what separates "the model answered nothing" (score 0.0) from "nothing checked it" (no score) ? collapsing those is what produced the original false rejections. Documented in `docs/reference/session-breakdown.md`.
- **`optimization_stack` entries** from warm-replay promotion carry `accuracy` (the replay score).

### Eval-context feasibility

When `--max-model-len` is lowered below what gsm8k needs, the accuracy gate cannot run ? that is an integration fault, not a kernel verdict. Prevents charging an unrunnable eval to the patch under test.

### R3 ? baseline/eval failure recording

Already satisfied by the existing `enablement` framework (`baseline_eval_kind`, `observed_accuracy`, evidence paths). No new code needed; confirmed across 1500 sessions.

### R4 ? the recipe KB is left alone

No change to the KB schema or its write paths, by design.

For transparency about the branch history: an earlier commit here (`1ae743126`) added a CLOSE-time `_champion_accuracy_ok` gate on the KB write, and a later one (`e14f5fb26`) removed it again after self-review found it covered only one of the two paths that write `best_config` and read a stack field two thirds of the promoting lanes never populate. Both commits are internal to this branch ? the gate never existed on `main`, and the net diff against `main` contains no KB change at all.

## No threshold changes, deliberately

`ACCURACY_THRESHOLD` stays at a 0.05 allowed drop for baseline-relative comparison: healthy run-to-run spread in the pool reaches 0.037, so tightening it would flag noise. The no-baseline path reuses the existing enablement floor rather than introducing a new number.

No threshold addresses the real failure mode anyway ? the fault is per-server-launch and bimodal, the same config measuring 0.0076 and 0.9393 within one session. Repeated sampling of high-risk configs is the follow-up for that and is deliberately not bundled here.

## Test plan

- [x] `test_warm_replay_accuracy_gate.py` ? 19 tests: degradation rejection, within-tolerance pass, warmup-round score location, no-baseline absolute floor pass/reject, and each way a score can be absent
- [x] `test_baseline_warmup_double_run.py` ? replay double-run forces `RUN_EVAL` on the warmup round only, and honours `--no-eval`
- [x] `test_eval_context_feasibility.py` ? 13 tests still pass
- [x] Exhaustive probe over every `parse_eval_results` path that yields no score, asserting `eval_ran` and the promote/reject decision: 11/11
- [x] Related suites on the merged tree ? 254 passed
- [x] Full inference_optimizer suite on pod ? 8 failed, 11238 passed; the same 8 fail on pristine `main` (real-ROCm / real-backend cases), so zero net regression
- [x] `ruff check` clean on changed files

Tests were run on Linux; the local Windows checkout cannot import `recipe_kb` (`fcntl`).


zengleixin-amd:
## CI status update (automated triage)

**Code-related checks on `d49a5ac17` — all green**
- test shards (py3.10/3.11 x 4), coverage, ruff, bandit, pylint, CodeQL, RTD

**Two failures — not caused by this PR's code changes**

### 1. E2E (`ci-e2e/run`) — platform infra
Both the PR push run and `/retest` failed immediately with:
`mkdir: cannot create directory '/shared_nfs/hyperloom-k8s/ci-pr-1216-...': Read-only file system`

Same error on PR #1220 E2E just now — cluster NFS mount is read-only, not a Hyperloom code issue.
Previous commit `e47a06421` had E2E green (~2h48m smoke run).

### 2. `build (sphinx html)` — flaky rocm_docs extension
`ExtensionError: add_source_buttons ... cannot unpack non-iterable NoneType object`

Reran 3x on the same workflow; all failed. Main branch docs builds pass; this PR does not touch docs.

**Next:** wait for NFS restore, then `/retest` E2E; sphinx may self-heal on next run or needs infra/rocm_docs fix.

ZhengGong-amd (approved):
Approved

PR #1221, opened 2026-08-18
Fix/geak revalidation dispatch

lishuoshuo-amd:
## Problem

A GEAK e2e win only reaches the headline after a same-harness rebench revalidates it. Across 68 audited sessions, 19 of 26 eligible candidates (73%) never completed that rebench, and the quality gate never once fired — the candidates were not rejected on merit, the rebench never ran to completion.

Two defects caused it:

1. **The phase boundary cancelled the rebench.** It runs as an `explore` task and no phase after KERNEL allows that kind, so `KERNEL_AGENT -> SWEEP` cancelled the task KERNEL had just dispatched — 475ms after creation in the sampled session.
2. **The idempotency key was a fixed string.** `"geak-revalidate"` is unique per session, so from macro-cycle 2 onward `create_or_return_existing` returned the settled first-cycle row, which KERNEL read as `rebench_unavailable`. All 20 affected sessions sat at `macro_cycle >= 2`.

Making the rebench survive and retry then exposes four more:

3. A rebench spared at every boundary also survives into `CLOSE`, where it runs a full serve-and-bench that starves the post-opt roofline of the GPU lane and can rewrite the headline after the report is written. The wall-clock timeout path is worse: it never reaches the CLOSE sequencer, so it cancels queued work but leaves a running rebench writing back through the grace window and the slot claiming a rebench is still coming.
4. Per-cycle keys plus no in-flight check allow two live rebenches; the orphan's result is taken at face value.
5. A cancelled attempt cannot be replaced within its cycle, because its key returns the cancelled row.
6. A candidate abandoned without revalidation produces no fact and no warning in the report — the session reads as if nothing was found. This predates the branch: `rebench_unavailable` was already invisible.

## Fix

1. `cancel_queued_not_allowed` takes a `spare_queued` predicate; the GEAK policy spares the rebench at every boundary except `CLOSE`. Deny-list, not allow-list: the phase set changes and a missing allow-list entry silently restores the bug, while a surplus costs one wasted bench that the gate in (4) refuses anyway.
2. Keys are scoped per macro-cycle.
3. Both wind-down paths — the CLOSE sequencer and the wall-clock closing phase — share one drain that cancels queued *and* running rebench, then settles the slot. Settling is driven by state, not by what the caller just cancelled, since the phase boundary has already cancelled the queued case by then; it edits the verdict fields in place so the candidate's self-reported numbers survive for the report.
4. `_enqueue_geak_revalidation` returns early while a rebench is in flight, and a result is applied only when `geak_pending` still tracks the finishing task. Applies to all four 2b decisions, including the inconclusive branch, which otherwise reached the 2a harness replay and cleared the tracked slot. Ignored results emit a `geak_rebench_result_ignored` observation.
5. Keys step past settled attempts (`-r1`, `-r2`, bounded per cycle). Pruning the explore family settles the slot instead of leaving KERNEL held open.
6. `rebench_unavailable` and the new `rebench_cancelled` both emit a warning naming the drop reason, in the final report and the mission view.

GEAK policy lives in `orchestrator/phases/geak_rebench.py`. `TaskRegistry` stays generic, receiving only the `spare_queued` hook and a `find_by_idempotency_key` lookup.

RCA: `Desgin-DOCS/hyperloom-geak-revalidation-dispatch-rca.md`

## Tests

26 regression tests in `test_geak_revalidation_dispatch.py` covering per-cycle keys, re-dispatch after cancel, the `SWEEP` spare and the cancel on both wind-down paths, duplicate-dispatch suppression, prune settling, audit fields surviving a settle, orphan and late rejection across all four decisions, and report rendering for dropped candidates.

```bash
pytest src/hyperloom -n 12
# 14339 passed, 21 failed
```

The 21 failures are pre-existing and environment-induced (credentials / gateway auth), each reproduced on a clean `origin/main` worktree. `test_objective.py::test_run_closing_phase_writes_report_on_time_exhausted` is a pre-existing flake driven by its 6ms `max_minutes`; measured 8 runs at pass=6 fail=2 on both this branch and `origin/main`.

## Breaking changes

None. `rebench_cancelled` is a new additive `geak_pending.status`; readers treat unknown statuses as non-pending. Sessions on the old key resume correctly — the slot holds a task id, which matches directly.


ZhengGong-amd (approved):
Approved

xiaofei-zheng (approved):

PR #1222, opened 2026-08-18
fix(gemm): stop stamping self-reported speedups as validated, enable tuned-config hit logging

chennyiiis:
## What this is

`#1188` made every GEMM tuning candidate measured end to end instead of promoted
from a self-reported ratio. This branch is the layer above that: **an end-to-end
number can be measured correctly and still be measuring the wrong thing.**

Three ways that happens, all observed:

- **The tuned table was never read.** When the artifact does not reach the
  runtime, the server loads its own bundled default. Our CSV is still on disk,
  so nothing looks wrong — the throughput difference is measuring drift.
- **There was nothing to read it with.** A scan of 60 production logs found
  `AITER_LOG_TUNED_CONFIG` set in none of them. Without hit lines, shape
  discovery falls back to `config.json`, which covered **0.4%** of the keys the
  runtime actually asks for.
- **One measurement cannot tell a gain from drift.** Three rounds of a single
  *unchanged* configuration spanned 58% on this fleet; one controlled repeat
  moved 16%.

## What it changes

| Area | Change | Blocking? |
|---|---|---|
| Apply verification | `measurement/apply_verification.py`, wired into the KEEP decision | **Yes** — see below |
| Hit logging | `AITER_LOG_TUNED_CONFIG=1` injected for every serving run (operator value wins) | No |
| Candidate filter | `partial_output` and `unverified_shapes` results now reach the e2e validator | No |
| Paired A/B | `_confirm_gemm_gain_paired` + `measurement/paired.py`, opt-in | No |
| Convergence judge | `measurement/convergence.py` | No — **not yet wired into the measurement rounds** |
| Audit trail | Per-tuner status/error/error_class preserved in the tuning row | No |
| MoE routing | Warns when routing is coarser than the log it read | No |

### The one behaviour change that can reject a KEEP

`_gemm_apply_verdict` can now block a KEEP that would previously have been
accepted. It blocks on **two positive findings only**:

- `not_merged` — the artifact is absent from the runtime's merge list, so the
  server is serving its bundled default;
- `zero_hit` — the artifact was merged, hit logging was on, and every lookup
  missed it.

Everything else (`inconclusive_no_hit_logging`, `no_lookups`, `unknown`) does
**not** block. "Cannot tell" is never treated as "failed", and a box without
the forge parser degrades to `unknown`/non-blocking.

### Environment variables

Both are documented in `docs/reference/environment-variables.md`:

- `AITER_LOG_TUNED_CONFIG` — default `1`, injected per serving run. An operator
  value wins; set `0` to turn hit logging off, at the cost of the demand list
  and the apply verdict going inconclusive.
- `HYPERLOOM_GEMM_PAIRED_PAIRS` — default `0` (off). Interleaved
  baseline/tuned pairs before a gain is reported as confirmed. When off, the
  gain is still promoted but labelled an unpaired block comparison.

## Rebase note

This branch was cut before `#1188` and originally carried a fix to
`_promote_gemm_tuning_keep`, teaching it not to stamp self-reported speedups as
validated. `#1188` deleted that function outright and solved the same problem
the stronger way. After rebasing, those hunks and their two tests are gone, and
one commit was dropped as fully obsolete. `cumulative_gain_validated` is
unaffected by this branch: `#1188` already made it come from a real
measurement.

## Test plan

- [x] `orchestrator/measurement` — 44 passed / 14 skipped without KernelForge,
      58 passed / 0 skipped with it (both checked on a Linux box)
- [x] `test_decision_framework.py`, `test_coordinator_gemm_promote_units.py`,
      `test_kernel_request_handlers_units.py` — 521 passed, 14 skipped
- [x] Ruff clean on every file this branch touches
- [x] 8-card MI355X, two boxes (sglang source checkout and vLLM wheel):
      demand derived from a real serving log (122 misses), `vllm_moe_triton`
      9/10 shapes improved at 1.783x, `fmoe_ck` 10/10 at 1.097x
- [ ] Full `inference_optimizer` suite on a Linux box (cannot run on Windows:
      `fcntl`)


chennyiiis:
## 8-GPU MI355X verification

Verified on two 8-card boxes (`vLLM 0.27.1+rocm723` and `sglang 0.5.17`).

### The injected flag produces the evidence the gate needs

`AITER_LOG_TUNED_CONFIG=1` was previously set in none of the 60 production logs
scanned during the RCA, which is why every apply verdict came back empty. On a
real vLLM run with the injection in place the server emitted **83 hit lines
against 122 misses**, and the parser returned `verdict="served"` with a 40.5%
hit ratio — the first non-empty apply verdict available. Without the flag the
same log reads as a total miss, which is exactly the case the gate must *not*
treat as a failed apply.

### KEEP is now refused when the artifact provably never applied

The e2e loop already computed reachability and then credited the gain anyway.
Two positive findings now block the promotion, and "cannot tell" deliberately
does not:

| verdict | blocks KEEP | why |
|---|---|---|
| shape keys never resolved | yes | the delta measures something else |
| table absent from the merge list | yes | the server ran its bundled default |
| `inconclusive_no_hit_logging` | no | hit lines need the flag; absence proves nothing |
| `served` / no evidence at all | no | nothing to accuse |

A related finding from the forge side reinforces why the merge check matters:
pointing `AITER_CONFIG_GEMM_BF16` at a bare tuned CSV makes aiter load only that
file and drop its 14 bundled tables, taking measured hits from 83 down to 52.

### Paired A/B is wired, and the ledger now says which it got

`base_tput` and `new_tput` are measured at different times, so their difference
carries whatever the machine did in between — 16% on one controlled repeat on
this fleet, 58% across three rounds of an unchanged configuration. Interleaving
separates the two but costs two extra benchmark rounds per pair, so it is opt-in
via `HYPERLOOM_GEMM_PAIRED_PAIRS` (default off, no change to session cost).

What is no longer optional is labelling: `measurement_basis` now records
`e2e_paired`, `e2e_paired_<reason>`, or `e2e_rebench_unpaired`, so a block
comparison cannot reach the breakdown as an unqualified "validated". **Whether
to turn the paired pass on by default is a fleet budget call worth making
explicitly** — the mechanism and its tests are in place either way.

### MoE stage detection

A serving log showing both 1-stage and 2-stage dispatch stays tunable; only an
exclusively-1-stage log suppresses CK tuning. Verified against the routing
predicate directly, since collapsing that to "saw 1-stage anywhere" forfeits the
token range 2-stage serves.

168 tests pass, 10 of them new.

ZhengGong-amd (approved):
Approved

PR #1227, opened 2026-08-18
fix(cli): reload SharedState before pre-Coordinator save

meinali-566:
## Summary
- Reload `SharedState` from disk before the pre-Coordinator launch-shape save introduced in #1197.
- Prevents T0 `warm_start_*` fields from being clobbered by a stale seed/resume in-memory object, which caused PRELUDE to skip warm replay with `no_warm_start_recipe` despite a successful KB download.

## Test plan
- [ ] Start a fresh remote-mode session with an existing KB champion; confirm `state.json` retains non-empty `warm_start_recipe` / `warm_start_ts` after startup.
- [ ] Confirm PRELUDE dispatches warm replay instead of reporting `no_warm_start_recipe`.
- [ ] Resume a session with `--robustness-disable-server-probe`; confirm robustness launch-shape fields still persist correctly.


ZhengGong-amd (approved):
Approved

PR #1234, opened 2026-08-19
fix(roofline): reduce the profile's sandbox-rebuild exposure, and record pod ownership

jiaqiang-dot-liu:
## Context

Session `Kimi-K3_20260818T062936Z_a62853e5` lost 3/3 rooflines, never produced an `analysis.md`, and skipped its whole KERNEL phase. Root cause is **outside this repo**: the agent harness saw `UND_ERR_SOCKET` on an in-flight sandbox call, hit its rebuild path, and issued `POST /workloads/{id}/stop` — which kills the optimizer and the in-flight profile along with the pod. It ran across 4 pods / 3 rebuilds, each rebuild landing inside a roofline trace-flush window.

Hyperloom cannot fix that trigger (it owns no sandbox lifecycle and never calls any workload-stop API). What it can do is stop *inviting* the trigger, stop paying for capture it cannot use, and make a rebuild diagnosable after the fact. This PR does those three things.

## Changes

**1. Forbid blocking waits in the monitor loop** (`SKILL.md`, prompt-only)

`## Monitoring` asked for a 5-minute polling cadence without saying how to wait for it. An agent whose only waiting tool is a sandbox bash call implements that as `sleep 110; sleep 110; sleep 80` — which is what the 08-18 session did. That call holds the sandbox connection open for its whole duration, so overlapping it with a trace flush (8 ranks serialising torch traces to NFS, minutes on a 2.8T MoE) is what dropped the connection. The cadence was never this skill's job: the Robustness Monitor two sections up already polls every 5 minutes from its own `setsid nohup` process. The section now states the division of labour (launch, check once, hand off, report), forbids `sleep` / wait loops / `tail -f`, keeps the sanctioned one-shot `sleep 30` health check explicit, and records *why* so the next editor does not reintroduce a wait.

This is the only change here that reduces the trigger probability rather than the blast radius.

**2. Drop annotation-only capture options when the TraceLens patch failed** (`_workload_envs.py`)

Without the server-side patch the trace carries no `kernel_shape_profiler` events at all (trace-health check 5), yet SGLang still requested `shape_discovery` / `detailed_annotations` over `/start_profile` — full capture and serialization cost for data nothing downstream can read, and a longer flush window to be interrupted in. The vLLM branch already gated its equivalent `detailed_trace_annotation`; SGLang did not, and the existing `"profile will omit annotation-only flags"` warning described only the CLI flag, not the `extra_body` options.

Gated on the degraded **reason**, not on `tracelens_patch_ok`: "patching disabled" (`HYPERLOOM_ENABLE_PATCH=0`, which a pre-patched image legitimately uses) is not "patching failed" and must keep the annotations. That also leaves the healthy and kill-switch paths byte-identical, so the profile golden lock is unchanged.

**3. Resolve the CUDA-graph flag contradiction** (`_workload_envs.py`)

Eager profiles carried both `--enable-profile-cuda-graph` and `--disable-cuda-graph`. An eager server captures no graph, so the former is setup cost for a no-op. The two flags arrive from different layers — the profile YAML template versus the eager fallback folded into `extra_server_args` — and only meet after the reference/server-args merges, so the resolution lives there.

**4. Record pod ownership across rebuilds** (`lock.py`, `session_paths.py`)

Every acquirer truncates `optimizer.lock`, so a rebuilt session keeps only its first owner (`manifest.json`) and its last (the lock body). The 08-18 session ran on 4 pods but reads as single-pod from the session directory alone — which is how the first pass at triage undercounted it as 2. Each acquisition now appends to `runtime/pod_history.jsonl`; append-only and best-effort, since a write failure must never fail an optimizer that already holds the lock.

## Non-goals, and what still needs an owner

- **The rebuild trigger itself is not fixed here.** Filed separately for the harness team: all three rebuilds fired at `consecutive_errors=1` even though the threshold is configured as 3. That one check would neutralise every trigger path — long bash, short poll, health probe — and is higher leverage than anything in this PR.
- **Change 3 is hygiene, not a defence.** A sibling Kimi-K3 session (`Kimi-K3_20260818T064529Z_f059f91d`, vLLM) hit the same failure with CUDA graph *enabled* (`enforce_eager=False`), so shortening the eager flush window does not prevent it.
- **Partial-trace salvage was considered and rejected.** The 2/8 traces that survived rebuild #2 are truncated mid-write — `gzip` fails with `EOFError: Compressed file ended before the end-of-stream marker was reached`. Salvage would feed a corrupt trace to TraceLens, or reject it on validation and salvage nothing. Machinery for no benefit in the observed case.

## Test plan

- [x] 7 new tests: annotations dropped when degraded / kept when healthy / kept when patching was never attempted; graph-capture flag dropped when eager / kept in graph mode; pod ledger accumulates one line per acquire; ledger write failure does not break acquire.
- [x] `pytest` on the touched areas: 293 passed, plus the `SKILL.md` consumer (`test_dotenv_precedence_recipes.py`) green after the prompt edit.
- [x] Profile golden lock (`test_workload_envs_golden_lock.py`) unchanged — the healthy and kill-switch paths are byte-identical.
- [x] `ruff check` clean on changed files. Remaining `ruff format` drift verified pre-existing (identical on `main`); not reformatted, to keep the diff reviewable.
- [ ] Pre-existing unrelated failure in this environment: `test_trace_analyze_handler_dry_run_returns_structured_result` (`ModuleNotFoundError: No module named 'hyperloom'` in a subprocess — reproduces identically on `main` without `pip install -e .`).

### DIFF
--- a/src/hyperloom/inference_optimizer/SKILL.md
+++ b/src/hyperloom/inference_optimizer/SKILL.md
@@ -1098,7 +1098,21 @@ guess). Edit the example before copying if defaults need to change.
 
 ## Monitoring
 
-Poll at most every 5 minutes unless debugging a startup failure.
+Each poll is one fast read of persisted state, and you poll only when you are
+next invoked. **Never block to reach the next poll**: no `sleep`, no wait loop,
+no `tail -f`. The only sanctioned wait is the one-shot `sleep 30` health check
+right after launch. Recurring 5-minute polling is the Robustness Monitor's job
+and already runs in its own `setsid nohup` process (see above) — do not
+reimplement it here.
+
+**Why this is load-bearing, not style.** A blocking call holds the sandbox
+connection open for its whole duration. Overlap it with a `roofline` trace flush
+— 8 ranks serialising torch traces to NFS, which takes minutes on a large TP=8
+MoE — and the connection can drop; the agent harness reads that as an
+unreachable sandbox and rebuilds it, which kills the optimizer and the in-flight
+profile with it. Session `Kimi-K3_20260818T062936Z_a62853e5` lost 3/3 rooflines,
+its `analysis.md`, and its whole KERNEL phase to a `sleep 110; sleep 110;
+sleep 80` progress loop. Launch, check once, hand off to the monitor, report.
 
 Resolve `$SESSION` the same way the Robustness Monitor does — never from
 `$USER_DATA_PATH`, which is the workspace root, not the session dir.
--- a/src/hyperloom/inference_optimizer/references/troubleshooting.md
+++ b/src/hyperloom/inference_optimizer/references/troubleshooting.md
@@ -103,3 +103,14 @@ Bypass with `--critic-mock` for offline / smoke runs. See
   family). Do not hand-edit `state.json`.
 - `stop_reason=time_exhausted`: resume same session (`--resume-from`); do not
   start fresh.
+- `dead_holder_reaped` (lease reclaimed because the holder pid vanished): the
+  optimizer process died without unwinding, which on a sandboxed runtime usually
+  means the pod itself went away mid-task. Read
+  `runtime/pod_history.jsonl` — one append-only line per optimizer-lock
+  acquisition (`acquired_at` / `hostname` / `pid`), so more than one line means
+  the session was rebuilt and each entry marks a takeover. Prefer it over
+  `manifest.json` (pins only the *first* owner) and over `runtime/optimizer.lock`
+  (holds only the *current* one); a multi-rebuild session reads as single-pod
+  from those two alone. Correlate the takeover timestamps with the task that
+  failed — a roofline reclaimed here typically leaves a partial `torch_trace/`
+  and no `analysis.md`.
--- a/src/hyperloom/inference_optimizer/session/lock.py
+++ b/src/hyperloom/inference_optimizer/session/lock.py
@@ -25,12 +25,19 @@ The lock file is intentionally never unlinked on release: unlinking a flock'd
 path races a concurrent acquirer (it would flock a now-unlinked inode while a
 newcomer creates and flocks a fresh file). A stale body is harmless — the next
 acquirer flocks the same inode and overwrites it.
+
+That overwrite keeps only the *current* owner, so each acquisition is also
+appended to ``<session_dir>/runtime/pod_history.jsonl``. A session whose sandbox
+is rebuilt mid-run otherwise records only its first owner (in
+``manifest.json``) and its last (in the lock body), and reads like a single-pod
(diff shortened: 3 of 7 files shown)


xiaofei-zheng (approved):

jiaqiang-dot-liu:
Thanks — acted on three of the four, and one clarification on the second.

**1. `_tracelens_patch_enabled()` docstring out of sync — fixed** (96764a3)

You're right that it read as contradicting the new SGLang behaviour. While writing it up I realised the two gates are not just stylistically different, they answer different questions, so the docstring now says that explicitly:

- A **server flag** only a patched argparse accepts (`--profiler-config.detailed_trace_annotation`, `--enable-shape-discovery-for-cuda-graph-profile`) must not be injected when patching is off — an unpatched server *rejects* it. That is a compatibility gate.
- The `PROFILE_EXTRA_BODY` annotations ride `/start_profile`, which an unpatched server accepts. Nothing breaks; you just pay for capture that produces nothing readable. That is a cost gate.

So the kill switch is also how a pre-patched image opts out of runtime patching while still supporting the annotations, which is exactly why that gate reads `HYPERLOOM_PROFILE_DEGRADED_REASON` (set only on the attempted-and-failed path) rather than `tracelens_patch_ok`.

**2. vLLM / SGLang gating style — agreed, not unifying here**, with one refinement to your reading: the two are not merely different expressions of the same rule. vLLM's `tracelens_patch_ok` gate protects against an argparse rejection; the SGLang gate avoids paying for unusable capture. Keying them differently is load-bearing, not incidental — unifying on `tracelens_patch_ok` would silently degrade a pre-patched image. That reasoning is now in the docstring so a future cleanup does not "fix" it.

**3. `pod_history.jsonl` had no consumer — addressed** (96764a3)

Added a `dead_holder_reaped` bullet to troubleshooting's run-time signals: read `runtime/pod_history.jsonl`, one line per lock acquisition, more than one line means the session was rebuilt. It also spells out why that beats `manifest.json` (first owner only) and `optimizer.lock` (current owner only) — those two are what made the original triage of the 08-18 session undercount 4 pods as 2.

**4. Redundant `import json` — fixed** (96764a3)

Removed the three in `test_profile_and_kernel_handlers.py`, which does have a module-level `import json`. Note the fourth occurrence, in `test_session_lock.py`, was *not* redundant — that file has no top-level `json` import — so I hoisted it to the top instead of deleting it, matching that file's all-top-level import style.

Re-ran after the changes: 205 passed, `ruff check` clean, `ruff format` drift back to the pre-existing baseline. The one remaining failure (`test_trace_analyze_handler_dry_run_returns_structured_result`) is the environment's missing `pip install -e .`, identical on `main`.

PR #1237, opened 2026-08-20
fix(coordinator): route target_reached through CLOSE so the close seq…

WhatGhost:
## Problem

When a run meets its `--target-gain`, the coordinator breaks out of the tick loop immediately, so the CLOSE phase never runs and the 7-step close sequencer is skipped. The session ends with only the CLI safety-net report:

```json
{ "report_complete": false, "safety_net": true, "stop_reason": "target_reached" }
```

```markdown
# Inference Optimizer — emergency final report
> **Auto-generated safety-net.** The CLOSE phase 7-step sequencer did not run to
> completion (process exited before phase transition, or `report` executor failed).
```

A successful run therefore loses the real `report` output, the post-optimization roofline (CLOSE step 0), `kernel_optimization_summary.json`, `robustness_postmortem.md`, and the CLOSE segment in `session_breakdown.json`.

## Root cause

`machine_state.py:228` registers `target_reached` as an "any phase -> CLOSE" terminal transition reason, but **nothing ever produced that transition** — `compute_next_phase()` never returns it. The only producer was the tick loop, which used it as a local variable and broke:

```python
if objective.reached(self.shared_state):
    stop_reason = "target_reached"   # local variable only
    break                            # phase machine never runs again
```

`_advance_phase_if_needed()` lives in the tick body, so `break` removes any further chance to reach CLOSE — and `shared_state.stop_reason` is still empty at that point, since it is only written in the `finally` block.

This also made the two success terminals inconsistent: `time_exhausted` (also in `_SUCCESS_STOP_REASONS`) routes through a closing path, so a timeout closed cleanly while a met target did not.

## Fix

Publish the stop reason and let the existing state machine drive the transition:

```python
if objective.reached(self.shared_state):
    if not self.shared_state.stop_reason:
        self.shared_state.set_stop_reason("target_reached")
        self.shared_state.save(self.session_dir)
        continue
    stop_reason = self.shared_state.stop_reason
    break
```

`_global_terminal()` (`machine_state.py:1645`) already picks this up — its comment reads *"Coordinator-set stop_reason takes precedence over phase exits"*. That path had simply never been fed a value. No new mechanism is introduced.

Bounded: after `continue`, the `if self.shared_state.stop_reason and not in_closing` check necessarily fires on the next tick, so the loop runs at most one extra tick and cannot spin regardless of whether CLOSE succeeded. The `else` branch preserves the previous behaviour when a stop reason is already published, leaving the timeout path untouched.

## Verification

**Unit test.** `test_target_reached_routes_through_close_phase` builds a run that necessarily meets its objective and asserts the phase reaches CLOSE. Reverting only the implementation confirms it catches the bug:

```
without fix:  FAILED  assert 'FRAMEWORK_AGENT' == 'CLOSE'
with fix:     PASSED
```

**Full suite.** `9 failed, 14360 passed, 32 skipped` — the same 9 pre-existing failures as `main`; passing count goes 14359 -> 14360 (the new test).

**End-to-end.** A real run (MI355X, vLLM 0.24.0, Qwen3-8B, TP=1, conc=64, ISL=OSL=1024, bf16, `--target-gain 15`) ending in `target_reached`:

```
close_sequence_done:       True
cumulative_gain_validated: 28.12%   (baseline 6736.07 tok/s/GPU)

PRELUDE   from=-         reason=prelude_done
EXPLORE   from=PRELUDE   reason=target_reached
CLOSE     from=EXPLORE   reason=(current)
```

`reason=target_reached` on the EXPLORE segment is the vocab entry that previously had no producer.

| | before | after |
|---|---|---|
| `final.md` heading | `— emergency final report` | `Inference Optimizer Report — <id>` |
| `final.json` | 2,869 B | 59,898 B |
| `report_complete` / `safety_net` | `false` / `true` | absent |
| CLOSE phase segment | missing | present |
| `kernel_optimization_summary.json` | missing | written |
| `robustness_postmortem.md` | missing | written |

### DIFF
--- a/src/hyperloom/inference_optimizer/tests/test_coordinator_runtime.py
+++ b/src/hyperloom/inference_optimizer/tests/test_coordinator_runtime.py
@@ -2520,5 +2520,8 @@ async def test_target_reached_routes_through_close_phase(session_dir):
         )
         assert reason == "target_reached"
         assert (c.shared_state.phase or "").upper() == "CLOSE"
+        # The close sequencer actually ran; this is what separates a real close
+        # from the cli safety-net path.
+        assert c.shared_state.close_sequence_done is True
     finally:
         await c.stop()
--- a/src/hyperloom/inference_optimizer/tests/test_coordinator_runtime.py
+++ b/src/hyperloom/inference_optimizer/tests/test_coordinator_runtime.py
@@ -5,10 +5,13 @@
 
 from __future__ import annotations
 
+import asyncio
 import json
+import time
 from pathlib import Path
 from types import SimpleNamespace
 from typing import Any
+from unittest import mock
 
 import pytest
 
@@ -2589,4 +2592,87 @@ async def test_target_reached_closes_despite_a_failed_state_save(session_dir):
         assert c.shared_state.close_sequence_done is True
     finally:
         c.shared_state.save = real_save  # type: ignore[method-assign]
+        await c.stop()
+
+
+@pytest.mark.asyncio
+async def test_target_reached_close_still_runs_the_post_opt_roofline(session_dir):
+    """A met target must not be treated as a wall-clock rescue.
+
+    ``_maybe_run_close_post_opt_roofline`` returns early on ``closing_phase``,
+    which means "the wall clock ran out, shed expensive work". Routing a success
+    terminal through that flag would drop the post-opt snapshot the
+    optimization-progress chart reads -- one of the artifacts skipping CLOSE
+    loses in the first place.
+    """
+    _write_marker_target_baseline(session_dir)
+    c = Coordinator(session_dir, backends=_silent_backends())
+    c.sub.register_executor("report", report_executor)
+    c.shared_state.baseline_tput = 100.0
+    c.shared_state.cumulative_gain_validated = 50.0
+    # A kernel-level optimization landed, so the roofline step applies.
+    c.shared_state.optimization_stack = [{"action": "integrate", "tput": 150.0}]
+    c.shared_state.save(session_dir)
+
+    ran: list[str] = []
+
+    async def _record_roofline() -> None:
+        ran.append(str(c.shared_state.closing_phase))
+
+    c.phase_close._maybe_run_close_post_opt_roofline = _record_roofline  # type: ignore[method-assign]
+    try:
+        reason = await c.run(
+            objective=TargetGainObjective(target_gain_pct=10.0),
+            max_minutes=0.0001,
+            max_ticks=6,
+        )
+        assert reason == "target_reached"
+        assert ran == ["False"], f"post-opt roofline saw closing_phase={ran}"
+    finally:
+        await c.stop()
+
+
+@pytest.mark.asyncio
+async def test_target_reached_close_is_not_cancelled_by_an_outer_bound(session_dir):
+    """The sequencer's own per-step timeouts are the budget, not an outer one.
+
+    ``CLOSE_POST_OPT_ROOFLINE_TIMEOUT_SEC`` alone allows 600s, so any outer
+    bound short enough to matter would cancel the step mid-flight and drop the
+    run onto the safety net -- the outcome this routing exists to avoid. With
+    the session bound already elapsed, an awaited step would be skipped outright
+    unless the terminal close lifts it.
+    """
+    _write_marker_target_baseline(session_dir)
+    c = Coordinator(session_dir, backends=_silent_backends())
+    c.sub.register_executor("report", report_executor)
+    c.shared_state.baseline_tput = 100.0
+    c.shared_state.cumulative_gain_validated = 50.0
+    c.shared_state.save(session_dir)
+
+    # The bound is armed and elapsed; the terminal close must ignore it.
+    c._run_deadline = time.monotonic() - 1.0
+    assert c._seconds_until_session_bound() is not None
+    c._terminal_closing = True
+    assert c._seconds_until_session_bound() is None, "terminal close must be unbounded"
+    c._terminal_closing = False
+
+    real_advance = c.phase_machine._advance_phase_if_needed
+    completed: list[bool] = []
+
+    async def _slow_advance() -> None:
+        # Any await at all is cancelled or skipped by an elapsed outer bound.
+        await asyncio.sleep(0.3)
+        await real_advance()
+        completed.append(True)
+
+    try:
+        with mock.patch.object(c.phase_machine, "_advance_phase_if_needed", _slow_advance):
+            reason = await c.run(
+                objective=TargetGainObjective(target_gain_pct=10.0),
+                max_minutes=0.0001,
+                max_ticks=2,
+            )
+        assert reason == "target_reached"
+        assert completed, "the close step was skipped or cancelled by an outer bound"
+    finally:
         await c.stop()
--- a/src/hyperloom/inference_optimizer/tests/test_coordinator_runtime.py
+++ b/src/hyperloom/inference_optimizer/tests/test_coordinator_runtime.py
@@ -29,6 +29,7 @@ from hyperloom.orchestrator.loop.coordinator_helpers import (
 )
 from hyperloom.orchestrator.loop.proposals import ProposalsCollaborator
 from hyperloom.inference_optimizer.protocol.intent import Intent, IntentType
+from hyperloom.orchestrator.state.objective import TargetGainObjective
 from hyperloom.orchestrator.state.shared_state import SharedState
 from hyperloom.orchestrator.loop.sub_agent_runner import (
     SubAgentRunner,
@@ -2494,5 +2495,30 @@ async def test_dispatch_audit_skips_kernel_owned_kind_under_no_kernel(session_di
         assert evidence.get("reason") == "policy_denied"
         assert evidence.get("rule") == "kernel_owned_by_kernel_agent"
         assert not await c.tasks.by_state("failed")
+    finally:
+        await c.stop()
+
+
+@pytest.mark.asyncio
+async def test_target_reached_routes_through_close_phase(session_dir):
+    """A met objective transitions to CLOSE instead of breaking out of the loop.
+
+    ``machine_state`` registers ``target_reached`` as an "any phase -> CLOSE"
+    transition reason, so a met target must reach the close sequencer rather
+    than leaving the run with only the cli safety-net report.
+    """
+    _write_marker_target_baseline(session_dir)
+    c = Coordinator(session_dir, backends=_silent_backends())
+    c.sub.register_executor("report", report_executor)
+    c.shared_state.baseline_tput = 100.0
+    c.shared_state.cumulative_gain_validated = 50.0
+    c.shared_state.save(session_dir)
+    try:
+        reason = await c.run(
+            objective=TargetGainObjective(target_gain_pct=10.0),
+            max_ticks=6,
+        )
+        assert reason == "target_reached"
+        assert (c.shared_state.phase or "").upper() == "CLOSE"
     finally:
         await c.stop()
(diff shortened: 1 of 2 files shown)


xiaofei-zheng:
## Code Review

### Summary

Solid fix. The root-cause analysis is accurate: `target_reached` was registered in `machine_state` as a terminal transition to CLOSE, but the coordinator only set a local `stop_reason` and `break`d, so `_global_terminal()` never fired and the 7-step close sequencer was skipped.

Publishing `stop_reason` and `continue` lets the existing state machine drive `any phase → CLOSE` on the next tick. The close sequencer runs synchronously inside `_on_enter_close()` during phase advance, before the loop-end `stop_reason` check — ordering is correct. The `else` branch preserves prior behavior when `stop_reason` is already set (resume path).

E2E evidence in the PR description is convincing (`close_sequence_done: True`, artifact sizes, CLOSE segment in `session_breakdown.json`).

---

### Findings

#### 1. [Medium] Unit test uses the wrong state field

In `test_target_reached_routes_through_close_phase`:

```python
c.shared_state.cumulative_gain_validated = 50.0
```

But `TargetGainObjective.reached()` reads `state.cumulative_gain`, not `cumulative_gain_validated` (see `objective.py`). The existing pattern in `test_objective.py::test_run_stops_on_objective_reached` correctly uses:

```python
c.shared_state.cumulative_gain = 50.0
```

With `cumulative_gain` left at the default `0.0`, `objective.reached()` stays false and the run exits via `max_ticks=6` with reason `"max_ticks"`, not `"target_reached"`. Please switch to `cumulative_gain`.

#### 2. [Low] Test assertion is weaker than the fix target

The test only asserts `phase == "CLOSE"`. It does not assert `close_sequence_done is True`, which is the actual outcome that distinguishes a real close from the safety-net path. Recommend adding:

```python
assert c.shared_state.close_sequence_done is True
```

#### 3. [Low] Resume with pre-set `stop_reason` still skips CLOSE (pre-existing)

`test_run_preserves_prior_stop_reason_when_loop_exits_without_new_reason` shows that a session resumed with `stop_reason="target_reached"` exits on the first stop check without entering CLOSE. Not introduced by this PR, but worth a follow-up if resume-after-interrupt is a common path.

#### 4. [Nit] One extra tick of reactor work after target is met

The `continue` design defers the CLOSE transition to the next tick, so reactors/dispatcher/framework pump still run once after the target is hit. Functionally fine; acceptable trade-off for reusing existing machinery.

#### 5. [Nit] `Coordinator.run()` docstring could mention CLOSE routing

The docstring still lists `target_reached` alongside `time_exhausted (via closing phase)` without noting that `target_reached` now routes through the CLOSE phase sequencer (a different mechanism from `closing_phase` grace). Minor doc sync.

---

### Verdict

**Approve with minor test fix.** Logic is correct and the change is minimal. Please fix `cumulative_gain_validated` → `cumulative_gain` in the new test (and consider asserting `close_sequence_done`) before merge.

WhatGhost:
> ## Code Review
> ### Summary
> Solid fix. The root-cause analysis is accurate: `target_reached` was registered in `machine_state` as a terminal transition to CLOSE, but the coordinator only set a local `stop_reason` and `break`d, so `_global_terminal()` never fired and the 7-step close sequencer was skipped.
> 
> Publishing `stop_reason` and `continue` lets the existing state machine drive `any phase → CLOSE` on the next tick. The close sequencer runs synchronously inside `_on_enter_close()` during phase advance, before the loop-end `stop_reason` check — ordering is correct. The `else` branch preserves prior behavior when `stop_reason` is already set (resume path).
> 
> E2E evidence in the PR description is convincing (`close_sequence_done: True`, artifact sizes, CLOSE segment in `session_breakdown.json`).
> 
> ### Findings
> #### 1. [Medium] Unit test uses the wrong state field
> In `test_target_reached_routes_through_close_phase`:
> 
> ```python
> c.shared_state.cumulative_gain_validated = 50.0
> ```
> 
> But `TargetGainObjective.reached()` reads `state.cumulative_gain`, not `cumulative_gain_validated` (see `objective.py`). The existing pattern in `test_objective.py::test_run_stops_on_objective_reached` correctly uses:
> 
> ```python
> c.shared_state.cumulative_gain = 50.0
> ```
> 
> With `cumulative_gain` left at the default `0.0`, `objective.reached()` stays false and the run exits via `max_ticks=6` with reason `"max_ticks"`, not `"target_reached"`. Please switch to `cumulative_gain`.
> 
> #### 2. [Low] Test assertion is weaker than the fix target
> The test only asserts `phase == "CLOSE"`. It does not assert `close_sequence_done is True`, which is the actual outcome that distinguishes a real close from the safety-net path. Recommend adding:
> 
> ```python
> assert c.shared_state.close_sequence_done is True
> ```
> 
> #### 3. [Low] Resume with pre-set `stop_reason` still skips CLOSE (pre-existing)
> `test_run_preserves_prior_stop_reason_when_loop_exits_without_new_reason` shows that a session resumed with `stop_reason="target_reached"` exits on the first stop check without entering CLOSE. Not introduced by this PR, but worth a follow-up if resume-after-interrupt is a common path.
> 
> #### 4. [Nit] One extra tick of reactor work after target is met
> The `continue` design defers the CLOSE transition to the next tick, so reactors/dispatcher/framework pump still run once after the target is hit. Functionally fine; acceptable trade-off for reusing existing machinery.
> 
> #### 5. [Nit] `Coordinator.run()` docstring could mention CLOSE routing
> The docstring still lists `target_reached` alongside `time_exhausted (via closing phase)` without noting that `target_reached` now routes through the CLOSE phase sequencer (a different mechanism from `closing_phase` grace). Minor doc sync.
> 
> ### Verdict
> **Approve with minor test fix.** Logic is correct and the change is minimal. Please fix `cumulative_gain_validated` → `cumulative_gain` in the new test (and consider asserting `close_sequence_done`) before merge.

@xiaofei-zheng  Thanks for you review~. I'll read and fix them.

WhatGhost:
@xiaofei-zheng  Thanks for the review. I've read and fixed them. Please review again.

**For 1** — This may be based on an earlier revision. On current `main`, `TargetGainObjective._current()` returns `cumulative_gain_validated` (`objective.py:141-143`), and `test_objective.py:239` sets that same field. I tried the suggestion locally: switching to `cumulative_gain` makes the run exit with `assert 'max_ticks' == 'target_reached'`. Keeping `cumulative_gain_validated` — happy to look again if you're seeing something different.

**For 2** — Agreed, real gap. Added `assert c.shared_state.close_sequence_done is True`, which is the flag the CLI `finally` block uses to decide on the safety-net write.

**For 3** — Reproducible, and a consequence of the `else` branch kept for backward compatibility. Pre-existing, so I'd rather not fold it into this PR. Happy to open a separate issue.


**For 5** — Fixed: `target_reached (via the CLOSE phase sequencer)`, to distinguish it from `time_exhausted`, which goes through `_enter_closing_phase()`.

ZhengGong-amd:
Request changes: this cannot merge yet because the fix relies on phase advancement in the next tick. If the target is reached at or just after the session bound, `_await_within_session_bound()` skips phase advancement; the next `stop_reason` guard then exits directly, so CLOSE is skipped and the CLI safety-net is used again. A `shared_state.save()` failure has the same outcome.

Please route the terminal request into CLOSE in the same tick (with a deadline-aware close mode), rather than relying on `continue`. Add regression coverage for: (1) target reached at the session deadline, and (2) a transient initial state-save failure. Both should assert that the intended close/finalization completion marker and artifacts are produced.

WhatGhost:
@ZhengGong-amd Hi, thanks for your reviews!
You're right on both counts — fixed in `5fb78f70b`.

**Same-tick close.** The transition no longer waits for the next tick. It arms the closing bound first, which swaps `_seconds_until_session_bound()` onto `_closing_deadline` — the mechanism its own docstring describes ("during CLOSE the session deadline has already passed... CLOSE work is not skipped") — then advances the phase through `_await_within_session_bound`, so the sequencer stays bounded.

The window is `max(grace_sec, 120s)`, not `grace_sec` itself: `grace_sec` is `0` on an unbounded run and `max_minutes * 60 * 0.02` otherwise, so it shrinks with the budget. A met target is the success path rather than a rescue, so the close needs a workable window regardless.

**Failed save.** Now best-effort — the terminal is already set in memory and CLOSE persists state itself. The old behaviour was worse than skipping CLOSE: that line sits outside the tick body's `try`, so the exception propagated straight out of `run()`.

**Regression coverage.** Two tests, both asserting `close_sequence_done is True`:

- `test_target_reached_at_session_bound_still_closes` — `max_minutes=0.0001` so the bound is already elapsed when the target is met
- `test_target_reached_closes_despite_a_failed_state_save` — first save after the terminal is set raises

Both confirmed to catch the defect: against the previous implementation they fail with `assert False is True` and an uncaught `OSError` respectively.

Full suite: `9 failed, 14362 passed` — same nine pre-existing failures as this branch's base, passing count up by the two new tests.

tsrikris:
@xiaofei-zheng this seems like a massive PR. Are we sure there is no leaner fix possible?

ZhengGong-amd:
Request changes:

1. Do not set `closing_phase=True` for `target_reached`. That flag selects the deadline-rescue path, so CLOSE skips the post-opt roofline. Use a distinct success-close bound/mode; reserve `closing_phase` for an actual wall-clock timeout.
2. Do not wrap the whole CLOSE sequencer in a fixed 120s timeout. Individual close tasks are explicitly allowed to wait up to 600s; on the 120s timeout this code cancels the sequencer, immediately breaks, and falls back to the CLI safety net.
3. Add coverage for a successful kernel-patch run writing the post-opt roofline, and for a close step exceeding 120s without cancelling the sequencer.
4. Please also fix the branch base: the PR now reports 1,269 changed files and GitHub cannot render the full diff.

WhatGhost:
> Request changes:
> 
> 1. Do not set `closing_phase=True` for `target_reached`. That flag selects the deadline-rescue path, so CLOSE skips the post-opt roofline. Use a distinct success-close bound/mode; reserve `closing_phase` for an actual wall-clock timeout.
> 2. Do not wrap the whole CLOSE sequencer in a fixed 120s timeout. Individual close tasks are explicitly allowed to wait up to 600s; on the 120s timeout this code cancels the sequencer, immediately breaks, and falls back to the CLI safety net.
> 3. Add coverage for a successful kernel-patch run writing the post-opt roofline, and for a close step exceeding 120s without cancelling the sequencer.
> 4. Please also fix the branch base: the PR now reports 1,269 changed files and GitHub cannot render the full diff.

@ZhengGong-amd  Thanks for review again!
the reviews are both correct — fixed in `5d2f09990`. Also rebased onto `66e7c6310`, so the diff is back to 2 files.

**`closing_phase`.** Dropped. A new `_terminal_closing` flag lifts the session bound for that one advance without claiming a rescue is under way, so `_maybe_run_close_post_opt_roofline` no longer takes its early return and the post-opt snapshot lands. Reset in a `finally`.

**Outer bound.** Gone. `_seconds_until_session_bound()` returns `None` while `_terminal_closing` is set, leaving the sequencer's own per-step timeouts as the budget — a 120s outer bound would cancel a step allowed 600s and drop the run onto the safety net, which is what this routing exists to prevent.

**Coverage.**

- `test_target_reached_close_still_runs_the_post_opt_roofline` — seeds an `integrate` entry so the roofline step applies, asserts it observes `closing_phase == False`
- `test_target_reached_close_is_not_cancelled_by_an_outer_bound` — asserts `_seconds_until_session_bound()` is `None` under the terminal close, then that an awaiting advance runs to completion with the session bound already elapsed

Against the previous implementation they fail with `assert ['True'] == ['False']` and `assert -1.0 is None` respectively.

Suite: `10 failed, 15060 passed` — same ten as this branch's base (six isolated-venv probes, three robustness heartbeat, one `test_setup_cli`), passing count up by the two new tests.

PR #1246, opened 2026-08-20
fix(writeback): stack every applied explore winner, not just the best…

WhatGhost:
# fix(writeback): stack every applied explore winner, not just the best one

## Problem

When one explore round produces several winners, they are all applied and all counted in the round's throughput, but only `best_variant` reaches `optimization_stack`. The other winners' args are dropped, so the reported final config cannot reproduce the throughput the report claims.

From a real 3h run:

```
explore_search.accepted        3 entries
optimization_stack             2 entries      <- one lost
final.extra_server_args        "--attention-backend ROCM_AITER_FA --kv-cache-dtype fp8"
                                              <- missing --async-scheduling ...
gain_per_stack_entry[0]        23.32%         <- 6638.2 x 1.2332 = 8186.4
stack[0].tput                  8186.4         <- the *second* winner's throughput
stack[0].extra_server_args     "--attention-backend ROCM_AITER_FA"   <- only the first
```

The second winner's throughput was credited to the first winner's config, and its own args vanished. `winners_count` per round confirms it: the round with 2 winners contributed 1 stack entry, the round with 1 winner contributed 1.

## Root cause

`_promote_explore` records every winner as accepted, but lifts only one:

```python
for winner in winners:
    self.shared_state.record_explore_accepted(accepted)   # every winner

# 4. Lift the best winner into current_best / optimization_stack.
if isinstance(best_winner, dict) and ...:
    promoted = self._lift_to_current_best("explore", float(best_tput), best_winner, ...)
```

The two arguments come from different things: `best_tput` is `result["output_throughput"]` — the whole round's cumulative throughput — while `best_winner` is `max(winners, key=gain_pct)`, a single variant.

Winners in a round are cumulative, not alternatives. In `explore.py`:

```python
keep_entry["tput"] = stack_rebench_tput    # 1780, overwritten after a stable KEEP
running_base_tput = stack_rebench_tput     # 1778, base advances with it
output_throughput = float(running_base_tput)   # 2016
# comment: "Each KEEP advances running_base_tput, so this is the final stack."
```

The second variant is stack-rebenched on top of the first, so the round's throughput reflects both. Crediting it to one variant's config is what breaks the report.

## Fix

Lift every winner, each with its own post-apply `tput`:

```python
for winner in winners:
    entry = dict(winner)
    entry_tput = entry.get("tput")
    if not valid(entry_tput):
        entry_tput = best_tput      # payloads without a per-winner tput
    if self._lift_to_current_best("explore", float(entry_tput), entry, ...):
        promoted = True
```

Since `keep_entry["tput"]` tracks `running_base_tput`, the per-winner values rise in application order, so each one clears the anchor check in `_lift_to_current_best` (`if anchor > 0 and best_tput <= anchor: return False`). `_lift_to_current_best` already accumulates `extra_server_args` from the previous `current_best`, so stack entries compose into the full config on their own.

The fallback keeps older payloads unchanged: with one such winner the behaviour is identical to before, and with several the anchor check keeps the first — the pre-existing outcome. `best_tput` is still used for `_update_cumulative_gain_validated`, which correctly wants the round's final stack throughput.

## Verification

**Unit test.** `test_promote_explore_stacks_every_applied_winner` feeds two winners whose args must both land in the stack. Reverting only the implementation confirms it catches the bug:

```
without fix:  FAILED  assert '--flag-b' in '--flag-a'
              stack: [{'variant_name': 'v1', 'extra_server_args': '--flag-a', 'tput': 130.0}]
              (tput is the cumulative value, args are one variant's)
with fix:     PASSED
```

**Full suite.** `9 failed, 14360 passed, 32 skipped` — the same 9 pre-existing failures as `main`; passing count goes 14359 -> 14360 (the new test).

### DIFF
--- a/src/hyperloom/inference_optimizer/tests/test_promote_shared_state_lock.py
+++ b/src/hyperloom/inference_optimizer/tests/test_promote_shared_state_lock.py
@@ -1616,3 +1616,43 @@ async def test_integrate_keep_lets_a_tuning_env_delta_win(session_dir):
     )
 
     assert s.current_best["extra_envs"] == {"KEEP_ME": "1", "TUNED": "new"}
+
+
+@pytest.mark.asyncio
+async def test_promote_explore_stacks_every_applied_winner(session_dir):
+    """Every winner applied in a round must reach optimization_stack.
+
+    Winners in one round are applied cumulatively: ``running_base_tput``
+    advances with each in-batch KEEP, so the second variant is stack-rebenched
+    on top of the first and the round's ``output_throughput`` reflects both.
+    Recording only ``best_variant`` therefore credits the round's cumulative
+    throughput to a config that is missing the other winners' args.
+    """
+    coord = _coord(session_dir)
+    s = coord.shared_state
+    s.baseline_tput = 100.0
+    w1 = {"name": "v1", "fingerprint": "f1", "tput": 120.0, "extra_server_args": "--flag-a"}
+    w2 = {"name": "v2", "fingerprint": "f2", "tput": 130.0, "extra_server_args": "--flag-b"}
+
+    await coord._promote_to_shared_state(
+        "explore",
+        {
+            "explore_search_update": {},
+            "winners": [w1, w2],
+            "round_id": "r1",
+            "best_variant": w1,
+            "output_throughput": 130.0,
+            "best_gain_pct": 20.0,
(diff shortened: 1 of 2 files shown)


WhatGhost:
colse this PR because of this issue has been fixed in main

PR #1249, opened 2026-08-20
fix(kernel-agent): stop the KERNEL phase acting on unusable profile traces

jiaqiang-dot-liu:
## Why

Four GLM-5.2 / Qwen3.8 sessions on MI355X/SGLang lost their whole KERNEL phase, each to a different defect, and in two of them nothing in the pipeline noticed. One burned 4.6 h — 40% of the session — analysing a trace that was never a workload timeline.

**Supersedes #1248**: its commit is in this branch's history unchanged, then extended. #1248 is closed so one PR carries the whole incident.

## Fix 1 — gate on the GPU compute share

Session `20260818T062821Z` profiled an 18186.60 ms window holding **725.85 ms of compute (3.99%)**, 0.02% idle, 95.99% exposed communication. One `aiter::cross_device_reduce_2stage` invocation ran **17456.08 ms** — 99.99% of all collective time — while the other 63 invocations of the same kernel had a median of **14.3 µs**.

That event is cross-rank arrival skew at the profiler-start boundary. Across the eight per-rank traces every rank exits that collective within 1 µs of the others while entry timestamps span **17.46 s**, and each rank's background-thread spans show the profiler windows ending together but starting 17.46 s apart. Steady state has no skew: after it retires, 726.7 ms of decode runs with ~2.4 ms of collectives total.

A custom all-reduce polls peer signals from inside the kernel, so the wait is charged as GPU-**busy**. Every existing gate defends only against "GPU too idle", so the window passed all three as maximally healthy:

| Gate | Judgement | Measured | Verdict |
|---|---|---|---|
| N25 `steady_state_chunk_empty` | `num_gpu_events > 0 and gpu_busy_duration > 0` | 73056 / 18183589 µs | pass |
| N36 `steady_state_chunk_low_quality` | `busy_ratio >= 0.05` | **0.99994** | pass |
| idle gate | `idle_pct <= 80.0` | **0.02%** | pass |

`trace_health_warnings` came back empty. `busy_ratio` is inverted in this regime: the worse the skew, the healthier the chunk looks.

Added a low-compute gate beside the idle gate — both answer the same question (can a kernel rewrite move end-to-end latency here) and a rewrite is bounded by the compute share. It reuses the proven idle-gate response rather than inventing a mechanism: suppress `hot_kernels[]`, emit `low_gpu_compute_pct`, let the Coordinator route to comm/params.

- Default 10%, via `HYPERLOOM_TRACELENS_MIN_COMPUTE_PCT_THRESHOLD`, deliberately below the 20% that would exactly mirror the idle gate.
- **Fails open.** Skipped under graph under-recording, and skipped when the share is unknown. Reading the number distinguishes absent / blank / unparseable from a real zero, because this gate fires *below* its threshold: a defaulted `0` would suppress the hot-kernel list on every trace, silently, with `status` still `ok`. Share-column spellings get the same tolerance the row labels already needed, and `float(None)` on a short row raises `TypeError`, which the inherited `except ValueError` did not catch.
- Wired into the deterministic route (`gpu_timeline.csv`) and the agent route (`analysis.md`).
- Numeric fields of *every* trace-health warning now render from one table, so the compact prompt line carries `compute=` / `exposed_comm=` instead of a bare code. Telling a comm-bound window from a host-bound one is exactly what the Coordinator uses those two for.

## Fix 2 — publish `gpu_pct` on one basis, and say which

`ops_summary.csv` normalizes against total *op* time (collectives excluded) while `parse_analysis_md` reports an end-to-end share — and both wrote `gpu_pct`. In round 1 of the same session that put `hipLaunchKernel` at "23.495%" (compute basis) above `main_kernel` at "4.681%" (e2e basis) in one ranked list; the true e2e shares were 0.54% and 4.68%, so the ranking was inverted.

`gpu_pct` is now normalized against the trace wall span when `gpu_timeline.csv` supplies it, and every candidate carries `gpu_pct_basis` (`e2e_window` / `sidecar_reported` / `listed_ops_sum`). A batch that ends up spanning two bases — reachable when only some ranking rows carry an absolute GPU time — is called out in the log, since that is the unrankable state the label exists to catch.

Two honest limits:

- **Admission is deliberately *not* moved onto the new basis.** `min_gpu_pct` (10%) is a tuned constant meaning "a large slice of the GPU work we know about"; re-pointing it at the wall span raises the bar by the reciprocal of the compute share, and on an ordinary 60%-compute trace with work spread over eight kernels that takes this recovery path from eight candidates to zero. The floor stays on the share it was tuned against; only the published number moves. Consequence: on `20260818T062821Z` the four fallback candidates are still *recovered*, now reported at their true 0.67 / 0.54 / 0.45 / 0.40% — that session ends up with an empty `hot_kernels[]` because of Fix 1, not because of Fix 2.
- **This bypasses rather than repairs the alias collision.** `_RANK_PCT_KEYS` still maps both `% of compute time` and `%e2e` onto one field, so a run with no `gpu_timeline.csv` falls back to the sidecar's number and carries the original mixed-basis hazard — now labelled `sidecar_reported` rather than silent. Splitting those aliases into separate fields is the real repair and is not done here.

## Fix 3 — classify capture sidecars by shape

Session `20260814T163244Z` had a **healthy** serving trace: 20096 reduce invocations per rank, max 0.05 ms, median 12–13 µs, no skew. It still lost its kernel phase, because Hyperloom analysed the wrong file.

Discovery ranks candidates into buckets and breaks ties by **descending size**. An SGLang without the profiler patch writes `graph_capture_profile/cuda_graph_capture-<runner>-TP-<n>`, matching neither the `capture_traces/` directory name nor a start-anchored filename test — the `cuda_` prefix defeats the anchor. The sidecars landed in the **default bucket alongside the workload traces**, and at 103.87 MB vs 20.03 MB the capture won the tie:

```
trace_input_type=capture_dir
trace_files=16
trace_gpu_kernel_events=186795 (probed=cuda_graph_capture-DecodeCudaGraphRunner-TP-3.json.gz=186795)
```

The probe stops on the first candidate with kernels, so it never reached the serving trace. The splitter cut a bs=1/conc=1 window with zero GPU events and N25 correctly refused — but the kernel track (`kernel_opt_attempts`, `last_gemm_tuning`, `last_fusion` all empty) was already lost.

The size tie-break exists precisely as a name-independent safety net; its docstring says *"ordering by descending size puts the right file first even for a fragment shape nobody has seen yet."* Here the unrecognised file is the larger one, so the safety net selected the wrong trace.

Matching is now by shape: `graph_capture` anywhere in a filename, `capture_traces` exactly or a name starting with `graph_capture` for a directory, keeping #1248's anchored `bs_<digits>` guard.

| Layout | before | after |
|---|---|---|
| `capture_traces/bs_64_rank7.json.gz` | detected | detected |
| `graph_capture_rank0.json.gz` (vLLM) | detected | detected |
| `graph_capture_profile/cuda_graph_capture-*` | **missed** | detected |
| `graph_capture_profile/DecodeCudaGraphRunner_bs_104_rank0.json.gz` | **missed** | detected (directory rule) |
| `<ts>-TP-0.trace.json.gz`, `rank_0.trace.json.gz`, `merged-*` | kept | kept |
| `torch_profiler_with_graph_capture/rank_0.trace.json.gz` | kept | kept (directory rule is anchored) |

The directory half is anchored while the filename half is not: a filename carries the `cuda_` prefix, but a directory is named for what it holds, and an unanchored token there would condemn a descriptive sibling — which the capture-only preflight's `all(...)` turns into a rejection of the whole input.

`discover_capture_folder` matches by the same shape, so `graph_capture_profile/` is *located* rather than merely demoted. Without that a run could pick the right workload trace and still lose `--capture_folder`.

## Fix 4 — reject a capture-only trace input (from #1248)

A directory holding nothing but capture sidecars is refused before analysis, instead of producing an empty chunk and a `trace_split_no_steady_state` that reads as "the window was too short" and sends the next reader to lengthen a capture that was never a workload timeline.

## Refactor — one capture rule, four call sites

Each trace-analysis path carried its own copy and the copies **had already drifted**: the bypass reader never learned `graph_capture` names at all, and `bypass_trace_analysis` was start-anchored so a `cuda_graph_capture-*` shard was skipped and its shapes never indexed. The rule now lives in `_capture_shapes` (stdlib-only, so the bypass reader stays independent of TraceLens, mirroring `_idle_gate`) and is used by `tracelens_analysis`, `_bypass_trace_reader`, `bypass_trace_analysis`, `tracelens_skill_runner`, `profile.py` and `request_handlers`.

## Blast radius

- `ta_result.status` stays `ok` with empty `hot_kernels` — identical in shape to what the idle gate has always produced, so no new retry path is reachable. `cuda_graph_attribution_degraded` is guarded on `per_kernel_attribution_degraded` (not merely `not hot`) and the multi-node compute-bound re-profile on `_trace_is_high_idle`, so neither misfires.
- Widening the capture classifier only ever *demotes* a candidate, and the capture-only preflight uses `all(...)`, so a mixed directory is never rejected.
- `_VARIANT_RE` in `bypass_trace_analysis` is unanchored too, so `DecodeCudaGraphRunner_bs_104_rank0` now yields the `bs_104` variant label instead of falling through to the bare stem.
- Low-compute gate is TraceLens-route only: `_bypass_report` reports `exposed_comm_pct: None`, so the bypass reader cannot measure a compute share and keeps the idle gate alone. Documented in the `_idle_gate` docstring.

## What this does *not* fix

- Two Qwen3.8 sessions (`20260814T175123Z`, `20260817T220229Z`) stay broken. Their root cause is **graph under-recording**: the profiler captured GPU activity for **1 of 128** graph replays (`graph_launches_with_kernels: 1`, GPU-side `step[DECODE]` annotation 1/128, 3,604 kernel events against ~293,000 on a comparable healthy trace). Hyperloom's own detector identifies this correctly and `cuda_graph_attribution_degraded` already carries the right remedy (`--enforce-eager`), but N25 raises first and `roofline.py` returns early on `status != "ok"`, so neither warning is ever emitted. Wiring that up is the obvious next PR.
- sglang's `/start_profile` has no cross-rank barrier before the first forward (Fix 1's upstream trigger). A barrier alone is not sufficient — it must be host-side and placed after the expensive profiler preparation, or the wait just moves into another spin-waiting kernel.
- Hyperloom's heaviest profiler options inflate per-rank start cost and therefore the skew.
- The Coordinator reads only compute-category reasoning-candidate blocks, so TraceLens's system-tier P-item — which diagnosed this arrival skew correctly and recommended trimming the window — is never consumed.

## Test plan

Whole suite, branch vs `origin/main`, same environment. `PYTHONPATH=$PWD/src` so subprocess-spawning tests can import the package (it is not pip-installed here); without it 9 further tests fail on both sides for that reason alone.

```
branch:  6 failed, 13318 passed, 32 skipped, 15 subtests passed in 878.48s
```

The 6 failures reproduce identically on `origin/main` (verified in a clean worktree at `2f681a05d`: `6 failed, 54 passed`), all in isolated-venv version detection this PR does not touch:

- `test_common_provenance.py::test_a_framework_in_its_own_venv_is_still_versioned`
- `test_common_provenance.py::test_a_venv_root_that_is_not_there_is_not_a_failure`
- `test_manifest_unit.py::test_manifest_versions_a_framework_installed_in_its_own_venv`
- `test_server_patcher_vllm_isolated_venv.py` (3 tests)

**No test regressions.** Coverage added by this PR:

- Fix 1: threshold resolution (default / env / bad / negative); warning shape with and without comm context; `gpu_timeline.csv` readers across every row-label and column spelling; gate fires on the real numbers (3.99% / 95.99%), passes a healthy window (99.3%), skipped when unknown; `analysis.md` extractors
- Fix 1 fail-open regression: unknown column, truncated row, blank cell and non-numeric cell each yield `None` and leave candidates untouched
- Fix 2: e2e normalization when the window total is passed; admission unchanged with and without it; a healthy 60%-compute / eight-kernel trace keeps all eight candidates; sidecar-only and listed-ops fallbacks labelled; mixed-basis batch logged
- Fix 3: classifier parametrized over every observed layout including the two workload shapes and the descriptive-sibling directory it must not touch; discovery regression pinning the 103 MB capture behind the 20 MB trace; `discover_capture_folder` finds `graph_capture_profile/` and ignores the descriptive sibling; bypass-route equivalent
- Renderer: `low_gpu_compute_pct` reaches the prompt with its numbers; both gates render side by side
- `ruff check` clean on every touched file

Known gap: **`main()` wiring for Fix 1 is untested.** Both route branches and the `graph_under_recorded` suppression sit inside `main()` and need a testability refactor first; Fix 3 has `_drive_main_over_capture_dir`, Fix 1 has no equivalent. The gate logic, both extractors and the fail-open contract are covered directly.

### DIFF
--- a/src/hyperloom/agents/kernel/tests/test_bypass_reader.py
+++ b/src/hyperloom/agents/kernel/tests/test_bypass_reader.py
@@ -297,6 +297,23 @@ def test_bs_named_fragment_without_subdir_is_deprioritized(tmp_path):
     assert resolved is not None and resolved.name == main.name
 
 
+def test_unpatched_sglang_capture_dir_is_deprioritized(tmp_path):
+    # Both routes now share one classifier, so the bypass reader also knows the
+    # SGLang-without-profiler-patch layout: a ``graph_capture_profile/`` holding
+    # ``cuda_graph_capture-*``. The reader's own copy of the rule only knew
+    # ``bs_<n>_rank<n>`` / ``capture_traces/`` and took the sidecar as a trace.
+    d = tmp_path / "torch_trace"
+    main = _write_main_tp_trace(d)
+    cap_dir = d / "graph_capture_profile"
+    cap_dir.mkdir(parents=True, exist_ok=True)
+    _write_capture_fragment(cap_dir, 512)
+    (cap_dir / "cuda_graph_capture-DecodeCudaGraphRunner-TP-3.json.gz").write_bytes(
+        (cap_dir / "bs_512_rank0.json.gz").read_bytes()
+    )
+    resolved = reader.resolve_trace_file(d)
+    assert resolved is not None and resolved.name == main.name
+
+
 def test_capture_traces_detection_is_relative_to_trace_root(tmp_path):
     # An unrelated ancestor dir named ``capture_traces`` above the trace root must
     # not flag the main trace; only a genuine subdir within the root marks shards.
--- a/src/hyperloom/agents/kernel/tests/test_bypass_trace_analysis.py
+++ b/src/hyperloom/agents/kernel/tests/test_bypass_trace_analysis.py
@@ -816,6 +816,38 @@ def test_discover_capture_shards_dedups_tp_ranks(tmp_path):
     assert len(shards) == 2
 
 
+def test_discover_capture_shards_dedups_a_runner_prefixed_batch(tmp_path):
+    # An SGLang without the profiler patch prefixes the runner name:
+    # DecodeCudaGraphRunner_bs_104_rank0. The batch token is still what
+    # identifies the variant, so the ranks must collapse the same way — an
+    # anchored read finds nothing here and falls back to the per-rank stem,
+    # which silently turns one variant into one-per-rank.
+    d = tmp_path / "graph_capture_profile"
+    d.mkdir()
+    for f in (
+        "DecodeCudaGraphRunner_bs_104_rank0.json",
+        "DecodeCudaGraphRunner_bs_104_rank1.json",
+        "DecodeCudaGraphRunner_bs_8_rank0.json",
+    ):
+        (d / f).write_text("{}", encoding="utf-8")
+    shards = bta._discover_capture_shards(str(d), str(d))
+    assert sorted(lbl for _p, lbl, _m in shards) == ["bs_104", "bs_8"]
+
+
+def test_discover_capture_shards_skips_a_whole_capture_per_rank_file(tmp_path):
+    # ``cuda_graph_capture-<runner>-TP-<n>`` is one file per rank holding every
+    # batch size at once, so it carries no variant identity. Indexing it mints a
+    # bogus variant per rank (TP=8 -> eight shape-identical entries) because the
+    # label falls back to the stem, which differs per rank. It is a capture
+    # sidecar — the shared classifier is right to demote it from analysis — but
+    # that is a wider question than "does this shard name a variant".
+    d = tmp_path / "graph_capture_profile"
+    d.mkdir()
+    for rank in range(3):
+        (d / f"cuda_graph_capture-DecodeCudaGraphRunner-TP-{rank}.json").write_text("{}", encoding="utf-8")
+    assert bta._discover_capture_shards(str(d), str(d)) == []
+
+
 # ── CUDA/HIP graph-mode under-recording ──────────────────────────────────────
 
 
--- a/src/hyperloom/agents/kernel/tests/test_idle_gate.py
+++ b/src/hyperloom/agents/kernel/tests/test_idle_gate.py
@@ -5,10 +5,10 @@
 # See LICENSE for license information.
 ###############################################################################
 
-"""Unit tests for the shared idle-gate helpers (_idle_gate).
+"""Unit tests for the shared trace-health gate helpers (_idle_gate).
 
 Locks threshold resolution (default + env override + bad-value fallback) and
-the high-idle warning shape.
+the warning shapes for both the high-idle and low-compute gates.
 """
 
 from __future__ import annotations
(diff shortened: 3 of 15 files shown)


xiaofei-zheng (approved):

tsrikris:
@devalshahamd I believe this sanity check was planned on TL anyways right?

PR #1252, opened 2026-08-20
fix(breakdown): put GEAK's kernel journey on the canonical streams

### LINKED ISSUE #1251: GEAK's kernel wins never reach the stream the dashboard reads (route_strategy=legacy_only)
## Summary

The TOP Model dashboard's **Kernel Agent Gain (GEAK)** column reads zero for every session, including sessions where GEAK optimized a kernel, the A/B passed, and the workload got faster. The kernels ran and won. Nothing wrote it down.

## The column and where it comes from

`Kernel Agent Gain (GEAK)` is `summary_by_source["kernel_agent"]["by_backend"]["geak"]`, built by `collect_recorded_optimizations` from the recorder's `operations` and `adoptions` streams.

`_record_geak_kernel_journey` (`src/hyperloom/orchestrator/phases/kernel.py`) replays GEAK's `kernel_journey.json` into those streams. All five of its recorder calls pass `route_strategy="legacy_only"`:

| line | call |
|---|---|
| 1660 | `record_kernel_discovery` |
| 1680 | `record_kernel_dispatch` |
| 1687 | `record_kernel_backend_result` |
| 1702 | `record_kernel_e2e` (accept path) |
| 1785 | `record_kernel_e2e` (revert path) |

Three of those entry points return **before** writing the operation and the adoption:

- `instrument.py:2767` — `record_kernel_discovery`
- `instrument.py:2918` — `record_kernel_dispatch`
- `instrument.py:3422` — `record_kernel_e2e`

The comment at 3422 states the consequence exactly:

> No operation and no adoption for this integrate. On a KEEP that is a change the workload carries with nothing on the ledger claiming it, which is precisely the shape a lost write leaves.

So a won kernel and a lost write are indistinguishable, and GEAK produces the first one on every session.

Every non-test `legacy_only` caller in the orchestrator is this function. No other producer is affected.

## Evidence — replay of real campaign sessions

Real `kernel_journey.json` from `/shared_nfs/hyperloom-claw/`, replayed through the **real** `instrument.record_kernel_*`, the **real** `assemble_parts`, and the **real** `collect_recorded_optimizations`. Only `route_strategy` differs between the arms.

| session | GEAK KEEP kernels | credited today | credited with route fixed |
|---|---:|---:|---:|
| `GLM-5.2-MXFP4/20260814T163244Z` | 3 | **0** | **3** |
| `Kimi-K3/20260816T122327Z` | 2 | **0** | **2** |
| `Llama-3.1-8B-Instruct/20260816T114410Z` | 1 | **0** | **2** |
| `MiniMax-M3-MXFP8/20260801T023259Z` | 4 | **0** | **4** |
| `Mixtral-8x7B-Instruct-v0.1/20260730T180726Z` | 2 | **0** | **2** |
| `Qwen3-0.6B/20260730T213628Z` | 3 | **0** | **3** |
| `Qwen3-0.6B/20260805T130336Z` | 1 | **0** | **1** |
| `Qwen3-14B-FP8/20260730T184453Z` | 1 | **0** | **1** |
| `Qwen3-14B-FP8/20260814T163051Z` | 2 | **0** | **2** |
| `Qwen3-14B-FP8/20260816T050457Z` | 3 | **0** | **3** |
| `Qwen3.5-122B-A10B-FP8/20260801T034513Z` | 2 | **0** | **2** |
| `Qwen3.5-122B-A10B-FP8/20260805T080211Z` | 2 | **0** | **2** |
| `gemma-4-26B-A4B-it/20260814T155153Z` | 2 | **0** | **2** |
| `gemma-4-26B-A4B-it/20260815T130915Z` | 5 | **0** | **5** |
| `gemma-4-26B-A4B-it/20260816T112750Z` | 2 | **0** | **2** |
| **total** | **35** | **0** | **36** |

Operations assembled: **0 in 15 of 15 sessions today**, 15 of 15 populate with the route fixed. `geak` total gain **+0.000 in all 15** today.

(Llama shows 1 → 2 because the KEEP column counts only strictly-positive gains while the collector counts every adopted step.)

The largest single win currently invisible is `dsa_sparse_attn_prefill_main_kernel` on `GLM-5.2-MXFP4/20260814T163244Z`, **+29.994%** e2e, parity pass.

## Fix

Drop the `route_strategy="legacy_only"` argument at all five sites and let the default `kernel_agent_forge` apply. The route names the *stream*, not the vendor; the `geak` backend label is resolved separately from the journey's `backend_result` and is unaffected.

The revert path (1785) must change with the others, or a kernel that was taken back out keeps credit it was given.

## Scope note

This makes GEAK **visible**. It does not by itself make the number **right**. The journey publishes `e2e_gain_pct` with no throughput pair (36/36 KEEP blocks), so the collector cannot state each win in points of the session baseline and falls back to summing percentages that were measured against a moving one. Route-fix-alone credits **+348.6 pp** across these 15 sessions, which is a unit error, not a gain. That half is a GEAK-side defect and is filed separately.

## Related

- Second, larger blocker: GEAK has produced **0 kept kernels since 17 August** (406 hot kernels found, none optimized). Fixing reporting alone will not lift the current window off zero. Being investigated separately.

iraj465:
Fixes #1251.

## What was wrong

`_record_geak_kernel_journey` replayed every GEAK kernel with `route_strategy="legacy_only"`. `record_kernel_discovery`, `record_kernel_dispatch` and `record_kernel_e2e` all return **before** writing the operation and the adoption on that route.

The dashboard's GEAK column is `summary_by_source["kernel_agent"]["by_backend"]["geak"]`, which `collect_recorded_optimizations` builds from exactly those two streams. So a session finished with a kept, validated, positive-gain kernel and nothing on the ledger naming it — the same shape a lost write leaves.

## The change

Five deletions. Drop `route_strategy="legacy_only"` and let the default `kernel_agent_forge` apply. The route names the *stream*, not the vendor; the `geak` backend label is resolved from the journey's `backend_result` and is unaffected.

The revert path changes with the accept path, or a kernel taken back out keeps credit it was given.

## A/B on real campaign logs

Real `kernel_journey.json` from `/shared_nfs/hyperloom-claw/`, through the **real** `instrument.record_kernel_*`, the **real** `assemble_parts`, and the **real** `collect_recorded_optimizations`. Only `route_strategy` differs between arms.

| session | GEAK KEEP kernels | credited today | credited after |
|---|---:|---:|---:|
| `GLM-5.2-MXFP4/20260814T163244Z` | 3 | **0** | **3** |
| `Kimi-K3/20260816T122327Z` | 2 | **0** | **2** |
| `Llama-3.1-8B-Instruct/20260816T114410Z` | 1 | **0** | **2** |
| `MiniMax-M3-MXFP8/20260801T023259Z` | 4 | **0** | **4** |
| `Mixtral-8x7B-Instruct-v0.1/20260730T180726Z` | 2 | **0** | **2** |
| `Qwen3-0.6B/20260730T213628Z` | 3 | **0** | **3** |
| `Qwen3-0.6B/20260805T130336Z` | 1 | **0** | **1** |
| `Qwen3-14B-FP8/20260730T184453Z` | 1 | **0** | **1** |
| `Qwen3-14B-FP8/20260814T163051Z` | 2 | **0** | **2** |
| `Qwen3-14B-FP8/20260816T050457Z` | 3 | **0** | **3** |
| `Qwen3.5-122B-A10B-FP8/20260801T034513Z` | 2 | **0** | **2** |
| `Qwen3.5-122B-A10B-FP8/20260805T080211Z` | 2 | **0** | **2** |
| `gemma-4-26B-A4B-it/20260814T155153Z` | 2 | **0** | **2** |
| `gemma-4-26B-A4B-it/20260815T130915Z` | 5 | **0** | **5** |
| `gemma-4-26B-A4B-it/20260816T112750Z` | 2 | **0** | **2** |
| **total** | **35** | **0** | **36** |

| metric | today | after |
|---|---:|---:|
| sessions assembling any operation | **0 / 15** | **15 / 15** |
| kept kernels credited to `geak` | **0** | **36** |
| kept kernels credited to `forge` | 0 | 0 |
| kept kernels left `unattributed` | 0 | 0 |

Largest win currently invisible: `dsa_sparse_attn_prefill_main_kernel` on GLM-5.2-MXFP4, **+29.994%** e2e, parity pass.

## Tests

New: `src/hyperloom/inference_optimizer/tests/test_geak_reaches_dashboard.py` (3 tests), pinning the path end to end through the real recorder and real collector.

| test | on `main` | on this branch |
|---|---|---|
| `test_kept_geak_kernel_reaches_the_kernel_agent_column` | **FAIL** (`geak keeps == 0`) | pass |
| `test_gain_is_stated_in_points_of_the_session_baseline` | **FAIL** (`geak keeps == 0`) | pass |
| `test_reverted_geak_kernel_is_not_credited` | pass | pass |

The third asserts an absence, so it holds both ways by construction; it guards the revert path against over-crediting once the other two pass.

Related suite: **552 passed, 1 failed**. The one failure is `test_coverage_gap_units.py::test_credentials_endpoint_resolution_and_geak_sync`, which asserts on `credentials._resolve_llm_endpoints` and fails identically with this change stashed. Pre-existing and unrelated.

## Scope

This makes GEAK **visible**, not yet **correctly sized**. The journey publishes `e2e_gain_pct` with no throughput pair (36/36 KEEP blocks), so the collector cannot state each win in points of the session baseline and falls back to summing percentages measured against a moving one — route-fix-alone reads **+348.6 pp** across these 15 sessions, which is a unit error. That half is GEAK-side and filed separately; the collector already has the arithmetic and only lacks the inputs.

Separately, GEAK has produced **0 kept kernels since 17 August**, so this will not lift the current window off zero on its own.

### DIFF
--- a/src/hyperloom/inference_optimizer/breakdown/collectors/optimizations.py
+++ b/src/hyperloom/inference_optimizer/breakdown/collectors/optimizations.py
@@ -85,9 +85,9 @@ def _empty_summary() -> dict[str, Any]:
         for source in _SOURCES
     }
     summary["kernel_agent"]["by_backend"] = {
-        "geak": {"keeps": 0, "total_gain_pct": 0.0},
-        "forge": {"keeps": 0, "total_gain_pct": 0.0},
-        "unattributed": {"keeps": 0, "total_gain_pct": 0.0},
+        "geak": {"keeps": 0, "total_gain_pct": 0.0, "non_attributable_keeps": 0},
+        "forge": {"keeps": 0, "total_gain_pct": 0.0, "non_attributable_keeps": 0},
+        "unattributed": {"keeps": 0, "total_gain_pct": 0.0, "non_attributable_keeps": 0},
     }
     return summary
 
@@ -1148,6 +1148,26 @@ def collect_recorded_optimizations(
         for attempt in attempts
         if attempt.get("adopted") and attempt.get("attribution_eligible") is False
     ]
+    # `entries` is the GAIN ledger and deliberately holds only attributable keeps, so a backend whose
+    # keeps are all non-attributable reads as zero keeps — indistinguishable from an optimizer that
+    # produced nothing. That is the same conflation the canonical-stream fix set out to remove, one
+    # layer further in. Count the keep from the attempts, keep the gain coming from the entries.
+    for attempt in non_attributable:
+        if str(attempt.get("agent") or "") != "kernel_agent":
+            continue
+        by_backend = summary_by_source.setdefault(
+            "kernel_agent", {"keeps": 0, "total_gain_pct": 0.0}
+        ).setdefault("by_backend", {})
+        backend = str(attempt.get("backend") or "unattributed")
+        if backend not in by_backend:
+            backend = "unattributed"
+        backend_bucket = by_backend.setdefault(
+            backend, {"keeps": 0, "total_gain_pct": 0.0, "non_attributable_keeps": 0}
+        )
+        backend_bucket["keeps"] += 1
+        backend_bucket["non_attributable_keeps"] = (
+            int(backend_bucket.get("non_attributable_keeps", 0)) + 1
+        )
     # An adopted step that contributes nothing to the total is a hole in the
     # accounting, not a zero. Counting it keeps the sum honest about what it
     # could not see.
--- a/src/hyperloom/inference_optimizer/breakdown/recorder/instrument.py
+++ b/src/hyperloom/inference_optimizer/breakdown/recorder/instrument.py
@@ -1279,6 +1279,33 @@ def _kernel_selection_operation_id(
     return _stable_id("op", "kernel_optimizer_selection", context_key, discriminator)
 
 
+# Canonical kernel routes: the route operation every kernel record hangs under, and the strategy
+# stamped on the kernel operation itself. Two entries today, and the pair is derived rather than
+# written at each call site because the forge identity used to be a literal in nine places — a tenth
+# reader was one edit away from silently parenting somebody else's kernels under Forge.
+#
+# A GEAK kernel replayed from its kernel_journey must NOT hang under the Forge route: the tree would
+# then assert it ran beneath a route that never dispatched it, and a reader walking parents to answer
+# "which optimizer produced this?" gets the wrong answer. The `geak` route operation already exists
+# (see the GEAK dispatch writer below); this makes the replay reachable to it.
+CANONICAL_KERNEL_ROUTES: dict[str, tuple[str, str]] = {
+    # route_strategy -> (route operation name, strategy stamped on the kernel operation)
+    "kernel_agent_forge": ("kernel_agent_forge", "forge"),
+    "geak": ("geak", "geak"),
+}
+
+
+def _canonical_route(route_strategy: str | None) -> tuple[str, str]:
+    """(route operation name, kernel strategy) for a canonical route.
+
+    Unknown values fall back to the forge pair, which is what every caller got before routes were
+    named, so an unrecognised string cannot silently drop a kernel off the streams.
+    """
+    return CANONICAL_KERNEL_ROUTES.get(
+        str(route_strategy or ""), CANONICAL_KERNEL_ROUTES["kernel_agent_forge"]
+    )
+
+
 def _kernel_route_operation_id(
     session_dir: Path | str,
     strategy: str,
@@ -1526,19 +1553,21 @@ def record_native_kernel_run_start(
     payload: Mapping[str, Any] | None = None,
     macro_cycle: int | None = None,
     route_operation_id: str = "",
+    route_strategy: str = "kernel_agent_forge",
     producer: str = PRODUCER_KERNEL_AGENT,
 ) -> None:
     """Upsert the selected native Kernel Agent plus Forge route as running."""
+    route_name, _ = _canonical_route(route_strategy)
     if not session_dir:
         trace_skip(reason="no session_dir", section="kernel_journey")
         return
     current_route_id = _KERNEL_ROUTE_CONTEXT.get(
         _kernel_route_context_key(session_dir),
         {},
-    ).get("route:kernel_agent_forge", "")
+    ).get(f"route:{route_name}", "")
     route_id = route_operation_id or _kernel_route_operation_id(
         session_dir,
-        "kernel_agent_forge",
+        route_name,
         macro_cycle=macro_cycle,
         run_discriminator=""
         if current_route_id
@@ -1549,19 +1578,19 @@ def record_native_kernel_run_start(
         operation_id=route_id,
         producer=producer,
         kind="kernel_optimizer_run",
-        name="kernel_agent_forge",
+        name=route_name,
         phase="KERNEL_AGENT",
         scope="run",
         strategy_group="kernel_optimizer",
-        strategy="kernel_agent_forge",
-        executor_class="deterministic",
+        strategy=route_name,
+        executor_class="llm_tool" if route_name == "geak" else "deterministic",
         status="running",
         started_at=_now_iso_safe(),
         inputs=dict(payload or {}),
         subject={
             "subject_id": _kernel_route_subject_id(
                 session_dir,
-                "kernel_agent_forge",
+                route_name,
                 route_operation_id=route_id,
             ),
             "subject_type": "kernel_optimizer_route",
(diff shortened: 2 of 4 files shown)


iraj465:
Paired with AMD-AGI/GEAK#427 (issue AMD-AGI/GEAK#426).

This PR makes GEAK's records **visible**; #427 makes them **composable**. Measured on the same 36 KEEP blocks in `/shared_nfs/hyperloom-claw`:

| | credited | `baseline_chain` steps |
|---|---:|---:|
| today | 0 | 0 |
| this PR alone | 36 | 0 |
| this PR + GEAK#427 | 36 | 23 |

Without #427 the collector falls back to summing percentages measured against different denominators. Both are needed before the dashboard column is non-zero **and** correct.

zihaoanllm (approved):

luochen-amd:
Thanks for putting together the canonical-stream fix. The visibility diagnosis is correct, and the KEEP/REVERT replay coverage is valuable. I found two items that should be addressed before merge:

1. **Do not make GEAK's local percentages attributable when the journey has no throughput pair.**

   The current campaign artifacts provide `e2e_gain_pct`, but 36/36 KEEP blocks do not provide `base_tput` and `new_tput`. Once the replay writes canonical adoptions, `collect_recorded_optimizations` falls back to projecting those local percentages through the session chain. These percentages were measured against moving GEAK-local baselines and are not additive as session-baseline percentage points; the PR's own replay reports `+348.6 pp`, which is known to be a unit error.

   The canonical operation/adoption should still be written so the KEEP is visible, but it must be marked non-attributable until the same-config throughput pair is available. The local GEAK gain should remain diagnostic evidence and must not contribute to `summary_by_source.kernel_agent.by_backend.geak.total_gain_pct`. The new test currently injects `base_tput/new_tput`, so it validates the future contract rather than the shape of today's real journey files. Please add coverage for a KEEP that has only `e2e_gain_pct`.

2. **Use an explicit canonical GEAK replay route instead of falling through to the default `kernel_agent_forge` route.**

   Removing `route_strategy="legacy_only"` causes `record_kernel_discovery` / `record_kernel_dispatch` to create a parent operation named and tagged as `kernel_agent_forge`. `record_kernel_backend_result` later updates the per-kernel operation strategy to `geak`, so the dashboard GEAK backend bucket can populate, but the resulting tree still claims that the GEAK kernel ran under a Forge parent route. Replays after process restart can also mint additional runtime-keyed Forge route operations.

   Please keep the canonical write, but route it under a stable GEAK parent/root operation with GEAK strategy/provenance. Add an assertion that replayed operations contain the GEAK kernel operation and adoption while containing no `kernel_agent_forge` route.

The intended split should remain:

- GEAK journey replay records kernel identity, attempts, local A/B evidence, and KEEP/REVERT state.
- Hyperloom's bundle-level revalidation remains the source of truth for the session's validated GEAK gain.
- A bundle containing several kernels/config changes may name its members, but its gain must not be copied onto each kernel.

With those two corrections, the PR should fix the zero-visibility defect without replacing it with incorrect gain magnitude or route provenance.

luochen-amd:
Thanks for the follow-up commit. The two main blockers from the previous review are now addressed:

- a KEEP without `base_tput/new_tput` is visible but non-attributable, so its local percentage is not added to the session gain;
- the accepted-kernel replay now uses a canonical GEAK route rather than the default Forge route.

I found two remaining route-provenance details:

1. **The REVERT replay still falls back to the Forge route.**

   The KEEP-side calls now pass `route_strategy="geak"`, but `_reject_geak_kernel_journey()` only removes `legacy_only` and does not pass the GEAK route. Therefore `record_kernel_e2e()` uses its default `kernel_agent_forge` route for a revalidation REVERT. The adoption is revoked correctly, but the kernel operation is re-parented under a synthetic Forge route.

   Please pass `route_strategy="geak"` on the REVERT call as well, and extend the revert test to assert that no `kernel_agent_forge` route operation is created.

2. **The GEAK route subject still names Forge.**

   `record_native_kernel_run_start()` now derives the operation name and strategy from `route_name`, but the subject payload still contains:

   ```text
   name = kernel_agent_forge
   ```

   A GEAK replay therefore produces `operation.name=geak` / `operation.strategy=geak` with `subject.name=kernel_agent_forge`. Please derive the subject name from `route_name` too, and assert the subject identity in the route-provenance test.

With those two small corrections, the canonical gain and route behavior look ready from the Session Breakdown side.

zihaoanllm:
@luochen-amd  Thanks! — both fixed in 752109c6.                                                                                                                                      
                                                                                                                                                                        
  1. Revert path now names the GEAK route. One correction: the symptom doesn't reproduce. record_kernel_e2e mints no route operation on that path — checked after a KEEP  replay and standalone as a restart would replay it; the only route present is geak either way. Real fallback, inert today. Passing it anyway, since a call that         doesn't name its route is one recorder change away from being wrong.                                                                                                  

  2. Route subject now derives its name from route_name — the literal was left behind when the operation's name and strategy were parameterised.

xiaofei-zheng (approved):

Issue #1239, opened 2026-08-20
GEAK rebench is cancelled by the KERNEL->SWEEP phase boundary 0-1s after enqueue


haishuok0525:
## Summary

After GEAK hands back an e2e optimization result, Hyperloom enqueues a rebench on its own harness to validate the gain. Only a passing rebench writes GEAK into `current_best` / `optimization_stack`.

In practice that rebench task is **cancelled within 0-1 second of being created** by the `KERNEL_AGENT -> SWEEP` phase transition. GEAK's gain is never validated and the session ends with `rebench_unavailable`. At the moment of cancellation the session still had **30 minutes to nearly 4 hours** of wall clock left, so this is not a "ran out of time" problem — the phase boundary kills it.

Commit `1a552873c` was intended to fix this, but it only guards the `skip_to_sweep` exit path. GEAK sessions always exit via the wall-clock budget path (`kernel_budget_cap`), which is completely unguarded.

## Symptom

Final session breakdown:

```
geak_pending.status             = "rebench_unavailable"
geak_pending.revalidation_error = "task settled before dispatch (cancelled)"
```

No `geak_e2e` entry in `optimizations`; GEAK contributes nothing to the headline gain.

### 1. The rebench is enqueued as an `explore` task

`src/hyperloom/orchestrator/loop/writeback.py:5067`

```python
task, existing = await self.tasks.create_or_return_existing(
    kind="explore",
    params=params_ps,
    idempotency_key="geak-revalidate",
)
```

`explore` is a kind no later phase (SWEEP / CLOSE) allows, so the phase transition sweeps it unconditionally.

### 2. Only one exit path checks for pending work

`kernel_work_pending()` is the only guard that can hold KERNEL open until the rebench lands, and it is wired only into the `skip_to_sweep` branch.

`src/hyperloom/orchestrator/phases/machine_state.py:2130` (`exit_normal_kernel`)

```python
if _pending_escalate_hint(state) == ESCALATE_HINT_SKIP_TO_SWEEP:
    if not kernel_work_pending(state):
        return "kernel_no_more_leverage", {
```

The two wall-clock budget branches in the same function return unconditionally and never ask whether a rebench is in flight.

`src/hyperloom/orchestrator/phases/machine_state.py:2173`

```python
if remaining is not None and remaining <= 0:
    return "kernel_phase_budget_exhausted", {
        ...
    }
[...]

PR #1258, opened 2026-08-21
fix(geak): close settled revalidation state

### LINKED ISSUE #1240: GEAK revalidate slot is a session-wide singleton; a second request collides with the already-succeeded task
## Summary

The GEAK rebench task uses a **module-level constant** idempotency key, so a session has exactly one revalidate slot for its entire lifetime.

When a second rebench is requested within the same session, `create_or_return_existing` returns the first task — which finished hours earlier — with `state == "succeeded"`. The consumer only whitelists `{queued, running}`, so it treats that as "the task was never dispatched", writes `rebench_unavailable`, and overwrites the correct verdict that had already been recorded.

Net effect: the validation actually ran and produced a verdict, but the final breakdown reports that validation was unavailable.

## Symptom

```
geak_pending.status             = "rebench_unavailable"
geak_pending.revalidation_error = "task settled before dispatch (succeeded)"
```

Note the difference from the phase-boundary bug: the parenthesised state is `succeeded`, not `cancelled`.

### 1. Fixed idempotency key

`src/hyperloom/orchestrator/phases/kernel.py:53`

```python
_GEAK_REVALIDATE_IDEMPOTENCY_KEY = "geak-revalidate"
```

The constant carries no macro_cycle, cycle index, or result fingerprint, and `_enqueue_internal_stack_rebench` uses it on every call:

`src/hyperloom/orchestrator/loop/writeback.py:5067`

```python
task, existing = await self.tasks.create_or_return_existing(
    kind="explore",
    params=params_ps,
    idempotency_key="geak-revalidate",
)
```

### 2. The consumer whitelist rejects settled states

`src/hyperloom/orchestrator/phases/kernel.py:836` (`_enqueue_geak_revalidation`)

```python
task_id = str(summary.get("task_id") or "") if isinstance(summary, dict) else ""
task_state = str(summary.get("task_state") or "queued").strip().lower() if task_id else ""
pending = dict(state.geak_pending) if isinstance(state.geak_pending, dict) else {}
if task_id and task_state in {"queued", "running"}:
    pending["status"] = "awaiting_rebench"
    pending["revalidation_task_id"] = task_id
    state.geak_pending = pending
    state.save(self.session_dir)
    return True

pending["status"] = "rebench_unavailable"
# Drop the reservation placeholder so no stale id outlives the slot.
if pending.get("revalidation_task_id") == _GEAK_REVALIDATE_IDEMPOTENCY_KEY:
    pending.pop("revalidation_task_id", None)
pending["revalidation_error"] = str(
    (summary or {}).get("reason") or f"task settled before dispatch ({task_state or 'unknown'})"
)[:500]
state.geak_pending = pending
state.save(self.session_dir)
```

`succeeded` falls into the else branch and gets described as "settled before dispatch", which is the opposite of what happened: the task was not undispatched, it had already completed.

### 3. Two ways a second request is triggered

A second rebench request does not only come from "GEAK ran another cycle". Both of these were observed:

**(a) GEAK genuinely produced a second result.** Session 97139 emits `geak_e2e_done` three times:

```
2026-08-01T18:18:30  status=ok       speedup=1.0047
2026-08-02T01:06:21  status=ok       speedup=1.0054
2026-08-02T06:26:31  status=no_gain  speedup=1.0
```

**(b) KERNEL is entered a second time and re-recovers the same `result.json` through the crash-recovery path.**

`src/hyperloom/orchestrator/phases/kernel.py:864`

```python
result_path = out_dir / "result.json"
recovered = _read_geak_result(result_path)
# Tombstone: a result already dropped by 2b as no-material must not be
# re-recovered from the stale (still status=ok) result.json, else each
# KERNEL entry re-enqueues a wasted rebench in a loop.
prev_geak = (
    self.shared_state.geak_result
    if isinstance(getattr(self.shared_state, "geak_result", None), dict)
    else {}
)
dropped_no_material = str(prev_geak.get("revalidation_status") or "") == "no_material"
if recovered.get("status") == "ok" and not self._geak_win_already_recorded() and not dropped_no_material:
```

Session 96572 is this case: the rebench completed and was adjudicated at 14:19:28, but `geak_pending` was overwritten with `rebench_unavailable` at **17:12:07**, and there is **no new `geak_e2e_done` event** at that time. The existing tombstone only blocks a previous `no_material` outcome; it does not block "already adjudicated as not beating current_best".

So the accurate statement is: **the revalidate slot is a session-wide singleton and does not support a second revalidate request, whether that request comes from a new GEAK cycle or from a repeated recovery of the same result.**

## Impact assessment

Key fact: the first rebench **ran to completion and was consumed normally**, and its verdict was valid. From the archived `coordinator.db` event stream:

| session | event | measured tput | current_best at the time | verdict |
|---|---|---|---|---|
| 96572 | `geak_no_promote` | 7876.4133 | 7876.9574 | 0.007% below, legitimately rejected by the KEEP guard |
| 96951 | `geak_no_promote` | 9766.7206 | 9860.9467 | 0.96% below, legitimately rejected |

Each session's `tasks` table has exactly **one** row with `idempotency_key='geak-revalidate'`, in state `succeeded`, confirming the singleton slot.

Impact therefore splits into two classes:

**(1) State corruption (primary, 12 of 14).** Validation completed and concluded "did not beat current_best", but the breakdown reports `rebench_unavailable`, implying validation machinery failure. This sends debugging in the wrong direction.

**(2) Real gain loss (secondary, 1 of 14).** Only when the first rebench validated and promoted, and a later GEAK cycle then produced a new win, can that new win never be validated because the slot is taken. Session 97139 is the only such case: its `geak_e2e` is already booked at 2.37%, while the second cycle's +0.54% (speedup 1.0054) was swallowed. Small in magnitude.

> One easily misread detail: in the archive, the `geak_revalidate` variant under `explore_search.tested` shows `outcome=KEEP` and `gain_pct` values such as 11.48% or 36.53%. Those are **cumulative-versus-baseline** figures, and the variant is measuring the then-current `current_best` config itself, so they do not represent a GEAK delta. For 96572, GEAK's real delta was -0.007%.

## Affected sessions

Selection criteria: `geak_pending.revalidation_error` contains `task settled before dispatch (succeeded)`.

| session | commit | started | model | contains `1a552873c` | rebench measured gain (cumulative basis) | variant verdict |
|---|---|---|---|---|---|---|
| 93356 | 78a168c30 | 07-25 07:51 | gpt-oss-120b | no | not in archive | — |
| 91727 | 78a168c30 | 07-25 15:22 | gpt-oss-120b | no | not in archive | — |
| 92072 | 78a168c30 | 07-25 15:23 | Qwen3.5-122B-A10B-FP8 | no | 23.04% | KEEP |
| 96555 | b9f7a56 | 07-30 18:16 | Qwen3-14B-FP8 | no | 5.48% | KEEP |
| 96572 | b9f7a56 | 07-30 18:34 | Qwen3-8B | no | 11.48% | KEEP |
| 96553 | b9f7a56 | 07-30 18:44 | Qwen3-14B-FP8 | no | 19.55% | KEEP |
| 96758 | 16eb8c4 | 07-31 16:33 | Qwen3-14B-FP8 | no | 7.89% | KEEP |
| 96951 | 16eb8c4 | 07-31 (ended 08-01 03:45) | gpt-oss-120b | no | 17.32% | KEEP |
| 97139 | 2c38027 | 08-01 07:37 | Qwen3-8B | no | 36.53% | KEEP |
| 97244 | ccceff025 | 08-02 08:21 | Qwen3-8B | no | 30.92% | KEEP |
| 97945 | c1f93bb | 08-05 13:27 | Qwen3-0.6B | no | 1.20% | KEEP |
| 98173 | 5b5458517 | 08-06 08:09 | Qwen3.5-397B-A17B-MXFP4 | no | 29.04% | KEEP |
| 101649 | 7a9d5c9 | 08-15 12:32 | Qwen3-14B-FP8 | **yes** | -1.44% | REVERT |
| 101682 | 7a9d5c9 | 08-16 11:43 | Qwen3-0.6B | **yes** | 0.70% | REVERT |

14 sessions in total (12 before the fix, 2 after). `1a552873c` is unrelated to this bug and neither improves nor worsens it.

The two post-fix cases were adjudicated REVERT, so they would not be adopted even with a corrected whitelist. The difference is that this should have been an **evidence-backed rejection** rather than the current "measured, but treated as never measured".

## Suggested fix

1. **Add a discriminator to the key.** Append `macro_cycle` or a content fingerprint of `result.json` to the idempotency key so each GEAK cycle gets its own slot.
2. **Accept settled-success in the whitelist.** On `succeeded`, do not write `rebench_unavailable`; read the task result back and run it through the existing `_promote_geak_from_candidate()` adjudication. Only `failed` / `cancelled` should produce `rebench_unavailable`.
3. **Fix the misleading message.** `task settled before dispatch (...)` is semantically backwards for `succeeded`; distinguish "already settled: succeeded" from "cancelled before dispatch".
4. **Extend the tombstone.** It currently covers only `no_material`. `_reject_geak_kernel_journey` (rebench did not beat current_best) should also be tombstoned so the next KERNEL entry does not re-recover the same `result.json` and re-enqueue.

Items 2 and 4 are the cheapest and most directly valuable; they can land first.

## How to verify

List affected sessions:

```sql
SELECT id, commit_sha, session_started_at, model_name,
       raw_data->'final'->'geak_pending'->>'revalidation_error' AS err
  FROM session_breakdowns
 WHERE deleted_at IS NULL
   AND raw_data->'final'->'geak_pending'->>'revalidation_error'
       LIKE '%succeeded%'
 ORDER BY session_started_at;
```

Then pull the session archive's `storage/coordinator.db` and check:

- `tasks` filtered on `idempotency_key='geak-revalidate'` should contain exactly one row in state `succeeded`
- `events` containing `geak_no_promote` shows the `measured_tput` versus `current_best_tput` comparison
- `events` containing `geak_e2e_done` tells how many times GEAK actually handed back within the session

Comparing `geak_pending.ts` with the rebench task's `updated_at`: if the former is clearly later and no new `geak_e2e_done` occurred in between, this is the repeated-recovery path from 3(b).

chao-xu-spec:
Fixes #1240.

## Problem

GEAK revalidation already uses macro-cycle-scoped idempotency keys, but two
settled-state paths could still reopen an adjudicated candidate:

1. `no_promote` cleared the pending slot without persisting a closed verdict.
   A later KERNEL entry could recover the stale successful `result.json` and
   enqueue the same candidate again.
2. When task creation returned an existing task that had already succeeded,
   KERNEL treated it as if it had settled before dispatch. This replaced the
   real verdict with `rebench_unavailable` and left misleading state.

## Change

- Persist `no_promote` on `geak_result`.
- Treat both `no_material` and `no_promote` as crash-recovery tombstones.
- Reconcile an existing succeeded task from the persisted GEAK verdict.
- If state writeback was interrupted after task completion, recover the
  persisted `delegated_result` and replay it through the normal GEAK
  adjudication path.
- If a succeeded collision has no recoverable result, report that condition
  without claiming the task never ran.
- Use the default explore cold/hot and confirmation path instead of a
  GEAK-specific multi-round protocol.
- Keep the change model-agnostic and free of session-specific data.

## Scope

This PR addresses #1240 only. It does not duplicate the #1239 phase-transition
fix or its budget-cap regression test from #1221.

## Tests

The results below are separate test runs and are not summed as unique tests.

- Focused #1240 recovery and tombstone tests: **5 passed in 0.76s**.
- Complete revalidation dispatch and gain-alignment suites: **71 passed in
  3.86s**.
- Default explore executor suite: **62 passed in 6.78s**.
- Main-overlap GEAK dashboard and FMoE suites: **54 passed in 0.51s**.
- Complete GEAK-selected regression suite after merging main: **284 passed,
  11037 deselected in 15.89s**.
- Modified-file diagnostics: **no errors**.
- Diff whitespace validation: **passed**.

The focused coverage verifies:

- `no_promote` is persisted as a closed verdict;
- stale successful results are not recovered and re-enqueued;
- macro-cycle keys and bounded retry keys step past settled attempts;
- a persisted result from an already-succeeded task is replayed and
  adjudicated;
- an existing succeeded task collision is reported honestly;
- cancelled work is reported as cancelled before completion;
- GEAK dispatch does not carry the removed multi-round or identity-guard
  parameters.


xiaofei-zheng:
## Right fix, but I think it's over-engineered

The bug is real — a **cold** revalidation compared against a **hot** `current_best` gives the false `no_promote`. But the root cause is that we compare a number from **GEAK's** measurement system against one from **Hyperloom's**. No amount of guarding (median-of-3, TTFT 3x, config digests, workload signature, protocol reconstruction) makes a cross-system comparison sound — it only patches around it.

**Cleaner fix:** run revalidation through the *same default cold/hot path* as every other explore variant, then compare to `current_best`. Fairness then holds by construction — warm-vs-warm or cold-vs-cold. This is essentially what the core `use_warm_decision` change already does; the rest is scaffolding on a separate revalidation path.

**This is safe because `current_best` is already always Hyperloom-measured:** `_record_geak_candidate` stores GEAK's number as audit-only (`self_reported_tput`) and doesn't touch `current_best`; only `_promote_geak_from_candidate(measured_tput=...)` writes it. So a re-measured candidate is always Hyperloom-vs-Hyperloom — promote if it wins, keep the incumbent if not, identical to a normal KEEP/REVERT.

**Suggested scope:**
1. Revalidate via the default path; drop the special `enable_stack_rebench=False` dispatch and the revalidation-only guards.
2. Trade-off: single-sample noise near the threshold — acceptable IMO, since normal explore already lives with it; reduce noise globally, not with a revalidation-only standard.
3. Please split the #1239 / #1240 state-machine + idempotency fixes into a separate PR — they're valid but unrelated to measurement fairness.

chao-xu-spec:
> ## Right fix, but I think it's over-engineered
> The bug is real — a **cold** revalidation compared against a **hot** `current_best` gives the false `no_promote`. But the root cause is that we compare a number from **GEAK's** measurement system against one from **Hyperloom's**. No amount of guarding (median-of-3, TTFT 3x, config digests, workload signature, protocol reconstruction) makes a cross-system comparison sound — it only patches around it.
> 
> **Cleaner fix:** run revalidation through the _same default cold/hot path_ as every other explore variant, then compare to `current_best`. Fairness then holds by construction — warm-vs-warm or cold-vs-cold. This is essentially what the core `use_warm_decision` change already does; the rest is scaffolding on a separate revalidation path.
> 
> **This is safe because `current_best` is already always Hyperloom-measured:** `_record_geak_candidate` stores GEAK's number as audit-only (`self_reported_tput`) and doesn't touch `current_best`; only `_promote_geak_from_candidate(measured_tput=...)` writes it. So a re-measured candidate is always Hyperloom-vs-Hyperloom — promote if it wins, keep the incumbent if not, identical to a normal KEEP/REVERT.
> 
> **Suggested scope:**
> 
> 1. Revalidate via the default path; drop the special `enable_stack_rebench=False` dispatch and the revalidation-only guards.
> 2. Trade-off: single-sample noise near the threshold — acceptable IMO, since normal explore already lives with it; reduce noise globally, not with a revalidation-only standard.
> 3. Please split the [GEAK rebench is cancelled by the KERNEL->SWEEP phase boundary 0-1s after enqueue #1239](https://github.com/AMD-AGI/Hyperloom/issues/1239) / [GEAK revalidate slot is a session-wide singleton; a second request collides with the already-succeeded task #1240](https://github.com/AMD-AGI/Hyperloom/issues/1240) state-machine + idempotency fixes into a separate PR — they're valid but unrelated to measurement fairness.

Thanks for the review. I have simplified and narrowed the change accordingly.

The GEAK-specific measurement protocol has been removed: there is no
median-of-three logic, TTFT-ratio gate, workload signature, config-file digest,
or benchmark-protocol reconstruction. GEAK dispatch also no longer forces
`enable_stack_rebench=False`; it now inherits the same default cold/hot and
single-confirmation behavior used by normal explore variants.

This PR is now scoped to #1240. The retained state-machine changes are:

- use macro-cycle-scoped idempotency keys and bounded retry keys;
- step past settled attempts instead of returning an old task row;
- persist `no_promote` as a closed verdict and treat it as a crash-recovery
  tombstone;
- reconcile an existing succeeded task from the recorded verdict;
- if state writeback was interrupted, replay the persisted
  `delegated_result` through the normal GEAK adjudication path;
- distinguish a succeeded collision from a task cancelled before completion.

#1239 is not included in this PR. It was fixed by #1221. I separately verified
the exact KERNEL budget-cap transition: a just-enqueued GEAK rebench remained
queued across `KERNEL_AGENT -> SWEEP` and its pending task ID remained intact.

Verification on the final worktree:

- focused #1240 recovery and tombstone tests: 5 passed;
- complete GEAK revalidation dispatch and gain-alignment suites: 71 passed;
- default explore executor suite: 62 passed;
- main-overlap GEAK dashboard and FMoE suites: 54 passed;
- complete GEAK-selected regression suite after merging main: 284 passed,
  11037 deselected;
- #1239 budget-cap scenario: passed;
- modified-file diagnostics and diff whitespace checks: passed.

xiaofei-zheng (approved):

PR #1261, opened 2026-08-21
fix(multi_node): cross-namespace rollback, kill exit code, and resume topology comparison

weilei0120:
Three defects in the multi-node control plane, from a review of the `multi_node/` package. Each was traced end to end before changing anything; several other items from the same review were checked and deliberately left alone (below).

Single-node paths are out of scope and untouched.

### `launch_multinode.py` — roll back on the node that owns the PID

The launch driver runs inside the Ray head pod and shares a PID namespace only with rank 0, but the spawn-failure rollback signalled every rank's PID locally. For each remote rank `os.getpgid()` raised `ProcessLookupError`, which the handler swallowed, so the stated purpose — *"so no half-started servers leak"* — never held for any rank but rank 0. Those servers kept running and kept their GPUs.

Sentinel PID `0`, written for vLLM worker ranks, was worse where reachable (vllm with ≥3 aggregated nodes, or PD with `prefill ≥ 2`): `os.getpgid(0)` resolves to the **caller's** process group, so the first rollback step SIGTERMed the driver itself and nothing was rolled back at all. Verified empirically:

```
>>> os.getpgid(0) == os.getpgrp()
True
```

Each rank's node id now travels with its PID, the kill is dispatched through a NodeAffinity-pinned actor, `pid <= 0` is skipped, and a rank that cannot be reached is logged rather than swallowed.

### `kill_multinode.py` — return non-zero when the kill did not finish

`still_alive` / `ports_busy` / `error` were collected into the JSON summary and then discarded: `main()` returned `0` unconditionally and no caller read the JSON. The caller sequences LAUNCH straight after, so a half-finished kill surfaced later as `port_base not available` or `TCPStore shut down too early` rather than as a kill failure.

Restarts that used to limp past an incomplete kill now fail and take the orchestrator's reclaim-and-retry — the same thing `launch_infera_node._kill_prior` already does when a prior server will not die.

`gpu_busy` stays advisory: its fallback wait is a coarse per-card threshold an unrelated co-tenant can trip, and it is already time-boxed. The port probe *is* safe to gate on — it binds without `SO_REUSEADDR`, so a released listen socket reports free immediately and there are no TIME_WAIT false positives.

### `restart-server` — compare the whole topology before resuming

The RayJob resume fast path compared `framework/model/tp/ep/pd_mode/extra_args` while persisting six more fields it never read back, so a round changing only the PD node split, per-role TP, transfer backend or IB device matched the previous launch and resumed it — and the benchmark then measured the previous topology under this round's config.

Replaced with a single topology record compared as a whole, written at the early checkpoint too since that is the one a poll-timeout retry reads. The per-field keys stay: the orchestrator reads them as PD argument fallbacks.

**This is latent rather than live on the orchestrated path** — PD topology there is resolved from state and env by `_resolve_pd_args` and does not vary per round. It is reachable through `restart-server`'s own flags, which the skill documents as a supported way to drive a cluster.

On infera the same comparator ignored the KV transfer backend, which reaches the pods as `--disaggregation-transfer-backend`, and never persisted it — leaving that tier of the orchestrator's three-tier fallback dead on this backend.

## Reviewed and deliberately not changed

- **Exit-code 1/2 inconsistency.** The sole programmatic caller (`_multi_node_server_lifecycle._restart_and_wait`) checks only `rc != 0` and bounds retries at exactly one, so nothing loops on a config error. Reporting terminal failures as `2` would instead make the agent path abandon a cluster on a transient dashboard blip.
- **`launch_infera_node`'s cmdline reaper scope.** Each infera pod has its own PID namespace and the engine port is hardcoded, so two engines cannot share a namespace and the reaper only ever sees its own pod's.
- **`known_hosts` refresh on host-key mismatch.** Pod re-provisioning with IP reuse is a plausible legitimate source of a changed key; ruling that out needs platform knowledge this change does not have.
- **Quoting the SSH interpreter.** The sandbox already holds the pod SSH key and `ssh_run_script` ships arbitrary script bodies to pods, so `$HYPERLOOM_MN_POD_PYTHON` grants no capability an attacker would not already have. Constraining it to a validated absolute path would be real hardening; shell-quoting alone only looks like one.

## Known gaps left open

Feature gaps rather than defects, not addressed here:

- RayJob drops `--pd-prefill-ep` / `--pd-decode-ep` / per-role extra-args (infera implements them).
- infera never reads `--pd-ib-device` at all (RayJob forwards it).

## Testing

`342 passed` across the multi-node, infera launcher, SSH client and grid-runner suites, rebased onto current `main`.

Five failures in `test_common_utils` / `test_coverage_gap_units` (claude config, credentials, LLM audit) are pre-existing — reproduced on a clean tree with these changes stashed.

### DIFF
--- a/src/hyperloom/inference_optimizer/multi_node/cli.py
+++ b/src/hyperloom/inference_optimizer/multi_node/cli.py
@@ -1735,6 +1735,48 @@ def _resume_probe_timeout_s() -> int:
         return _DEFAULT_RESUME_PROBE_TIMEOUT_S
 
 
+def _rayjob_topology_fingerprint(args: argparse.Namespace, nnodes: int) -> dict[str, Any]:
+    """Every field that changes what the RayJob launcher spawns, as one record.
+
+    The resume fast path compares this record as a whole, so any field that
+    reaches the launcher belongs here: one left out lets a round that changed
+    it resume the previous launch and benchmark the previous topology under
+    this round's config.
+
+    Args:
+        args (argparse.Namespace): Parsed ``restart-server`` arguments.
+        nnodes (int): Node count this launch targets.
+
+    Returns:
+        dict[str, Any]: The comparable topology record.
+    """
+    pd_mode = (getattr(args, "pd_mode", "") or "aggregated").lower()
+    fingerprint: dict[str, Any] = {
+        "framework": str(args.framework),
+        "model": str(args.model),
+        "tp": int(args.tp),
+        "ep": int(getattr(args, "ep", 1) or 1),
+        "nnodes": int(nnodes),
+        "pd_mode": pd_mode,
+        "extra_args": _normalize_extra_args(getattr(args, "extra_args", "")),
+    }
+    if pd_mode == "disaggregated":
+        # Only meaningful under PD; leaving them out when aggregated keeps a
+        # stale value from an earlier PD run out of the comparison.
+        fingerprint.update(
+            {
+                "pd_prefill_nodes": int(getattr(args, "pd_prefill_nodes", 0) or 0),
+                "pd_decode_nodes": int(getattr(args, "pd_decode_nodes", 0) or 0),
+                "pd_prefill_tp": int(getattr(args, "pd_prefill_tp", 0) or 0),
+                "pd_decode_tp": int(getattr(args, "pd_decode_tp", 0) or 0),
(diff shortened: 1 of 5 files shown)


ZhengGong-amd (approved):
validated

PR #1263, opened 2026-08-21
Get every hot kernel the trace found to a backend attempt

jiaqiang-dot-liu:
A run could finish "successfully" having optimized nothing. Each commit here
removes one reason the trace's evidence never reached a kernel backend. The
first pass fixed finding the source; the later ones fixed everything between a
resolved candidate and a measured rewrite.

Measured end to end on a MiniMax-M3-MXFP4 8k/1k run (vLLM, MI355X gfx950,
TP8/EP8): the KERNEL phase selects and dispatches on its own — no orchestration
request on the bus at all — and its rewrites now reach the optimization stack.
Three candidates dispatched, all three verified faster by forge's
3-measurement gate; `_gqa_sparse_fwd_kernel` (2.55x micro, SNR 51.8 dB) then
cleared end-to-end integration at **+1.11% throughput**, taking the session from
30.60% to **32.05%**. The two prior runs on this workload ended the KERNEL phase
with nothing integrated.

## Finding the source (6 commits)

`KNOWN_SEARCH_ROOTS` pinned one container layout, so on a wheel-install host
every search root was absent: the grep tier resolved nothing, the LLM tiers got
an empty shortlist, and the run reported success with zero routable kernels —
indistinguishable from a trace with nothing worth optimizing. Roots are now
discovered through the orchestrator's centralised resolver, so this tool agrees
with PolicyGate and patch application about where framework source lives. A
host with no searchable root now says so, in the log and as a trace-health
warning. Measured on a recorded session: 0 of 16 hot kernels routable → 6 of 16,
covering 13.74% of GPU time.

The two tool-free model tiers are replaced by one agent session that audits the
candidate table and is handed **paths rather than contents**, so it can open a
file and confirm it defines the kernel it is credited with rather than ranking
paths by keyword. Bounds on the added freedom: proposals only (what the trace
measured stays the trace's), a revised path must resolve under a known
framework root, a candidate already resolved by the active finder is not
overridable, and the candidate sources are fingerprinted around the session so
a review that edited the tree is discarded. Faults are contained — losing the
audit costs some candidates, failing the run costs the hours of benchmarking
behind the trace. Artifacts split into `kernel_candidates.raw.json`,
`kernel_candidates.json` and `kernel_candidates_revisions.json`, so a bad
dispatch can be traced to the stage that caused it.

Per @tsrikris's review: the review now reads only `analysis.md`, not the
internal `category_data/` sidecars. Almost nothing is given up — `analysis.md`
carries the same operand dims and the same launcher attribution, and the CSV
lookups only ran once that came back empty.

## Getting operand dims to the driver (4 commits)

A graph replay has no cpu_op parent, so the profiler records no arguments and
`shapes` arrives empty — on this model, for every hot kernel. Empty is not a
neutral state: with no shapes the tuning backend picks its own and cannot see
the serving configuration, so a prefill kernel serving 8192 tokens was tuned at
sequence lengths 512 and 1024, measured 1.1155x, and lost 3.05% end to end once
integrated. The question was never measured dims versus invented ones — it was
which side invents. The review reads the model config, the serving arguments and
the run's own artifacts; the backend reads none of them.

`shapes` and `input_dtypes` therefore leave `IMMUTABLE_FIELDS`, which now holds
only what cannot be revised for a reason. Dims carry their own provenance and a
session cannot claim `torch_trace`. The correction is pinned to
`invocation_spec_*.json`, which is what forge builds its driver from, and only
the reviewed table can be resolved as the candidate source. Two bugs had to go
for any of it to fire: revisions were only read on a rewrite that moved the
path, which dropped dims for exactly the rows that had them, and the
re-derivation check asked only whether `source_file` moved.

The predispatch gate then refused shapeless rows outright. That does not produce
a measured shape — it only makes a captured model's hottest kernels permanently
unoptimizable, and the backend already covers the gap: the invocation spec lists
the absent operands under its `missing` list and driver preparation recovers
them from the kernel source, the tests the spec names and the deployment context
it carries. Provenance is still validated, but only for a shape that is there.
The escape hatch goes with the gate it bypassed.

## Getting a dispatch to happen (5 commits)

- **The phase's own dispatch sat at the end of the GEMM branch**, so
  `INFERENCE_OPTIMIZER_SKIP_GEMM_TUNING=1` removed it too. A run then optimized
  nothing for 65 minutes waiting on an orchestration `run_optimization` request
  that never came, and exited with every candidate untried. Both entry routes
  now end in one shared tail, each stage consulting only its own switch.
- **The budget read as twice what it buys**, since forge-loop reserves half the
  window for finalize. At 60 minutes a campaign finished one iteration —
  planning alone took 16 of its 30 usable minutes — and terminated on
  `budget_exhausted`, which the ledger records as a kernel that could not be
  improved rather than one that was tried once. Now 90, which the verifying run
  spent reaching a third or fourth iteration on every candidate it dispatched.
  The request template also told the agent to send `budget_minutes: 60`
  verbatim, which outranks the shipped default, so that field is gone from the
  template; and the same template advertised a 10% hot-kernel floor that had not
  matched the shipped 5% for some time, which is how an orchestration agent came
  to nominate four sub-floor kernels in a row. **The floor itself stays at 5%.**
- **A veto that was only an echo.** The table handed to the review carries
  `reusable_native_kernel`, and every unresolved row carries false. Sessions
  returned it unchanged alongside a reason arguing the file *is* editable,
  which refused every kernel the review had just located. A `false` with no
  stated reason is now dropped with an audit note.
- **A gate rejection was filed as a failed attempt**, spending the source's
  retry quota on a decision no backend made and reading in the report as a
  technical failure when the cause was a threshold.
- **The floor-parity test never called the second gate it was named for**, so it
  could not observe drift, and what it did assert passed for an unrelated
  reason: its two straddling rows shared a source file, so the op-fanout dedup
  merged the weaker away no matter where the floor sat. It asserted a 10%
  boundary at a 5% default without ever failing. Now verified by mutation.

## Getting the rewrite deployed (1 commit)

A producer rooted below the consumer's package exports patch entries that
resolve under more than one ancestor of the traced source, and the ancestor walk
that resolves the deploy root has no information to break the tie. aiter carries
two different copies of `gemm/basic/gemm_afp4wfp4.py` — one under `ops/triton`,
one under `ops/triton/_triton_kernels` — and forge roots its worktree at the
deeper one, so the exported entry matched both. The walk requires exactly one
match, found two, and returned no root: a rewrite measured correct at 1.077x
with SNR 99 dB was reported as an unresolvable artifact and never reached
integrate.

Refusing was right — the two files differ, so deploying to the wrong one patches
something nothing imports and spends a full re-baseline proving it changed
nothing. But the tie was never real: the trace had already resolved which file
defines the kernel, and that answer sits in `target_file`. Removing the entry
that describes it from its own absolute path leaves the root the producer was
rooted at, with nothing guessed. The anchor is tried after an explicit root and
before the walk, longest tail first so a basename collision cannot claim it
ahead of the full relative path, and it still has to satisfy the same preimage
check — a descriptor set that does not belong to the traced tree falls through
to the walk exactly as before.

## Review response (4 commits)

- **The operator identity carried the launch call.** A trace names a candidate
  after both rows it occupies, so the same kernel reads
  `hipModuleLaunchKernel->_gqa_sparse_fwd_kernel` in an analysis whose trace
  paired the two and `_gqa_sparse_fwd_kernel` in one whose trace did not — one
  session produced both, from two profiles of the same configuration four
  minutes apart. Forge keys its experience store on that name, so a warm-start
  read of either reported no prior record. Reduced with
  `native_operation_key`, which already owns this normalization for the
  task-group identity.
- **Re-derivation now runs only where the source moved.** The stage cleared and
  rebuilt every source-derived field on any revision, including one supplying
  nothing but shapes. Eighteen of those nineteen keys have no producer in that
  pass, so clearing them where the path stood still dropped them for good.
  Stated plainly: on the twelve candidate tables recorded so far none of those
  eighteen is populated, so nothing observed was lost — the argument is that a
  field with no producer must not be cleared by a revision that gave no reason
  to doubt it.
- **The review session is bounded on stream inactivity**, not just the 900-second
  total. It reuses the TraceLens runner's two tiers rather than adding a second
  implementation: the SDK is silent between the block that launches a tool and
  the result that ends it, so a single bound tight enough to catch a dead stream
  kills a working tool call. Extracting the two drivers into one shared runner,
  which the review also asked for, is left as its own change.
- **Docs brought back to what the code does.** The environment-variables chapter
  still described the two retired completion tiers and three variables with no
  reader left; `HYPERLOOM_KERNEL_OPT_MIN_GPU_PCT` was in no document at all; and
  the conceptual doc still drew the source-level dispatch as the tail of the GEMM
  route, which is the shape that made the skip-GEMM bug possible.

Two findings were not taken as written. `collect_source_candidates_via_grep` and
`_relaxed_candidate_keywords` are deleted as dead, but the `find_benchmark_files`
early return is not equivalent to the main path — it returns before the
same-library guard and the multi-GPU demotion, so merging them would change
behaviour rather than tidy it; a genuinely redundant `if`/`else` beside it is
gone. And the reviewer's expected consequence of the re-derivation bug — a
prebuilt kernel dispatched for a 90-minute run that edits a wrapper — does not
follow on this data: `classify_patchability` rejects on `source_type`,
`vendor_dispatch_wrapper` and `runtime_generated_kernel`, all three recomputed,
and the five vendor binaries of the verifying run were refused at analysis time.

## Not wasting the trace (2 commits)

`last_profile_workload` has two writers with different shapes, so the staleness
check could never match a record written by the roofline path. Every session
paid about fifty minutes re-profiling and re-analysing a configuration that had
not moved. The comparison now projects both sides onto the fields that say
*what* was profiled, not how the profile task was parameterized. A real
configuration change is still caught, pinned by a parameterized test over five
cases — nothing moved, a new server argument, a new environment variable, a
changed value, a changed workload — of which only the first is forgiven.

## Prompt (1 commit)

The kernel-opt prompt was written for GEAK and handed to forge unchanged: 530
lines, of which forge uses about 40. Most of the remainder is merely redundant
(the source is inlined although forge is given the path; the runtime-metadata
block is labelled for GEAK's parser although forge reads the invocation spec).
Two sections are worse than redundant:

- The **deliverable contract** told the agent to write `optimization_report.md`
  and a copy of the kernel under `optimized_versions/`. Those land as new
  untracked paths in the measurement workspace, which forge's guard refuses
  outright, reverting the iteration for a protected-integrity violation with its
  measurements discarded. Hyperloom writes both files itself from forge's
  published manifest and git diff, so nothing ever read the agent's copies.
  Observed on two kernels of the verifying run, each losing its first iteration
  this way; in an earlier 60-minute run that was the only iteration, and it took
  a rewrite that had already measured correct and 6.7% faster.
- The **A/B recipes** walk the agent through building a standalone hipcc program
  or a `cpp_extension.load` module to time against — a second benchmark beside
  the driver that forge's own in-session gate scores. GEAK needs them because it
  brings no benchmark; for forge a number from the wrong benchmark is worse than
  no number.

Forge now gets the trace evidence it cannot derive plus the two
source-attribution guards that keep a rewrite off a `@compile_ops` wrapper and
on the right `__global__`. GEAK gets everything, unchanged.

## Test plan

- [x] `agents/kernel/tests/` + `common/tests/`: 1977 passed, 27 skipped
- [x] `inference_optimizer/tests/`: 11411 passed. Four pre-existing failures in
      `test_common_provenance` / `test_server_patcher_vllm_isolated_venv` fail
      identically on `origin/main` — they assume no vLLM is installed on the host
- [x] Root discovery unit-tested for drop-if-absent, package-dir-only, empty and
      explicit-checkout cases; verified on a wheel-install host (3 roots)
- [x] Candidate review agent's own suite (measured fields, invented paths,
      edited tree)
- [x] Floor-parity test verified by mutation: lowering one side's floor fails
      the parity assertion, and restoring the shared source file fails the
      skip-reason assertion
- [x] Deploy-root anchor verified by mutation: without it the ambiguous case
      returns no root. Both copies of the duplicated path resolve to their own
      root, and the two vLLM kernels of the same session still resolve by the
      walk
- [x] Re-profile boundary parameterized over five configuration cases
- [x] Deterministic stage replayed on a recorded trace: 0 → 6 of 16 routable
- [x] Agent review exercised end to end on a live run
- [x] Forge dispatch consumes the reviewed table, reaches committed rewrites
      (2.55x / 1.11x / 1.077x, all gate-verified), and one of them clears
      end-to-end integration at +1.11%

## Known, not addressed here

- The forge workspace guard still refuses any new untracked file, so an agent
  that writes a report of its own accord still loses that iteration. Dropping
  the instruction removes the cause Hyperloom controls; permitting the two paths
  is a KernelForge change.
- `_gqa_sparse_decode_kernel` integrated at −0.50%, inside the noise band this
  fleet shows on a repeated identical configuration, and was correctly parked at
  NEEDS_REVIEW rather than reverted.
- KERNEL spent about three hours idle before its exit guard fired, because a
  single orchestration LLM call hung for 71 minutes and the guard only advances
  on ticks it is reached by. Separate from this PR.

## Rebased

Rebased onto `main` @ 434b26b95 (was 116 commits behind). One conflict, in
`test_kernel_request_handlers_helpers.py`: `main` removed
`_source_escapes_reusable_roots` together with its test, and this branch's base
still had both. Resolved by taking the removal and keeping only the three tests
this branch adds.


src/hyperloom/agents/kernel/tools/tracelens_analysis.py
```diff
@@ -6947,6 +6559,257 @@ def build_audit_summary(
...
+            "hot_kernels": candidates,
+            "routable_kernels": routable,
+        },
+    )
+    artifacts["kernel_candidates_raw"] = str(raw_path)
+
+    source_paths = [str(c.get("source_file") or "") for c in candidates if isinstance(c, dict)]
+    before = source_fingerprint(source_paths)
+
+    reference_paths = {
```
tsrikris, line 7049: @jiaqiang-dot-liu Not all that these files are not meant to be read in practice. The only source of truth is analysis.md. If you continue to keep reading that we will add a rm -rf to clear out all files execept analysis.md

src/hyperloom/agents/kernel/tools/tracelens_analysis.py
```diff
@@ -2560,112 +2577,183 @@ def find_repo_root(source_file: str) -> str:
...
     (
-        ("rmsnorm_quant", "add_rmsnorm_quant", "rmsnorm", "add_rmsnorm"),
         (
-            "/sgl-workspace/aiter/op_tests/test_rmsnorm2dFusedAddQuant.py",
-            "/sgl-workspace/aiter/op_tests/test_rmsnorm2d.py",
-            "/sgl-workspace/aiter/op_tests/op_benchmarks/triton/bench_rmsnorm.py",
-            "/sgl-workspace/sglang/sgl-kernel/benchmark/bench_rmsnorm.py",
-            "/sgl-workspace/aiter/op_tests/triton_tests/normalization/test_rmsnorm.py",
-            "/sgl-workspace/aiter/op_tests/triton_tests/normalization/test_fused_add_rmsnorm_pad.py",
+            "rmsnorm_quant",
```
tsrikris, line 2543: @jiaqiang-dot-liu Can you comment on how this segment is different from the kernel name to launcher detection? We've seen that these paths significantly move around across versions of AITER, vLLM, SGLang

tsrikris:
@jiaqiang-dot-liu Two clarifying questions on this PR:
1. Deriving missing args, launcher path: A trace under GraphLaunch/Compile will only yield the kernel name as you've pointed out. Deriving the other components (args, launcher path) can only be handled agentically. The question here is should this layer be present within Hyperloom or inside Kernel Forge? Per our discussions with @sdubagun-amd and @amd-ethany, GEAK handles this within. Clarifying this since we can offload this layer into the kernel tuning agent and mitigate the codebase increase.
2. A deterministic TraceLens Agent path is being built that should help further offload this corner case.

src/hyperloom/agents/kernel/tools/tracelens_analysis.py
jiaqiang-dot-liu: Thanks for raising this — you're right.

My earlier change here only fixed half of the problem. The table used to hold
absolute paths, e.g. `/sgl-workspace/aiter/op_tests/test_rmsnorm2d.py`. When the
framework is installed from a wheel, none of those paths exist on the host. The
code still put them into `benchmark_files` and passed them downstream, where a
non-empty list reads as a test file someone can actually run. I changed the paths
to be checkout-relative and added an existence check, so only files that are
really there are kept. That stops us from reporting paths that do not exist.

But it does not address what you are pointing at: the table is essentially a copy
of AITER's directory layout kept inside Hyperloom. The moment upstream
reorganizes those directories this goes stale, and there is no mechanism that
would notice.

On your question about how this differs from launcher detection: they solve
different problems. Launcher detection resolves `source_file` — which file
defines the kernel. This table populates `benchmark_files` — which test or
benchmark can exercise the kernel.

There is one thing the table does that a grep cannot: it maps a device-side
kernel symbol to a Python-level test file. For example, the kernel symbol is
`add_rmsnorm_quant_kernel` and the source file is `rmsnorm_quant_kernels.cu`, so
those are the only two names a grep has to search for. But `test_rmsnorm2d.py`
calls the Python API, so neither name appears in it and the grep finds nothing.
That mapping is worth keeping; the hardcoded paths are not.

This PR is a set of unrelated bug fixes, so adding a change to how harnesses are
discovered does not belong in it. I will send that as its own PR: drop the
hardcoded path list, keep only the mapping from kernel-name markers to op names,
use those op names as grep keywords, and discover every path at runtime.

jiaqiang-dot-liu:
> @jiaqiang-dot-liu Two clarifying questions on this PR:
> 
> 1. Deriving missing args, launcher path: A trace under GraphLaunch/Compile will only yield the kernel name as you've pointed out. Deriving the other components (args, launcher path) can only be handled agentically. The question here is should this layer be present within Hyperloom or inside Kernel Forge? Per our discussions with @sdubagun-amd and @amd-ethany, GEAK handles this within. Clarifying this since we can offload this layer into the kernel tuning agent and mitigate the codebase increase.
> 2. A deterministic TraceLens Agent path is being built that should help further offload this corner case.

@tsrikris Thanks for raising this — I agree with you. Deriving the information a kernel is
missing sits better inside the KERNEL AGENT, and it keeps that complexity out of
the Hyperloom layer.

In the near term we will move the KernelForge implementation into Hyperloom in
its entirety, as a self-contained KERNEL AGENT module. That changes only where
the code lives, not the responsibility boundary: this derivation remains the
KERNEL AGENT's job.

Once that migration is done, we will move the agent-driven derivation of missing
kernel information into the KERNEL AGENT itself.

ZhengGong-amd (approved):
Validated

PR #1267, opened 2026-08-24
fix(provenance,patcher): a declared framework venv wins over the orchestrator's interpreter

WhatGhost:
## Problem

`--framework-env isolated` is the default for vLLM, so `$VLLM_VENV_ROOT` names the vLLM a run actually serves with. But two places consulted the orchestrator's own interpreter first and only fell back to that venv when the lookup failed:

```python
# provenance.py — version recorded in the session manifest
try:
    return (_im.version(dist) or "").strip()   # this interpreter
except Exception:
    pass
if not venv_path:
    return ""
...                                            # venv, only as a fallback

# _server_patcher.py — which vLLM gets the TraceLens patch
try:
    import vllm                                # this interpreter
    ...
except Exception as e:   # any import failure → isolated-venv fallback
    probed = _probe_isolated_vllm()
```

On any host where vLLM is also installed alongside the orchestrator, that fallback is dead code — the interpreter always answers first. Consequences:

- the session manifest records the version of an environment the run never used
- the TraceLens patch is applied to that same wrong tree, so the served vLLM stays unpatched (no profiler annotations) while an unrelated install is modified

The version case is worse than the `vllm unknown` it replaced: `unknown` is honest, a wrong version looks right.

`provenance.py` also could not express "a venv was declared but is unusable" — `_venv_site_packages()` returned `[]` both when no venv was declared and when one was declared but empty, so the caller had no way to tell them apart.

## Fix

Both sites now branch on what the run **declared**, not on what happens to be importable:

```python
# provenance.py
if venv_path is not None:        # a venv was declared -> authoritative
    ...scan it; not found is ""  # never falls back to the interpreter
try:                             # no venv declared -> shared environment
    return (_im.version(dist) or "").strip()
except Exception:
    return ""

# _server_patcher.py
if os.environ.get("VLLM_VENV_ROOT", "").strip():
    probed = _probe_isolated_vllm()
    if probed is None:
        return None              # skip the patch rather than patch another env
    ...
else:
    import vllm                  # shared environment
```

`_venv_site_packages()` returns `None` (no venv declared) instead of `[]`, so `[]` now unambiguously means "declared but holds nothing" — and the `venv_path is not None` test keeps that case out of the interpreter branch.

A declared-but-unusable venv reports `unknown` / skips the patch. For the patcher that is deliberate: patching the wrong tree is worse than not patching, since the served vLLM stays unpatched either way and the wrong one gets modified.

## Verification

**Behaviour matrix.** Both functions, all four combinations, before vs after:

`detect_stack_fingerprint()["vllm"]`

| vLLM in interpreter | venv declared | before | after |
|---|---|---|---|
| no | no | `unknown` | `unknown` |
| no | yes | `9.9.9` | `9.9.9` |
| yes | no | `0.24.0+rocm723` | `0.24.0+rocm723` |
| yes | yes | `0.24.0+rocm723` | **`9.9.9`** |

`_discover_vllm_install()`

| vLLM in interpreter | venv declared | before | after |
|---|---|---|---|
| no | no | `None` | `None` |
| no | yes | venv `0.21.0` | venv `0.21.0` |
| yes | no | system `/opt/venv` | system `/opt/venv` |
| yes | yes | **system `/opt/venv`** | **venv `0.21.0`** |

Six of eight cells are unchanged. The two that move are exactly the defect. The first two rows of each table are the CI environment (no vLLM installed), so CI behaviour is identical.

**Tests.** Six existing tests that fail on any host with vLLM installed now pass:

```
test_common_provenance.py::test_a_framework_in_its_own_venv_is_still_versioned
test_common_provenance.py::test_a_venv_root_that_is_not_there_is_not_a_failure
test_manifest_unit.py::test_manifest_versions_a_framework_installed_in_its_own_venv
test_server_patcher_vllm_isolated_venv.py::test_isolated_venv_discovers_plan
test_server_patcher_vllm_isolated_venv.py::test_isolated_venv_missing_returns_none
test_server_patcher_vllm_isolated_venv.py::test_isolated_venv_no_profiler_sentinel_returns_none
```

They pass on CI today only because the runner has no vLLM; on a bare-metal Hyperloom host — the environment this code targets — they fail.

**Full suite.** `3 failed, 14758 passed` — down from `9 failed` on `main`. The remaining three are pre-existing robustness heartbeat failures (`assert 2 == 1`), unrelated to this change and equally present on a clean `main`.

`ruff check` passes. `ruff format` and `mypy` each report findings that are already present on `main` for these files, so they are left alone.

## One existing test was modified

`test_isolated_venv_missing_returns_none` gained `monkeypatch.setitem(sys.modules, "vllm", None)`.

Without `$VLLM_VENV_ROOT` the run is in the shared-environment case, where patching the interpreter's vLLM is the *correct* answer — so asserting `None` there only holds on a machine with no vLLM installed. The added line makes the test state the case it actually means (no venv declared **and** nothing importable here), which is what the assertion was always about, and makes it independent of the host.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,11 +99,22 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
   of a shared install the run never served with when one happened to be present.
   `_resolve_framework_build` already walks the candidate interpreters and imports
   the framework to find the right one, but `_check_serving_framework` only
-  printed the winner; it is now published as `$HYPERLOOM_FRAMEWORK_PYTHON` and
-  the fingerprint reads its `site-packages`. The installer-written
-  `$VLLM_VENV_ROOT` is deliberately not consulted: it is only ever written, never
-  cleared, so it survives an isolated-to-shared move and would name an
-  environment the run no longer serves with.
+  printed the winner; it is now published as `$HYPERLOOM_RESOLVED_FRAMEWORK_PYTHON`
+  (paired with `$HYPERLOOM_RESOLVED_FRAMEWORK`, since the scan answers for one
+  framework and `sglang` is the default) and the fingerprint reads its
+  `site-packages`. The installer-written `$VLLM_VENV_ROOT` is no longer read by
+  the fingerprint directly: it is only ever written, never cleared, so on its own
+  it cannot say whether the tree it names still holds vLLM. It still leads
+  preflight's candidate list and is probed there, which is what the recorded
+  version now follows. A prefix that yields no `site-packages` — a system Python keeps its
+  packages in `dist-packages` — is treated as a failed derivation and falls back
+  to this process, not as an authoritative "not installed".<br/>
+  **Operator note**: the framework check returns before publishing when
+  `$HYPERLOOM_SKIP_FRAMEWORK_CHECK` is set, when `$BENCHMARK_BASE_URL` points at
+  a remote server, on external multi-node, and for scriptable frameworks (xDiT,
+  custom) that own their entrypoint — serving is not local on those paths, so
+  the fingerprint falls back to this process rather than reading a venv root
+  that describes some other host.
 
 - **Shell and loader hijack names are rejected from the `extra_envs` argument to
   `materialize_config_with_envs` before the config is persisted.** The predicate
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,21 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ### Fixed
 
+- **The recorded framework version now comes from the interpreter preflight
+  resolved, not from whatever the orchestrator's own process happens to have.**
+  `--framework-env isolated` is the default for vLLM, whose ROCm wheel pins its
+  own torch, so the framework is installed where `importlib.metadata` in this
+  process cannot see it — and `detect_stack_fingerprint` probed this process
+  first, recording `unknown` on the default bare-metal vLLM path, or the version
+  of a shared install the run never served with when one happened to be present.
+  `_resolve_framework_build` already walks the candidate interpreters and imports
+  the framework to find the right one, but `_check_serving_framework` only
+  printed the winner; it is now published as `$HYPERLOOM_FRAMEWORK_PYTHON` and
+  the fingerprint reads its `site-packages`. The installer-written
+  `$VLLM_VENV_ROOT` is deliberately not consulted: it is only ever written, never
+  cleared, so it survives an isolated-to-shared move and would name an
+  environment the run no longer serves with.
+
 - **Shell and loader hijack names are rejected from the `extra_envs` argument to
   `materialize_config_with_envs` before the config is persisted.** The predicate
   was `valid_env_key`, a key-shape check that let `LD_PRELOAD`, `PYTHONPATH` and
--- a/src/hyperloom/common/provenance.py
+++ b/src/hyperloom/common/provenance.py
@@ -33,6 +33,7 @@ import hashlib
 import os
 import re
 import subprocess  # nosec B404 - best-effort, guarded provenance probes only.
+import sys
 from importlib import metadata as _im
 from pathlib import Path
 from typing import Any, Mapping
@@ -58,15 +59,29 @@ _STACK_FINGERPRINT_ENVS: dict[str, tuple[str, ...]] = {
 # ``importlib.metadata`` here cannot see it, and the version silently degraded to
 # "unknown" on the default bare-metal vLLM path.
 #
-# Preflight already walks the candidate interpreters and imports the framework to
-# find this one, so it describes the runtime this session resolved. The installer's
-# ``$VLLM_VENV_ROOT`` is deliberately not consulted: it is host state that is only
-# ever written, never cleared, so it survives an isolated -> shared move and would
-# name an environment the run no longer serves with.
-_FRAMEWORK_PYTHON_ENV: str = "HYPERLOOM_FRAMEWORK_PYTHON"
+# Preflight already walks the candidate interpreters and locates the framework
+# under one of them, so it describes the runtime this session resolved. The
+# installer's ``$VLLM_VENV_ROOT`` is no longer read here directly: it is host
+# state that is only ever written, never cleared, so on its own it cannot say
+# whether the tree it names still holds the framework. It stays in play through
+# preflight, which leads with it and probes it -- and that is the right answer
+# either way, because ``_derive_runtime_paths`` also puts that root at the head
+# of ``PATH`` and vLLM starts as a bare ``vllm serve``, so a root that still
+# holds a working vLLM *is* what serves. A root that no longer does fails the
+# probe and the scan moves on, where reading it directly would have recorded a
+# version from an environment the run never touched.
+#
+# Distinct from ``HYPERLOOM_FRAMEWORK_PYTHON``, the per-attempt YAML key a
+# framework-agent attempt uses to force its own launch interpreter: that one is
+# a per-benchmark override, this one is the session-wide resolution.
+#: Published by preflight; read here. Public so the two stay in one place.
+RESOLVED_FRAMEWORK_PYTHON_ENV: str = "HYPERLOOM_RESOLVED_FRAMEWORK_PYTHON"
 
-# Components served by that interpreter. Others live in this process.
-_FRAMEWORK_COMPONENTS: frozenset[str] = frozenset({"vllm", "sglang"})
+#: The framework that interpreter was resolved for. The scan answers for one
+#: framework only, so the pair has to travel together: an SGLang session
+#: publishes an interpreter that knows nothing about vLLM, and reading it as the
+#: vLLM answer reports "unknown" for a vLLM this process can see.
+RESOLVED_FRAMEWORK_ENV: str = "HYPERLOOM_RESOLVED_FRAMEWORK"
 
 #: Runtime-arch overrides only. ``PYTORCH_ROCM_ARCH`` is deliberately absent:
 #: it names the archs a wheel is *compiled* for, not the installed device, and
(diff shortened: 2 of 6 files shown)


ZhengGong-amd:
Thanks for addressing the isolated-vLLM case. I don’t think this is safe to merge yet.

This changes discovery priority in one path, but runtime selection is still independently derived from `VLLM_VENV_ROOT`, `VLLM_PYTHON`, and `PATH` across preflight, patching, provenance, and build probes. In particular, a stale isolated-vLLM root after an isolated → shared vLLM transition can still be selected, while `VLLM_PYTHON` can point to a mismatched environment. This can silently make Hyperloom patch or probe a different vLLM installation than the server process.

Please centralize runtime resolution through the existing `FrameworkRuntime` model and have all consumers use the resolved interpreter/root. At minimum, remove `VLLM_PYTHON`, handle vLLM mode transitions explicitly, and add coverage for stale-root and server/interpreter mismatch cases.

WhatGhost:
@ZhengGong-amd Thanks for your reviews!
You're right about the patcher — dropped in `f6c9aed8c`. The PR is now `provenance.py` only.

**Patcher.** Reproduced the stale-root case — `$VLLM_VENV_ROOT` left over from an isolated → shared transition, pointing at a tree that no longer holds vLLM. On `main`, `_discover_vllm_install()` patches the shared install (correct); with my change it returned `None` and nothing was patched. That is worse than the problem I was fixing.

**Provenance.** The same stale root reads differently there. On `main` the probe answers `0.24.0+rocm723` — a version belonging to an environment the run never used; with this change it answers `unknown`. For a provenance record the latter is the better answer: a plausible-looking wrong version is believed, an `unknown` gets checked. The original fallback weighed `unknown` against "the extra directory scan", not against "a version from the wrong environment".

**`FrameworkRuntime`.** Not sure it fits: its docstring scopes it to the bench subprocess's attempt venv (`$SESSION_DIR/enablement/stacks/...`) and its consumers are all under `framework/`. `$VLLM_VENV_ROOT` is a different layer — where setup installed the baseline framework on this host. Happy to be corrected if you had a specific shape in mind.

**Centralizing.** Agreed the derivation is scattered: `$VLLM_VENV_ROOT` is read independently in `preflight.py`, `_server_patcher.py`, `apply_kernel_patch.py`, `provenance.py`, `framework/paths.py` and `env_safety.py`, while `install_baremetal.sh` writes both it and `VLLM_PYTHON`. But which layer owns the resolved runtime, and whether a stale root should fail loudly or fall back, are design calls I'd rather not make unilaterally here. An issue with the intended ownership model would help.

Suite: `6 failed, 14755 passed` — three version probes now pass, the three patcher tests are back to their pre-PR state, three heartbeat failures unrelated.

ZhengGong-amd:
Thanks for splitting the patcher back out — dropping it was the right call, and your reproduction of the stale-root case is exactly the problem I was pointing at.

On `FrameworkRuntime`: you're right, it's attempt-scoped and this isn't its layer. I withdraw that suggestion.

But I don't think `$VLLM_VENV_ROOT` can be authoritative for provenance either, for the same reason it couldn't be for the patcher. It is host state the installer wrote, not the runtime this session resolved: `write_runtime_dotenv` only ever writes it (`install_baremetal.sh:1760-1763`, no `else`), and `restore_persisted_framework_env` reads it back from `.env` when no flag is passed (`:1412-1418`). So after an isolated → shared transition it survives. Your test covers a root that is gone (`unknown`, good); the dangerous case is a root that still exists and still holds vLLM 0.27 while the run serves shared vLLM 0.30 — this change would confidently record 0.27.

The resolution you need already exists and is currently thrown away. `_resolve_framework_build` (`cli/preflight.py:879`) walks the candidate interpreters, imports the framework, and verifies a ROCm build; `_check_serving_framework` then only prints the winner (`:956-962`). Every consumer afterwards re-derives it more weakly — and with four different precedence orders (`preflight:740`, `_server_patcher:559`, `_grid_runner:235`, `framework/paths:298` and `:431`), so the recorded version, the patched tree and the probed tree can be three different installs.

Suggested scope for this PR:

1. Publish what preflight already resolved: `os.environ["HYPERLOOM_FRAMEWORK_PYTHON"] = found` in `_check_serving_framework`.
2. In `provenance.py`, drop `_STACK_VENV_ROOT_ENVS` / `_venv_site_packages` and derive `site-packages` from that interpreter instead. Fall back to the current interpreter when unset (today's `main` behaviour), not to `$VLLM_VENV_ROOT`.

That records the isolated venv on the default vLLM path — your original goal — without a value that can go stale, and it keeps the version and the tree tied to one interpreter.

Two more things, either here or as follow-ups:

- `_derive_runtime_paths` (`cli/preflight.py:270-272`) prepends `$VLLM_VENV_ROOT/bin` to the front of `PATH` with no existence check and no framework gate. Since Magpie launches a bare `vllm serve`, PATH's head *is* the server. A stale root shadows the shared bin even for an SGLang session, so provenance accuracy alone doesn't close this.
- The PR currently shows 1,298 files changed (+235,844/-62,606). Looks like unrelated history came in with a merge — worth fixing the branch base before this is reviewable.

WhatGhost:
> Thanks for splitting the patcher back out — dropping it was the right call, and your reproduction of the stale-root case is exactly the problem I was pointing at.
> 
> On `FrameworkRuntime`: you're right, it's attempt-scoped and this isn't its layer. I withdraw that suggestion.
> 
> But I don't think `$VLLM_VENV_ROOT` can be authoritative for provenance either, for the same reason it couldn't be for the patcher. It is host state the installer wrote, not the runtime this session resolved: `write_runtime_dotenv` only ever writes it (`install_baremetal.sh:1760-1763`, no `else`), and `restore_persisted_framework_env` reads it back from `.env` when no flag is passed (`:1412-1418`). So after an isolated → shared transition it survives. Your test covers a root that is gone (`unknown`, good); the dangerous case is a root that still exists and still holds vLLM 0.27 while the run serves shared vLLM 0.30 — this change would confidently record 0.27.
> 
> The resolution you need already exists and is currently thrown away. `_resolve_framework_build` (`cli/preflight.py:879`) walks the candidate interpreters, imports the framework, and verifies a ROCm build; `_check_serving_framework` then only prints the winner (`:956-962`). Every consumer afterwards re-derives it more weakly — and with four different precedence orders (`preflight:740`, `_server_patcher:559`, `_grid_runner:235`, `framework/paths:298` and `:431`), so the recorded version, the patched tree and the probed tree can be three different installs.
> 
> Suggested scope for this PR:
> 
> 1. Publish what preflight already resolved: `os.environ["HYPERLOOM_FRAMEWORK_PYTHON"] = found` in `_check_serving_framework`.
> 2. In `provenance.py`, drop `_STACK_VENV_ROOT_ENVS` / `_venv_site_packages` and derive `site-packages` from that interpreter instead. Fall back to the current interpreter when unset (today's `main` behaviour), not to `$VLLM_VENV_ROOT`.
> 
> That records the isolated venv on the default vLLM path — your original goal — without a value that can go stale, and it keeps the version and the tree tied to one interpreter.
> 
> Two more things, either here or as follow-ups:
> 
> * `_derive_runtime_paths` (`cli/preflight.py:270-272`) prepends `$VLLM_VENV_ROOT/bin` to the front of `PATH` with no existence check and no framework gate. Since Magpie launches a bare `vllm serve`, PATH's head _is_ the server. A stale root shadows the shared bin even for an SGLang session, so provenance accuracy alone doesn't close this.
> * The PR currently shows 1,298 files changed (+235,844/-62,606). Looks like unrelated history came in with a merge — worth fixing the branch base before this is reviewable.

@ZhengGong-amd  Thanks for review again. 

Adopted your approach in `c8a94a9d2`. You were right — my test only covered a root that was gone; the case you describe (root present, holding an older vLLM) would have been recorded confidently. `provenance.py` no longer reads `$VLLM_VENV_ROOT` at all.

`_check_serving_framework` now publishes what it resolved as `$HYPERLOOM_FRAMEWORK_PYTHON`; `detect_stack_fingerprint` derives `site-packages` from it and falls back to this process when unset. `_STACK_VENV_ROOT_ENVS` / `_venv_site_packages` are gone.

| case | recorded |
|---|---|
| shared | this process |
| isolated | venv version |
| after isolated → shared, stale root present | this process |
| resolved venv since removed | `unknown` |

**Two deviations.** I gated the publish on `probe.verdict is not False` — `_resolve_framework_build` can return a candidate it has *refuted*, and publishing that hands every consumer a known-wrong interpreter. And `_FRAMEWORK_COMPONENTS` covers `sglang` too; happy to narrow it to `vllm` if you'd rather not widen the contract.

Note `$HYPERLOOM_FRAMEWORK_PYTHON` is already read by `bypass_runner` and written per-attempt by `_grid_runner`. Setting it in `os.environ` means bypass paths without a per-attempt override now inherit the resolved interpreter — per-attempt values still win, but it is wider than provenance alone.

Tests: the four existing cases construct the resolved interpreter instead of a venv root, plus one asserting `$VLLM_VENV_ROOT` alone is *not* consulted.

Suite: `7 failed, 15059 passed` — three version probes fixed; the rest pre-existing on this base 

**PATH.** Confirmed, and worse than provenance: `_derive_runtime_paths` prepends `$VLLM_VENV_ROOT/bin` with no framework gate, and `vllm_mi355x.sh:89` runs `setsid vllm serve` — a bare name — so a stale root means the run serves with the wrong vLLM regardless of what is recorded. SGLang differs: `sglang_mi355x.sh:80` runs `python3 -m sglang.launch_server`, so the interpreter is shadowed rather than the binary. A missing directory is harmless (the shell skips it), so the risk is specifically a stale root that still exists. Issue or follow-up PR, whichever you prefer.

ZhengGong-amd:
Approach looks right now. One correction to my earlier review first: I claimed a stale root holding an older vLLM would be recorded while the run served the shared one. That case isn't reachable — `_derive_runtime_paths` puts `$VLLM_VENV_ROOT/bin` at the head of `PATH` and vLLM starts as a bare `vllm serve`, so a stale root that still holds a working vLLM *is* what serves. Recording it is correct. Ignore the regression test I asked for.

Two blockers on the implementation.

**1. `.resolve()` walks out of the venv.** In `_framework_site_packages`:

```python
venv_root = Path(python_exe).resolve().parent.parent
```

`bin/python` is a symlink chain into the base interpreter, so this lands on `/usr`:

```
venv/bin/python -> python3 -> /usr/bin/python3
resolve()               /usr/bin/python3.12
resolved parent.parent  /usr
glob on resolved        []
glob on literal         ['venv/lib/python3.12/site-packages']
```

An empty list isn't `None`, so `_probe_pkg_version` stays on the venv branch and returns `""` without falling back to this process. Net effect on a real host: `vllm` is `unknown` in the isolated case (the bug this PR set out to fix) *and* in the shared case, where `main` currently answers correctly via `_im.version`.

The tests can't catch it because `_installed()` returns a path that doesn't exist on disk, and `Path.resolve()` doesn't follow symlinks for a non-existent path — it just normalises and stays inside `tmp_path`.

Fix: drop `.resolve()`. The published value is already a literal path, and `sys.executable` inside a venv is the unresolved venv path, so `.parent.parent` is correct for both. The test fixture needs a real `python -m venv` (or at least the symlink) or it will keep passing regardless.

**2. `sglang` in `_FRAMEWORK_COMPONENTS` loses its version.** `_framework_site_packages` doesn't check which framework the interpreter was resolved *for*. On a vLLM run, preflight publishes the isolated vLLM venv, sglang is then looked up there, finds nothing, and records `unknown` — where `main` reads it correctly from this process.

Fix: narrow to `vllm`, or publish the resolved framework alongside the interpreter and only use it for the matching component.

Two smaller things, non-blocking:

- `_check_serving_framework` returns before publishing for `$HYPERLOOM_SKIP_FRAMEWORK_CHECK`, `$BENCHMARK_BASE_URL` and external multi-node. Those paths previously got the right version from `$VLLM_VENV_ROOT` and now fall back to this process — worth a line in the CHANGELOG.
- `_resolve_framework_build` returns early on `probe.timed_out` with `verdict is None`, which passes the `is not False` gate. The comment says "proven to import here", but a timeout proves nothing.

Also consider renaming: `HYPERLOOM_FRAMEWORK_PYTHON` is already the per-attempt YAML override key. Nothing reads the new one from `os.environ` today (`bypass_runner` reads `bench_envs`), so there's no live conflict, but the two layers now share a name with different meanings.

WhatGhost:
@ZhengGong-amd Both blockers fixed in `d37a6ecb8`.

**`.resolve()`** — dropped. Reproduced it on a real venv: the resolved `parent.parent` is `/usr`, glob empty, so isolated *and* shared both degraded to `unknown`. `_installed()` now builds the real `bin/python -> python3.12 -> <base>` chain, and restoring `.resolve()` fails it.

**`sglang`** — took your second option: `$HYPERLOOM_RESOLVED_FRAMEWORK` is published with the interpreter and only the matching component reads it. Related: a prefix yielding no `site-packages` is now a failed derivation rather than an authoritative empty answer, since a system Python keeps its packages in `dist-packages`.

**CHANGELOG** added, plus a fourth early return you didn't list — `framework_registry.is_scriptable`. Renamed to `HYPERLOOM_RESOLVED_FRAMEWORK_PYTHON`, which also drops the `bypass_runner` inheritance I flagged last round.

**`probe.timed_out`** — you're right that `find_spec` doesn't prove an import, and the comment is fixed. I left the gate at `verdict is not False` though: the timeout is in the ROCm verdict, `_check_serving_framework` keeps serving with that interpreter after a warning (`:979-984`), and withholding it would record the orchestrator's own vLLM for a run the venv served. Reading a version wants `dist-info` metadata, which doesn't require an import. Happy to gate it if you disagree.

**Your correction** caught an error in mine too: several comments still said `$VLLM_VENV_ROOT` "would name an environment the run no longer serves with", and read as though it were ignored — when `_framework_probe_interpreters:744-747` leads the candidate list with it. Reworded: provenance doesn't read it directly because on its own it cannot say whether that tree still holds vLLM; preflight probes it, and a root that still does is what serves.

10 new cases cover the publish contract and the `None`-vs-authoritative split. The autouse fixture now pops the two vars on teardown — `delenv(raising=False)` records no undo for an absent key, so they were outliving the file.

Suite: `7 failed, 15015 passed`; all seven fail identically on `main`.

ZhengGong-amd (approved):
Approved

PR #1268, opened 2026-08-24
fix(gemm): drop unreachable backend branch and misleading precision filter

chennyiiis:
## Summary

- Remove the backend branch in `_gemm_tuning_required_before_kernel_opt()` and the legacy FP8 + SGLang filter it applied. That branch cannot execute, so this is a **no-op for every configuration**.
- Eligibility is now a supported framework (`sglang` / `vllm` / `vllm-aiter`) and nothing else, with the reachability argument documented inline.
- Add two invariant tests: the backend order's two owners are exact complements (including `forge,geak`, which is not an exact match and so falls back to GEAK), and every precision forge supports clears the gate.

## Why the branch is unreachable

Backend selection is a **single switch** — `forge_explicitly_enabled()`, i.e. exactly `KERNEL_OPT_BACKEND_ORDER=forge` (case-insensitive). Both functions that matter here read that one switch and nothing else:

- `geak_selected()` → `_raw_kernel_backend_order()` returns `["forge"]` when it is true and `["geak"]` otherwise.
- `_resolve_gemm_tuning_backend()` returns `"forge"` when it is true and `"geak"` otherwise.

They are two views of the same switch, hence strict complements. `_on_enter_kernel()` hands the entire KERNEL_AGENT phase to the GEAK delegate and returns *before* this gate whenever GEAK owns the order. So arriving at the gate already implies forge, and re-resolving the backend inside it can only answer `forge` — no input could ever select the `else` branch.

## Why it was worth deleting anyway

Its comment read as a live precision filter ("legacy FP8 + SGLang only"), which points a reader at the wrong explanation in precisely the case they will investigate: a session where GEMM tuning produced nothing *looks* precision-filtered, when in truth the phase went to GEAK and tuning was never reached. That cost real time in a production investigation before the early return was spotted.

No precision / MoE pre-filter replaces it. Real end-to-end KEEPs span bf16/fp16/fp8/fp4/mxfp4, dense as well as MoE — bf16 *dense* alone was worth +11.1%. Filtering here would silently block a category that can optimize, and forge already reports `no_improvement` when a shape cannot be beaten.

## Note on an earlier RCA

A prior RCA attributed missing GEMM tuning to precision filtering at this gate. That was wrong on both counts: the filter never ran, and Forge does tune when a pipeline opts in (2026-08-09 forge batch: +24.51% GEMM gain in the session breakdown). The recent absence of forge batches is Conductor-side scheduling, not orchestrator gate logic.

For the same reason, a non-exact value such as `KERNEL_OPT_BACKEND_ORDER=perfskills` (used by one pipeline) falls back to GEAK like any other non-`forge` value. That is the documented contract, not a defect.

## Test plan

- [x] `pytest src/hyperloom/inference_optimizer/tests/test_coordinator_sync_helpers_coverage_unit.py -k gemm`
- [x] Unit / lint / CodeQL / docs / packaging CI green
- [ ] E2E smoke run (in progress)

### DIFF
--- a/src/hyperloom/inference_optimizer/tests/test_coordinator_sync_helpers_coverage_unit.py
+++ b/src/hyperloom/inference_optimizer/tests/test_coordinator_sync_helpers_coverage_unit.py
@@ -520,6 +520,49 @@ def test_gemm_tuning_required_before_kernel_opt(coord: Coordinator, monkeypatch)
     assert coord._gemm_tuning_required_before_kernel_opt() is False
 
 
+def test_only_forge_can_reach_the_gemm_gate(monkeypatch) -> None:
+    """The backend order decides the phase before the gate is consulted.
+
+    ``_on_enter_kernel`` hands the whole phase to GEAK and returns without
+    calling the gate whenever GEAK owns the order, so the gate is reachable
+    only under forge. A backend branch inside it was therefore dead, and its
+    comment ("legacy FP8 + SGLang only") read as a live precision filter --
+    which invites reading a skipped model as precision-filtered when GEMM
+    tuning was in fact never reached at all.
+    """
+    from hyperloom.orchestrator.kernel.request_handlers import (
+        _resolve_gemm_tuning_backend,
+        geak_selected,
+    )
+
+    for order, forge_owns in (("forge", True), ("geak", False), ("", False),
+                              ("forge,geak", False), ("FORGE", True)):
+        monkeypatch.setenv("KERNEL_OPT_BACKEND_ORDER", order)
+        # GEAK owning the phase and forge owning it are exact complements.
+        assert geak_selected() is not forge_owns, order
+        if forge_owns:
+            # ... and where the gate is reachable, the backend is always forge,
+            # so no branch on it can do anything.
+            assert _resolve_gemm_tuning_backend({}) == "forge", order
+
+
+def test_the_gemm_gate_does_not_filter_on_precision(coord: Coordinator, monkeypatch) -> None:
+    """Every precision forge supports must clear the gate.
+
+    Filtering here would silently drop a category that can optimize: real
+    end-to-end KEEPs span bf16/fp16/fp8/fp4/mxfp4, and bf16 dense alone was
+    worth +11.1%.
+    """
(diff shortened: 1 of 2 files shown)


chennyiiis:
Closing: this is a pure cleanup with **zero runtime behavior change**, and it does not solve any real problem. Not worth reviewer time on its own.

Recording what the investigation established, since these facts are expensive to re-derive:

**Backend selection is one switch.** `forge_explicitly_enabled()` — exactly `KERNEL_OPT_BACKEND_ORDER=forge`, case-insensitive — drives both `geak_selected()` (via `_raw_kernel_backend_order()`) and `_resolve_gemm_tuning_backend()`. They are two views of the same switch, hence strict complements. `_on_enter_kernel()` hands the whole KERNEL_AGENT phase to GEAK and returns before the GEMM gate whenever GEAK owns the order, so the gate is reachable only under forge, where the backend can only resolve to `forge`. That is why the deleted `else` branch could never execute. Any non-exact value (`perfskills`, `forge,geak`, unset) falls back to GEAK — documented contract, not a defect.

**GEAK does no GEMM tuning of its own.** `agents/kernel/tools/backends/geak_runner.py` contains no GEMM path at all; it is kernel rewriting only. A GEAK session can still reach `run_gemm_tuning` through an LLM-issued REQUEST (dispatched inline regardless of backend), which lands on the legacy `_run_geak_gemm_tuning` wrapper.

**The FP8 + SGLang rule still lives in the prompt.** After this branch and `policy/gate.py::_validate_fp8_only_action` are out of the picture, the only remaining copies are in `orchestrator/prompts/orchestration.md` — "valid only for FP8 SGLang workloads" and "On FP8 SGLang run `run_gemm_tuning` first". Those two are the ones that actually take effect, since they decide whether the LLM issues the request at all. Note this sits oddly with `_run_geak_gemm_tuning`'s own docstring, which states Hyperloom does not decide precision/framework applicability and lets GEAK decide. Left alone deliberately for now.

**Forge GEMM tuning itself is healthy.** Verified against the production checkout `/shared_nfs/hyperloom/KernelForge` (`e8810708`, 2026-08-23) on an MI355X box: `forge-gemm-tune plan` resolves Llama-3.1-8B-Instruct as dense bf16 and selects the `sglang_dense_bf16` tuner (`[RUN]`, rc=0).

PR #1269, opened 2026-08-24
fix(breakdown): record reproduced warm replay as an adopted step

luochen-amd:
## Summary

Fixes a Session Breakdown accounting gap where a reproduced warm replay's gain
landed in `final` but not in `optimizations.entries`, so a session that had
measurably gained reported an empty optimization ledger and surfaced its whole
gain as a `reconciliation_gap_pct`. A live scan found 22 such sessions since
Aug 21 (e.g. session 102144, DeepSeek-V4-Pro, +198.9%), covering both
successful and failed `stop_reason`s.

### Root cause

`_promote_to_shared_state` mirrored `replay_warm_recipe` into the canonical
recorder streams **before** the keep decision was made. A warm replay's
executor always settles on `succeeded`, which is not in
`{kept, promoted, adopted}`, so every replay was recorded as `discarded` with
no adoption. `_promote_warm_replay` then ruled the replay reproduced, pushed it
onto `optimization_stack`, moved `cumulative_gain_validated`, and journalled a
KEEP — leaving four places saying "adopted" and only the canonical adoption
stream saying "discarded". The collector builds `entries` from the canonical
adoptions, so the gain had no adopted step to attach to.

### Changes

- Remove `replay_warm_recipe` from the pre-verdict mirror; add
  `_mirror_warm_replay_verdict`, invoked from `_promote_replay_warm_recipe`
  **after** `_promote_warm_replay` rules. A reproduced replay is recorded as a
  keep with `attribution_eligible=True` and the baseline/after throughput pair
  the gain was measured against; drift/failed replays stay discarded.
- `validated` is keyed off a present `replay_accuracy`, not `eval_ran`: a replay
  admitted with no usable score records `keep_verdict_unscored`, a scored
  passing one records `accuracy_pass`.
- Stamp `baseline_tput_anchor` on the warm-replay outcome so the mirror can
  state the throughput pair rather than a bare percentage.
- Extract the repeated v4 workload construction into `_v4_workload()`.

Recording is idempotent by `operation_id` (action + task_id), so a resume that
re-enters promote with the stack entry already present still upserts the same
canonical adoption.

## Test plan

- [x] `pytest src/hyperloom/inference_optimizer/tests/test_sbd_optimizations.py` (collector contract: reproduced replay is an adopted ledger entry; drift stays out; unscored keep is recorded as such) — 34 passed locally
- [x] `ruff check` / editor lints clean on all changed files
- [ ] CI: `test_warm_replay.py` end-to-end cases (reproduced reaches the ledger, drift stays out, eval-ran-but-unscored records `keep_verdict_unscored`, scored records `accuracy_pass`) — require `fcntl`, run on Linux CI


ZhengGong-amd (approved):
Approved

xiaofei-zheng (approved):

PR #1270, opened 2026-08-24
Fix framework build routing starvation and phase_history markers.

lishuoshuo-amd:
Scan oldest unrouted targeted builds instead of returning when the newest build is already accounted for, and ack failed builds only after rearm. Record framework phase markers through append_phase_history_event so rows use the canonical schema, ts_unix, and history cap.


ZhengGong-amd (approved):
Approved

ZhengGong-amd (approved):

xiaofei-zheng (approved):

PR #1271, opened 2026-08-24
Tighten PolicyGate freeform wave validation and ProposalScorer contract (P2-36, P2-37)

BaoYunkai:
## Problem

PolicyGate's freeform `tasks[]` wave gate skipped non-dict entries and empty descriptions whenever any valid item remained, so partially invalid or entirely empty waves could pass validation without surfacing the failing index. ProposalScorer keyed results by display name (allowing silent overwrites), parsed the first `scores` JSON envelope in a reply, accepted boolean coercions as numeric scores, and could invoke duplicate model slugs on repeated API calls.

## Fix

Wave validation now rejects the entire dispatch on any malformed `tasks[i]` via a shared `validate_freeform_wave_task` helper used by PolicyGate and explore fan-out; domain dispatches also reject negative `max_turns`, and SpecialistRunner explicitly maps `max_turns=0` to the default unbounded cap. ProposalScorer assigns stable `proposal_<n>` IDs for prompting and parsing, rejects duplicate proposal names and model slugs plus boolean scores at input, and selects the last valid tail `scores` envelope while still emitting display-name keys for existing render paths.

### DIFF
--- a/src/hyperloom/common/jsonio.py
+++ b/src/hyperloom/common/jsonio.py
@@ -201,6 +201,57 @@ def extract_first_json_with_key(
     return found
 
 
+def extract_last_json_with_key(
+    text: str,
+    required_key: str | None = None,
+) -> dict[str, Any] | None:
+    """Return the last JSON object in *text* (by start offset) that qualifies.
+
+    Scans fenced blocks and every bare ``{`` opener in document order so a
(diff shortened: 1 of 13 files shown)


ZhengGong-amd (approved):
Approved

PR #1272, opened 2026-08-24
fix(breakdown): make reported counts and evidence match their contracts

zengleixin-amd:
Fixes P2-38, P2-39 (PR-4) and P2-42, P2-43, P2-44 (PR-5).

The breakdown pipeline stated four contracts in comments and enforced none of
them. Both ends leaked, each in its own way, and no test asserted any of the
four — which is how they survived.

| Contract, as written | What the code did |
|---|---|
| `schema.py`: `keeps` = "kernels adopted at integrate (NOT micro-only KEEP)" | counted exactly those |
| `recorder.py`: "safe across processes and on network filesystems" | `record_upsert_*` is read-merge-write under an in-process lock |
| `base.py`: skipped sections "emit a one-line not-run note" | never implemented |
| `llm_prompt.py` SYSTEM_PROMPT: "do not invent numbers" | plain prose, zero enforcement |

## What changed

**P2-42 — `collectors/timeline.py`, `schema.py`**
A lane could report `keeps=8, status=kept` for three kernels, none of which
ever passed integrate. Three separate faults: micro-only KEEPs counted as
adoptions; invocation rows counted instead of distinct kernels; `task_id` never
reached the de-dup key because `change` is always set, so two tasks sharing a
wall-clock second collapsed into one row. Micro-only KEEPs now have their own
tally and no longer make a lane read as `kept`.

**P2-43 — `reporters/cross_section.py`, `reporters/compose.py`**
A skipped section's key facts and warnings were discarded, making "this never
ran" indistinguishable from "this ran clean" — the reader sees no bad news and
concludes there is none. That evidence now reaches `data_quality_flags`.

Sections nothing produces (`decision_journal`, `kernel_profiling`,
`kernel_decision_path`, `data_provenance`) are held apart from sections with no
data *this session*: they are always skipped, so four permanent flags would
have drowned the ones that describe the run — and "no data recorded this
session" is the wrong explanation for a feature that was never wired up.

`geak_invocations` / `forge_invocations` rendered but had no `SECTION_GROUPS`
entry, so the report quoted their adoption counts while showing none of the
attempts behind them. Fixing the counts without this would have left the
corrected numbers invisible.

**P2-44 — `reporters/llm_prompt.py`**
Model prose was pasted verbatim beneath a heading the composer owns. A line
opening a heading, code fence or block element re-parents every deterministic
block that follows — a wrong paragraph becomes a wrong document. Such lines are
stripped; prose past the length the prompt asks for is dropped whole rather
than truncated (half a sentence reads worse than none). Both cases fall back to
the deterministic path that already existed, so no new code path was needed.
Number-checking is deliberately out of scope: it needs the response schema made
explicit first, and over-eager matching would flag ordinary phrasing.

**P2-38, P2-39 — `recorder/recorder.py`**
The docstring now names which methods are actually cross-process safe. No lock
was added: `agents/**` and `multi_node/**` contain no writers today, and an
advisory lock is unreliable on the network filesystem this spool lives on — a
guard test enforces that instead, so the promise cannot quietly become false.

Keyless items drew two sequence numbers, so a filename never matched its own
envelope and a `seq=N` trace line pointed at a file that did not exist. Stable
filenames now carry a digest of the untouched key, because sanitizing folded
`a/b`, `a:b` and `a b` onto one file. Fragments already written under the old
name keep it, so resuming a session updates its fragment instead of forking a
second one.

## Tests

26 new cases. Three guard against future drift rather than today's bug:

- the dead-section list is **recomputed** from the real exporter + recorder key
  space, so wiring up a producer (or adding another dead section) fails the suite
- registry and `SECTION_GROUPS` are compared **both ways**, so nothing can render
  without reaching the report again
- the import guard **ships with a test that it can fail** — a rule that cannot
  fail protects nothing

`test_forge_lane.py`'s capability assertion pinned the
micro-only-counts-as-adopted behaviour; it now supplies the integrate verdict
that makes its KEEP a real adoption.

## Verification

- breakdown suite: **158 passed, 2 failed**; both failures are pre-existing
  Windows path-separator assertions, reproduced on `main` at the same commit
- `ruff check` clean; `ruff format --check` clean **for the files this PR
  touches** — four files under `breakdown/` were already unformatted on `main`
  and were deliberately left alone rather than swept into this diff
- unit-level only: no end-to-end run against a real session breakdown

### DIFF
--- a/docs/reference/session-breakdown.md
+++ b/docs/reference/session-breakdown.md
@@ -445,10 +445,27 @@ T+90 min" charts.
 ## `capability_summary` — `CapabilitySummary`
 
 One card per live capability (`geak`, `forge`, `explore`, `sweep`,
-`specialist`) with: `status`, `attempts`, `keeps`, `tested`,
-`best_gain_pct`, `reason`. Legacy `backends`, `params`, and
-`validate_stack` rows can appear when archived sessions are rebuilt.
-Drives the per-session UI cards in Primus-Claw.
+`specialist`) with: `status`, `attempts`, `keeps`, `micro_only_keeps`,
+`pending_integrate`, `reverts`, `e2e_gain_pct`, `tested`, `best_gain_pct`,
+`reason`. Legacy `backends`, `params`, and `validate_stack` rows can appear
+when archived sessions are rebuilt. Drives the per-session UI cards in
+Primus-Claw.
+
+For the kernel lanes (`geak`, `forge`) these counts are not interchangeable:
+
+- `keeps` — **distinct kernels adopted at integrate**, i.e. end-to-end
+  verified. A kernel re-tried across runs counts once.
+- `micro_only_keeps` — kernels that cleared the micro benchmark but never
+  reached integrate. Not adoptions: a faster kernel in isolation does not
+  imply a faster service.
+- `pending_integrate` — kernels whose integrate verdict is `NEEDS_REVIEW` or
+  not yet recorded. Undecided, not successful.
+- `reverts` — kernels integrate rejected (end-to-end regression).
+- `attempts` — **invocation rows**, not distinct kernels: how many tries the
+  lane made. Deliberately a different unit from `keeps`.
+
+The `specialist` row uses `keeps` / `attempts` differently: see
+`CapabilitySummary` in `schema.py`.
 
 ---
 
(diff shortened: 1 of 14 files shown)


ZhengGong-amd (approved):
Approved

PR #1273, opened 2026-08-24
fix(orchestrator): P2-33/34/35 budget+roofline+kernel-ledger numbers, P2-45/47/48 lock+trace+receipt durability

chennyiiis:
## Summary

Audit follow-up covering **PR-2 (P2-33, P2-34, P2-35)** and **PR-6 (P2-45, P2-47, P2-48)** from the P2 review list. Both clusters are cases where a *reported* number or record was wrong, or where a persisted record could be torn — no new behaviour.

### Reported numbers — P2-33 / P2-34 / P2-35

**P2-33 · a 0-second budget disabled the budget gate**
`conc_sweep`'s `total_budget_sec` used one value for two meanings. Clamping the configured budget against the remaining session clock (minus the 120 s CLOSE reserve) could produce `0`, and downstream `has_budget = total_budget_sec > 0` read that as "budget gate off" — so a sweep that should have skipped instead ran the whole CONC ladder unbounded, past `--max-hours`.

- `None` is now the wire value for "no budget gate"; `<= 0` means "no time left".
- `phases/sweep.py` declines to enqueue when the clamp leaves nothing and records its own terminal skip (`session_time_budget`), instead of enqueueing a 0.
- `kernel/conc_sweep.py` returns a `no_time_budget_remaining` skip before booting a server. It does **not** stamp `budget_exhausted`, so `conc_sweep_declined_to_run()` still reads it as a decline rather than a sweep that spent its budget.
- The executor no longer coerces an explicit `None` to `0`.

**P2-34 · roofline within% could exceed 100, and the two sides' ceilings were assumed equal**

- `within_roofline_pct` is capped at 100 so `gap_to_roofline_pct` can no longer go negative. The uncapped ratio is kept as `within_roofline_pct_uncapped` with a `roofline_ceiling_exceeded` flag — beating the modelled ceiling means the ceiling is wrong (stale dtype / quantization assumption), which the report should say rather than round away.
- New `ceilings_comparable()`: when a runtime dtype / quantization change moves the theoretical ceiling, the two sides' within% have different denominators. In that case the `within` / `gap` deltas are withheld (`None`), `ceilings_comparable: False` lands on the comparison block, the table prints **both** ceilings instead of one "single-source ceiling" line, and the report's caveat text switches accordingly.
- Both new snapshot fields are mirrored into the breakdown section + schema.

**P2-35 · a rejected grouped kernel was reported as IN_FLIGHT**
A grouped task's members are intentionally left out of `rejected_kernel_ids` (synthetic ids, members can be re-dispatched), but `attempt_summary` classified only on that set, so a terminally rejected group task read as `IN_FLIGHT`. `_classify_attempted` now also reads the task row's terminal decision (`integration_status`, `integration_rejected_reason` / `rejected_reason`) — the same rule `machine_state` already uses — and the rejection-reason buckets accept either spelling.

### Concurrency and persistence — P2-45 / P2-47 / P2-48

**P2-45 · local KB lock contract + unlocked graph mutation**

- Every store operation constructed a fresh `_CidLock`, so its `threading.Lock` was a new, uncontended object: two threads in one process got no mutual exclusion from it at all. The mutex is now shared per lock-file path through a process-wide registry.
- `KGClient.emit_fact` was an unguarded read-modify-write over a whole markdown page. It now serialises per page slug and verifies the write by re-reading: a page that a concurrent writer replaced is retried (up to 3 attempts), while an unchanged page is treated as a read that has not caught up (no pointless retry, still reported as written).

**P2-47 · call pairing, reasoning tokens, flush semantics**

- Token and conversation halves paired on `(component, task_id, dyn_id, tick, turn, role, model, utc_second)`. A call whose two rows land either side of a second boundary never paired, and two calls inside one second could cross-pair. Both records now carry a `call_id`, stamped by the backend that produced the turn (Claude / Codex / codex-agent metadata) and threaded at the critic / proposal-scorer call sites; `pair_key` prefers it and falls back to the old key when absent, so out-of-process shards keep working unchanged.
- `reasoning_output_tokens` reaches the unified ledger (it was parsed and then dropped by `from_metadata`) and is reported as its own Langfuse usage key — folding it into `output_tokens` would misreport the visible reply.
- `flush_session` ran its reconcile steps in one `try` and set `_flushed = True` in `finally`, so one failing step permanently skipped the rest. Steps are now tracked individually: a retry re-runs only what failed, and `_flushed` is set only once every step has succeeded. The receipt reports `flush_steps_done`.

**P2-48 · non-atomic Langfuse receipt**
The receipt is the cross-process idempotency record for the `session_start` and breakdown pushes, but was written with a plain `write_text()` — a crash mid-write could either replay a push or suppress one forever. It is now written via `atomic_write_json` (temp file + rename + fsync) and carries a `payload_sha256`.

## Test plan

- [x] `ruff check` clean.
- [x] New unit tests, all green locally:
  - budget gate: `None` runs unbounded, `0` skips without booting a server, clamp ≤ 0 declines the enqueue, clamp math (`5 min → 180 s`), executor keeps `None`.
  - roofline: overshoot capped with the uncapped value preserved, `ceilings_comparable` detects a moved ceiling and a unit change, saturation delta withheld while ceiling-independent deltas survive, table renders both ceilings.
  - kernel summary: grouped terminal rejection / integration classified from the task row, reason readable under either key.
  - KB: `_CidLock` mutex shared per path and serialising threads, `emit_fact` retries a lost update, stops on a stale read, reports failure when every write is lost.
  - trace: `call_id` pairs across a second boundary, distinct ids in one second do not cross-pair, reasoning tokens reach `usage_details`, a partial flush retries only the failed step, receipt written atomically with a hash.
- [x] Regression check against `origin/main`: the related test files were run per-file on both branches and the failure sets match (the only difference is `test_local_recipe_store.py`, which fails 2 on main and 0 here). The remaining failures are pre-existing Windows-only ones (`fcntl` / `resource` / POSIX paths); CI on Linux is the real gate.

## Notes for reviewers

- **Contract change:** `total_budget_sec` now distinguishes `None` from `0`. The only producers are `phases/sweep.py` and the CLI default; the old "`<= 0` disables" test was updated to the new split.
- **File overlap with PR-3 (P2-36/37):** the `call_id` threading touches `scoring/proposal_scorer.py` and `specialists/runner.py` (one parameter / one field each). @baoyunkai heads up on a possible small conflict.
- Reasoning tokens are recorded in the ledger but deliberately not folded into the breakdown's spend rollup — reasoning output is priced differently, so that is a separate decision.


chennyiiis:
Review addressed in `df29cc82a`. Every item below was a failure path the first pass left open, so each fix comes with the test the review asked for.

**Blocking**

1. **`_CidLock` leaked the shared mutex on a failed setup** — the mutex is taken per path and lives for the process, so a failed `mkdir` / `os.open` blocked every later write to that cid forever. The whole acquisition now runs under one handler that closes any fd and releases the mutex before propagating. Test: first `os.open` raises, the mutex is observably free and the next acquisition succeeds.

2. **Roofline overshoot was persisted but never rendered** — a 130%-of-ceiling run still read as "Within 100% / Gap 0%". Both the single-snapshot and Base/Opt tables now emit a `Within roofline % (uncapped)` row plus a `> **Ceiling model exceeded (Base|Opt)**` warning that names the ratio and points at the dtype / quantization / GPU-spec inputs behind the understated peak. Tests are on the rendered markdown (single, before/after, and the no-overshoot case that must stay clean), not just the snapshot fields.

3. **`client.flush()` sat outside the step loop** — a failed final flush still set `_flushed` and stranded the SDK buffer. It is now `("client_flush", self._flush_client)`, the last step, and raises instead of swallowing. Test: a fake client whose `flush()` raises once — `counts_final` is `false` after the first `flush_session()` and `true` only after the retry.

4. **A failed Generation was already dropped from `_pending`** — nothing could retry it. `_emit_generation` now returns whether the send happened; `_flush_pending_halves` re-queues unsent halves (`_requeue_parts`, which never overwrites a newer live row) and raises so the step stays unfinished. The live `_buffer` path re-queues too, including the error-row branch. Test: the first send fails, the row is still buffered and the step is not done; the retry emits exactly one Generation.

5. **`emit_fact` trusted an unconfirmable write** — `put_page`'s in-band error was ignored, and both "verify read failed" and "page unchanged" returned `True`. It now checks `_rpc_failed(res)`, confirms through `_fact_confirmed` (bounded re-read with a small linear backoff, for eventually-consistent reads), retries an unconfirmed write, and returns `False` when the attempts are exhausted. The test that pinned the old behaviour is replaced: an unreflected write is now `False` after 3 attempts, and a new test covers the in-band error being retried into a success.

**Reporting consistency**

6. **Reasoning tokens only reached the ledger from the metadata backends** — the same model billed differently per call path. Reading them is now one helper, `parse_usage.reasoning_output_tokens` (handles the canonical key, `reasoning_tokens`, and the OpenAI `completion_tokens_details` nesting as mapping *or* SDK object), used by the critic accumulators, the proposal scorer and the forge-log scrape. `parse_forge_usage` carries the counter through instead of dropping it in `normalize_usage`.

7. **`grand_total` claimed "all-in" without reasoning** — it now includes it, with `total_reasoning_out` rolled up as its own family beside visible and cache tokens; `total_out` still counts only what the model said. `TokenBucket` / `DecisionTokens` / `TokenUsageBucket` carry the field, and `test_decision_token_rollup_unit.py` pins the arithmetic (including that a non-reasoning session's `grand_total` is unmoved).

**Minor**

8. **The receipt hash was a dead field** — `read_receipt` now recomputes it and ignores a receipt whose body no longer matches, so a corrupted file cannot replay or suppress the one-shot `session_start` / breakdown pushes. A receipt with no hash is still accepted (written before the field existed). Test: tamper with a count, the receipt reads back as `None`.

9. **The durability claim was stronger than the write** — `atomic_write_text` / `atomic_write_json` gained an opt-in `fsync_dir`, and the receipt uses it, so the rename itself is durable rather than only the temp file's contents. It is a no-op where the platform has no directory fd.

10. **Docs were out of sync** — `docs/reference/environment-variables.md` now describes the step-wise flush (with `flush_steps_done` / `counts_final` and the receipt hash) and the four counter families behind `grand_total` / `total_in_out` / `total_reasoning_out`.

Verification: `ruff check` and `ruff format --check` clean; 721 tests pass across the affected files. The remaining local failures (`test_common_utils`, `test_breakdown_exporter_unit::test_json_default_path`, `test_critic_agent_backend::test_critic_agent_real_runtime_clears_proposal`, `test_relative_to_session`) fail identically on `origin/main` — they are Windows-only (`os.getuid`, POSIX paths); Linux CI is the gate.

chennyiiis:
Second round addressed in `6d1164613`.

**1. Blocking — conc-sweep lease reclaimed valid work.** The lease came from the configured `conc_sweep_total_budget_sec`, which is the wrong bound in exactly the cases this PR introduced: a non-positive config on a long session produces a *clamped* budget larger than 9000 (5 h left → 17880), and an unbounded sweep has no budget at all. The lease now comes from the budget the task was actually enqueued with, plus a 600 s grace for server teardown and the report flush (`_conc_sweep_lease_ttl_sec`). An unbounded sweep gets `0`, the registry's documented "no lease" encoding (`reclaim_expired_running` skips `ttl <= 0`) — opting out of the watchdog beats being reclaimed at an arbitrary deadline. Tests cover both: `clamped == 17880 → lease 18480`, and `unbounded → lease 0`.

**2. Important — rejection totals vs breakdown.** Correct: the bucketing only ran in the top15 loop. Extracted as `_rejection_bucket(reason)` and called from both loops. Test asserts `sum(rejection_breakdown.values()) == totals.rejected` with one top15 kernel and one ledger-only kernel (the latter also exercising the `max_failures_*` collapse).

**3. Important — ext shard sends were dropped silently.** Fixed, and made durable rather than in-memory: `_flush_ext_shards` now keeps a **per-shard row cursor** persisted in the receipt as `ext_rows_sent`. The cursor only advances over contiguous successes, so a failed row stops that shard instead of being stranded behind later ones; the step raises while anything is outstanding, so it is not marked done. Two tests: a failed second row leaves the step unfinished and the retry emits exactly the missing row (no duplicate); and a *new emitter* for the same session restores the cursor, skips the sent row, and still picks up a row appended afterwards.

**4. Important — receipt was not a cross-process claim.** Right, `read → emit → write` is not a CAS. `record_session_start` / `record_session_breakdown` now take an exclusive-create marker (`reports/trace/.session_start.claim`, `O_CREAT|O_EXCL`) and stand down on `EEXIST`, so the filesystem arbitrates instead of the receipt. Documented trade-off: a claim holder that dies before emitting loses that one observation, which is the accepted cost against a duplicate on every concurrent shutdown; an unwritable claim path falls back to the receipt check rather than dropping the push. Test simulates the race by deleting the receipt after the first emit — the second emitter does not re-emit.

**5. Important — `flush_steps_done` not restored.** I went the other way on this one, and updated the docs instead of the behaviour, because restoring the step set would have broken two things: `pending_halves` drains an *in-process* buffer (a step another process finished is not the same work — restoring it would strand this process's buffered calls), and a resumed session writes *new* ext shards, which a restored `ext_shards` step would skip entirely. So the durable unit is finer than a step: `ext_rows_sent` (item 3) is what crosses processes, plus the exclusive claims (item 4) for the one-shot pushes. `_flushed` / `flush_steps_done` stay per process by design, and the docs now say exactly that — including the remaining gap, which I'd rather state than paper over: the audit backfills (recipe-KB, specialist intel, forge steps, GEMM tuning, decision scores) are **not** cursor-tracked, so a second process reaching CLOSE for the same session re-emits those spans. Giving them cursors means threading a success signal through `record_kb_span` (which swallows its own errors today); happy to do it here or as a follow-up, your call.

Verification: `ruff check` / `ruff format --check` clean; 453 tests pass across the affected files, with only the pre-existing Windows-only `test_relative_to_session` (POSIX path separator) failing — it fails identically on `origin/main`.

ZhengGong-amd (approved):
approved

PR #1276, opened 2026-08-25
Set --pids-limit in docker-mode launch templates (fixes silent hangs + OpenBLAS crashes under Ray + TP>1)

skariapaul:
## Summary

Ray's prestarted worker pool (`--num-cpus` workers) plus a TP>1 framework server (one heavy process per GPU, each with its own OpenBLAS thread pool) reliably exceeds Docker/Podman's default `pids.max` cgroup limit (2048) before the server finishes booting in docker-mode. Depending on which thread loses the race for a PID slot, this surfaces either as a loud OpenBLAS `pthread_create failed: Resource temporarily unavailable` crash in the server, or as complete silence: the orchestrator logs the benchmark launch (`baseline_executor: launching Magpie cmd=[...]`) and then never produces a completion, failure, or exception anywhere — because its own subprocess-management thread lost the race for a PID slot instead of the server.

Verified end-to-end running `moonshotai/Kimi-K3` (TP=8) on 8x AMD Instinct MI325X (gfx942) via Podman:
- Confirmed via `cat /sys/fs/cgroup/pids.current /sys/fs/cgroup/pids.max` inside the container: `pids.current` was already at 1659/2048 from Ray alone before vLLM even tried to start.
- A synchronous, out-of-band reproduction of the exact same subprocess call (bypassing the async orchestration layer) reproduced the underlying symptom directly: `OpenBLAS blas_thread_init: pthread_create failed ... Resource temporarily unavailable`.
- Recreating the container with `--pids-limit -1` fixed it immediately: real `Magpie`/`vllm serve` processes appeared, `ray status` showed `8.0/8.0 GPU` in use, and the checkpoint loaded normally.

## Changes

- `examples/hyperloom-custom-advanced/SKILL.md`, `examples/hyperloom-qwen3-8b-3h/SKILL.md`, `examples/hyperloom-qwen3-14b-fp8-12h/SKILL.md`: add `--pids-limit "${HYPERLOOM_PIDS_LIMIT:--1}"` to the docker-run template, plus an explanatory note (why it's required, not cosmetic — mirrors the existing sibling note for the fd-limit/Ray raylet issue).
- `docs/reference/troubleshooting.md`: three new entries, matching the style of the existing sibling entry for the analogous Ray fd-limit problem:
  1. The pids-limit issue itself (symptom/cause/fix, with the `pids.current`/`pids.max` diagnostic command).
  2. A related gotcha found in the same debugging session: a stale `enablement.inflight_task_id` in `state.json` survives `--resume` even with `--enablement off` (which only prevents *new* authoring rounds, not reaping an already in-flight one) — includes a snippet to clear it directly.
  3. `baseline_accuracy_failed` / Explore never proposing anything — for models without an accuracy-eval task mapping yet, `--no-eval` is the correct fix (not a workaround for a bug), documented alongside how to distinguish that from a genuinely idle orchestrator under a very small workload + reduced flag set.

## Test plan

- [x] Reproduced the silent-hang failure on a fresh docker-mode container (default `pids-limit`) with Ray (128 prestarted workers) + vLLM TP=8.
- [x] Confirmed root cause via `pids.current`/`pids.max` inspection and a synchronous out-of-band repro of the exact subprocess call.
- [x] Recreated the container with `--pids-limit -1` and confirmed the benchmark subprocess launches and progresses normally (checkpoint loads, GPUs show 100% utilization, real throughput numbers recorded).
- [x] Ran a full custom-advanced optimization session end-to-end afterward (baseline established, framework-agent correctly screened and rejected an inapplicable upstream PR candidate) with zero crashes.


tsrikris:
@skariapaul Can you pls confirm why so many changes are required across files? 

@xiaofei-zheng there are a significant number of changes on TraceLens, etc. which I'm not sure how to sign off on, it should require a much shorter fix and not involve changing 60% of the files

PR #1287, opened 2026-08-25
Factor in  AMD compute-partition modes (SPX/DPX/QPX/CPX) as an optimizer lever

rpoornac:
## Summary

Makes AMD compute partitioning a first-class optimizer lever instead of something an operator does by hand around the loop.

- **The mode list drives the search.** `--compute-partition-modes spx,dpx,qpx,cpx` expands into one env-only explore variant per mode, seeded ahead of the grid. Explore already stacks a KEEP'd variant's envs onto everything after it, so the winning mode becomes the topology the remaining knobs are explored inside, and each later mode has to beat the best mode so far rather than the original baseline.
- **Infeasible modes are dropped before they cost a run.** A partition gets a fraction of the card's HBM while every stream on it holds a full copy of the weights, so the narrow modes OOM first. `read_hbm_gib()` asks the card its capacity and `prune_infeasible_modes()` drops what cannot fit, reporting the arithmetic for each drop.
- **The lever is discoverable.** It is off unless a flag names the modes, so it now appears in `reports/final.md` (with the flag, the privilege requirement, and the latency cost) and as a `compute_partition` row in the session report's *Capabilities not attempted*.

## Where it deliberately refuses

- **Scriptable frameworks only.** `plan_partition_run` is called by the scriptable runner and not the serving path, so a serving framework would take the env, change nothing, and file the number under a mode the card was never in. Refused at launch, with the generator declining as a second line.
- **A set is not a set until it reads back.** `amd-smi set` reports success for a change that has only been staged, so every mutation re-reads the mode and compares.
- **Capacity outside SPX is unknown, not guessed.** `amd-smi` reports VRAM per device, and under a split mode a device is a partition, with nothing in the payload saying which. Dividing an already-divided figure again would understate capacity eightfold and prune every mode that fits.
- **Unknown footprint or capacity prunes nothing.** Wrongly dropping a mode costs the optimization the configuration that would have won. The weights bound is used strictly as a lower bound, so a drop is always a proof.

## Known limitation

No in-tree harness reports peak memory yet. The measured path is plumbed end to end (`peak_gib_per_stream` -> measurement -> `current_best`) and prunes correctly the moment a report carries the number, but on a stock report the axis prunes on weight bytes alone. That bound is sound but loose for activation-heavy models: a 1.26B diffusion model carries ~2.5 GiB of weights against a ~20.7 GiB real footprint. Closing it is a harness-side change.

## Test plan

- [x] Full suite: **15,021 passed**, 59 skipped.
- [x] The 5 failures in `test_external_multi_node.py` are pre-existing and unrelated - confirmed identical on a clean `HEAD` worktree.
- [x] `read_hbm_gib()` verified against a live MI355X: reads 287.98 GiB, and `layout_for` now reports per-partition memory (`CPX (8 x 32 CU, 36 GiB)`).
- [x] Pruning verified live: a 20.7 GiB/stream footprint at 2 streams drops CPX only, matching the 13 OOMs observed in an earlier sweep.

### DIFF
--- a/docs/reference/session-breakdown.md
+++ b/docs/reference/session-breakdown.md
@@ -467,6 +467,12 @@ For the kernel lanes (`geak`, `forge`) these counts are not interchangeable:
 The `specialist` row uses `keeps` / `attempts` differently: see
 `CapabilitySummary` in `schema.py`.
 
+`compute_partition` (the AMD SPX/DPX/QPX/CPX lever) is present only on
+frameworks that can apply it, and is the one row whose `not_attempted`
+means *nobody offered it* rather than *the search declined it* — the
+lever is off unless `--compute-partition-modes` names the modes. Its
+`reason` carries the flag needed to turn it on.
+
 ---
 
 ## `kernel_lifecycle` — `KernelLifecycle`
--- a/src/hyperloom/common/gpu_partition.py
+++ b/src/hyperloom/common/gpu_partition.py
@@ -31,6 +31,14 @@ Two invariants this module exists to enforce:
   devices 7 and 8, not 0 and 1. Selecting by index measures a full card and
   reports it as a partition -- a wrong number with no error attached. Callers
   get :func:`partition_device_predicate` for a CU-count test instead.
+* **The card decides what it supports, not this table.** A mode's name being one
+  of the four known ones says nothing about whether this board offers it, and
+  :data:`MODE_PARTITION_COUNTS` is an assumption about the ladder's width.
+  ``amd-smi partition -a`` states both, so :func:`read_partition_profiles` asks
+  and :func:`partition_count_conflicts` checks the assumption against the
+  answer. The query needs the same privilege as the set and degrades to *no
+  answer* -- never to *supports nothing* -- so an unelevated session stays
+  usable and is told its request went unvalidated.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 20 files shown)


tsrikris:
@rpoornac can you pls check to see why major changes have been added to the TraceLens files? I'm not sure if aspects related to TraceLens need to be handled in this PR

src/hyperloom/agents/kernel/tools/tracelens_skill_runner.py
tsrikris, line 1: Recommend reverting these changes

src/hyperloom/agents/kernel/tools/tracelens_analysis.py
tsrikris, line 1: Recommend reverting these changes

tsrikris (changes requested):

rpoornac:
Closing this in favour of #1288 - same work, but based on the live `main`.

@tsrikris your question was the right one to ask, and the answer is that the TraceLens files were never part of this change. This branch was cut from a local ref for `main` that had since been force-pushed, so its base was a lineage `main` no longer has. Git had no way to show that as "your base is stale" — it reported a tidy 7 commits ahead locally, and the tests passed, because every local measurement was taken against that same dead ref. GitHub compared against the real `main` and correctly attributed the whole diverged month to the PR: 1169 commits and 1308 files, including `tracelens_analysis.py` and `tracelens_skill_runner.py`. Nothing in them was mine to revert.

Rebasing the seven feature commits onto the live `main` brings it back to what it should have been: **20 files, +3452/-31**. One conflict, in `explore.py`, where this branch had refactored an inline block that `main` had not otherwise touched.

The bot findings split the same way. The ones against `prompt_builder.py` and `codex_session.py` went away with the stale history. Three were genuinely ours and are fixed: an undocumented empty `except`, five implicit string concatenations in a list literal (where the suggested "add a comma" would have split a markdown bullet in two), and two `unreachable code` false positives from CodeQL not modelling `pytest.raises` as suppressing.

The rebase also surfaced that `ruff format --check .` — a hard gate — was already failing on two files from the original commits. `pytest` and `ruff check` don't see formatting differences, so only the repo-wide `ruff format` catches it. That's fixed too, and CI is now green apart from the `e2e` smoke run, which is failing across several unrelated branches in the same window.

Opening fresh rather than force-pushing this one so the review starts from the actual diff instead of 20 comment threads anchored to commits that no longer exist.

PR #1292, opened 2026-08-26
Trim the profiler-start transient before steady-state splitting

jiaqiang-dot-liu:
## Problem

`torch.profiler.start()` is a local, unfenced call whose cost varies by seconds across ranks, and SGLang does not barrier after it. Ranks whose profiler comes up first reach the first collective of the step and spin there until the slowest peer arrives. A spin-waiting collective is charged as GPU-busy time, so that wait lands in the trace as one enormous kernel at the head of the window.

On the `Qwen3.8-2.4T-A95B-Quark-MXFP4` reference capture (session `20260825T085359Z-74dc572c`), a single `cross_device_reduce_2stage` held **15747 ms of a 16805 ms window**.

### It is profiler-start skew, not communication

Measured per rank across both captures in that session:

| | run1 | run2 |
|---|---:|---:|
| profiler-open spread across 8 ranks | 30.88 s | 24.57 s |
| spinning ranks | 5/8 | 6/8 |
| all spinning ranks retire within | 0.8 us | 0.6 us |
| `spin_r` vs `(last_open - open_r)` residual | < 1 ms | < 1 ms |

Every spinning rank goes from profiler-open to entering the barrier in 4.08–5.71 ms (the six kernels ahead of the embedding all-reduce). The earliest rank was ready and waited; the others were still inside `start()`.

The window is not structurally special either — all 128 steps are `step[DECODE bs=64]` with `g_sqsk` incrementing by 64, so step 0 is an ordinary mid-stream decode step that happens to be the first one after the profiler opened.

Control: `Kimi-K3` on the same host, same image, launched 90 s apart, same `TP`/`CONC`/`ISL`/`OSL`/`start_step`/`num_steps`/`with_stack` — profiler-open spread **0.98 ms**, longest kernel 0.5 ms. It runs with `shape_discovery: false` and no `kernel_shape_profiler`.

### Consequences, which ran the length of the session

- Compute% read **5.75%**, tripping the 10% low-compute gate, which suppressed the entire hot-kernel candidate list at PRELUDE (`hot_kernels: []`, `source resolution: 0/0`). TraceLens had in fact already ranked the three BF16 `aiter::gemm_a16w16` decode shapes as **P1** with `aiter/tuned_gemm.py` resolved — the same three a specialist rediscovered by hand five hours later for the session's only validated gain (+2.58%).
- Every kernel's share of the window was deflated **15.9x** (4.2x on the second capture), so `gpu_pct` understated every dispatch candidate.
- The before/after Compute%/Comm% deltas in the final report (`+17.9` / `-18.0`) tracked the size of the barrier rather than any optimization. Across the 15 valid rank pairings the reportable delta ranges `+0.0` … `+20.5` pt.

## Fix

The collective is self-healing — releasing it re-synchronises every rank — so contamination is confined to the first step. The following 127 steps measure **32.94 ms ± 0.3**:

```
step 0        : 15781.320 ms      <- 479x the median
steps 1..31   : mean 32.944  min 32.649  max 33.153
steps 32..127 : mean 32.990  min 32.627  max 33.230
```

`pretrim_startup_transient()` cuts the leading outlier steps before the trace reaches `split_inference_trace_annotation`. TraceLens and the splitter are untouched.

**Instrumentation is untouched.** `shape_discovery` and the `with_stack` frames its shapes ride on are kept — `pseudo_op::moe_flydsl_stage1/2` carry their `perf_params` via `kernel_shape_profiler.py(364): impl`, a `python_function` frame, so disabling either would cost the MoE perf models (35% of compute).

### The cut is per timeline, not a single timestamp

The host runs a step ahead of the device, so the first kept step's launches are issued *while the dropped step still owns the GPU*:

```
dropped step 0, device : [4687499639913, 4687515421233]   15.78 s
kept    step 1, host   :         [4687514582330, 4687514585612]   <- 839 ms INSIDE the above
kept    step 1, device :                        [4687515421306, 4687515454269]
```

No single boundary separates them:

| cut at | drops step 0 whole | keeps step 1 whole |
|---|---|---|
| device boundary | ✅ | ❌ strips step 1's host ops, and with them the `kernel_shape_profiler` frames its shape attribution rides on |
| host boundary | ❌ leaves step 0's post-barrier kernels as orphans | ✅ |
| **per timeline** | ✅ | ✅ |

Device-timeline categories (`kernel`, `gpu_memcpy`, `gpu_memset`, `gpu_user_annotation`) cut at the device boundary; everything else cuts at the host boundary; flow events follow the endpoint they sit on (`ph:"s"` host, `ph:"f"` device). Result on the reference trace: **127 device and 127 host step annotations with identical name sets** — no step half-present.

### Other design notes

- Detection is on **step duration vs median**, not on any kernel name, so the same guard catches other head-of-window transients (first-replay JIT, KV-pool growth).
- Only a **leading** run is cut. A slow step mid-window is real behaviour the splitter and health gates should still see.
- More than `_PRETRIM_MAX_DROPPED_STEPS` (4) leading outliers means the run never reached steady state — surfaced, not trimmed away.
- `ph:"M"` metadata is preserved regardless of timestamp. torch stamps it with the profiler-open time, always before the cut, so a plain `ts` filter strips all 68 records. TraceLens's own splitter drops these downstream anyway, but the pretrimmed trace is also a readable artifact and should not be the thing that loses them.
- `raw_trace_path` still points at the untrimmed capture, so whole-run graph-capture health is read from the full trace.
- Report goes to `tracelens/pretrim.json` and is registered as an artifact rather than parked in the status file, which every later step rewrites.
- **No new switches.** The outlier multiple is a module constant: healthy decode steps hold inside 2% of the median while the transient runs 479x over it, so there is no operating point between them worth exposing — and a trace whose measurement window is dominated by a synchronisation artifact is not one any caller should be electing to analyse.

### Downstream step count is unchanged

The splitter caps its window at `--num-steps`, so it still hands TraceLens exactly that many steps — only which ones changes:

```
[decode_only] Longest pure decode-only run: [0, 127) (127 steps). Selected [0, 32) (32 steps, capped at num_steps=32).
```

128 → 127 stays far above the 32 requested. Derived counts hold with it: `operation_count` for the MoE ops stays 92 layers × 32 steps = 2944, and `Percentage (%)` is against the compute total rather than the window. The trim is refused outright when fewer than `--num-steps` would remain.

## Validation on the reference trace

`runs/roofline/99e9c7e5.../torch_trace/1787655005.0139868-TP-3.trace.json.gz`

| | before | after |
|---|---:|---:|
| dropped | — | 1 step / 15781.3 ms @ 478x median 32.99 ms |
| device / host step annotations | 128 / 128 | 127 / 127, identical name sets |
| longest kernel | 15747.04 ms | **0.16 ms** |
| window `[0,32)` | 16804.8 ms | **1056.4 ms** |
| `ph:"M"` metadata | 68 | 68 |

Every event category retains 98.6–99.2%, i.e. one step's worth removed uniformly:

```
python_function     1054079 / 1064036   (99.1%)      cpu_op            80746 /  81920  (98.6%)
ac2g                 319381 /  321921   (99.2%)      cuda_runtime       5711 /   5761  (99.1%)
kernel               312545 /  315008   (99.2%)      user_annotation    1016 /   1024  (99.2%)
kernel_shape_profiler   3048 /    3072  (99.2%)      gpu_user_annotation 507 /    511  (99.2%)
```

Runtime 26.6 s on a 1.79M-event / 25 MB gz trace, against a 1413–2210 s TraceLens run.

Expected downstream: Compute% 5.75% → ~91.6%, Comm% 94.22% → ~8.4% (matching TraceLens's own steady-state estimate in the same report), and `priority_data.baseline_ms` 16804.84 → ~1054 so the GEMM impact score goes 0.55 → ~8.7.

## Tests

11 new tests in `test_tracelens_csv.py`. The fixture reproduces the host-leads-device offset, since that is what makes the single-timestamp cut wrong.

Barrier trimmed · clean trace no-op · mid-stream outlier preserved · too-many-leading-outliers refusal · short-trace skip · unreadable-trace degradation · no-step-annotations · default-threshold boundary · host side of first surviving step preserved · dropped step's device work does not leak past the host cut · enough steps left for the splitter.

`ruff check` / `ruff format --check` clean. Full file: 276 passed, 1 failed — `test_run_tracelens_skill_uses_hermetic_claude_env`, which fails identically on `main` (verified via `git stash`).

### DIFF
--- a/src/hyperloom/agents/kernel/tools/tracelens_analysis.py
+++ b/src/hyperloom/agents/kernel/tools/tracelens_analysis.py
@@ -978,6 +978,251 @@ def count_gpu_kernel_events(trace_file: Path, max_events: int = 1_000_000) -> in
     return count
 
 
+#: How many times the median step a leading step must exceed before it counts as
+#: a profiler-start transient rather than real work. Observed transients run
+#: two-to-three orders of magnitude over the median (479x on the reference
+#: capture) while healthy decode steps hold inside 2% of it, so anything in this
+#: range separates them with a wide margin either way.
+_PRETRIM_OUTLIER_FACTOR = 10.0
+
+#: Fewest step annotations a trace must carry before the median is worth
+#: trusting. Below this a single slow step skews the median enough that the
+#: comparison stops meaning anything, so the trim is skipped rather than guessed.
+_PRETRIM_MIN_STEPS = 8
+
+#: Ceiling on how many leading steps may be dropped. A transient is one step in
+#: every capture examined; needing more than a handful means the run never
+#: reached steady state, which the splitter and the health gates should see
+#: rather than have quietly trimmed away.
+_PRETRIM_MAX_DROPPED_STEPS = 4
+
+
+#: Categories that live on the device timeline. Everything else in a torch
+#: trace -- ``cpu_op``, ``python_function``, ``user_annotation``,
+#: ``cuda_runtime`` -- is host-side.
+_GPU_TIMELINE_CATS = frozenset({"kernel", "gpu_memcpy", "gpu_memset", "gpu_user_annotation"})
+
+
+def _step_annotation_spans(events: list[Any]) -> list[tuple[float, float, str]]:
+    """``(ts, dur, name)`` for every GPU-side step annotation, ordered by start.
+
+    Args:
+        events (list[Any]): ``traceEvents`` from a torch-profiler trace.
+
+    Returns:
+        list[tuple[float, float, str]]: One entry per ``step[...]`` annotation on
+            the GPU timeline. Empty when the trace carries no step annotations.
+    """
+    spans: list[tuple[float, float, str]] = []
+    for ev in events:
+        if not isinstance(ev, dict):
+            continue
+        if ev.get("cat") != "gpu_user_annotation":
+            continue
+        name = ev.get("name")
+        if not isinstance(name, str) or not name.startswith("step["):
+            continue
+        ts = ev.get("ts")
+        dur = ev.get("dur")
+        if isinstance(ts, (int, float)) and isinstance(dur, (int, float)):
+            spans.append((float(ts), float(dur), name))
+    spans.sort()
+    return spans
+
+
+def _host_step_starts(events: list[Any]) -> dict[str, float]:
+    """Earliest host-side start for each ``step[...]`` annotation, keyed by name.
+
+    The host runs ahead of the device -- a step's launches are issued while the
+    previous step is still executing -- so a step's host span begins before its
+    device span. Step names embed the cumulative sequence length, so they pair
+    the two timelines exactly.
+
+    Args:
+        events (list[Any]): ``traceEvents`` from a torch-profiler trace.
+
+    Returns:
+        dict[str, float]: Step annotation name to its host-timeline start.
+    """
+    starts: dict[str, float] = {}
+    for ev in events:
+        if not isinstance(ev, dict):
+            continue
+        if ev.get("cat") != "user_annotation":
+            continue
+        name = ev.get("name")
+        ts = ev.get("ts")
+        if not isinstance(name, str) or not name.startswith("step["):
+            continue
+        if not isinstance(ts, (int, float)):
+            continue
+        prev = starts.get(name)
+        if prev is None or ts < prev:
+            starts[name] = float(ts)
+    return starts
+
+
+def pretrim_startup_transient(
+    src: Path,
+    dst: Path,
+    *,
+    outlier_factor: float = _PRETRIM_OUTLIER_FACTOR,
+    min_steps: int = _PRETRIM_MIN_STEPS,
+    max_dropped_steps: int = _PRETRIM_MAX_DROPPED_STEPS,
+) -> tuple[bool, dict[str, Any]]:
+    """Drop the profiler-start transient from the head of a torch trace.
+
+    ``torch.profiler.start()`` is a local, unfenced call whose cost varies by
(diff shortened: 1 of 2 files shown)


src/hyperloom/agents/kernel/tools/tracelens_analysis.py
```diff
@@ -978,6 +978,251 @@ def count_gpu_kernel_events(trace_file: Path, max_events: int = 1_000_000) -> in
...
+            continue
+        if not isinstance(ts, (int, float)):
+            continue
+        prev = starts.get(name)
+        if prev is None or ts < prev:
+            starts[name] = float(ts)
+    return starts
+
+
+def pretrim_startup_transient(
```
tsrikris, line 1188: @jiaqiang-dot-liu Can you comment why this solution needs to be present on Hyperloom? Isn't the right source for it on TraceLens's trace splitter? 

I also believe @devalshahamd has been enhancing the trace splitting. 

I would hold off this PR unless absolutely necessary

tsrikris (changes requested):

src/hyperloom/agents/kernel/tools/tracelens_analysis.py
kyle-hoffmeyer: I think this should be something handled in TraceLens's trace splitter

xiaofei-zheng (approved):

tsrikris:
@xiaofei-zheng can you pls comment why you have merged this PR? This is a sloppy implementation and adds so much code for a simple fix. Suggest reverting this PR ...

CC: @chaojhou

PR #1293, opened 2026-08-26
fix(breakdown): mark unreadable sweep reports as failed

siliangchen-amd:
## Description
Phase 2 P0 (`benchmark-integrity`): `_shape_sweep_point()` defaulted
`status="ok"` and only flipped it when a report was missing or
`success is False`. A truncated / empty / non-object
`benchmark_report.json` hit neither branch, so the variant was written
as `ok` with all metrics null. Downstream dashboards then treated a
corrupt file as a successful sweep point.
This PR:
- Derives status instead of defaulting it: no report → `skipped`;
  unreadable report → `failed` with `error`; readable
  `success is False` → `failed`; otherwise `ok`
- Adds `error` to every `SweepPoint` row (`null` when there is no
  failure reason) so table-shaped consumers see a stable key set
- Points the sweep renderer at the schema field names (`ok`,
  `output_throughput_tok_s`, `ttft_mean_ms`, `e2el_mean_ms`) so
  `session_report.md` counts successes and fills metric columns
- Documents the status semantics in `session-breakdown.md`
Not in this PR: the other Phase 2 P0 (`security-secrets`). No
`schema_version` bump.

## Linked issue(s)
No GitHub issue. Source: `RELEASE-BLOCKERS-Phase2.md` P0
`benchmark-integrity` (`_shape_sweep_point()`).

## Downstream / frontend
`session_breakdown.json` is the consumer contract. This change is
additive and tightens `ok`; it does not rename or remove fields.
- Existing `SweepPoint` keys are unchanged in name and type
- `status` values are still `ok` / `failed` / `skipped`
- New optional field: `error` (present on every row; `null` on
  ok/skipped)
- `schema_version` is unchanged
- `reports/final.json` and `reports/final.md` are not produced by
  this path and are unaffected
The one behavioral change: a variant whose report cannot be read is
now `failed` instead of `ok`. Consumers that filter on `status == "ok"`
will stop including rows with null throughput. That is the intended
fix.
Renderer changes only affect `session_report.md` (operator CLI).
Markdown table headers are unchanged (`conc`, `isl`, `osl`, `status`,
`tput`, `ttft`, `e2el`, `error`).

## Tests
Added/updated:
- `test_breakdown_report_integrity.py` — status matrix (ok /
  success:false / missing / truncated / empty / top-level list);
  stable key-set lock across ok/failed/skipped; collector output
  fed to the renderer (success=1, metrics present)
- `test_breakdown_renderers.py` / `test_low_coverage_renderers_unit.py`
  — fixtures use the schema shape (`ok`, `output_throughput_tok_s`,
  `ttft_mean_ms`, `e2el_mean_ms`)
Commands run:
PYTHONPATH=src python -m pytest
src/hyperloom/inference_optimizer/tests/test_breakdown_report_integrity.py
src/hyperloom/inference_optimizer/tests/test_breakdown_renderers.py
src/hyperloom/inference_optimizer/tests/test_low_coverage_renderers_unit.py
src/hyperloom/inference_optimizer/tests/test_breakdown_exporter_unit.py
src/hyperloom/inference_optimizer/tests/test_reporters_smoke.py
-q -p no:randomly
166 passed. `ruff check` / `ruff format --check` on the changed files
passed.

## Breaking changes
No fields renamed or removed. `ok` now excludes unreadable reports
(previously those rows were fabricated as success). That is a
correctness fix, not a wire-shape break. Consumers that treated
`status=="ok"` as "metrics are present" will now match reality.


siliangchen-amd:
Thanks for the review. Pushed `3a75a936f` on this branch (no rebase). Item-by-item:

**P1 — singular `error` drops native `errors[]` — accepted.** `_shape_sweep_point` now reads singular `error`, then Magpie `errors[]` (joined), then a fixed fallback so a `failed` row never ships `error=None`. Covers `{"success": false}` with neither key as well.

**P1 — renderer folds `skipped` into failed — accepted.** Key fact is now `success=N, failed=N, skipped=N`. The collector→renderer test that previously only asserted `success=1` now asserts all three counts.

**P2 — no bump / no CHANGELOG — accepted as documentation.** `schema_version` is still not bumped: no field is renamed or removed, `error` is additive, the `status` value set is unchanged. The change restores the existing guarantee that missing measurements are not fabricated as success. That rationale is now in `CHANGELOG.md` `[Unreleased]` and in `session-breakdown.md`.

**`require_dict=True` — accepted.** `_load_json_safe` gained a keyword-only `require_dict=False` so telemetry/sessions/kernels/roofline callers are unchanged. Sweep passes `True`; truncated / empty / top-level-list reports now land in `warnings[]` as well as on the row.

**`abort_reason.json` — accepted, and it was larger than the comment.** Sweep goes through `run_grid`, which writes the marker on ~13 abort paths (`no_benchmark_workspace`, `yaml_build_error`, `server_init_dead`, …), several of which never produce a report. Those variants were `skipped`. No report + `abort_reason.json` is now `failed` with `"<error_class>: <error>"`. No per-`error_class` branching.

Boundary, since you asked: the reap path writes the same marker but sets `VariantResult.status="skipped"`. The collector still maps that marker to `failed`. That is intentional — two vocabularies, and a reaped variant did run.

**Duplicate renderer tests — accepted.** `test_low_coverage_renderers_unit.py` was a byte-identical 13-function subset of `test_breakdown_renderers.py` with no unique tests. Deleted it. Status matrix is now parametrized and includes `{}`, `success:true` with no metrics, `success:true` throughput 0, `errors: [...]`, `success:false` with no reason, abort-marker-only, and warnings assertions on unreadable reports.

**`is_valid_measurement()` — not adopted.** Three reasons:

1. It reads `HYPERLOOM_AGENTX` at call time (`agentx_enabled()`). Collectors are documented as pure functions over `session_dir` / `state` / `manifest`. Wiring it in would make the same session directory export different breakdowns depending on the exporter's env.
2. It answers a different question: "usable for **selection**" in the optimizer loop. `SweepPoint.status` answers "did this measurement land". `success:true` with throughput 0 is a readable, self-reported success; it is not a missing measurement.
3. Adopting it would flip `success:true, throughput:0` from `ok` to `failed` — a larger semantic change than this PR, which collides with the same review's P2 on bumping for semantic changes.

That boundary is now a regression test (`success:true` + throughput 0 stays `ok`).

Prompt / `SKILL.md`: agreed, no change needed.

Tests: 248 passed across `test_breakdown_report_integrity.py`, `test_breakdown_renderers.py`, `test_breakdown_exporter_unit.py`, `test_reporters_smoke.py`, `test_geak_breakdown_unit.py`, `test_sbd_optimizations.py`. Ruff check/format on the changed files passed.

siliangchen-amd:
Follow-up `e708b87cb`:

- Renderer only selects a best point among `ok` rows with `_ot(v) > 0`. Empty-shell `ok` variants still count as success, but no longer emit `Best point: tput=None` or `kind=kept`.
- Unreadable `benchmark_report.json` now prefers a sibling `abort_reason.json` for `error` (still `failed`; parse failure still goes to `warnings[]`).
- SweepPoint `error` is `redact_secret_values`'d then clipped to 500 chars (collector convention). Did not import `_report_errors_summary` from `_grid_runner` — collectors should not depend on the executor package.
- Throughput 0 is no longer asserted as `tput is None`. Status stays `ok`; the `or`-chain collapse is an `xfail(strict=True)` known defect, not expected behaviour.

73 passed, 1 xfailed on the sweep integrity/renderer tests.

ZhengGong-amd (approved):
Approved

PR #1299, opened 2026-08-27
Feat/rpoornac/partition mode sweep

rpoornac:
#1288 merged, so this now reviews directly against `main`; the
`hyperloom.common.gpu_partition` module it imports is already there.

## What this is

#1288 made the partition mode a fixed session property: the optimizer reads it,
records it, refuses a session that cannot work in that shape, and never sets it.
That is the right split, but it leaves the obvious question unanswered — *which*
shape does this workload want? — and answering it by hand is four sessions and a
spreadsheet.

This is the other half of that boundary. It sets each mode on one card in turn,
runs the same benchmark on every partition the mode creates, sums the throughput,
prints the comparison, and restores the card's entry mode on the way out. The
privileged `amd-smi set` lives here and nowhere else: between benchmarks, in a
script an operator ran deliberately, rather than inside a loop that also runs
agent-authored code. `optimize` still never changes a mode.

```
mode   parts    CU    throughput  vs entry   worst lat
------------------------------------------------------------------------------
SPX        1   256        2560.0     1.00x   1000.0 ms
DPX        2   128        2560.0     1.00x   2000.0 ms
QPX        4    64        2560.0     1.00x   4000.0 ms
CPX        -     -             -         -           -   2 x 20.7 GiB will not fit a 36.0 GiB partition
```

## The fan-out is the substance, not a detail

A benchmark that loads one partition and ignores the other seven measures an
eighth of the card. A sweep built that way would report `CPX` as catastrophically
slow while looking exactly like a partitioning comparison — which is the
misleading number #1288 promised not to produce, and the reason it shipped
without a sweep. So every partition is launched before any is waited on, and a
mode is reported only when all of its partitions came back. Six of eight summed
understates by a quarter while still looking like a result, so that is
"unmeasured", not "slow".

The table above is the validation. The synthetic workload's throughput scales
with CU count, so a *correct* fan-out shows a constant aggregate across modes —
SPX 1×2560, DPX 2×1280, QPX 4×640 — while a broken one would show DPX at half
and QPX at a quarter. Run on an 8-card MI355X node. The latency column moving
1000 → 2000 → 4000 ms is the cost being paid for it.

## The enumeration rule, now measured rather than argued

#1288 justified publishing the *shape* instead of a device list by asserting that
HIP enumerates whole cards before partitions. That is now verified on hardware,
and the two tools genuinely disagree. With card 0 of 8 in `CPX`:

| | partitions of card 0 | whole cards |
|---|---|---|
| `amd-smi` (orders by PCI address) | devices 0-7 | devices 8-14 |
| HSA/HIP (`rocminfo`) | devices **7-14** | devices 0-6 |

So a device list computed with one tool and handed to the other is wrong, and
wrong invisibly — the benchmark runs to completion, on the wrong silicon. Under
DPX the partitions are HIP devices 7 and 8, exactly as
`partition_device_predicate`'s docstring predicted.

That also settles the loose end #1288 disclosed: `partition_device_predicate` had
no caller in the tree. It has one now, and this script is the fan-out consumer
the published env contract was written for. It sets the same
`HYPERLOOM_PARTITION_*` variables a session sets, so an entrypoint written
against that contract behaves identically under either.

Selection matches CU count **within the swept card's PCI bus**. Both halves
matter: the CU match distinguishes a partition from a whole card, and the bus
match is what stops `SPX` — whose "partition" is a whole card — from collecting
all eight cards on the node.

## Refusals instead of failures

The launch-time feasibility check from #1288 is reused per mode, so a mode whose
partitions provably cannot hold the configured streams is skipped with the
arithmetic shown rather than run into an out-of-memory failure. On MI355X that
correctly refuses `CPX` for a 20.7 GiB stream at two per partition: 36 GiB does
not hold 41.4.

The card is restored on every exit path, including a failure mid-sweep and a
Ctrl-C, and a failure to restore is its own exit code (`3`) rather than a line in
the log — a card left in a shape nobody asked for silently mislabels whatever
runs next. The sweep also refuses to start when processes hold GPU contexts,
since repartitioning evicts them; `--allow-busy` is there for when they are
yours.

## Three traps found on the hardware

All three were real failures during bring-up, and each is now a named test.

- **`rocminfo` puts `BDFID` before `Compute Unit`** in each agent block, so a
  line-at-a-time parser that emits on `BDFID` attributes every agent the
  *previous* one's CU count. It turned the first CPX partition into a 256-CU card
  and would have pointed the benchmark at whole silicon while labelling it CPX.
- **`amd-smi` reports an idle card's `process_list` as the string
  `"No running processes detected"`**, not an empty list, so a length check finds
  one process on every idle card and the sweep refuses to start on a free node.
- **An inherited `HIP_VISIBLE_DEVICES` alongside the ROCR mask applies as a
  second mask indexing into the first**, quietly redirecting every partition's
  work onto one device — every process succeeds and the mode reads as uniformly
  slow. Both it and `CUDA_VISIBLE_DEVICES` are removed rather than set.

## Scope and shape

- `scripts/partition_mode_sweep.py`, plus `scripts/tests/` — 88 tests, none
  needing a GPU. Follows the `platform_audit.py` conventions: argparse, `Usage::`
  docstring, `raise SystemExit(main())`, tests loaded via `importlib`.
- The mode-control machinery (readback verification, busy/drain retry, the
  profile query) is recovered from `ff451d840`, the revision of #1288 that had it
  before the mutation surface was removed from the optimizer. Its principle
  survives intact: **a set is not a set until it reads back.** `amd-smi set`
  reports success for a staged change and exits zero on some permission
  failures, so trusting the exit code means benchmarking the old topology under
  the new mode's name.
- `--dry-run` prints the plan and touches nothing.
- Takes either `--benchmark-config` (the in-tree Magpie invocation) or
  `--benchmark-command` with `{device}` / `{output_dir}` placeholders, substituted
  per already-split token so no shell is involved and a path with a space stays
  one argument.

## What it does not do

It chooses a mode; it does not tune one. The intended sequence is: sweep to pick
the shape, then run `optimize --compute-partition-mode <winner>` so the session
refuses to start if the card is not actually in it. The report says so in as many
words, because a throughput table is easy to mistake for a tuning result.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,7 +55,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
   benchmarks in a script an operator ran on purpose, and unreasonable inside an
   optimization loop that also runs agent-authored code, so `optimize` continues
   to only read the mode. Together the two halves are a boundary: the sweep
-  chooses the shape, the session asserts it.
+  chooses the shape, the session asserts it.<br/>
+  Because that set evicts work, the check standing in front of it fails closed:
+  an `amd-smi` process listing in a shape the parser does not model is a refusal,
+  not an empty one, since the only wrong answer that destroys anything is reading
+  a busy node as free. It is scoped to the card being swept, so a neighbour's
+  benchmark on a shared node no longer forces `--allow-busy` and with it the loss
+  of the guard on the target card. Every exit from a started sweep runs the
+  restore and the report, including on an error the script does not model — which
+  exits `4`, keeps the modes already measured, and still yields `3` if the card
+  could not be put back.
 - **The card's compute-partition shape is now recorded, checked, and published.**
   An MI300-series card can be split into independent partitions (`SPX`, `DPX`,
   `QPX`, `CPX`), and splitting one trades per-request latency for aggregate
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,27 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
   hipBLASLt, and the fusion backend. `deploy/` is also absent -- every file in
   it targets the retired repository.
 
+- **`scripts/partition_mode_sweep.py` measures which compute-partition mode a
+  workload wants.** Sets each mode on one card in turn, runs the same benchmark
+  on every partition that mode creates, sums the throughput, and restores the
+  card's entry mode on the way out — including after a failure or a Ctrl-C.
+  Modes whose partitions provably cannot hold the configured streams are skipped
+  with the arithmetic shown rather than run into an out-of-memory failure.<br/>
+  The fan-out is the substance of it. A benchmark that loads one partition and
+  ignores the rest measures a fraction of the card, which reports `CPX` as eight
+  times worse than it is; every figure here is the sum over a mode's partitions
+  with all of them loaded together, and a mode is reported only when every one of
+  its partitions returned a measurement. Partitions are selected by matching CU
+  count within the swept card's PCI bus, never by device index: `amd-smi` orders
+  by PCI address while HSA/HIP enumerates whole cards first, so on an 8-card
+  MI355X node with card 0 in `CPX` the two tools disagree about which devices the
+  partitions are — 0-7 against 7-14.<br/>
+  This is where the privileged `amd-smi set` lives, and the only place it does.
+  A card-wide mutation that evicts every GPU context is reasonable between
+  benchmarks in a script an operator ran on purpose, and unreasonable inside an
+  optimization loop that also runs agent-authored code, so `optimize` continues
+  to only read the mode. Together the two halves are a boundary: the sweep
+  chooses the shape, the session asserts it.
 - **The card's compute-partition shape is now recorded, checked, and published.**
   An MI300-series card can be split into independent partitions (`SPX`, `DPX`,
   `QPX`, `CPX`), and splitting one trades per-request latency for aggregate
--- a/README.md
+++ b/README.md
@@ -86,6 +86,12 @@ feedback on how to improve Hyperloom by completing the
 - Main agent instructions: [`src/hyperloom/inference_optimizer/SKILL.md`](src/hyperloom/inference_optimizer/SKILL.md)
 - CLI entry point: `python -m hyperloom.inference_optimizer.cli optimize`
 - Operator tools: `python -m hyperloom.inference_optimizer.tools.*`
+- Compute-partition sweep: `python3 scripts/partition_mode_sweep.py` — sets each
+  AMD partition mode (`SPX`/`DPX`/`QPX`/`CPX`) on one card in turn, runs the same
+  benchmark on every partition that mode creates, sums the throughput and restores
+  the entry mode. Answers which shape a workload wants before a session commits to
+  one; `optimize` itself only ever reads the mode. Needs privilege for the set, so
+  it is a script rather than part of the loop.
 - Platform tuning audit: `python3 scripts/platform_audit.py` — checks the host CPU
   tuning that silently changes benchmark results. Judges Core Performance Boost and
   the cpufreq governor against [AMD's BIOS & Workload Tuning Guide for EPYC 9004][58011];
--- a/docs/reference/environment-variables.md
+++ b/docs/reference/environment-variables.md
@@ -534,8 +534,18 @@ second indexes into the first.
 The privileged `amd-smi set` lives here and nowhere else. An operator-run script
 between benchmarks is a reasonable place for a card-wide mutation that evicts
 every GPU context; an optimization loop that also runs agent-authored code is
-not. Exit codes: `0` swept, `1` nothing measurable, `2` refused before anything
-changed, `3` swept but the card could not be restored to its entry mode.
+not. Before setting anything it refuses if a process holds a context on the card
+being swept — and only that card, since no other card is repartitioned. A
+neighbour's benchmark on a shared node is not a reason to stop. If `amd-smi`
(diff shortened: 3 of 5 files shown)


rpoornac:
Rebased onto current `main` (`d5f3561d4`), no content change.

#1288 was squash-merged, so the commit this branch was stacked on
(`ff6ff66f0`) is no longer an ancestor of `main`. GitHub retargeted the base
here automatically, which left the PR showing 18 files and +4218 — all of
#1288 re-presented on top of merge conflicts. Dropping the stale commit and
replaying the sweep's own two onto `main` puts it back to what it should be:

    CHANGELOG.md                               |   21 +
    README.md                                  |    6 +
    docs/reference/environment-variables.md    |   36 +
    scripts/partition_mode_sweep.py            | 1123 +
    scripts/tests/test_partition_mode_sweep.py |  609 +
    5 files changed, 1795 insertions(+)

88 sweep tests pass against the merged tree, `ruff check` and `ruff format`
clean. Worth noting the one thing that could have broken: #1288's last review
round trimmed `__all__` in `common/gpu_partition.py`, and this script imports
`MODE_PARTITION_COUNTS`, `layout_for` and `partition_device_predicate` from
there. Explicit imports are unaffected by `__all__`, and the tests confirm it.

The two earlier CodeQL notes on `_as_float` are stale — that helper is gone,
replaced by the shared `hyperloom.common.coerce.to_float`, which uses
`math.isfinite`.

ZhengGong-amd:
Re-reviewed the current head (`56d0b4382`). The three "must-fix" items from the earlier pass are still present — the only new commit since then (`use the canonical to_float`) is unrelated.

1. **`resident_processes` fails open on schema drift.** `rows = payload if isinstance(payload, list) else []`, and per-row key mismatches just `continue`. A wrapped/renamed `amd-smi process` payload silently becomes "no busy GPUs" and the sweep proceeds to `amd-smi set`, which evicts real processes. This guards a destructive op — it should raise `SweepError` (exit 2) on unexpected shape, not degrade to empty.

2. **Busy check is node-wide, but `set` only touches `--gpu`.** `occupied = {gpu: n for gpu, n in busy.items() if n}` isn't filtered to `args.gpu`. On a shared node, any card being busy blocks scanning an idle target card, and the only escape (`--allow-busy`) also disables protection for the target card itself. Filter to `{args.gpu}`.

3. **An unexpected exception mid-sweep skips reporting and breaks the exit-code contract.** Only `(SweepError, PartitionError)` are caught inside the loop; anything else (verified with a `KeyError` from `set_mode`) propagates out of `main()` past the `finally`. Restore is attempted, but `render()`/`sweep_summary.json` never run, and if restore itself also fails, the process exits with Python's default `1` instead of the documented `3` for "could not restore". Results from already-completed modes are lost.

Would like these addressed before merge.

rpoornac:
All three addressed in `49528a0ec`. One theme in each: a guard in front of the
one destructive operation here answered "fine" where it should have answered
"I don't know".

**1. `resident_processes` fails open on schema drift.** Fixed, and it was worse
than reported. The parser is strict now — a payload that is not a list of
per-GPU rows, a row missing `gpu` or `process_list`, or either field carrying a
type it does not model all raise `SweepError`, which is exit 2 before anything is
set. Three drifted payloads through the old parser and the new one:

| payload | before | after |
| --- | --- | --- |
| dict-wrapped rows | `{}` — a busy node reads as free | refused |
| entry without the `process_info` wrapper | `{0: 0}` — a real process reads as idle | `{0: 1}` |
| `process_list` as the bare idle string | `{0: 29}` — one per character | `{0: 0}` |

The third is the inverse failure and was also live: `for entry in
row.get("process_list") or []` iterates a string character by character, so the
documented idle sentinel at that depth counted as 29 processes and refused the
sweep on a free node — the exact failure `_NO_PROCESS_MARKER` exists to prevent,
reintroduced one level up. Both depths are recognised now.

**2. Busy check is node-wide, but `set` only touches `--gpu`.** Scoped to
`args.gpu`. A target card absent from the listing is now its own refusal rather
than an assumed-idle card: an absent row is "unknown", and unknown does not
license an eviction. `--allow-busy` and `--dry-run` skip the check entirely
instead of running it and ignoring the result — neither has a set for it to
guard, which makes the documented "`--dry-run` touches nothing" true on a busy
node and leaves `--allow-busy` as the deliberate way past a payload the parser
refuses.

**3. An unexpected exception mid-sweep skips reporting.** Reproduced exactly as
you described it, with a `KeyError` out of `set_mode`. Unexpected exceptions are
caught per mode now and stop the sweep by breaking rather than propagating, so
the restore, the table and `sweep_summary.json` all still run and the modes
already measured survive.

The contract needed a fourth code. Reporting a crashed sweep as `0` because an
earlier mode measured would be a lie, and `1` already means "ran, measured
nothing", so an unmodelled failure is `4` — and `3` outranks it, because a card
left in the wrong shape mislabels whatever runs on the node next, not just this
sweep. Two related holes closed alongside: the restore is now a helper that
raises nothing, since a non-`SweepError` out of it used to escape the `finally`
and take the report with it, and it attempts the set even when the read-back that
would have proved it necessary is what broke, rather than skipping the restore
because the check failed.

Tests 88 → 125. The `main()` control flow above cannot be reached from the
pure-function tests, so there is a fake node now — no GPU, no subprocess — and 29
of the new tests fail against `56d0b4382`. `ruff check` and `ruff format` clean.
The CHANGELOG and the environment-variables reference both state the scoping and
the new exit code.

rpoornac:
Rebased onto current `main` (`120cb3262`) to clear a `CHANGELOG.md` conflict in
the `Unreleased` section — `main`'s KernelForge entry landed in the same place.
Both entries kept, no other file conflicted, and the three fixes above are
unchanged; they are `3cf6b3413` now rather than `49528a0ec`.

PR #1302, opened 2026-08-27
fix(security): redact tool-command intel and reject non-http gbrain URLs

siliangchen-amd:
## Description
Two independent fixes.
1. Specialist tool summaries (Bash / WebSearch) were clipped into the intel ledger without redacting credentials, so Bearer tokens and `API_KEY=` assignments were stored verbatim. `_summarize_tool_input()` now runs `redact_secret_values()` on the full string and then clips. Claude and Codex both go through this helper.
2. `GbrainPageClient` documented that `urlopen` was only reached after a scheme check, but the check was never called. `__init__` now uses the existing `require_http_url()`. A bad `GBRAIN_BASE_URL` (`file://`, `ftp://`) is logged and explore skips GBrain (the env builder returns `None`); constructing the client directly still raises.
## Threat model and coverage
The intel ledger this PR redacts *is* in the exported session package (the trace directory's JSONL files are on the allowlist). The raw stream-json / Codex JSONL logs under the specialist run tree are not: the only file selected there is the done-marker JSON. So whoever can read the session directory on the shared filesystem still has the plaintext; whoever only receives the exported package does not. Caveat: with `include_transcripts=true` the done-marker JSON inlines the full transcript. Default is `false`; redacting that path touches the breakdown JSON and is a separate change.
What this helper masks today:
- `KEY=value` / `KEY="value"` / `KEY='value'` assignments whose name contains `API_KEY` / `TOKEN` / `SECRET` / `PASSWORD` / `CREDENTIAL`, except the tokenizer / max-tokens family (`TOKEN(?!IZER|S\b)`)
- `Authorization: Bearer …`
- `ak-` / `sk-` / `pk-` prefixes, `ghp_` / `github_pat_` shapes, AWS access-key ids (`AKIA…`), compact JWTs
Known boundaries (not masked, not claimed):
- space-separated flags (`--password hunter2`)
- `curl -u user:pass`
- `Authorization: token …` (Bearer only)
- JSON-key form (`{"password": "x"}`) — widening the name side also masks `{"input_tokens": 1234}`
## Linked issue(s)
None.
## Tests
- Bearer tokens and `OPENAI_API_KEY=` values are redacted in tool summaries, and redaction runs before clipping
- Non-http GBrain URLs raise; the env builder returns `None` for `file://`
201 passed on the targeted tests. `ruff check` / `ruff format --check` on the changed files passed.
## Breaking changes
No. Intel `query` strings may now contain `[REDACTED]` in place of credential values. A non-http `GBRAIN_BASE_URL` skips GBrain instead of calling `urlopen`.


siliangchen-amd:
According to five comments, two are fixed and three are answered below without a code change. This push also merges current main (the branch was 257 commits behind; no conflicts).
### 1) Redaction did not cover quoted assignments — fixed
Confirmed, all four of your inputs reproduced. Two notes.
One shape is worse than reported. When a tool input has none of the query-ish keys the summarizer prefers, the whole input is serialized with `json.dumps` first — the Write tool always takes that path, since its keys are the file path and the content. The value then arrives as `\"hunter2\"`, and the old pattern matched only the backslash, so the output was `MYAPP_PASSWORD=[REDACTED]"hunter2\""`: a `[REDACTED]` marker sitting right next to the readable secret. That is worse than no match at all.
I did not use `["']?[^\s,;'"]+` verbatim. It consumes the opening quote, so the result is `=[REDACTED]"` with the quoting unbalanced — a serialized summary stops being valid JSON — and it still misses the escaped form above. Instead the opening quote, optionally backslash-escaped, is matched separately and re-emitted: `(\\?["'])?[^\s,;'\"]+` with `\1\2[REDACTED]`. Result: `export MYAPP_PASSWORD="hunter2"` becomes `export MYAPP_PASSWORD="[REDACTED]"`, single quotes likewise, the escaped form masks the value, and a bare value is byte-identical to before — so existing behaviour and tests are untouched.
Deliberately not done:
- `curl -u admin:pass` and `Authorization: token abcdef` are new rules rather than the existing hole. `token` especially: the bearer rule is not anchored to `Authorization:`, so adding `token\s+` would hit token-count prose. Doing it safely means anchoring on the header name, which is its own rule with its own tests.
- The JSON-key form `{"password": "x"}` is still unmasked, because the quote between the name and the colon breaks the name-to-separator adjacency. Widening the name side also masks `{"input_tokens": 1234}` — I measured it — and `redact_secret_values` has a dozen-plus call sites carrying diagnostics, so that needs a numeric exemption and should land on its own.
### 2) Raw logs remain plaintext — definite answer, no code change
The export is an allowlist, not a filter. The trace directory's JSONL files are selected, and the specialist intel ledger lives there, so the ledger this PR redacts is in the bundle. Under the specialist run tree only the done-marker JSON is selected: the raw stream-json and Codex JSONL logs match no pattern and are never packaged.
So the threat model should read: whoever can read the session directory on the shared filesystem still has the plaintext, and whoever only receives the exported package does not. This PR closes the only exported copy, and I have put that in the PR description rather than leaving it implied.
One caveat, since the done-marker JSON *is* packaged: with `include_transcripts=true` it inlines the full transcript. The default is `false`, and redacting that path touches the breakdown JSON, so it is tracked as its own change.
### 3) `except` too narrow — fixed
Confirmed: `urlparse("http://[::1")` raises `ValueError`, and `require_http_url` does not wrap it. Of the two callers, `enumerate_pr_kb` is unguarded and would have raised instead of skipping the source, while the phase-audit path is already inside a broad `except`. So there was exactly one exposed caller, and yes, it is a failure surface this PR introduced by moving the check to construction time. `build_gbrain_page_client_from_env` now catches `(GbrainPageError, ValueError)`, with a test for `http://[::1`.
I kept it local rather than converting `ValueError` inside `require_http_url`: that helper has four other callers — the PR patch-source resolver, the InferenceX client, the Primus Cortex source and the critic KB client — and changing what it raises for all of them is wider than this PR should carry. Say the word if you would rather have it at the helper.
### 4) Regex cost on large inputs — agreed, deliberately not in this PR
Measured on a 1 MB string: `redact_secret_values` takes 41 ms against 1.8 ms for `json.dumps`, so the regex is roughly 95% of the cost, and the path is real for the reason given in item 1.
Pre-clipping to a loose bound first is safe: 4096 is far above the 240 limit, so a secret straddling the clip boundary is still whole inside the window, and even a secret longer than the bound still matches, every pattern being `{8,}` or `+`. I left it out to keep this PR purely security-semantic; it is first in the follow-up.
### 5) Neighbour sites — the provenance answer is "not trusted"
You made this argument yourself in a2046b8af ("restrict candidate diff fetches to http and https"): *"That URL is not operator input: it is built from the html_url field of a PR-KB or GitHub search response, so honouring the scheme let remote data read the local filesystem."* That commit fixed `_fetch_diff_to_path` and consolidated the five verbatim copies into the shared `require_http_url`.
`_fetch_diff_url` on the phase-audit path is the copy that did not get updated, and on current main it still accepts `file://`. The chain: a candidate's `diff_url` is assembled as `f"{html_url}.diff"`, and `html_url` comes straight off the remote payload in the Primus Cortex source (`item.get("html_url") or item.get("url")`) and in the GitHub search source; only the PR-KB source builds it locally. For contrast, the patch-source resolver does call `require_http_url` on `diff_url`.
So it is not deliberate asymmetry, it is a missed sixth site. Different module and different feature from this PR, so I would rather fix it in its own change — happy to open it.
### Verification
735 targeted tests pass across every suite that touches `redact_secret_values` — env safety, tool-input summaries, trace utils, shared state, breakdown recorder, failure evidence, specialist runner helpers, the Langfuse emitter, PR KB, recipe sanitization and explore. Ruff clean.

siliangchen-amd:
Items 1, 2 and the scan bound are in bbccf5e25; item 3 is a partial take with two corrections; item 4 is now in the PR description.
### 1) TOKEN substring false positives — fixed
Confirmed, and the quoted `--tokenizer="/models/x"` case is a regression this PR introduced. Bare `max_tokens=4096` already false-positive on main, as you said.
`TOKEN(?!IZER|S\b)` on the name side. Rechecked against the shapes you cited: `--tokenizer=` (quoted and bare), `--max-tokens="4096"`, `max_tokens=4096`, `num_speculative_tokens=5`, `TOKENIZERS_PARALLELISM=false` are all left intact; `HF_TOKEN=` and `ANTHROPIC_AUTH_TOKEN="…"` still mask. Same exemption the env-name classifier already had for `TOKENIZER`; the assignment pattern just had not caught up.
I did not switch the name side to an exact env-name allowlist. That list is ten known process env vars, so `export MYAPP_PASSWORD="hunter2"` — the hole the last push closed — would leak again. The specialist intel ledger is exactly shell commands, not just the process environment.
I do not have the session log you measured 199 hits on in this workspace. Against every shape you named, the assignment pattern now hits 0.
### 2) Unbalanced quotes on the escaped form — fixed, slightly different value class
Confirmed: `PASSWORD=\"hunter2\"` left the trailing backslash in the value, so the summary was `\"[REDACTED]""`.
Dropping `\` from the value class wholesale splits `PASSWORD=C:\foo\bar` into `PASSWORD=[REDACTED]\foo\bar` and leaks the rest. A backslash is now excluded only when it precedes a quote, so the JSON-escaped closer stays in place (`\"[REDACTED]\"`) and a value that contains backslashes is still masked whole. The quoted-assignment test now asserts the balanced form, not just `"hunter2" not in`.
### 3) The other redactor — AKIA / JWT copied across; allowlist and Authorization: token not
The "keep in sync" comment is real, and this PR had only updated one side. Two of the extra shapes are now in the shared helper: AWS access-key ids and compact JWTs. `aws --secret-key AKIA…` is masked by shape even with no assignment.
I am not replacing the name side with that allowlist, and I would not call that other redactor strictly better:
- Its key list is those same ten env names. `MYAPP_PASSWORD="hunter2"` does not match it at all. Adopting it would reopen the hole this PR exists to close.
- `_AUTHORIZATION_RE` does not actually cover `Authorization: token abcdef…`. Measured: `'Authorization: token abcdef1234567890'` becomes `'Authorization: [REDACTED] abcdef1234567890'` — it eats the scheme word `token` and leaves the credential. That is a separate bug in that redactor, and I would rather fix it on its own than copy it.
Merging the two implementations is a coverage-surface change across every diagnostics call site. Happy to do that as a follow-up with its own review.
### 4) Remaining shapes — documented as known boundaries
Agreed, not all claimed, now in the PR description. Same four you listed, plus the JSON-key form from last round (`{"password": "x"}`), which still needs a numeric exemption.
### 5) Last round's two leftovers
Raw logs: definite answer, in the PR description. The export is an allowlist. The trace directory's JSONL files (including the ledger this PR redacts) are selected. Under the specialist run tree only the done-marker JSON is selected; the raw stream-json / Codex JSONL logs match no pattern. Threat model: session-directory readers still have plaintext; package-only recipients do not. Caveat: `include_transcripts=true` inlines the transcript into the done-marker JSON (default `false`).
Scan cost: done in this push. The summarizer now scans at most 4096 characters before the 240-char clip. 4096 is far above 240, so a secret straddling the clip is still whole in the window; a secret longer than 4096 still matches (`{8,}` / `+`). A megabyte Write dump is no longer run through every pattern first.
### Verification
739 targeted tests pass across every suite that touches `redact_secret_values`. Ruff clean. AWS / JWT fixtures are concatenated at runtime so secret-scan does not flag the tests.

siliangchen-amd:
`TOKEN(?!IZER|S\b)` does not cover `tokenized_requests` (next letters are IZED) or `stop_token_ids` (next is `_`). A half-redacted list (`stop_token_ids=[REDACTED], 154827, 154829]`) is worse than leaving the knob: it looks like a secret was there. Both shapes already false-positive on main; this round was already in that class, so they are fixed in 1c1e57937.
I did not add another exemption, and I did not switch the name side to the specialist transcript's ten-name allowlist. That list is process env vars. The intel ledger records shell commands. Adopting it reopens the hole this PR closed (`MYAPP_PASSWORD="hunter2"`), and also drops `ANTHROPIC_AUTH_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN`, `AMD_API_KEY`, `api_key=`, `AWS_SECRET_ACCESS_KEY`, `TOKEN=`, `MYAPP_TOKEN=`. Same line you cited is the demonstration: `api_key=` must keep matching, and it is not one of those ten names.
TOKEN is now the last segment of the identifier (optional `_` + digits, so `HF_TOKEN_2` still matches). `API_KEY` / `SECRET` / `PASSWORD` / `CREDENTIAL` still allow trailing name characters. That ends the whack-a-mole for this family: `tokenizer`, `max_tokens`, `tokenized_requests`, `stop_token_ids`, `token_budget`, `tokens_per_second` are all left intact; `HF_TOKEN`, `ANTHROPIC_AUTH_TOKEN`, `MYAPP_PASSWORD`, `api_key=` still mask.
Re-scanned 56 session benchmark logs on this machine (your 301 are not here; smaller corpus, same lm_eval `model_args` shape). `tokenized_requests=[REDACTED]`: 36/36 files that carry the blob → 0. `api_key=`: still 36. `stop_token_ids` does not appear in these 56; covered by a fixture that used to produce the half-redacted list and now does not.
The two redactors are still not in sync, and should not be: one is an env-name allowlist for transcripts, the other is fragment matching for diagnostic text. Merging them is a coverage-surface change of its own.
739 targeted tests. Ruff clean.

xiaofei-zheng (approved):

PR #1305, opened 2026-08-27
Fix vLLM native TraceLens profiler sentinel name

BaoYunkai:
## Problem

Hyperloom's vLLM >=0.26 native profiler gate looked for capture_torch_profiler_dir in profiler.py, but upstream vLLM exposes capture_torch_profiler (bool). Native detection always failed, TraceLens flags were not injected, and capture_traces/ was missing.

## Fix

Restore capture_torch_profiler as the sentinel and re-inject --profiler-config.capture_torch_profiler True when the patch gate passes. Add regression tests for v0.28 profiler.py layout.


BaoYunkai:
No need after PR #1318.

PR #1313, opened 2026-08-27
Add additive SBD V6 metadata and timeline projection

luochen-amd:
## Summary
- add an additive SBD V6 projection while preserving the existing V5 schema version and execution behavior
- expose V6 `metadata`, `outcome`, `timeline`, and top-level `close` data
- record fail-open `install` and `model_gate` telemetry without changing preflight, gate, or resume outcomes
- implement the merged `framework_agent` timeline for configuration, source, Critic, macro-cycle, exit, and failure evidence
- project Framework-internal candidate discovery success, empty, and failure outcomes in execution order
- omit phase summaries and all token, pricing, and monetary cost fields

## Review Fixes
- persist hard preflight failures even when they occur outside an instrumented install step
- acquire the session lock before writing resume-failure artifacts and fall back to an isolated failed-attempt session when the target session is busy
- preserve specialist source phases and exclude non-Framework specialist evidence from `framework_agent` timeline creation
- make legacy fixed-file timeline migration resumable after a partial migration

## Compatibility
- keeps V5 output and state-machine behavior intact
- treats V6 collection as best-effort telemetry so collection failures do not affect optimization execution
- retains legacy Explore evidence only as input compatibility for the merged Framework Agent stage

## Validation
- focused blocker reproductions: 4 passed
- SBD, Framework Agent, specialist, enablement, and attribution regression selection: 266 passed
- preflight regression selection: 237 passed
- Ruff checks passed for all changed Python files
- Ruff format checks passed for all changed Python files
- `git diff --check` passed

## Environment Note
- Windows test runs inject a no-op `fcntl` compatibility module so POSIX-only KB imports can load; the production lock behavior remains covered by Linux CI and the focused busy-session test uses an explicit lock double

### DIFF
--- a/src/hyperloom/inference_optimizer/breakdown/collectors/__init__.py
+++ b/src/hyperloom/inference_optimizer/breakdown/collectors/__init__.py
@@ -186,6 +186,11 @@ from .geak import (
     _geak_reconstruct_from_disk as _geak_reconstruct_from_disk,
     collect_geak as collect_geak,
 )
+from .v6 import (
+    collect_v6_metadata as collect_v6_metadata,
+    collect_v6_outcome as collect_v6_outcome,
+    collect_v6_timeline as collect_v6_timeline,
+)
 
 __all__ = [
     "collect_attribution",
@@ -209,6 +214,9 @@ __all__ = [
     "collect_sweep",
     "collect_telemetry",
     "collect_token_usage",
+    "collect_v6_metadata",
+    "collect_v6_outcome",
+    "collect_v6_timeline",
     "collect_workload",
     "collect_model_info",
     "collect_recorded_optimizations",
--- a/src/hyperloom/inference_optimizer/breakdown/collectors/v6.py
+++ b/src/hyperloom/inference_optimizer/breakdown/collectors/v6.py
@@ -376,12 +376,6 @@ def _framework_windows(
             active["exit_row"] = row
             active = None
             continue
-        if from_framework and from_phase == to_phase:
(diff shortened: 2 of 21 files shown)


luochen-amd:
Addressed the three blocking review findings:

- Persist hard preflight/install failures by creating a safe failure session, writing the install event, and materializing `session_breakdown.json` before re-raising the original exception.
- Preserve repeated `install` and `model_gate` executions in an ordered per-event timeline while keeping `install.json` and `model_gate.json` as latest-event compatibility projections. Legacy fixed-file sessions are migrated on the next write.
- Exclude Kernel-phase `explore` rebench operations from the Framework Agent timeline and require matching cycle/time window evidence.

Regression coverage now includes hard preflight failure persistence, fresh-to-resume event retention, legacy event migration, and Kernel Explore exclusion.

Local validation:
- SBD V6 focused tests: 26 passed
- Related exporter/model-gate suites: 215 passed, 1 Windows-only path assertion deselected
- Ruff check/format and `git diff --check`: passed

luochen-amd:
This combined PR has been split to keep the review boundaries explicit:

- SBD V6 metadata, outcome, install, and model-gate timeline: #1331 (based on `main`)
- Merged Framework Agent timeline and specialist phase attribution: `feat/sbd-v6-framework-agent`

The Framework branch is intentionally kept separate while #1301 and #1331 are still open. A focused Framework PR can be opened after the dependency bases are aligned, avoiding duplicate core changes in review.

PR #1315, opened 2026-08-27
fix(server-args): keep +custom_ops JSON valid across variant composition

hmahmad26:
## Summary
- Variant composition ran every server-arg string through POSIX `shlex.split`, which stripped inner JSON quotes from `--compilation-config`.
- The bareword repair then failed on vLLM custom-op toggles such as `+fused_rms_norm_gated` (`+` was outside the charset), so variants died at argv parse with `Invalid JSON: key must be a string` while baseline (which does not go through `compose_server_args`) stayed valid.
- Tokenize quote-preservingly in `remove_server_args`, accept an optional leading `+`/`-` on JSON barewords, and log if composition still damages a JSON flag.

## Test plan
- [x] Round-trip checks: `+fused_rms_norm_gated` survives `remove_server_args` / `compose_server_args` in append, replace, and flag-removal modes
- [x] Plain `'kimi_k3'` wrappers are stripped so Magpie does not see them on argv
- [ ] CI: `test_sign_prefixed_custom_op_survives_removal`, `test_compose_preserves_json_for_every_args_mode`, `test_shell_quoted_plain_operand_loses_its_wrappers`

### DIFF
--- a/src/hyperloom/inference_optimizer/tests/test_grid_runner.py
+++ b/src/hyperloom/inference_optimizer/tests/test_grid_runner.py
@@ -2371,6 +2371,65 @@ class TestRemoveServerArgsPreservesJson:
         out = _grid_runner.remove_server_args(raw, ["--port"])
         assert json.loads(out.split("--compilation-config ", 1)[1].strip()) == {"cudagraph_mode": "FULL"}
 
+    def test_sign_prefixed_custom_op_survives_removal(self):
+        """A ``+``/``-`` prefixed custom-op value must survive the round trip.
+
+        Live regression: every variant of an explore round died at vLLM argv
+        parse with ``Invalid JSON: key must be a string`` -- including a control
+        leg that added one env var and zero args, because
+        ``strip_benchmark_harness_flags`` routes EVERY composed variant through
+        ``remove_server_args`` with a non-empty denylist. The old POSIX
+        ``shlex.split``/rejoin stripped the JSON quotes, and the bareword repair
+        heuristic could not re-quote ``+fused_rms_norm_gated`` (its ``+`` was
+        outside the charset), so the corruption reached the server verbatim.
+        """
+        raw = (
+            "--compilation-config "
+            '{"mode":3,"custom_ops":["+fused_rms_norm_gated"],'
+            '"cudagraph_capture_sizes":[1,2,3]} --port 8888'
+        )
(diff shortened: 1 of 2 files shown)


chaojhou:
## Blocking findings (reviewed at `94a4e9a`)

The core intent is right — I round-tripped 10 JSON-valued-flag cases against base and 5 that base corrupts now survive. But re-tokenizing with `shlex.split(..., posix=False)` after compaction is not safe, because **JSON compaction only strips structural whitespace — it never touches whitespace inside string values.** Any JSON blob whose string value contains a space still shatters into multiple tokens.

### 1. `remove_server_args` leaks argv fragments when a JSON value contains a space

```python
remove_server_args('--hf-overrides {"chat_template":"a b"} --port 8888', ["--hf-overrides"])
```

| | result |
|---|---|
| base | `--port 8888` ✅ |
| PR   | `b"} --port 8888` ❌ |

`_split_args_preserving_json` (line 122, `posix=False` at line 137) yields `['--hf-overrides', '{"chat_template":"a', 'b"}', '--port', '8888']`. The removal loop drops the flag and *only the first* value token; `b"}` survives into the output.

End-to-end this reaches `compose_server_args` and produces argv that is no longer parseable at all:

```
b"} --max-num-seqs 20 --hf-overrides {"chat_template":"c d"}
```

→ `shlex` raises `ValueError: No closing quotation` downstream. This is a hard launch failure, and it's a **regression** — base handles this input correctly.

### 2. Multi-space inside a JSON string value is silently collapsed

```python
{"tmpl":"a   b"}   ->   {"tmpl":"a b"}
```

Same root cause, but here the result still parses, so nothing raises and the tripwire never fires. Base is *loudly* broken on this input; the PR is *silently* wrong. For a field like `chat_template` that's a data-corruption path with no signal — strictly worse than a crash.

### Suggested direction

Don't re-tokenize the compacted string. Substitute each JSON blob with an opaque placeholder before `shlex.split`, then restore after — the blob stays one indivisible token regardless of its interior whitespace, and neither failure above is reachable.

### Also blocking-adjacent

- The tripwire's `before` snapshot is taken too late to catch the case above (it never observes the pre-mutation argv for this path). Either move it to `compose_server_args` entry, or drop it — as written those 34 lines have zero coverage and zero effect here.
- The branch is 399 commits behind `main`; please rebase before merge.

Happy to re-review once the tokenization is placeholder-based — the rest of the change looks good.

PR #1317, opened 2026-08-28
Force vLLM V1 on profile so CUDA graph traces are captured

BaoYunkai:
## Problem

vLLM V2 does not enable torch profiler during CUDA graph capture. Profile/roofline therefore never writes `capture_traces/`, and kernel-level analysis is empty.

## Fix

On vLLM profile/roofline materialization, set `VLLM_USE_V2_MODEL_RUNNER=0` so graph capture uses V1, which records those traces. Baseline is unchanged.


BaoYunkai:
No need after PR#1318, closed.

PR #1321, opened 2026-08-28
fix(geak): forward the run's actual GPU pin in the handoff

### LINKED ISSUE #1312: GEAK uses first GPU device regardless of top-level agent selection for xDiT run
### Target release

Unsure

Local

### Optimization domain

Inference

### Issue type

Crash / failure

### Which phase failed?

Optimization loop

### Short Summary

GEAK uses GPU device 0 regardless of what was requested of the agent

### Job ID (hosted UI runs only)

_No response_

```shell
Run a 8-hour Hyperloom xDiT optimization session for Tongyi-MAI/Z-Image-Turbo on the last available AMD GPU device. Use when the user wants a short xDiT diffusion-model demo (image generation, not LLM serving) on the local AMD ROCm environment.
```

### Expected behavior

GEAK will leverage the same GPU asked of the agent when the optimization run is kicked off.

### Actual behavior

GEAK used GPU device 0.  In this particular case that GPU was already occupied so that lead to test failures.

### Steps to reproduce

_No response_

### Logs

```shell
Agent summary:

    **`geak/handoff.json`'s `gpu_ids` field is hardcoded to `0` regardless of
    the run's actual `ROCR_VISIBLE_DEVICES` pin, and GEAK's `bench_e2e.sh`
    (`GPU=${GPU:-0}`) uses it to set `ROCR_VISIBLE_DEVICES=0` /
    `HIP_VISIBLE_DEVICES=0` for every full server it launches** (baseline,
    profile, and later validation/capture-selection legs) — confirmed via
    `/proc/<pid>/environ` on the actually-launched `xdit` server subprocess,
    even though the top-level `geak_runner.py`/`run_e2e.py` processes
    correctly inherit the real pin. This silently collides with physical GPU
    0 whenever something else on the host is using it. On a shared host, an
    unrelated long-running `vllm-server` container occupying ~93-99% of GPU
    0's VRAM caused this to reproduce 3 times in a single 2026-08-26 run
    (before that container existed, the 2026-08-19/08-21 runs never hit
    this, because GPU 0 was idle). The collision manifests as
    `torch.OutOfMemoryError: HIP out of memory... GPU 0 has a total capacity
    of 191.98 GiB of which 0 bytes is free` in
    `geak/e2e_cycle0/baseline/replica_*/attempt_*/server.log` and
    `.../profile/round_*/server.log`, and also shows up as GEAK's own
    config-tuning specialist explicitly declining to A/B a candidate ("the
    serving GPU (GPU 0) is held by a 186GB foreign co-tenant... must not be
    touched") in its own workflow reasoning log. **The isolated per-op
    opbench microbenchmark harness used for kernel-authoring bake-offs is
    unaffected** — it runs in-process on the already-correctly-pinned GPU;
    only GEAK's own full-server-launch paths (baseline/profile/config-tuning
    validation) are degraded. Before trusting a `no_gain`/`incomplete` GEAK
    baseline on a shared host, check `rocm-smi --showpids` for a foreign
    tenant on physical GPU 0 and cross-check
    `python3 -c "import json;print(json.load(open('geak/handoff.json'))['gpu_ids'])"`
    against the run's actual `ROCR_VISIBLE_DEVICES` — a mismatch confirms
    this bug rather than a real accuracy/perf regression. There is currently
    no operator-facing override; it would need a Hyperloom-side fix to
    forward the orchestrator's actual GPU pin into the handoff instead of
    the literal `0` default.
```

_No response_

### Data handling

- [x] I confirmed the attached logs/artifacts contain no sensitive information

zihaoanllm:
Fixes #1312.

## Problem

`geak/handoff.json` never told GEAK which cards the run owns. `phases/kernel.py` resolved
`gpu_ids` from `HIP_VISIBLE_DEVICES` / `CUDA_VISIBLE_DEVICES` only — `ROCR_VISIBLE_DEVICES`,
the canonical ROCm pin honoured everywhere else in this repo (`bus/gpu_pool.py`,
`policy/gate.py`, `cli/preflight.py`), was not consulted. So a ROCR-pinned run fell back to
`0..tp-1` and every server GEAK launches (baseline, profile, config-tuning validation) went to
physical GPU 0, OOM-ing against whatever else holds that card and reporting a plausible-looking
`no_gain`.

## Fix

- Resolve the mask with the repo's `ROCR` → `HIP` → `CUDA` precedence, recipe `benchmark.envs`
  before process env.
- Add `gpu_pin` to the handoff (`schema_version` 3): the **absolute** ids plus the var they came
  from, for consumers that write `ROCR_VISIBLE_DEVICES` themselves. Omitted when nothing is
  pinned — that means "whole machine", not "card 0".
- `gpu_ids` keeps its existing HIP-level meaning (HIP indexes into the ROCr-visible set), since
  GEAK's sglang/vllm adapters export it as `HIP_VISIBLE_DEVICES`. Only new behaviour: it is now
  clamped to the mask when `tp` overshoots it.
- The breakdown collector records both fields, so this failure is visible in the artifacts.

GEAK-side follow-up (not this repo): launch paths that write `ROCR_VISIBLE_DEVICES` should read
`gpu_pin["value"]`, as `adapters/launchers/magpie.sh` already does.

## Tests

New `test_geak_handoff_gpu_pin.py` plus an end-to-end case in `test_geak_resume_recovery.py`
(`ROCR=7` → `gpu_pin.ids=[7]`, `gpu_ids="0"`). `pytest src/hyperloom/orchestrator` 91 passed /
14 skipped; ruff and `pylint --errors-only` clean. Not verified on hardware — I have no repro of
the xDiT run, so the "server actually lands on the pinned card" leg is untested.


src/hyperloom/orchestrator/loop/coordinator_helpers.py
```diff
@@ -1546,6 +1546,117 @@ def _geak_sweep_measured_tput(res: dict[str, Any]) -> float | None:
...
+        environ: Environment mapping to read; defaults to ``os.environ``.
+
+    Returns:
+        ``{"var", "value", "ids", "source"}`` for the winning mask, where
+        ``ids`` are ABSOLUTE device ids and ``source`` is
+        ``"baseline_recipe"`` or ``"process_env"``. ``{}`` when no mask is set
+        anywhere — meaning "whole machine visible", not "pinned to 0".
+    """
+    env = os.environ if environ is None else environ
+    for source, table in (("baseline_recipe", recipe_envs or {}), ("process_env", env)):
```
xiaofei-zheng: **Recipe-first precedence defeats the fix for HIP/CUDA-pinned runs.**

`materialize_config_with_envs` unconditionally synthesizes `ROCR_VISIBLE_DEVICES=0..tp-1` into `benchmark.envs` (`_workload_envs.py:943-956`), and that materialized recipe is what `state.baseline_config_path` points at by the time KERNEL runs. So the `baseline_recipe` source always wins, and it always carries a synthetic `0..tp-1` mask.

Repro: run with `HIP_VISIBLE_DEVICES=4,5`, `TP=2`, no ROCR anywhere. `_resolve_gpu_pin` returns `{'var': 'ROCR_VISIBLE_DEVICES', 'value': '0,1', 'ids': [0, 1], 'source': 'baseline_recipe'}` (verified by executing the helper). `handoff["gpu_ids"]` becomes `"0,1"` where the pre-PR code emitted `"4,5"`, and `gpu_pin["value"] = "0,1"` tells a ROCR-writing consumer to hard-pin physical cards 0 and 1 — recreating exactly the foreign-tenant card-0 collision #1312 is meant to fix, as a new regression for HIP users.

The recipe mask is only meaningful as a pin when it was *authored*, not when it was autofilled. Either skip the autofilled ROCR key, or consult the process env first and use the recipe only as a fallback.

src/hyperloom/orchestrator/loop/coordinator_helpers.py
```diff
@@ -1546,6 +1546,117 @@ def _geak_sweep_measured_tput(res: dict[str, Any]) -> float | None:
...
+        for var in _VISIBLE_DEVICE_VARS:
+            raw = table.get(var)
+            if raw is None or str(raw).strip() == "":
+                continue
+            value = str(raw).strip()
+            return {"var": var, "value": value, "ids": _parse_device_list(value), "source": source}
```
xiaofei-zheng: **The documented `{}` ("whole machine") case is unreachable in production.**

Same root cause as above: because the materialized recipe always carries an autofilled `ROCR_VISIBLE_DEVICES=0..tp-1`, a genuinely unpinned run still produces a truthy pin. With no mask anywhere in the process env and `TP=4`, this returns `{'ids': [0,1,2,3], 'source': 'baseline_recipe', ...}` (verified by executing the helper), so `handoff["gpu_pin"]` is emitted.

That contradicts the docstring above and the table in `docs/components/geak.md` ("omitted entirely when no mask is set anywhere, which means whole machine visible, not pinned to card 0"). A GEAK launcher that writes `ROCR_VISIBLE_DEVICES = gpu_pin["value"]` now restricts a whole-machine run to cards 0-3, where before it inherited everything.

src/hyperloom/orchestrator/loop/coordinator_helpers.py
```diff
@@ -1546,6 +1546,117 @@ def _geak_sweep_measured_tput(res: dict[str, Any]) -> float | None:
...
+
+    Returns:
+        ``{"var", "value", "ids", "source"}`` for the winning mask, where
+        ``ids`` are ABSOLUTE device ids and ``source`` is
+        ``"baseline_recipe"`` or ``"process_env"``. ``{}`` when no mask is set
+        anywhere — meaning "whole machine visible", not "pinned to 0".
+    """
+    env = os.environ if environ is None else environ
+    for source, table in (("baseline_recipe", recipe_envs or {}), ("process_env", env)):
+        for var in _VISIBLE_DEVICE_VARS:
```
xiaofei-zheng: **Loop nesting makes source precedence dominate variable precedence.**

The source loop is outer and the variable loop inner, so *all three* vars are checked under `baseline_recipe` before `process_env` is reached. That means a recipe-level `CUDA_VISIBLE_DEVICES`/`HIP_VISIBLE_DEVICES` silently overrides a real process-level `ROCR_VISIBLE_DEVICES` pin.

Example: a hand-authored recipe carries `benchmark.envs.CUDA_VISIBLE_DEVICES: "0"` (a common leftover on a CUDA-derived YAML) while the run is launched with `ROCR_VISIBLE_DEVICES=6,7`. This returns `{'var': 'CUDA_VISIBLE_DEVICES', 'ids': [0]}`, and `_resolve_handoff_gpu_ids` takes the non-ROCR branch and emits `gpu_ids="0"` verbatim — pinning GEAK to physical card 0 while the run owns 6 and 7.

The docstring's claimed "`ROCR_VISIBLE_DEVICES` before `HIP`/`CUDA`" precedence only holds *within* a single source. If that ordering is meant to be global, the loops need to be swapped (or the precedence documented as source-major).

src/hyperloom/orchestrator/loop/coordinator_helpers.py
```diff
@@ -1546,6 +1546,117 @@ def _geak_sweep_measured_tput(res: dict[str, Any]) -> float | None:
...
+    Args:
+        gpu_pin: The :func:`_resolve_gpu_pin` result (``{}``/``None`` = unpinned).
+        tp: Tensor-parallel size; ``<= 1`` is treated as 1.
+
+    Returns:
+        A comma-separated device list, never empty.
+    """
+    width = max(int(tp or 1), 1)
+    ids = list((gpu_pin or {}).get("ids") or [])
+    if not ids:
```
xiaofei-zheng: **A non-numeric but non-blank mask yields a truthy pin with empty `ids`, and falls back to card 0.**

`_resolve_gpu_pin` returns early on any non-blank string, but `_parse_device_list` returns `[]` for anything non-numeric. Two real cases:

* ROCm accepts GPU UUID masks: `ROCR_VISIBLE_DEVICES=GPU-a1b2c3,GPU-d4e5f6`.
* YAML parses `ROCR_VISIBLE_DEVICES: [0,1]` as a list, and `str()` of it is `"[0, 1]"`.

In both, `handoff["gpu_pin"]` is present with `ids: []`, this branch sees `not ids` and returns `"0,...,tp-1"` — landing every GEAK server on card 0, the exact default the PR exists to eliminate. Meanwhile a consumer reading `gpu_pin["ids"]` gets nothing, and one writing `gpu_pin["value"]` exports the literal `"[0, 1]"`.

`test_gpu_ids_never_empty_for_a_blank_mask` enshrines the fallback as intended, but "never empty" and "never silently card 0" are different goals. Suggest treating a non-blank-but-unparseable mask as an error/warning rather than as unpinned.

src/hyperloom/orchestrator/loop/coordinator_helpers.py
```diff
@@ -1546,6 +1546,117 @@ def _geak_sweep_measured_tput(res: dict[str, Any]) -> float | None:
...
 
 
+#: Visible-device env masks, in the repo's ROCm precedence order.
+#: ``ROCR_VISIBLE_DEVICES`` is canonical on ROCm (the CLI preflight drops
+#: ``HIP_VISIBLE_DEVICES`` when ROCR is set); HIP/CUDA cover CUDA-style and
+#: legacy pins. Same order as ``gpu_pool._visible_device_mask`` /
+#: ``policy.gate.detect_gpu_count`` so every layer agrees on "the pin".
+#: (Kept local rather than imported from ``gpu_pool``: this module is the pure
+#: helper layer and ``gpu_pool`` drags in the SQLite connection.)
+_VISIBLE_DEVICE_VARS: tuple[str, ...] = (
```
xiaofei-zheng: **The "every layer agrees on the pin" claim does not hold for a present-but-empty mask.**

`gpu_pool._visible_device_mask` returns `([], present=True)` for `VAR=""` and `gate.detect_gpu_count()` derives 0 GPUs from it. This resolver instead skips a blank value and falls through to the next variable.

Divergence: `ROCR_VISIBLE_DEVICES=""` together with a stale `HIP_VISIBLE_DEVICES="2,3"`. The orchestrator believes it has zero GPUs, while `_resolve_gpu_pin` returns `{'var': 'HIP_VISIBLE_DEVICES', 'ids': [2, 3]}` and hands GEAK two cards. `test_pin_skips_blank_values` locks the divergent behaviour in.

Either match the other layers (blank = "zero visible", stop) or reword this comment so it doesn't claim an agreement that isn't there.

src/hyperloom/orchestrator/loop/coordinator_helpers.py
```diff
@@ -1546,6 +1546,117 @@ def _geak_sweep_measured_tput(res: dict[str, Any]) -> float | None:
...
+    "ROCR_VISIBLE_DEVICES",
+    "HIP_VISIBLE_DEVICES",
+    "CUDA_VISIBLE_DEVICES",
+)
+
+
+def _parse_device_list(raw: Any) -> list[int]:
```
xiaofei-zheng: **Fourth copy of the mask parsing/precedence rules.**

`_parse_device_list` is a line-for-line duplicate of `gpu_pool._parse_gpu_list`, and `_VISIBLE_DEVICE_VARS` is now the fourth copy of the precedence tuple (`gpu_pool.py:113`, `gate.py:182`, `preflight.py:1276`).

The stated justification — `gpu_pool` drags in the SQLite connection — is a real constraint, but it argues for lifting `_parse_gpu_list` and the var tuple into a dependency-free shared module (e.g. `hyperloom/common/`) that all four import, not for a fourth copy. As it stands the same rules live in four places with three subtly different empty-mask semantics (see the comment on `_VISIBLE_DEVICE_VARS`), and any future change has to be applied four times or the layers drift further apart.

src/hyperloom/orchestrator/phases/kernel.py
```diff
@@ -746,7 +761,14 @@ async def _run_geak_kernel_phase(self, *, from_phase: str) -> None:
...
         # (process-env fallback). Only resolved keys are sent.
-        bench_protocol = self._resolve_bench_protocol(str(getattr(state, "baseline_config_path", "") or ""))
+        _recipe_path = str(getattr(state, "baseline_config_path", "") or "")
+        bench_protocol = self._resolve_bench_protocol(_recipe_path)
+        # The run's ACTUAL GPU pin (issue #1312). GEAK launches full servers
+        # out-of-process and writes its own visible-devices mask for each one;
+        # with no pin in the handoff it defaults to physical GPU 0 and collides
+        # with whatever else holds that card. Recipe mask first (what Hyperloom
+        # benched with), then the process mask; {} means "whole machine".
+        gpu_pin = _resolve_gpu_pin(recipe_envs=self._read_recipe_bench_envs(_recipe_path))
```
xiaofei-zheng: **The recipe YAML is now read and parsed twice per handoff.**

`_resolve_bench_protocol(_recipe_path)` on the previous line already calls `_read_recipe_bench_envs`, and this line calls it again on the same path — two `Path.is_file()` + `read_text()` + `yaml.safe_load()` round trips.

Hoisting it to a single local and passing it into both consumers removes the second parse and guarantees both fields see the same snapshot:

```python
_envs = self._read_recipe_bench_envs(_recipe_path)
bench_protocol = self._resolve_bench_protocol(_recipe_path)  # or take _envs directly
gpu_pin = _resolve_gpu_pin(recipe_envs=_envs)
```

xiaofei-zheng (commented):
Code review of the GPU-pin handoff change. The direction is right — `ROCR_VISIBLE_DEVICES` genuinely was the missing link, and forwarding an absolute pin is the correct shape. But as written I believe the recipe-first precedence defeats the fix for HIP/CUDA-pinned runs and makes the documented "unpinned" contract unreachable in production, because `materialize_config_with_envs` autofills `ROCR_VISIBLE_DEVICES=0..tp-1` into every materialized recipe. Details inline; the first two are the ones I would block on.

One cross-cutting note not anchorable to the diff: `handoff["tp"]` is still read raw from `$TP` while `gpu_ids` is now clamped to the mask width, so the two fields can disagree (see the comment on `kernel.py`).

src/hyperloom/orchestrator/phases/kernel.py
```diff
@@ -795,13 +819,16 @@ async def _run_geak_kernel_phase(self, *, from_phase: str) -> None:
...
-            # Pin the serving GPU set: explicit visibility mask, else 0..tp-1.
-            "gpu_ids": (
-                os.environ.get("HIP_VISIBLE_DEVICES")
-                or os.environ.get("CUDA_VISIBLE_DEVICES")
-                or ",".join(str(i) for i in range(int(os.environ.get("TP", "1") or 1)))
-            ),
+            # The serving/optimization device set, as HIP-level ids (what the
+            # consumer exports as HIP_VISIBLE_DEVICES). Logical positions inside
+            # an inherited ROCR mask, a HIP/CUDA mask verbatim, else 0..tp-1.
+            "gpu_ids": _resolve_handoff_gpu_ids(gpu_pin=gpu_pin, tp=int(os.environ.get("TP", "1") or 1)),
```
xiaofei-zheng: **`gpu_ids` logical indices are computed against the recipe mask, but the child inherits the process mask.**

The GEAK subprocess is launched with `runner_env = dict(os.environ)` (line ~1171); nothing overrides `runner_env["ROCR_VISIBLE_DEVICES"]` from the resolved pin. So whenever the recipe mask and the process mask differ, the logical indices resolve against the wrong set.

This is exactly the case the new `test_pin_prefers_recipe_over_process_env` encodes: recipe `ROCR_VISIBLE_DEVICES="6"`, process `ROCR_VISIBLE_DEVICES="0"`. The handoff advertises `gpu_pin.ids=[6]` and `gpu_ids="0"`, but the child inherits `ROCR=0`, so HIP index 0 is physical card 0 — the handoff claims card 6 while every server GEAK launches sits on card 0. If the recipe mask is to be authoritative, the phase should also export it into `runner_env`.

**Separately, `tp` and `gpu_ids` can now disagree.** `handoff["tp"]` a few lines up is still `int(os.environ.get("TP"))` raw, while `gpu_ids` is clamped to the mask width. With `TP=8` exported on a 4-GPU pod the materializer clamps the recipe to `TP=4` / `ROCR="0,1,2,3"` (`test_baseline_param_overrides.py:385`), so `gpu_ids` becomes `"0,1,2,3"` while `tp` stays `8` — GEAK launches sglang with `--tp 8` and four visible cards and fails to load weights. Pre-PR both fields derived from the same `$TP` and could not disagree; `tp` should now come from the same resolved mask/recipe as `gpu_ids`.

src/hyperloom/inference_optimizer/breakdown/collectors/geak.py
```diff
@@ -671,6 +671,11 @@ def _load_json(p: Path) -> dict[str, Any]:
             "workload": handoff.get("workload"),
             "accepted_flags": handoff.get("accepted_flags"),
             "raw_baseline_tput": _to_float(handoff.get("raw_baseline_tput")),
+            # Device set + the run's absolute pin: a GEAK baseline that reads
+            # `no_gain`/`incomplete` because its servers landed on a foreign
+            # tenant's card is otherwise indistinguishable from a real result.
+            "gpu_ids": handoff.get("gpu_ids"),
```
xiaofei-zheng: **These fields are only recorded on the crash-recovery path, so the stated goal is not met.**

`_geak_reconstruct_from_disk` has exactly one call site (line ~906), guarded by `if not has_result`. A GEAK run that finishes normally and writes `geak_result={'status': 'no_gain'}` takes the `has_result` path at line ~900 and returns without ever calling it.

So for every completed run — including the `no_gain` outcome this comment names — the breakdown contains no `gpu_ids`/`gpu_pin`, and a foreign-tenant collision still can't be told apart from a real `no_gain`. Only a crashed run with no committed result gets the fields. The same two keys need to be recorded on the `has_result` path too.

src/hyperloom/inference_optimizer/breakdown/collectors/geak.py
```diff
@@ -671,6 +671,11 @@ def _load_json(p: Path) -> dict[str, Any]:
...
+            "gpu_pin": handoff.get("gpu_pin"),
```
xiaofei-zheng: **Writes an explicit `null` for unpinned runs and for every v1/v2 handoff.**

`handoff.get("gpu_pin")` yields `None` for any pre-v3 handoff on disk — still produced by any session resumed from before this deploy — and also for a genuinely unpinned v3 run. A breakdown reader then cannot distinguish "no pin was set" from "this handoff predates the field" from "the pin resolved empty".

This also contradicts the writer's own `if gpu_pin:` guard in `kernel.py` and the "omitted when nothing is pinned" contract documented in `docs/components/geak.md`. Suggest inserting the keys only when present, mirroring the writer.

src/hyperloom/inference_optimizer/tests/test_geak_resume_recovery.py
```diff
@@ -226,3 +226,49 @@ def _runner_resolved(_name: str) -> Path:
...
+    no pin in the handoff it defaults to physical GPU 0, so a run pinned to the
+    last card silently benchmarks on card 0 and OOMs against whatever else holds
+    it. ``gpu_ids`` stays HIP-logical (the consumer inherits the ROCR mask);
+    ``gpu_pin`` carries the absolute mask for consumers that write ROCR
+    themselves.
+    """
+    coord = Coordinator.__new__(Coordinator)
+    coord.session_dir = tmp_path
+    coord.shared_state = SharedState(baseline_tput=100.0, model_path="/models/m", gpu_type="mi355x")
+    coord.phase_kernel._record_geak_kernel_journey = lambda _result: None
```
xiaofei-zheng, line 248: **The end-to-end case never exercises the recipe branch — the branch that wins in production.**

This `SharedState` leaves `baseline_config_path` unset, so `_read_recipe_bench_envs` returns `{}`, `_resolve_gpu_pin` falls through to the process env, and the `source == "process_env"` assertion passes.

In every real run `baseline_config_path` points at a materialized recipe whose `benchmark.envs` always contains an autofilled `ROCR_VISIBLE_DEVICES`, so the asserted `process_env` path is effectively dead code in production — which is why the HIP-pin regression flagged on `_resolve_gpu_pin` is invisible to the suite.

A case that writes a real materialized YAML (with the autofilled ROCR mask) alongside a `HIP_VISIBLE_DEVICES` process pin, and asserts the HIP pin survives, would fail today. That is the test this PR most needs.

docs/components/geak.md
```diff
@@ -38,6 +38,29 @@ GEAK's `e2e_workflow` recursively drives `kernel_workflow` to author and tune th
...
+## GPU pinning in the handoff
+
+GEAK launches full servers out-of-process (baseline, profile, config-tuning
+validation) and writes a visible-devices mask for each one, so the handoff has
+to say which cards the run owns. Two fields carry that, in two different
+coordinate systems:
+
+| Field | Coordinate system | Value |
+|-------|-------------------|-------|
+| `gpu_ids` | HIP-level device list — HIP indexes into the ROCr-visible set | logical positions inside an inherited `ROCR_VISIBLE_DEVICES` mask, capped at `tp` (`ROCR=6` → `"0"`); a `HIP`/`CUDA` mask verbatim (`HIP=4,5` → `"4,5"`); `0..tp-1` when the run is unpinned |
```
xiaofei-zheng: **Two inaccuracies in this row.**

1. "capped at `tp`" is only true for the ROCR branch (`min(len(ids), width)` in `_resolve_handoff_gpu_ids`). The HIP/CUDA branch forwards the mask with no cap: `HIP_VISIBLE_DEVICES=4,5,6,7` with `TP=2` yields `gpu_ids="4,5,6,7"` — four devices for a two-way tensor-parallel launch.
2. "verbatim" is not accurate either — the ids go through `_parse_device_list`, which deduplicates and re-serializes. `HIP_VISIBLE_DEVICES=" 4, 4 ,5"` produces `"4,5"`, not the original string. The `_resolve_handoff_gpu_ids` docstring makes the same "VERBATIM" claim.

zihaoanllm:
Hi @xiaofei-zheng ,

Validated on real hardware. Full Hyperloom e2e run at 110189ed on an MI355X node (Qwen3-8B, sglang, TP=1 deliberately pinned to a non-zero card, ROCR_VISIBLE_DEVICES=6). Run completed cleanly (rc=0, no crashes) through PRELUDE → FRAMEWORK → EXPLORE → KERNEL (GEAK, 73 min) → SWEEP → CLOSE.

The handoff this run produced:

"schema_version": 3, "tp": 1, "gpu_ids": "0",
"gpu_pin": {"var":"ROCR_VISIBLE_DEVICES","value":"6","ids":[6],"count":1,"source":"process_env"}

Audited across the whole KERNEL phase (not sampled):

- Every GEAK serving dispatch: TP=1, GPU=0 — no exceptions.
- torch inside the adapter's exact wiring sees 1 device; logical 0 → PCI 0003:00:03.0 = rocm-smi GPU[6].
- Physical GPU 0 stayed at its idle floor for the entire run — never touched.
- Zero preflight refusals, zero HIP_VISIBLE_DEVICES contains more devices than ROCR_VISIBLE_DEVICES.

GEAK's own strategy.md records the invariant as HIP_VISIBLE_DEVICES=0 (logical), consistent with _resolve_handoff_gpu_ids emitting logical positions inside an inherited ROCr mask. Also: 26 unit tests pass.

src/hyperloom/orchestrator/loop/coordinator_helpers.py
```diff
@@ -1546,6 +1546,209 @@ def _geak_sweep_measured_tput(res: dict[str, Any]) -> float | None:
...
+    # process-env ROCR mask is inherited and its ids are logical; a mask that
+    # only exists in the recipe is not, ROCr shows every card, and the absolute
+    # ids are the correct HIP indices.
+    if str(pin.get("var") or "") == "ROCR_VISIBLE_DEVICES" and str(pin.get("source") or "") == "process_env":
+        # Token count, not len(ids): a UUID mask parses to zero numeric ids but
+        # still exposes that many cards to the child.
+        visible = int(pin.get("count") or len(ids) or 0)
+        if visible > 0:
+            return ",".join(str(i) for i in range(min(visible, width)))
+    if not ids:
```
xiaofei-zheng: **Non-numeric HIP/CUDA mask collapses to `0..tp-1`, re-creating the card-0 fallback this PR exists to remove.**

With `HIP_VISIBLE_DEVICES="GPU-a1b2c3,GPU-d4e5f6"` (the ROCm UUID form) and `tp=2`, `_resolve_gpu_pin` yields `ids=[]`, `count=2`. `_resolve_handoff_gpu_ids` skips the ROCR branch (`var != ROCR`), falls into `if not ids:` and returns `"0,1"`. GEAK then exports `HIP_VISIBLE_DEVICES=0,1` and every server it launches lands on physical cards 0/1.

Pre-PR the handoff forwarded the UUID string verbatim, so this is a regression on that path. The ROCR branch guards exactly this case with `count`; the HIP/CUDA branch does not. (Verified by executing the new helpers directly on this branch.)

src/hyperloom/orchestrator/loop/coordinator_helpers.py
```diff
@@ -1546,6 +1546,209 @@ def _geak_sweep_measured_tput(res: dict[str, Any]) -> float | None:
...
+        not "pinned to 0".
+    """
+    env = os.environ if environ is None else environ
+    recipe = dict(recipe_envs or {})
+    for var in _VISIBLE_DEVICE_VARS:
+        for source, table in (("process_env", env), ("baseline_recipe", recipe)):
+            raw = table.get(var)
+            if raw is None:
+                continue
+            value = str(raw).strip() if not isinstance(raw, (list, tuple)) else ",".join(str(p) for p in raw)
```
xiaofei-zheng: **A present-but-empty mask is treated as "unpinned", so a run with zero visible GPUs is reported as owning the whole machine.**

`ROCR_VISIBLE_DEVICES=""` is read as "zero devices visible" by both `policy/gate.detect_gpu_count` and `bus/gpu_pool._visible_device_mask`. Here it is skipped instead: `_resolve_gpu_pin` returns `{}`, `gpu_pin` is omitted from the handoff, and `gpu_ids` becomes `"0,1,..."` — GEAK launches on physical card 0, which is the precise collision #1312 is about.

The module comment documents the divergence deliberately, but the chosen fallback is the unsafe one. Emitting a pin with `count=0` would at least let the consumer refuse to launch.

src/hyperloom/orchestrator/loop/coordinator_helpers.py
```diff
@@ -1546,6 +1546,209 @@ def _geak_sweep_measured_tput(res: dict[str, Any]) -> float | None:
...
+#: pins nothing, so it falls through to the next variable.
+#: (Kept local rather than imported from ``gpu_pool``: this module is the pure
+#: helper layer and ``gpu_pool`` drags in the SQLite connection. The duplication
+#: is real — lifting the tuple and the parser into a dependency-free shared
+#: module is the right cleanup, but it touches all four call sites and does not
+#: belong in a bugfix.)
+_VISIBLE_DEVICE_VARS: tuple[str, ...] = (
+    "ROCR_VISIBLE_DEVICES",
+    "HIP_VISIBLE_DEVICES",
+    "CUDA_VISIBLE_DEVICES",
```
xiaofei-zheng: **Only three mask variables are enumerated, so runs pinned via `HSA_VISIBLE_DEVICES` or `GPU_DEVICE_ORDINAL` are still unfixed.**

`common/env_safety.GPU_MASK_ENV_NAMES` lists both alongside the three handled here, and `multi_node/scripts/launch_multinode.py:429` strips `GPU_DEVICE_ORDINAL` as a real mask. A pod pinned with `HSA_VISIBLE_DEVICES=7` gets `gpu_pin` omitted and `gpu_ids="0"`, and every GEAK-launched server lands on physical card 0 — the identical failure, in a change whose whole premise is that a mask variable was overlooked.

src/hyperloom/orchestrator/loop/coordinator_helpers.py
```diff
@@ -1546,6 +1546,209 @@ def _geak_sweep_measured_tput(res: dict[str, Any]) -> float | None:
...
+        try:
+            idx = int(tok)
+        except ValueError:
+            continue
+        if idx >= 0 and idx not in out:
+            out.append(idx)
+    return out
+
+
+def _is_autofilled_rocr(*, value: str, recipe_envs: Mapping[str, Any]) -> bool:
```
xiaofei-zheng, line 1577: **`_is_autofilled_rocr` bails out to `False` when the recipe carries no `TP` key, so the synthetic mask still poses as a real pin.**

This is the regression the second commit was added to fix, but it only closes when `TP` is present. Verified on this branch:

```
_resolve_gpu_pin(recipe_envs={"ROCR_VISIBLE_DEVICES": "0,1"},
                 environ={"HIP_VISIBLE_DEVICES": "4,5"})
-> {var: ROCR, ids: [0, 1], source: baseline_recipe}
```

Any recipe reaching `state.baseline_config_path` without a `TP` key — a hand-authored reference recipe, a `_grid_server_args` variant where TP came from neither env nor YAML, an on-disk recipe written by an older materializer — re-pins GEAK to cards 0,1 while the run actually owns 4,5.

src/hyperloom/orchestrator/loop/coordinator_helpers.py
```diff
@@ -1546,6 +1546,209 @@ def _geak_sweep_measured_tput(res: dict[str, Any]) -> float | None:
...
+    Returns:
+        ``{"var", "value", "ids", "count", "source"}`` for the winning mask,
+        where ``ids`` are ABSOLUTE device ids (empty for a UUID mask, hence
+        ``count``) and ``source`` is ``"process_env"`` or ``"baseline_recipe"``.
+        ``{}`` when no mask is set anywhere — meaning "whole machine visible",
+        not "pinned to 0".
+    """
+    env = os.environ if environ is None else environ
+    recipe = dict(recipe_envs or {})
+    for var in _VISIBLE_DEVICE_VARS:
```
xiaofei-zheng, line 1737: **When both ROCR and HIP are set, the inherited HIP mask is silently discarded and `gpu_ids` regresses.**

`ROCR_VISIBLE_DEVICES="4,5,6,7"`, `HIP_VISIBLE_DEVICES="2,3"`, `tp=2`: the run actually serves on physical 6,7. Pre-PR `gpu_ids="2,3"` (correct at the HIP level). Now the ROCR pin wins and `gpu_ids="0,1"` → physical 4,5, i.e. different cards than the run uses.

`cli/preflight._unset_hip_visible_devices` only strips the duplicate for CLI-launched runs; the SDK/orchestrator-embedded path, the Ray path and the tests all still reach the KERNEL phase with both set.

src/hyperloom/orchestrator/loop/coordinator_helpers.py
```diff
@@ -1546,6 +1546,209 @@ def _geak_sweep_measured_tput(res: dict[str, Any]) -> float | None:
...
+                source == "baseline_recipe"
+                and var == "ROCR_VISIBLE_DEVICES"
+                and _is_autofilled_rocr(value=value, recipe_envs=recipe)
+            ):
+                continue
+            return {
+                "var": var,
+                "value": value,
+                "ids": _parse_device_list(value),
+                "count": len(_mask_tokens(value)),
```
xiaofei-zheng: **`count` is `len(_mask_tokens(...))`, which counts duplicate, negative and non-numeric tokens, so the pin can yield logical indices for devices that do not exist.**

`ROCR_VISIBLE_DEVICES="3,3,2"` (or `"a,-1,2"`) with `tp=3` gives `count=3` but only 2 (resp. 1) real devices. `_resolve_handoff_gpu_ids` returns `"0,1,2"`; GEAK exports that into a 2-device ROCr set and the server aborts on an invalid device ordinal.

The docstring's claim that `ids` is empty only for a non-numeric mask ("hence `count`") is also not accurate here — `ids` and `count` diverge for a purely numeric mask.

src/hyperloom/orchestrator/loop/coordinator_helpers.py
```diff
@@ -1546,6 +1546,209 @@ def _geak_sweep_measured_tput(res: dict[str, Any]) -> float | None:
...
+    # only exists in the recipe is not, ROCr shows every card, and the absolute
+    # ids are the correct HIP indices.
+    if str(pin.get("var") or "") == "ROCR_VISIBLE_DEVICES" and str(pin.get("source") or "") == "process_env":
+        # Token count, not len(ids): a UUID mask parses to zero numeric ids but
+        # still exposes that many cards to the child.
+        visible = int(pin.get("count") or len(ids) or 0)
+        if visible > 0:
+            return ",".join(str(i) for i in range(min(visible, width)))
+    if not ids:
+        return ",".join(str(i) for i in range(width))
```
xiaofei-zheng: **The HIP/CUDA path re-serializes from the deduplicated `ids`, changing device cardinality.**

`HIP_VISIBLE_DEVICES="4,4"` with `tp=2` gives pin `count=2` but `_resolve_handoff_gpu_ids` returns `"4"` — a single id. GEAK receives `tp=2` alongside one device id.

This is the mirror image of the `count` inflation above: the two fields dedup on opposite sides. The docstring describes the normalization as intentional, but it is changing the device count, not just whitespace. Pre-PR the raw mask was forwarded untouched.

src/hyperloom/orchestrator/loop/coordinator_helpers.py
xiaofei-zheng: **A YAML-sequence mask is joined without stripping its elements, so `value` can carry embedded whitespace.**

`ROCR_VISIBLE_DEVICES: [' 4', 5]` in a recipe yields `value=" 4,5"` (verified). `ids` parses correctly to `[4, 5]` because `_mask_tokens` strips, but `gpu_pin["value"]` is documented as the mask to re-export as-is, and a leading space in the first entry is exactly the token form ROCm's parser rejects.

The scalar branch strips; the list branch does not. The asymmetry looks unintended.

src/hyperloom/orchestrator/loop/coordinator_helpers.py
```diff
@@ -1546,6 +1546,209 @@ def _geak_sweep_measured_tput(res: dict[str, Any]) -> float | None:
...
+#: The order agrees; the empty-mask semantics deliberately do not. Those layers
+#: read ``VAR=""`` as "zero devices visible" because they are counting capacity;
+#: this resolver is answering "where is the run pinned", and an empty string
+#: pins nothing, so it falls through to the next variable.
+#: (Kept local rather than imported from ``gpu_pool``: this module is the pure
+#: helper layer and ``gpu_pool`` drags in the SQLite connection. The duplication
+#: is real — lifting the tuple and the parser into a dependency-free shared
+#: module is the right cleanup, but it touches all four call sites and does not
+#: belong in a bugfix.)
+_VISIBLE_DEVICE_VARS: tuple[str, ...] = (
```
xiaofei-zheng: **Reuse: `_VISIBLE_DEVICE_VARS` and `_parse_device_list` are a fifth copy of a tuple/parser pair already in the repo.**

- `bus/gpu_pool._parse_gpu_list` (L60-81) is character-identical to the numeric core of `_parse_device_list`, and `_visible_device_mask` (L113) holds the same tuple
- `policy/gate.detect_gpu_count` (L182) holds it again
- `actions/executors/_ray_serving._VISIBLE_DEVICE_ENV_KEYS` (L58) is a literal duplicate
- `common/env_safety.GPU_MASK_ENV_NAMES` (L217) is the dependency-free shared module this belongs in

The in-code comment acknowledges the duplication and defers it, but precedence ordering now has to stay in sync across five sites, and the semantics have already diverged (empty-mask handling, see the comment above).

src/hyperloom/orchestrator/phases/kernel.py
```diff
@@ -795,13 +837,16 @@ async def _run_geak_kernel_phase(self, *, from_phase: str) -> None:
...
-            # Pin the serving GPU set: explicit visibility mask, else 0..tp-1.
-            "gpu_ids": (
-                os.environ.get("HIP_VISIBLE_DEVICES")
-                or os.environ.get("CUDA_VISIBLE_DEVICES")
-                or ",".join(str(i) for i in range(int(os.environ.get("TP", "1") or 1)))
-            ),
+            # The serving/optimization device set, as HIP-level ids (what the
+            # consumer exports as HIP_VISIBLE_DEVICES). Logical positions inside
+            # an inherited ROCR mask, a HIP/CUDA mask as-is, else 0..tp-1.
+            "gpu_ids": _resolve_handoff_gpu_ids(gpu_pin=gpu_pin, tp=_tp),
```
xiaofei-zheng: **`gpu_ids` is clamped to the ROCR mask width but `handoff["tp"]` is not, so the handoff can ship `tp: 2` next to a single device id.**

Process `ROCR_VISIBLE_DEVICES="6"` with `TP=2`: `_workload_envs.materialize_config_with_envs` sees `len(rocr_devices)=1 < TP=2`, expands the recipe to `ROCR="0,1"` and writes `envs["TP"]=2`. `_resolve_gpu_pin` prefers the process env → `count=1`; `_resolve_handoff_gpu_ids` returns `min(1,2)` → `"0"`; `_tp` reads the recipe → `2`. Handoff is `{tp: 2, gpu_ids: "0"}` and GEAK launches sglang `--tp 2` against one visible card, failing to load weights.

This is the disagreement the adjacent comment states is impossible. The same case also makes `gpu_pin` report card 6 while the measured baseline actually ran on the recipe's cards 0,1.

src/hyperloom/orchestrator/phases/kernel.py
```diff
@@ -795,13 +837,16 @@ async def _run_geak_kernel_phase(self, *, from_phase: str) -> None:
...
         }
+        if gpu_pin:
+            # ABSOLUTE ids + the var they came from, so a consumer that writes
+            # ROCR_VISIBLE_DEVICES itself re-applies the same pin instead of
+            # resetting the child to card 0.
+            handoff["gpu_pin"] = gpu_pin
```
xiaofei-zheng, line 874: **`gpu_pin["value"]` and `gpu_ids` are only mutually consistent when `var == ROCR` and `source == process_env`.**

For every other pin, a consumer following the documented contract (write `ROCR=value`, export `HIP=gpu_ids`) produces out-of-range HIP indices. With `HIP_VISIBLE_DEVICES="4,5"`, `tp=2` → `gpu_pin={var: HIP, value: "4,5", ids: [4,5]}`, `gpu_ids="4,5"`. `adapters/launchers/magpie.sh` (named in the PR body as the consumer) writes `ROCR_VISIBLE_DEVICES=4,5`, so the child sees 2 devices; exporting `HIP_VISIBLE_DEVICES=4,5` then indexes positions 4 and 5 of a 2-element set. Identical for a recipe-sourced ROCR pin (absolute `gpu_ids="6,7"` under `ROCR=6,7`).

`docs/components/geak.md` documents the asymmetry, but nothing in the payload lets the consumer detect which case it is holding — `source` is the only signal, and it is not called out as load-bearing.

xiaofei-zheng (commented):
Review of the handoff GPU-pin change. 13 findings; the ones marked below under "still falls back to card 0" mean the scenario in #1312 remains reachable on several paths.

Findings 1, 4, 5, 6, 7, 8 and the YAML-sequence one were confirmed by executing the new helpers directly against this branch, not just by reading.

Two non-findings for the record: `_resolve_bench_protocol` moving from `@staticmethod` to `@classmethod` has no external call sites, so nothing breaks; and the new `_read_recipe_bench_envs` is not registered in `coordinator.py`'s delegation map (~L996) next to `_resolve_bench_protocol` — harmless today since it is only reached via `self.` inside `PhaseKernel`, but inconsistent with the surrounding registry.

src/hyperloom/orchestrator/phases/kernel.py
```diff
@@ -746,7 +767,26 @@ async def _run_geak_kernel_phase(self, *, from_phase: str) -> None:
...
+        # fallback; {} means "whole machine", not "card 0".
+        gpu_pin = _resolve_gpu_pin(recipe_envs=_recipe_envs)
+        # tp must come from the SAME place as gpu_ids or the two can disagree.
+        # The materializer clamps TP to the visible GPU count, so a stale
+        # $TP=8 on a 4-card pod would otherwise ship `tp: 8` alongside four
+        # gpu_ids and GEAK would launch sglang with --tp 8 and fail to load.
+        try:
+            _tp = int(str(_recipe_envs.get("TP") or "").strip() or os.environ.get("TP", "1") or 1)
+        except (TypeError, ValueError):
+            _tp = int(os.environ.get("TP", "1") or 1)
```
xiaofei-zheng: **The `except (TypeError, ValueError)` recovery re-runs an unguarded `int()` and can raise from inside the handler.**

Recipe `benchmark.envs.TP: "auto"` (or `2.0`, which `int(str(...))` also rejects) plus `TP=auto` in the process env: the try body raises `ValueError`, then the handler evaluates `int(os.environ.get("TP", "1") or 1)` and raises again during handling, propagating out of `_run_geak_kernel_phase` and killing the KERNEL phase.

The fallback exists precisely because the primary parse is untrusted, so it should not repeat the same untrusted parse bare.

src/hyperloom/inference_optimizer/breakdown/collectors/geak.py
```diff
@@ -1003,9 +1033,23 @@ def _rel_if_under(p: Any) -> Any:
         else:
             accepted_kernels = []
 
+    # The cards GEAK was told to use. Read from the handoff on disk rather than
+    # from ``geak_result``, which never carried them — and recorded on THIS
+    # path, not just the crash-recovery one, because the outcome that needs
+    # disambiguating (`no_gain`) is a completed run.
+    _gpu_fields = _handoff_gpu_fields(
```
xiaofei-zheng: **`on_error` fires on a missing `geak/handoff.json`, adding a spurious warning to healthy runs.**

`FileNotFoundError` is an `OSError`, and `common/jsonio.read_json` invokes `on_error(exc)` for `OSError` before returning the default. Any resumed pre-v3 session, or a run whose GEAK artifacts live under an external `exp_root`, reaches this unconditional read on the SUCCESS collect path and gets `geak: handoff read failed: [Errno 2] No such file or directory` appended to the breakdown warnings.

It is also unconditional disk I/O on a path that previously touched no files, for two optional fields — worth gating on the file existing, or passing an `on_error` that ignores `FileNotFoundError`.

zihaoanllm:
@xiaofei-zheng
Addressed all 25 inline findings from both review rounds.

Changes:

- Widened pin resolution to the full ROCm precedence chain (`ROCR_VISIBLE_DEVICES`, `HSA_VISIBLE_DEVICES`, `HIP_VISIBLE_DEVICES`, `CUDA_VISIBLE_DEVICES`, `GPU_DEVICE_ORDINAL`), so a run pinned with a legacy spelling is no longer reported as unpinned.
- Made a set-but-empty mask an explicit classification rather than an absence: it resolves to a pin with `count: 0` and `gpu_ids_space: "none"`, and is logged at ERROR. `gpu_ids` deliberately still carries a placeholder, because GEAK's `gpu_ids or ",".join(range(tp))` fallback turns an empty string back into card 0.
- Derived `count` and `ids` from a single `effective_mask_tokens()` pass, so cardinality can no longer disagree across the two sides of the pin record.
- Forwarded a HIP-level mask nested inside a ROCr-level pin as `pin["inner"]`, dropping ordinals outside the advertised width, and forwarded a non-numeric (UUID) mask's own tokens instead of collapsing to `0..tp-1`.
- Added the `gpu_ids_space` field and documented the consumer renumbering rule, so an inherited ROCr pin's logical ids are not read as absolute device ids.
- Capped handoff `tp` to the advertised device count via `_resolve_handoff_tp`, and guarded every `tp` candidate through `_coerce_tp` so no coercion raises out of its own fallback.
- Fixed the autofill heuristic to fall back to the token-count shape when the recipe carries no usable `TP`, and stripped YAML sequence elements before they reach the pin value.
- Centralized the mask names and parsers in `hyperloom/common/visible_devices.py` and migrated all five call sites. `COUNTING_VISIBLE_DEVICE_VARS` is derived from the chain minus a named exclusion set rather than re-listed, so the counting subset and the precedence chain cannot silently drift; its value and order are unchanged.
- Guarded the collector's handoff read with `is_file()`, so only a handoff that exists and fails to parse produces a warning.
- Cleared the CI lint failure and both code-scanning notes: reflowed the over-wrapped test signature; replaced the `except ValueError: pass` fallthrough in `effective_mask_tokens` with an explicit `_is_negative_ordinal()` predicate (alert 2733); declared `__all__` on `visible_devices`, which defines `COUNTING_VISIBLE_DEVICE_VARS`for its three consumers and never reads it itself (alert 2742).

Validation completed successfully:

- Every repro from the review comments replayed against live `_resolve_gpu_pin` / `_resolve_handoff_gpu_ids` / `_resolve_handoff_gpu_ids_space` / `_resolve_handoff_tp`: all produce the behaviour described above.
- Focused suites `test_visible_devices.py` and `test_geak_handoff_gpu_pin.py`: 60 passed, including a partition test asserting every chain member is classified as counted or explicitly uncounted exactly once.
- Full `inference_optimizer` + `agents/kernel` sweep: 13644 passed, 15 failed. Those 15 fail identically on the base tree (real-ROCm build recipes, missing `claude_agent_sdk`), verified by re-running them stashed.
- `ruff check` and `ruff format --check`: clean.
- CI E2E on `158ca2cb`: succeeded (Qwen3-0.6B, 1× GPU, TP=1, 187m).

xiaofei-zheng:
### Blocking: a recipe-sourced ROCr pin emits absolute `gpu_ids` — a regression against the merge base

`_pin_is_inherited_rocr` (`coordinator_helpers.py:1823`) decides "are these ids logical?" by asking whether the **GEAK process itself** inherits the ROCr mask. It doesn't, when the mask lives only in the recipe — the phase launches GEAK with `dict(os.environ)`, so that part of the reasoning is right. But the servers GEAK launches are started **from `launch_recipe`**, and that recipe carries the ROCr mask. So the renumbering does happen, one level down, and `gpu_ids` is in the wrong coordinate system for it.

**Repro.** Workload YAML pins via `benchmark.envs` instead of the process env:

```yaml
benchmark:
  envs:
    ROCR_VISIBLE_DEVICES: "4,5,6,7"
    TP: 4
```

The materializer keeps that value verbatim (`_workload_envs.py:1084` — `len(rocr_devices) >= resolved_tp`, so no autofill), and `_is_autofilled_rocr` correctly declines to treat it as noise. Calling the resolvers directly on `d812fcfa`:

```
pin     = {var: ROCR_VISIBLE_DEVICES, value: "4,5,6,7", source: baseline_recipe}
gpu_ids = "4,5,6,7"
space   = "absolute"
```

`handoff["launch_recipe"]` points at that same YAML, and per the PR description the launcher also writes `gpu_pin["value"]` into `ROCR_VISIBLE_DEVICES`. Either path leaves the server with a 4-device ROCr set; exporting `gpu_ids=4,5,6,7` as `HIP_VISIBLE_DEVICES` then indexes 4..7 into a 4-element set and the server dies on an invalid device ordinal.

**Why this is blocking rather than an unfixed edge case:** it is a regression. At the merge base (`86d2ad5d4`) `gpu_ids` read `HIP`/`CUDA` only, so this config produced `0..tp-1` — which composes correctly with the recipe-applied ROCr mask and lands on cards 4-7. After this PR the same config breaks.

The `gpu_ids` docstring does warn that a consumer re-applying `gpu_pin["value"]` as ROCr "must use `0..count-1`, NOT these ids". But note that for the *inherited* case `gpu_ids` already **is** `0..count-1`, so re-applying is safe there; only the recipe-sourced case requires the consumer to silently switch which field it reads, and no field in the payload expresses that switch (`gpu_ids_space` describes the coordinate system of `gpu_ids` itself, not what to substitute after rewriting ROCr).

Suggested direction: base the logical/absolute decision on whether the ROCr mask will be in force for the *launched servers* (true for both `process_env` and `baseline_recipe` sources), not on whether the GEAK process inherits it.

**Same root cause, `coordinator_helpers.py:1809`:** `_resolve_inner_hip_mask` gates only on `is_rocr_level(var)`, so a recipe-sourced pin also gets a `gpu_pin["inner"]`:

```
env HIP_VISIBLE_DEVICES=1 + recipe ROCR_VISIBLE_DEVICES=4,5,6,7
→ pin.inner = {var: HIP_VISIBLE_DEVICES, value: "1", ...}
  gpu_ids   = "4,5,6,7"   space = "absolute"
```

`_resolve_handoff_gpu_ids` takes the absolute branch and ignores `inner` entirely, so the payload advertises a nesting it does not itself use. A consumer reading `inner` as documented ("the HIP-level mask nested inside a winning ROCr-level pin") composes ROCR=4,5,6,7 + HIP=1 and lands on card 5. Worth fixing together with the above.

---

Nothing else I found rises to blocking. For the record, four things I looked at and decided *not* to report: `HIP_VISIBLE_DEVICES=-1` surviving into `gpu_ids` (via `_mask_tokens` rather than `effective_mask_tokens`), a duplicated ordinal in a nested HIP mask inflating `tp`, `parse_device_list("0,00") == [0, 0]`, and an empty ROCr mask being outranked by a later HIP/CUDA mask. All reproduce, but the first two match merge-base behaviour and the last two need a hand-constructed mask; happy to file them separately if you want them tracked.

zihaoanllm:
@xiaofei-zheng Fixed.

xiaofei-zheng (approved):

PR #1322, opened 2026-08-28
test: pin verify_cpp_itfs_rebuilt freshness branches (PRPUNDIT-7)

### LINKED ISSUE #1319: verify_cpp_itfs_rebuilt fails open when build_dir is missing: Path("") is Path(".")
## Problem

`verify_cpp_itfs_rebuilt` is the gate that stops the integrate KEEP/REVERT path from scoring a **stale binary**. Its first guard is meant to bail out when the cpp_itfs build dir is missing:

https://github.com/AMD-AGI/Hyperloom/blob/main/src/hyperloom/agents/kernel/tools/apply_kernel_patch.py#L1448-L1456

```python
build_dir = Path(cache_backup.get("build_dir", ""))
...
if not str(build_dir) or not build_dir.exists():
    return {"verified": False, "status": "stale", ...}
```

That guard never fires for a missing/empty `build_dir`, because `Path("")` is `Path(".")`:

```
>>> from pathlib import Path
>>> str(Path(""))        # '.'  -> truthy, so `not str(build_dir)` is False
'.'
>>> Path("").exists()    # True -> so `not build_dir.exists()` is False
True
```

## Impact

A `cache_backup` with `is_cpp_itfs=True` but **no `build_dir` key** skips the stale branch and falls through to the glob, which then runs against the **process working directory**. If `module_names` is also empty the pattern degrades to the catch-all `"*/lib.so"`:

```python
globs = [f"{md}_*/lib.so" for md in module_names] or ["*/lib.so"]
```

So any unrelated, recently-built `lib.so` under the CWD satisfies the freshness check and the function returns `verified: True`. That is a **fail-open**: a stale binary passes the very gate this function exists to guard, and the measured gain gets scored as real.

The docstring states the intent explicitly — *"If no such fresh `lib.so` exists, the server dlopened a stale binary (or the kernel was never exercised) and the measured gain is meaningless"* — so this is a defect against stated behavior, not a judgement call.

## Suggested fix

Reject the empty case before constructing the `Path`, e.g.:

```python
raw_build_dir = str(cache_backup.get("build_dir") or "").strip()
if not raw_build_dir:
    return {"verified": False, "status": "stale",
            "reason": "cpp_itfs build dir missing from invalidation record"}
build_dir = Path(raw_build_dir)
if not build_dir.exists():
    return {"verified": False, "status": "stale",
            "reason": f"cpp_itfs build dir absent after re-baseline: {build_dir}"}
```

Worth considering alongside it: make the `module_names`-empty `"*/lib.so"` fallback fail closed rather than widening the search, since a record with neither `build_dir` nor `module_names` carries no evidence at all.

## Notes for whoever picks this up

There is **no test coverage** for this function anywhere in the repo — grep finds only the definition and the single caller in `orchestrator/kernel/request_handlers.py`, with zero hits across the test suite for the function name or for any of its distinctive literals (`fresh_lib_so`, `cpp_itfs build dir absent`, `non-cpp_itfs target`) or the caller-side `cpp_itfs_rebuild_not_verified` error class. Please add tests with the fix. Three things to get right:

1. **The stale fixture must set `mtime` more than 1.0s before `invalidated_unix`.** The check is `mtime + 1.0 >= since`, so a file 0.5s older still counts as fresh and a naive test asserts the wrong thing.
2. **Assert on `reason`, not `status`** — the absent-build-dir branch and the no-fresh-`lib.so` branch both return `status="stale"`.
3. **Add a case with `build_dir` missing entirely**, which is the hole above.

The producer `_invalidate_aiter_cpp_itfs_cache` has a `build_dir_override` test hook, so the record can be built by production code rather than hand-assembled — the stronger version of the test.

<sub>Found while adjudicating the test-gap review queue at `3934874d0924` ([row](https://test-gap-review.slai-app.amd.com/r/AMD-AGI__Hyperloom/3934874d0924ac8d439a0092626e79ae3c27bfce?show=all#cand:16adfb605026f498016b5c628f623e2a)). Confirmed still present on `main` @ 86d2ad5d4. The queue row covers the missing tests; this issue is the live defect.</sub>

jiagaoxiang:
## Summary
- Adds unit coverage for `verify_cpp_itfs_rebuilt` KEEP/REVERT freshness branches: non-cpp_itfs skip, absent/empty `build_dir`, stale vs fresh `lib.so`, 1s mtime slack, and empty `module_names` glob fallback.
- Tightens the empty `build_dir` guard so `Path('')` cannot glob the process CWD (reviewer yihanwan found this live hole).

Closes test gap **PRPUNDIT-7**.
Fixes #1319.

## Test plan
- [x] `PYTHONPATH=src pytest src/hyperloom/agents/kernel/tests/test_apply_kernel_patch_isolated_aiter.py -k verify_cpp_itfs_rebuilt`


zoroyihan7 (approved):
Reviewed with an independent verification pass: ran the touched test file on this branch and mutation-tested the production code to confirm the new tests actually catch regressions (not just pass).

Mutations tried and caught: reverting only the production hunk (restoring the Path("") guard) turns test_verify_cpp_itfs_rebuilt_rejects_absent_and_empty_build_dir red; widening the slack to 3.0s, tightening it to 0.5s, and flipping >= to > each turn exactly one side of the boundary red; collapsing the two stale reason strings is also caught.

No regressions in the surrounding suite (pre-existing failures verified identical on origin/main). LGTM.

zoroyihan7:
/retest

(e2e failed on the `Dispatch + monitor single-GPU smoke run` step, which is failing on unrelated PRs too — retriggering. Auto-merge is armed, so this will land on its own once e2e is green.)

jiagaoxiang:
PR description now includes `Fixes #1319`, so GitHub will close that issue when this merges. Left `skip-e2e-test` off so this gets a real e2e run.

jiagaoxiang:
@zoroyihan7 you're right — #1325's `workflow_dispatch` cancelled this PR's run 33463143636. CI E2E concurrency falls through to `github.ref` (`refs/heads/main`) when `pull_request.number` is empty, so every dispatch shares one group.

Fix: #1367 (key on `inputs.head_sha` before `github.ref`; same pattern on Forge E2E). `forge-kernel-bench.yml` already serializes with `cancel-in-progress: false`, so I left it.

#1325's dispatch has finished, so I'm re-dispatching CI E2E for this SHA now. Until #1367 is on `main`, a second dispatch from `main` would still cancel this one — I'll hold #1325's next dispatch until this run completes.

HTTP 403-only `skipped-status` (your non-blocking 5xx note) already landed in #1357.

PR #1327, opened 2026-08-28
test: pin platform_fingerprint degrade paths (PRPUNDIT-12)

jiagaoxiang:
## Summary
- Exercises `platform_fingerprint` three-tier degrade: probe `None` → unavailable, outer exception → `status=error`, and per-block gpu/stack `status=error`.

Closes test gap **PRPUNDIT-12**.

## Test plan
- [x] `PYTHONPATH=src pytest src/hyperloom/common/tests/test_platform_probe.py -k platform_fingerprint`


zoroyihan7 (changes requested):
Thanks — all three degrade tiers are reached and the assertions are on the real returned dict (including the exact `"no host CPU sysfs on this machine"` literal), no mock-asserts-the-mock anywhere.

Two things mutation testing turned up.

**1. Per-block independence isn't actually pinned.** `test_platform_fingerprint_gpu_and_stack_blocks_degrade` breaks **both** `amdgpu_device_count` and `detect_stack_fingerprint` in the same call, so it can never observe a healthy block surviving its sibling's failure — which is the "the rest of the record survives" half of the contract. Mutating the stack handler to also overwrite `record["gpu"]` and `record["cpu"]` leaves all 15 tests green. Worse, swapping which key each handler writes (GPU handler setting `record["stack"]` and vice versa) is undetectable, and in the single-block-failure case silently drops the `"gpu"` key entirely.

Splitting into two cases — gpu raises / stack healthy, and stack raises / gpu healthy — and asserting the surviving block's real content in each (e.g. `got["stack"] != {"status": "error"}`, `got["gpu"]["gfx_arch"]` present) closes both.

**2. The outer net is pinned for only one exception type.** Narrowing `except Exception as exc` to `except RuntimeError as exc` stays green, because the new test happens to raise `RuntimeError`. An `OSError`/`AttributeError` escaping `probe_cpu_platform` would then propagate to the crash-path caller with no test noticing — and per `exporter.py:1010-1013` there's no second net there, so `reports/final.json` would be lost for a run that already died. Raising a non-`RuntimeError` in one of the cases fixes it.

Smaller ones:

- `"multi_node_session": multi_node` → `bool(multi_node)` stays green, which destroys the docstring's explicit None-vs-False distinction ("None records that nobody established it ... rather than an unearned False") — no test calls `platform_fingerprint` with `multi_node` left unset on an ok record.
- No test covers the all-healthy `status="ok"` record, so the gpu sub-dict shape (`host_count` / `gfx_arch` / `amdgpu_driver`, the `"unknown"` fallbacks) has no pin — deleting or renaming any of those keys keeps the suite green. Deleting the `"host": socket.gethostname()` line is green too.
- The new tests monkeypatch module globals, which the module docstring (lines 24-25) discourages in favour of the injectable `root`. It's justified here — `platform_fingerprint` takes no `root` and the error tiers aren't reachable through a fake sysfs tree — but a one-line comment saying so would stop the next reader "fixing" it.

jiagaoxiang:
Split GPU-raises/stack-healthy vs stack-raises/GPU-healthy (asserting the surviving block's real content), and the outer net now raises `OSError`. Added the all-healthy `status=ok` gpu sub-dict (`host_count` / `gfx_arch` / `amdgpu_driver`, including `"unknown"` fallbacks), `host=gethostname()`, and `multi_node is None` vs `False`. Comment on the tests explains why these cases monkeypatch module globals (`platform_fingerprint` has no injectable `root`).

zoroyihan7 (approved):
Re-reviewed. Splitting into two single-block-failure cases is exactly what was needed — the surviving-block content assertions are what make the key-swap mutations detectable.

Five mutations, all red now:

- `except Exception` → `except RuntimeError` on the outer net (L263) — **red**
- gpu handler writing `record["stack"]` instead of `record["gpu"]` (L256) — **red**
- stack handler writing `record["gpu"]` instead of `record["stack"]` (L261) — **red**
- `multi_node` → `bool(multi_node)` (L236) — **red**
- `host` key renamed (L235) — **red**

17 passed. LGTM.

PR #1328, opened 2026-08-28
test: cap _SiteStats.record shape and caller sets (PRPUNDIT-13)

jiagaoxiang:
## Summary
- Calls `_SiteStats.record` past `_MAX_CONTAINER_WIDTH` / `_MAX_CALLERS_PER_SITE` and asserts count/wall/nbytes and first/last keep moving while the sets stop growing.

Closes test gap **PRPUNDIT-13**.

## Test plan
- [x] `PYTHONPATH=src pytest src/hyperloom/inference_optimizer/tests/test_framework_rewrite_evidence.py::test_site_stats_record_caps_shape_sigs_and_callers_while_tallies_grow`


zoroyihan7 (changes requested):
Thanks — the caps are solidly pinned: pushing past both bounds, asserting the counters keep accumulating, and explicitly asserting the overflow entries are absent from the sets is exactly the right shape, and removing either `len(...) < CAP` condition turns this test red.

One gap: **`first_s` is only half-pinned.** Mutating the sentinel init at `hl_host_probe.py:223`:

```python
self.first_s = 0.0      # was: -1.0
```

kills the latch outright — `if self.first_s < 0` never fires again, so `first_s` is permanently `0.0` and never records the real first timestamp — and **all 73 tests stay green**.

The reason is that the loop's first call passes `at_s=0.0`, which is exactly the degenerate value, so `assert stats.first_s == 0.0` can't distinguish "latched the first timestamp" from "never latched anything". (Deleting the `if` so `first_s` is overwritten every call *is* caught — that's the other half.)

One-token fix: drive the loop with `at_s=float(i) + 1.0` and assert `first_s == 1.0`. That kills both the sentinel mutation and the overwrite mutation.

Two nits while you're in here:

- The falsy-input branch of the guards (`if shape_sig and ...` / `if caller and ...`, lines 242/244) isn't pinned — replacing them with bare `if len(...) < CAP:` leaves the file green. Passing `shape_sig=""` / `caller=""` and asserting the empty string is absent would be two more lines. Outside the contract this PR claims, so genuinely optional.
- Re-asserting `first_s` after the post-cap record would make the latch/overwrite contrast explicit next to the `last_s == 99.0` assertion.

jiagaoxiang:
`at_s=float(i)+1.0` with `first_s == 1.0` pins the sentinel-init mutation. Re-asserted `first_s == 1.0` after the post-cap `last_s == 99.0` record, and added a case that records empty `shape_sig`/`caller` while the sets still have room.

zoroyihan7 (approved):
Re-reviewed. The `at_s` offset closes the sentinel hole.

Mutations red now:

- `self.first_s = -1.0` → `0.0` in `_SiteStats.__init__` (L223), which kills the latch outright — **red** (2 tests)
- `if self.first_s < 0:` → always true, i.e. overwrite on every call — **red**
- `if shape_sig and` guard dropped (L242) — **red**
- `if caller and` guard dropped (L244) — **red**

Both halves of the latch contract are pinned now. 74 passed. LGTM.

FYI, not for this PR: the identical sentinel at L290 belongs to a different class (`depth` / `arg_samples` / `strict_sigs`) and stays green under the same mutation. Might be worth its own ticket.

jiagaoxiang:
Leaving the `_CallStats` sentinel (`hl_host_probe.py:290`) out of this PR as discussed. Filed #1356.

PR #1324, opened 2026-08-28
test: SharedState collective save-failure rollback (PRPUNDIT-8, PRPUNDIT-9)

jiagaoxiang:
## Summary
- Forces `SharedState.save` to fail during `record_collective` and `record_collective_integration`.
- Asserts in-memory `last_collective` / `collective_attempts` restore and on-disk `state.json` is unchanged.

Covers test gaps **PRPUNDIT-8** and **PRPUNDIT-9** (merged from #1323 into this PR).

## Test plan
- [x] `PYTHONPATH=src pytest src/hyperloom/inference_optimizer/tests/test_collective_lane.py::test_record_collective_rolls_back_when_save_fails src/hyperloom/inference_optimizer/tests/test_collective_lane.py::test_record_collective_integration_rolls_back_when_save_fails`

PR #1329, opened 2026-08-28
test: union and dedup prior refs in merge_staged_sections (PRPUNDIT-14)

jiagaoxiang:
## Summary
- Stages a framework patch onto a section that already holds replayed refs and locks union/dedup order.

Closes test gap **PRPUNDIT-14**.

## Test plan
- [x] `PYTHONPATH=src pytest src/hyperloom/inference_optimizer/tests/test_remote_recipe_v2.py::test_merge_staged_sections_unions_and_dedups_prior_refs`


zoroyihan7 (changes requested):
Thanks — this is the right target, and it does pin dedup, insertion order, and before-preservation.

One problem though: **the union half — the actual point of the branch — isn't pinned.**

The test makes `after` a *subset* of `before`:

```python
value = {"framework": {"patches": [prior_patch, staged_ref], ...}}   # before
# staged.knowledge["patches"] == [staged_ref]                        # after
```

so the expected result `[prior_patch, staged_ref]` is byte-identical whether the code unions, ignores `after` entirely, or skips the branch. I confirmed by mutation at `values.py:1054`:

```python
if not before and after:      # was: if before or after:
```

i.e. newly staged refs are silently discarded whenever the section already holds prior replayed refs. **478 tests stay green, including this one.**

That mutation is worth taking seriously because in production it's silent data loss rather than an error: `files.adopt()` has already copied the staged patch into the files tree, the ref never lands in `value`, and `files.prune_superseded(knowledge)` (`values.py:1200`) then deletes the orphaned file — so a newly authored patch disappears from the published recipe with no exception raised.

The subset shape is also the opposite of the real producer shape: `_adopt_replayed_prior` (`values.py:854-867`) writes refs at the **replay overlay index** while `stage_patches` writes at the **stack index**, so in production `before` and `after` are normally **disjoint** and the append is the whole reason the branch exists.

Fix is small — stage two patches and duplicate only one into `before`. That gives you union, dedup and order in a single case:

```python
before = [prior_patch, staged_ref_a]      # prior + one overlapping
after  = [staged_ref_a, staged_ref_b]     # overlapping + one new
expect = [prior_patch, staged_ref_a, staged_ref_b]
```

Two minor notes, non-blocking:

- `prior_artifact = "framework/artifacts/prior.bin"` exercises a shape no real producer creates — `_adopt_replayed_prior` only ever writes `value[owner]["patches"]`, never `"artifacts"`. Still a valid pin of the generic per-key loop, just noting the docstring's framing only strictly applies to the patches half.
- Asserting exact insertion order is slightly stronger than the downstream contract, since `_patch_timeline` re-sorts by `(stack_index, member_index)`. Fine to keep — just an implementation-order lock rather than an observable-behavior one.

jiagaoxiang:
Follow-up for the union-half pin: `before = [prior, staged_ref_a]`, `after = [staged_ref_a, staged_ref_b]`, expect `[prior, a, b]`. Confirmed locally that mutating `if before or after:` → `if not before and after:` at `values.py:1054` now fails this test (drops `staged_ref_b`).

zoroyihan7 (approved):
Re-reviewed. The union shape is fixed and now genuinely pinned — `before = [prior, staged_a]`, `after = [staged_a, staged_b]` — and asserting `after` itself as a precondition is a nice touch.

Re-ran the mutations on `values.py:1055`; all four are red now:

- `if before or after:` → `if not before and after:` (the silent-data-loss one) — **red**
- dedup removed (`list(dict.fromkeys(...))` → `list(...)`) — **red**
- `[*before, *after]` → `[*after]` — **red**
- operands reversed — **red**

123 passed, no regressions. LGTM.

zoroyihan7:
Heads-up — I updated this branch from `main` so it would pick up the `skipped-status` fix from #1357, and that surfaced a break. `skipped-status` is green now, but `ruff` went red:

```
F821 Undefined name `FrameworkAgentKB`
 --> src/hyperloom/inference_optimizer/tests/test_remote_recipe_v2.py:2020
```

`FrameworkAgentKB` was removed somewhere in the 207 commits this branch was behind — it doesn't exist anywhere on `main` any more. It's now `PatchKB` in `orchestrator/knowledge/agent_kb.py`, and the pre-existing test right above yours (line 2002) already uses it:

```python
refs = PatchKB(sections).stage_patches(...)
```

So it should be a one-word change on line 2020:

```python
staged_refs = PatchKB(sections).stage_patches([patch_a, patch_b], stack_index=2)
```

`PatchKB` is already imported in the file post-merge, so nothing else should need touching. I confirmed it's a real runtime break, not just lint — locally the test errors with `NameError: name 'FrameworkAgentKB' is not defined` (1 failed, 125 passed).

I ran the other four updated branches too and they're all clean (#1324 51 passed, #1327 17, #1328 74, #1330 44) — this is the only one the merge caught out.

Happy to push the rename myself if you'd rather (`maintainerCanModify` is on) — just say the word, otherwise it's yours.

zoroyihan7:
Pushed the fix to this branch (`acf74f665`) — it turned out to be more than the one-word rename I described above, so here's exactly what changed and how I checked it.

Two renames landed in the 207 commits this branch was behind, not one:

| was | is now |
|---|---|
| `FrameworkAgentKB` | `PatchKB` (`orchestrator/knowledge/agent_kb.py`) |
| section `"framework"` | section `"patch"` (`PatchKB.SECTION`) |

So after fixing the class name the test still failed, first with `NameError`, then `AttributeError: 'NoneType' object has no attribute 'knowledge'` (because `sections.staged("framework")` no longer resolves), then `KeyError: 'framework'`. The diff is 8 lines, all inside your test — the class name, the two `prior_*` ref prefixes, the `staged(...)` lookup, the `value` key, and the three assertions.

`merge_staged_sections` itself was also refactored on main, which matters for what the test is pinning. It used to be:

```python
combined = {...before...}
if before or after:
    combined[ref_key] = list(dict.fromkeys([*before, *after]))
```

and is now seeded from the staged side instead:

```python
combined = dict(staged.knowledge)
for ref_key, before in current.items():
    after = combined.get(ref_key, [])
    ...
    combined[ref_key] = list(dict.fromkeys(str(ref) for ref in [*before, *after] if str(ref)))
```

I re-ran the mutation set against the new shape to make sure the test still catches a regression rather than just passing:

- `after = []` (drop the newly staged side — the silent-data-loss direction) — **red**
- `current = {}` (skip the union loop, drop prior refs) — **red**
- dedup removed — **red**
- `[*before, *after]` → `[*after]` — **red**
- operands reversed — **red**

126 passed, ruff check and format both clean.

One thing I deliberately left alone: the comment above `after_patches` still describes the old `if not before and after:` shape, which no longer exists. The test's behaviour is right, but that comment now points at code that isn't there. Wanted you to word that one rather than guess at it.

The other four branches I updated are unaffected — #1324, #1327, #1328 are all green and #1330 is nearly there.

PR #1332, opened 2026-08-28
Stop discarding a framework's JSON args and its measured accuracy

fengshaoyi-amd:
## What happened

An ATOM GLM-5.2-MXFP4 run (16h budget, TP8/EP1, ISL 8192 / OSL 1024, conc 64, MXFP4 on 8x MI355X) stopped at 6h09m with `stop_reason=conc_sweep_failed`, no accuracy anywhere in `final.json`, and ~9.8h of budget unspent. Three separate hardcoded, vLLM-shaped assumptions each failed silently on a framework whose flags do not match the vLLM spelling.

Every `conc_sweep` launch died like this:

```
openai_server.py: error: argument --online_quant_config/--online-quant-config:
  invalid loads value: '{global_quant_config:ptpc_fp8,exclude_layer:[lm_head,model.embed_tokens,*.mlp.gate,*expert*]}'
```

## Root cause, narrowed to one hop

`remove_server_args` tokenized with a POSIX `shlex.split` and space-joined without re-quoting, so a JSON-valued flag lost its inner double quotes. `_reserialize_json_blobs` on the return path is supposed to recover that via `_repair_unquoted_json`, but that heuristic's bareword pattern (`_JSON_BAREWORD`) excludes `*`, so atom's `exclude_layer` wildcards (`*.mlp.gate`, `*expert*`) could not be re-quoted, `json.loads` kept failing, and the damaged blob was retained verbatim.

Reproduced as a pure-function chain against unmodified `main`:

```
ORIG                     : --online_quant_config '{"global_quant_config": "ptpc_fp8", "exclude_layer": [...]}'
compact_json_server_args : json.loads OK
remove_server_args       : json.loads FAIL   <- the only damage point
compose / dedup / lift   : FAIL (inherited)
```

The other three join points the incident report flagged are safe: `merge_server_args` never splits, and `dedup_vllm_server_args` / `_shell_safe_dedupe` both go through `tokenize_server_args_preserving_json` (`posix=False`, quotes preserved).

**The hop is unconditional.** `compose_server_args` always ends in `strip_benchmark_harness_flags`, whose removal list is non-empty by construction, and `_lift_to_current_best` re-strips both sides before merging. So no operator `remove_args` is needed to trigger it.

**Where it actually bit, per the session artifacts.** The `EXPLORE` variant YAMLs carry the value intact — `{"global_quant_config":"ptpc_fp8","exclude_layer":[...]}`, which `json.loads` accepts — and those variants benchmarked fine. The damage appears once `SWEEP` lifts `current_best` through the strip, and every `conc_sweep` YAML from that point on carries the unquoted form. An earlier revision of this description said all 12 variants were affected; the artifacts do not support that, and the corrected scope is the `conc_sweep` launches.

A/B against the two real trees — the incident's own checkout and this branch — confirms the hop and reproduces the rejected string byte for byte:

| stage | incident tree (`7b50bee1e`) | this branch (`f3884adb9`) |
|---|---|---|
| `compact_json_server_args` | parses | parses |
| `strip_benchmark_harness_flags` | **`{global_quant_config:ptpc_fp8,exclude_layer:[lm_head,model.embed_tokens,*.mlp.gate,*expert*]}`** — identical to the string in `server.log` | parses |
| `compose_server_args` | corrupted (inherited) | parses |

## Changes

**1. `remove_server_args` is tokenized the way the transport is.** Magpie expands `EXTRA_*_ARGS` through an unquoted shell wrapper, so the shell word-splits on whitespace and performs **no quote removal** — a quote byte reaches the server as part of the value. `shlex`, POSIX or not, disagrees with that sink about where the tokens are, and every disagreement showed up as a removal that deleted or rewrote the wrong bytes. Three review rounds each found a different instance:

- a span tracking JSON brace depth could enter a blob that never balanced (`--foo a}`) and run to the end of the string, so removing one flag silently deleted the whole tail of the config;
- quoted operands were compared byte-for-byte, so `remove_args: ['--foo bar']` was a no-op against `--foo "bar"` — a reverse regression against main, in the spelling operators and LLMs write most often;
- `" ".join` reassembly collapsed whitespace inside retained JSON string values, so `{"chat_template": "a  b"}` reached the server as `"a b"`.

`290901c96` replaces the tokenizer with a whitespace split that keeps its separator pieces. Reassembly is then byte-exact for everything retained, comparison runs on a quote-normalized logical value, and **there is no untokenizable path left** — earlier revisions of this PR had a `shlex` `ValueError` fallback that whitespace-split and rejoined lossily, and the PR text claimed it was deleted when it was not. It is deleted now, by removing the reason it existed rather than the branch.

**1a. SEMANTIC NOTE — a pair spec removes the whole operand list.** `remove_args: ['--cuda-graph-bs 1']` against `--cuda-graph-bs 1 2 4` removes all three operands, not just `1`. This is a deliberate widening over main and is called out here because it affects attribution: a gain or regression measured after such a removal belongs to all the operands, not the one the spec named. Deleting `1` alone is not an option — the leftover `2 4` are bare argv words, which `validate_server_args_shell_safe` rejects outright and which the paths that skip it hand to the server as positionals. It is no longer silent: a pair spec matching a multi-operand span logs a WARNING naming the flag and the operand count.

**2. `SPACE_VALUE_FLAGS` deleted rather than extended.** It had no reader left anywhere in the tree (only a definition, an alias, a `_grid_runner` re-export and an `__all__` entry), yet its docstring still promised dedup-time protection for enrolled flags. That stale promise is why atom's `--online_quant_config` looked covered while nothing guarded it. Value shape, decided by parsing, is the property that matters. Note this removes a name whose comment claimed out-of-tree test use; flagging explicitly in case that matters to anyone.

**3. The accuracy gate is unconditional.** `EXPLORE` parsed a variant's eval result only when `is_high_accuracy_risk` matched a list of vLLM/SGLang flag names and `VLLM_*`/`SGLANG_*` env keys by substring. It matched **none** of the 12 ATOM variants: atom spells the knob `--kv_cache_dtype` (underscores, never matching `--kv-cache-dtype`), uses `AITER_*` env vars, and `--online_quant_config` — which changes numeric precision directly, the highest-risk class there is — was never enrolled at all.

`bench-eval-bench` runs the eval on every `warmup_round` regardless, so all 12 scores were already on disk. Confirmed by calling `parse_eval_results` on the winning variant's slot after the fact: `accuracy=0.9689158453373768`, `task=gsm8k`, `metric=exact_match,strict-match`. The predicate's only effect was discarding a number already paid for. `baseline_accuracy=0.9704321455648218` was in `state.json` the whole time, so keeping that precondition is sufficient.

Recovered from disk, GSM8K `exact_match,strict-match` vs baseline 0.9704:

| variant | accuracy | delta |
|---|---|---|
| vendor-proven-stack-mtp3 | 0.9704 | 0.0000 |
| ci-online-ptpc-fp8-nonexpert (winner) | 0.9689 | -0.0015 |
| cudagraph-mode-full | 0.9659 | -0.0045 |
| max-num-seqs-64 | 0.9636 | -0.0068 |

Worst drop 0.0068 against `ACCURACY_THRESHOLD = 0.05`, so unconditional gating would not have rejected a single variant here. The predicate and both lists are removed.

**Blast radius, and what bounds it.** Gating everything means a broken eval path REVERTs a whole round where it used to cost a few variants. Two things bound that. The gate only arms when a reference exists — `baseline_accuracy > 0` for serving — and the baseline is already where a missing accuracy result halts the run, so an armed gate is itself evidence the eval path produced a score at least once this session. Second, the round now reports `accuracy_gate.eval_infrastructure_suspected` when not one gated variant produced a verdict. That is the distinction no single variant can make: its own config breaking its own eval looks identical to the eval path being broken for everyone, until you count the round. It is observability, not a relaxation — every gated variant is still judged fail-closed on its own evidence.

**4. `conc_sweep_failed` is no longer terminal.** The short-circuit sat ahead of `should_reloop_to_explore()`, so one failure ended the whole optimization. conc_sweep is a closeout concurrency scan, not the optimization itself, so its failure carries no evidence about whether the remaining budget could still find gain. It now flows into the normal reloop decision and stays the `stop_reason` when reloop is blocked, so the honest outcome survives without the budget being forfeited too. Reloop evidence gains `sweep_exit_reason` so a downgraded failure is still distinguishable after the fact.

**Known cost, taken deliberately.** A *deterministic* conc_sweep failure now retries once per macro-cycle until the budget, `max_cycles` or R7 convergence stops it, and each retry is a full `FRAMEWORK_AGENT -> KERNEL -> SWEEP` cycle. `8d5b37434` added a cap for exactly this and `f5c0a6016` took it back out: both signals a cap could read here mean something other than what it needs (`phase_history` is capped and resume-restored; OPTIMIZE's `plateau` flag is what an ordinary cycle looks like), and reading either reproduced the original incident from a second branch. The right shape is a durable `SharedState` counter, incremented on failure and reset on success, and it wants its own change. Until then the trade is explicit: budget spent re-searching beats budget forfeited outright, and the JSON fix removes the only trigger this failure mode was ever observed on.

**5. The graded round runs its eval instead of being trusted to have run it** (`06b2de054`). Change 3 above rests on "the score is on disk either way". That was an assumption, not a guarantee: whether the eval runs is resolved from the materialized YAML — base config, reference envs, the variant's own `extra_envs`, the process env — none of which the explore executor controls. A stale config or a variant-supplied `RUN_EVAL=false` would leave no score on disk and fail an otherwise good variant closed as `accuracy_unavailable`. `RUN_EVAL` is now forced on for exactly the rounds the gate reads, set after the variant's own envs so a proposal cannot switch off the eval its own KEEP is gated on. `--no-eval` still wins, since it means no reference was asked for and the gate is skipped entirely.

**6. The measured baseline outranks a proposed one** (`reconcile_baseline_accuracy`). `accuracy_baseline` is offered to the LLM in the action schema, and `explore` used to let it win over `SharedState.baseline_accuracy`. Every in-tree writer only ever copies the state, so a proposed figure that disagrees is a hallucination rather than a second opinion — and change 3 raises what it costs: one bad number used to reach the handful of variants a flag catalogue called risky, and now reaches every variant in the grid. `integrate_patch` is reconciled the same way, since that value is what its KEEP gate grades against. A genuinely re-measured reference still has a channel — the baseline writeback promotes it onto the state — and the log line says so rather than only saying the value was ignored.

**7. An equals-joined flag owns its trailing operands.** The span in change 1 was computed for the space-separated spelling only, so `remove_args: ['--cuda-graph-bs']` against `--cuda-graph-bs=1 2 4` dropped one word and stranded `2 4` as bare argv words — the outcome the operands-as-a-unit rule in 1a exists to prevent, reached through the one spelling that rule did not look at. The shape is malformed at the source (argparse reads the extras as positionals, not as the flag's list) and `validate_server_args_shell_safe` refuses it either way, so nothing launchable regressed; but the invariant is documented unconditionally and was not. The span is computed once now, ahead of both branches. With no trailing operand — the overwhelmingly common case — the end index is `i+1` and the behaviour is byte-identical.

**8. A JSON value's fragments are read as a value, not as option names** (`ce493ec13`). Dropping the depth state in change 1 traded one bug for another. `_span_end` decided where a value ended by asking `_is_flag_token` alone, so a JSON string value carrying a dash-prefixed word ended the span inside itself: `--tmpl {"t":"Answer --now please"}` removed by name left `--now please"}` behind, which is not argv, so `validate_server_args_shell_safe` aborts the launch in `_finalize_framework_server_args` — on a string main removed cleanly. The same blind spot let a fragment be *removed* as if it were a flag: a denylist name inside a string value is matched by `strip_benchmark_harness_flags`, which rides on every compose, and cut the blob in half with no spec asking for it.

The span may now run past an option name while the JSON scan is unbalanced, and only as far as the word that balances it. Nothing balancing it keeps the flag-blind scan, so `--foo a}` still cannot run the span off the end of the string — the bug change 1 dropped the depth state to fix. The removal loop, the spec parser and the widening warning all read one `_words_inside_json`, so the two sides of a comparison cannot disagree about what is a flag, and the warning cannot name a byte of somebody's value as an option. That set clamps a negative depth to closed; carrying it forward would mark every word after a stray `}` as nested and silently disable every removal after it.

## Test plan

- [x] New `test_json_server_args_roundtrip.py` (120 cases at the head) locks the compose -> dedup -> lift chain, asserting on `json.loads` rather than exact strings so any join point that stops re-quoting fails regardless of normalization shape. The three cases added by `4e45fad3f` cover the shared-string case above; all three fail on `f3884adb9` and pass on `4e45fad3f`.
- [x] **Discriminating**: run against unmodified `main`, the suite as first written (17 cases) is 7 failed / 10 passed. The failures are all the ATOM wildcard cases; the vLLM `--compilation-config` cases pass on both because their values have no `*` and the repair heuristic recovers them. That split is exactly why vLLM never hit this and ATOM did.
- [x] `4e45fad3f` checked the same way against its own parent: the server-args / grid-runner / accuracy suites fail on an identical set of 41 pre-existing cases on both commits, zero new and zero incidentally fixed.
- [x] Regression checked by diffing failure *sets* (not counts) against a clean `main` worktree: server-args + accuracy suites 39 vs 39 identical; phase/termination suites 185 vs 185 with zero new; explore suites 45 vs 45. Pre-existing failures are a missing `pytest-asyncio` in the local venv, unrelated to this change.
- [x] `ruff check` clean, `ruff format` applied.
- [x] Updated `test_sweep_closes_on_failed_conc_sweep_even_when_reloop_available`, which locked the old terminal behaviour, and added a case asserting `conc_sweep_failed` still surfaces as the `stop_reason` when reloop is blocked, so the downgrade cannot launder the outcome.
- [x] `06b2de054` checked the same way against its own parent: the accuracy / explore / phase suites fail on an identical set of 3 pre-existing cases on both commits, zero new.
- [x] **CI caught one this enumeration missed.** `test_optimize_loop_walkthrough.py::test_both_arms_dry_walks_the_rest_of_the_chain` failed on `assert 'FRAMEWORK_AGENT' == 'CLOSE'`. It reached CLOSE through the short-circuit change 4 removes; without it, SWEEP consults R1/R7 like every other exit, and the two arms the test dries up are per-cycle while 178 of its 180 minutes remained — so the reloop was the correct answer to the state as written. The setup now also sets `no_gain_cycle_streak`, the field R7 actually reads for "this RUN has converged", and the walkthrough arrives at CLOSE through the convergence wind-down instead — which is also the path that carries `conc_sweep_failed` out as the stop_reason. Chain, reasons and assertions unchanged. Two reasons the sweep above missed it: it enumerates suites by name and this walkthrough is in none of those families, and it is an `asyncio` test, which the same bullet notes could not run in that local venv. CI has both.
- [x] `TestEqualsJoinedOperands` covers change 7: the three removal spellings against `--block-size=16`, a non-matching value left byte-exact, the trailing-operand span, a neighbour removal leaving the equals form verbatim, and the widening warning. The shape matrix gains `("--eqlist=1", "2 4")`, so the four mechanical invariants now run over it too.
- [x] `TestAnOptionLookalikeInsideAValue` covers change 8: the span, the sink's verdict on the result, the lookalike being unremovable in its own right, a neighbour removal leaving the value byte-exact, and both directions of the unbalanced fallback. The shape matrix gains `("--tmpl2", '{"t":"a --b c"}')`, so the four mechanical invariants run over a value carrying an option lookalike too. The file is 8 failed / 112 passed on `b2740b1cf` and 120 passed on `ce493ec13`.
- [x] Full local re-run on the then-head `b2740b1cf`: 767 passed across the twelve suites reaching `remove_server_args` / `compose_server_args` / the phase machine, plus `test_optimize_loop_walkthrough.py` end to end.
- [x] `ce493ec13` checked against its own parent the same way: 35 failed / 325 passed over the five suites reaching this sink, an identical failure *set* on both commits — zero new, zero incidentally fixed. Those 35 are the local venv's missing `pytest-asyncio` and POSIX-only `fcntl`, unrelated to this change. `ruff check` and `ruff format --check` clean.

## Hardware validation

A 12h ATOM GLM-5.2-MXFP4 run on 8x MI355X, same workload as the incident (TP8/EP1, ISL 8192 / OSL 1024, conc 64, MXFP4), on `06b2de054`. Session `20260828T112643Z-4e79d3d5`, 11:26 to 19:51 UTC on 2026-08-28. Baseline 946.8 -> 2699.0 tok/s/GPU, validated gain 185.07%, `crash_count=0`, winning stack `warm_replay -> mtp_spec3 -> mtp_spec3_relaxed -> dp_attention`.

**What this run covers, and what it does not.** It ran on `06b2de054`, three days before `290901c96` replaced `remove_server_args`'s `shlex` tokenizer with the whitespace split change 1 now describes. Fixes 3, 4 and 5 are unaffected — none of that code moved since. Fix 1 is the exception, and change 8 moved it again since: the *outcome* below is real, but the function that produced those intact values is not the function on this branch, so read it as evidence for the fix's premise rather than for its current implementation.

Substituted with a mechanical A/B rather than a second 12h run. `remove_server_args` was driven over 414 `(args, removal-spec)` combinations — 20 token shapes at head / middle / tail, up to seven removal spellings each including one aimed at every neighbour — against a clean `main` worktree, classifying each output by whether `validate_server_args_shell_safe` accepts it:

| outcome | `b2740b1cf` | head `ce493ec13` |
|---|---|---|
| identical output | 237 | 255 |
| main unlaunchable -> this branch launchable | 81 | 81 |
| both launchable, text differs (short options now removable; quoted operands keep their bytes) | 57 | 66 |
| neither launchable, text differs | 12 | 12 |
| **main launchable -> this branch not** | **27** | **0** |

**An earlier revision of this description reported that last cell as 0 over 520 combinations, and it was wrong.** That sweep had 15 token shapes and not one of them carried a JSON value containing a dash-prefixed word, which is the only input the regression fires on — so the matrix could not see it however many combinations it enumerated, and enumerating more of the same shapes would not have helped. Three such shapes are enrolled now (`{"tpl":"Answer --now please"}`, `{"t":"a --b c"}`, `{"a":{"b":"x --y z"},"c":[1,2]}`) and all 27 regressions land on them, nine each, and on nothing else. Change 8 is what takes the cell back to 0, this time on a matrix that can fail.

The 66 text differences are the two intended widenings and nothing else: `-tp` and friends were silently un-removable on main because the scan only recognised `--` names, and a non-matching `--parser 'hermes'` used to come back with its quotes stripped. The 12 in the row above are `--half 'unclosed` and `--cuda-graph-bs=1 2 4`, which main cannot produce a launchable result for either; the count and the shapes are the same before and after change 8, so nothing in that row is this branch's doing.

**Fix 1 (JSON survives) — confirmed.** `exclude_layer` is the complete `["lm_head","model.embed_tokens","*.mlp.gate","*expert*"]` at every layer of `optimization_stack` and in the winning `effective_extra_server_args`; 48 intact occurrences in `state.json`. `invalid loads value` — the string the incident died on — appears **zero** times in the run log and in every `server.log` under the session, so nothing was launched with a corrupted value.

**Fix 3 + 5 (accuracy is recorded) — confirmed, and this is what the run was for.** The accuracy gate fired 10 times with five distinct real scores (0.9704, 0.9651, 0.9500, 0.9492, 0.9484) spanning 11:40 to 19:25, four of them inside the second macro-cycle's `EXPLORE`. All four layers of `optimization_stack` carry an `accuracy` field (0.9704, 0.9704, 0.9492, 0.9484). `accuracy_unavailable` appears zero times, so the failure mode `06b2de054` guards against did not occur.

**Fix 4 (reloop) — partially confirmed; the specific label is still unit-test-only.** conc_sweep did run this time (15:02:46 to 17:10:25, `successful_pairs=2 failed_pairs=6 best_speedup=2.3221`) and `SWEEP -> FRAMEWORK_AGENT (reason=cycle_reloop)` opened a second macro-cycle at 17:11:01. But conc_sweep **succeeded**, so this exercised the success branch, which the deleted short-circuit never guarded. The reordering that matters for `conc_sweep_failed` is therefore still covered by unit tests only. Now that the JSON corruption is fixed there is no natural trigger to observe it, and that is stated rather than papered over.

The run ended 3.5h before its 23:26:43 deadline for `reloop_blocked = all_directions_saturated` (`serving_specialist` at `within_pct=100.0` against a 95% threshold), not for lack of budget. Target was 500% gain against 185% achieved.

## Known issues found while validating, not fixed here

- **`last_profile_args` and `last_profile_workload.server_args` still hold the unquoted form.** Both carry `'{global_quant_config:ptpc_fp8,...}'` in `state.json`. This is record-only in this run: all three profile passes (11:59, 12:45, 19:12) completed and produced traces, and the zero `invalid loads value` count above holds. It is a latent hazard rather than an observed failure — anything that ever replays a launch from those fields would use the broken value — so it wants its own change rather than a late addition here.
- **`stop_reason` reported `time_exhausted` when time was not exhausted.** The real cause was direction saturation, and the phase records hold the honest version (`sweep_budget_cap`, `reloop_blocked=all_directions_saturated`), so no information was lost. Same "report disagrees with reality" family as this incident but with no budget wasted, and no causal link to any fix here; keeping it out so this PR's causal chain stays clean.

## Not changed, and why

The incident report also suggested making `SPACE_VALUE_FLAGS` value-shape-driven (moot, see change 2) and supporting `RUN_EVAL` from `.env`. The latter is by design: `RUN_EVAL` is listed in `BLOCKED_EXTERNAL_ENV_NAMES` because workload/benchmark keys are owned by the CLI flags (`--eval` / `--no-eval`), so honouring it from `.env` would contradict that ownership. The warning is correct; only its wording could be clearer. The warm-replay `REPRODUCED` label inconsistency is a separate subsystem and is left for its own PR.


fengshaoyi-amd:
All 4 blocking findings reproduced locally before touching anything, all 4 fixed, plus the 3 strongly-recommended and the 6 non-blocking. Pushed as `290901c96`, `f5c0a6016`, `3c1af43ab`.

Before the itemised replies, the thing worth saying: I stopped patching two of these areas and changed the approach instead, because the pattern across four rounds says the patches were the problem.

**`_grid_server_args.py` (findings 1, 2, 5, 6, 7) — one root cause, not five bugs.**

`remove_server_args` modelled its input with `shlex` and reassembled with `" ".join`. Neither matches the sink. Magpie expands `EXTRA_*_ARGS` through an unquoted shell wrapper, so the shell word-splits on whitespace and performs **no quote removal** — a quote byte reaches the server as part of the value. Every disagreement between that model and the transport surfaced as a removal that deleted or rewrote the wrong bytes, which is exactly findings 1, 2 and 7, and why finding 5's fallback existed at all.

So the tokenizer is now a whitespace split that keeps its separator pieces, which is not an approximation of the sink but *is* the sink. That single change gives:

- byte-exact reassembly of everything retained → **finding 7** (`{"chat_template":"a  b"}` no longer collapses)
- comparison on a quote-normalized logical value → **finding 2** (`['--foo bar']` now removes `--foo "bar"`)
- span = "up to the next option name", with no JSON depth state that can run off the end → **finding 1** (`'--foo a} --tp 8 --max-num-seqs 64'` → `'--tp 8 --max-num-seqs 64'`)
- no untokenizable path at all, because nothing is left for a quote to unbalance → **finding 5**. You were right that the PR text and the code disagreed; the resolution is that the branch is gone rather than documented. An unbalanced quote is now attributed to its owning option name like any other token, so the `<unparseable>` sentinel is gone too.
- `_json_scan` no longer has a second caller → **finding :148**, along with `_is_flag_token` now being imported from `_canonical_fingerprint`.

**Finding 6** (single-operand spec removes the whole multi-value span): kept, because splitting an operand list leaves bare argv words that `validate_server_args_shell_safe` rejects and the paths skipping it hand to the server as positionals. But your objection about attribution is right and it is no longer silent — a pair spec matching a multi-operand span now logs a WARNING naming the flag and the operand count. Declared in the docstring under `SEMANTIC NOTE` and in the PR body.

**`machine_state.py` (findings 3, 4) — both guards deleted, not fixed.**

You are right that these are not implementation bugs. Both guards read state that cannot carry the meaning they need:

- **Finding 3:** confirmed at `machine_state.py:2731` — `exit_normal_optimize` stamps `plateau: True` on the routine `source_dry and config_dry` end of a cycle. That is what an ordinary cycle looks like, and a new cycle re-seeds both arms. There is no "the arms are permanently spent" signal in the state to gate on, so gating reproduced the incident from a second branch. Deleted.
- **Finding 4:** confirmed — `_PHASE_HISTORY_CAP = 100` and `phase_history` is restored from `state.json`, so the scan is both forgetful in long runs and over-remembering on resume, exactly as you describe. A retry cap divined from a capped, resume-restored log cannot work at any scan bound; noting that the sibling function is bounded was the right catch but the bound was never the fix. Deleted.

`conc_sweep_failed` now never short-circuits the reloop, which is the branch's stated purpose. Capping deterministic retries is still worth doing and needs a durable `SharedState` counter (increment on failure, reset on success); I'll file that separately rather than smuggle a third attempt in here.

**Remaining non-blocking:**

- `_accuracy_gate.py:108` — the direction is now stated at the call site. An unreadable config reports "not disabled", which leaves the gate **on**, so the corrupt-config case keeps grading. That is the direction that cannot ship a regression, but you were right that nothing said so.
- `integrate_patch.py:3207` — `or None` dropped; a legitimate `0.0` baseline survives.
- `explore.py:1266` — confirmed, not just plausible. Scriptable frameworks gate against a fixed `1.0` reference, so the forced eval pass fed nothing. Now `accuracy_gate_enabled and not scriptable`.
- `:220` — cache evicts its oldest entry when full instead of clearing wholesale (which re-armed every option already reported), and keys on payload size as well as option names.
- `:103` — down to one `_reserialize_json_blobs` pass over the args plus one per spec; the trailing re-serialize is gone since the input is normalized before the split.

**On why this took four rounds.** The findings were consistently adjacent edges of the same unspecified contract, and my verification was a matrix I wrote myself from memory of the previous round's bug — so it could only contain cases I had already been bitten by. Neither "quoted operand on the *spec* side" nor "unbalanced brace in a *retained* neighbour" was ever in it. The fix for that is mechanical, not more care: the removal tests now enumerate the token shapes that reach this sink (short options, negative numbers, multi-operand lists, compact JSON, JSON with whitespace, unbalanced braces, quoted operands with and without whitespace, lone edge quotes, double spaces inside JSON strings) and assert invariants over the product at head/middle/tail with every removal spelling. It found two cases on its first run. For the phase machine, findings 3 and 4 were both "reasoned from the consumer, never opened the producer", so any field a guard reads now has to have its writer read and its durability (cap, reset, resume) established first.

Local: 347 targeted tests pass, plus 308 in the adjacent grid-runner/phase files. One pre-existing failure in `test_grid_runner.py::TestSessionBudgetWarmupRounds::test_the_budget_is_re_checked_after_the_uncapped_server_restart` reproduces on `8d5b37434` untouched, so it is not from this change.

xiaofei-zheng (approved):

PR #1334, opened 2026-08-28
refactor: remove the post-KEEP confirmation round

zengleixin-amd:
An `explore` variant and an `integrate_patch` candidate were each benched once
more after they had already been graded, and that second measurement — the
post-KEEP confirmation round — overwrote the first as the reported number.
Both now report the round that graded them. The shared helper the two used,
`executors/_stack_rebench.py`, is gone, and with it the `enable_stack_rebench`
and `rebench_stable_threshold_pct` task params.

**22 files, +89 / −1258.**

## Why, per site

**`explore`** ran its confirmation as a third measurement on the server its
warmup and decision rounds had already warmed. It therefore carried more cache
than the round it overwrote — and because a stable confirmation also advanced
`running_base_tput`, the inflated value became the anchor the next in-batch
variant was graded against, so the bias rode the KEEP chain rather than staying
with one variant. A search now stops at the decision round, which also saves a
full benchmark per KEEP and makes the round's own budget estimate
(`fit_required_sec`) match what it actually spends — it never counted the third
round.

**`integrate_patch`** measured its confirmation on a server of its own, so the
cache argument does not apply there and removing it costs two things worth
stating plainly:

- A patch that cleared the bar on one measurement is no longer asked to clear
  it again before `_git_commit_kept` puts it in the framework tree. The KEEP
  threshold (1.0%, the grid noise floor) is now the only bar.
- `delta_pct` is read off the same measurement that selected the patch. An
  independent re-measure of a selected candidate reads lower than the
  measurement that selected it, so the reported number goes up.

This was removed because it matches the shape being removed — a confirmation
round appended after a KEEP — not because it was measuring the same thing
`explore`'s was.

## Blast radius

**GEAK's same-harness revalidation** (`_enqueue_internal_stack_rebench`)
dispatched an `explore` that inherited the confirmation round by omitting the
flag — the one production caller that did. It now measures like every other
explore, so its throughput meets the engagement (`baseline * 1.02`) and
current-best gates colder. Expect more `fallback` (2a harness replay) and
`no_promote` verdicts from it. A `no_promote` closes that GEAK candidate for
good, so this is a real change in what the kernel phase promotes.

**Removed from the session record:** the `KEEP_UNSTABLE` outcome, the
`keep_unstable_in_stack` result key, and `stack_rebench_tput` /
`stack_rebench_workspace` / `stack_rebench_warnings`. Every consumer read these
through `.get()` or an allowlist membership test, and `writeback` already
collapsed `KEEP_UNSTABLE` into `OUTCOME_REVERT`, so nothing breaks on their
absence. `cumulative_gain_validated` records `e2e_decision_round` as the
measurement basis for explore promotions.

**Kept deliberately:** the breakdown readers for `keep_unstable_count`
(`timeline.py`, `capability_summary.py`, the schema field) and `writeback`'s
`KEEP_UNSTABLE` branch. They read persisted state, including what warm replay
injects, so a session recorded before this change still renders. Only the
producers are gone.

**Not touched:** `_enqueue_internal_stack_rebench` itself. Despite the name it
is not an appended round — it dispatches a whole explore to re-measure the
stack, and it is what produces `cumulative_gain_validated`. The specialists'
`rebench` CLI helper is a different thing that shares the word.

## Validation

Full suite on Linux, this branch vs `main`:

```
main         14845 passed, 58 failed
this branch  14830 passed, 58 failed
```

Failure sets compared item by item — identical; no new failures. The 58 on
`main` are pre-existing. The 15 fewer passes are the deleted tests.

Twelve tests were deleted rather than adapted, because each one's subject was
the confirmation round: the floor-follows-the-anchor test, the unstable-eviction
test, the two threshold-arithmetic pins, the two reaped-confirmation tests, and
six in `test_integrate_patch_coverage_unit.py` driven by mocking
`_confirm_stack_rebench`. Where a test only passed `enable_stack_rebench` to
switch the round off, the line went and the test stayed.

Not covered by tests: whether the reported number moves on real hardware.
`run_grid`'s auto-warmup is disabled under pytest, so the cache effect this
addresses cannot be reproduced in the suite — only which round becomes the
headline can. A real session after merge would settle it.

## Design note lost with the file

`_stack_rebench.py` was the only place recording why the stability floor was
clamped to `keep_threshold / 2`: with a fixed 0.5% band and a
`decaying_keep_threshold_pct` that falls toward 0.1%, the confirmation floor
inverts past the first cycle and rejects variants the same round admitted.
Noting it here so the reasoning survives the deletion.


ZhengGong-amd (approved):
Approved

PR #1338, opened 2026-08-29
Vendor KernelForge in-tree as Hyperloom's built-in kernel-opt agent

xiaofei-zheng:
## Description: what and why

Vendors KernelForge into this repo so that forge becomes Hyperloom's built-in
kernel-opt agent. Installing Hyperloom now gives you forge — there is no
checkout to clone, no `$FORGE_PATH` to export, and no second repo whose version
you have to keep aligned by hand.

The chain this replaces was fragile. `install.sh` carried a comment recording a
2026-07-28 incident where `FORGE_PATH` went unexported, forge raised
`ModuleNotFoundError`, and *every* kernel attempt that day silently reverted.
The Dockerfile needed an SSH key mounted just to clone a private repo. That
whole class of failure is now gone: `kernelforge` is an ordinary package inside
the wheel.

What is deliberately unchanged: the orchestrator's kernel-agent dispatch path,
and forge's standalone CLI.

Landed as P0–P10, one phase per commit, each independently verifiable. Two of
them are worth calling out to a reviewer:

- **`812a61642` (P3) changes no code** — 90k lines reflowed from KernelForge's
  line-length 100 to Hyperloom's 120. `git diff -w` is *not* empty here, because
  reflowing joins lines rather than just re-indenting them; the check that does
  hold is an AST comparison, which reports all 335 Python files semantically
  identical across the commit (the 4 that differ byte-wise after `ast.parse`
  differ only in a docstring gaining a leading space). Skip it.
- **`bcbe1fa74` / `b401b91bc` / `885d4376e` delete a lot** — the unused
  `data/knowledge_base` tree, a `local_knowledge` sync that drops 720 docs to
  213 matching an upstream cleanup, and then the `intellikit` kernel backend
  with its `languages/asm/` tree (see below). Large deletion count, no
  behaviour change for any reachable path.

### Notable decisions

| | |
|---|---|
| Package name | `kernelforge` (top-level, not folded into `hyperloom.*`) — the old `kernel_agents` collided with this repo's own `hyperloom/agents/kernel` |
| CLI | **one** `kernelforge` binary, 4 subcommands. `forge-gemm-tune` is gone as a separate console script; `kernel-agents` survives one release as a deprecated alias |
| History | snapshot copy, no git history |
| Upstream | Hyperloom becomes the sole source |
| `fellow` → `kernel_backend` | a colleague's coinage for what is really the kernel's backend. "kernel backend" in prose, `kernel_backend` in code |
| Backends | 8: CK, FlyDSL, Triton, Gluon, AITER, HIP, hipBLASLt, fusion. `intellikit` did not come across |

### Silent-failure fixes found along the way

The migration surfaced several bugs that produced *no error* — the reason each
one is fixed here rather than deferred:

- `resources.py` used to return a non-existent path instead of raising, so a
  missed data tree meant forge-loop running against an empty knowledge base
  forever. It now raises, and the tests assert a **file-count floor** (`.exists()`
  does not catch an empty directory).
- Three call sites wrote into `site-packages` unconditionally. An autouse
  fixture now fails any test that writes under the installed package.
- `_forge_loop_constant()` swallowed a lookup miss and substituted a default
  wall-clock budget — a missed module rename would have meant forge running on
  the wrong timeout with nobody the wiser. It warns now.
- `49e444804` (the `fellow` rename) had spaced out 12 `kernel_backend` string
  keys into `"kernel backend"`. None of them raise: the playbook lookup fell
  through to `"aiter"` forever and the trace reader handed every op an empty
  backend. Fixed in `d8cacb216`, with a grep guard so it cannot recur.
- `_prepare_worktree()` treated "inside a git repo" as "tracked by it". A
  framework tree hosting a repo that indexes only part of itself produced a
  worktree *without* the kernel, surfacing far downstream as "prepared kernel
  does not exist". Fixed in `ced24080e`.
- Seven shipped `run_example.sh` still passed `--fellow`, a flag `cli.py` no
  longer declares. `forge-loop` is a `TolerantCommand`, so they did not fail —
  they ran an *inferred* backend and looked like they worked. The rename guard
  missed them because its exemption globbed `data/*` instead of `data/*.md`,
  swallowing runnable scripts along with the prose it meant to protect. Fixed
  in `d18bd9f22`, and the narrowed glob caught an eighth site immediately.
- `ensure_rocprof_compute()` was gated on a `$FORGE_PATH` checkout that nothing
  set, so the `forge-profiling` extra never installed and forge profiled on the
  thin PMC path on every pod, silently. Now unconditional, with
  `SKIP_FORGE_PROFILING=1` as the escape hatch — an opt-out rather than an
  opt-in, because an opt-in is what the broken gate effectively was.
- `FORGE_DISABLE_COMPILED_FELLOWS` was renamed, but deleting the old name does
  not retire it: `FORGE_` is on env_safety's dotenv prefix allowlist, so a stale
  value is still forwarded and then read by nothing. That variable switched the
  compiled kernel backends *off*; ignoring it switches them back *on*. It is now
  detected and warned about once per process — refused rather than honoured, so
  the retired vocabulary does not survive.
- Three `except ImportError` sites treated a missing kernelforge as a supported
  configuration and handled it three different ways (info log / bare return /
  nothing). It ships in this wheel now, so that is a broken install: all three
  warn and name the fix.
- `_resolve_serving_patches_root()` logged an override that *resolved* but fell
  through to the packaged tree in silence when it did not, so a mistyped root
  looked exactly like no root and the patches applied were not the ones asked
  for.
- `check_wheel_contents.py` required 3 entries under `serving_patches/` where
  the tree holds 1 — a floor above the truth, which fails a correct wheel. The
  file that must be present is now named, not counted.

### Dead code the vendoring carried in

A vendored snapshot brings along whatever the upstream repo had stopped
calling. An AST pass over `src/kernelforge` — count each top-level name's
identifier occurrences across the tree, subtract its own definition — found ten
that nothing reached, `run_ab()` (the coarse decode A/B, ~160 lines with
`_bench_one_batch_cmd` and `_run_arm`) being the largest. The scan was re-run
after each deletion, because removing a caller exposes the next layer: three
more names surfaced in round two, none in round three. Their five tests went
with them.

The same idea applies to the rename allowlists. An exemption that matches no
tracked line is not inert — it stays on the list and pre-approves whatever later
lands on that path and matches that regex, which is exactly the miss the greps
exist to catch. Three such entries are deleted, and a new guard fails the suite
on any allowlist entry with zero matches, so deleting exempted code now forces
the exemption out too.

Coverage after both: 90.36% against `fail_under = 90`.

### What the first CI run caught that this host could not

Six red checks, six different reasons the local run was green:

- **anthropic.** `pyproject` asks for `anthropic>=0.40`, so CI resolves **1.2.0**
  while this host sits on 0.120.0. 1.x moved to **httpx2** and type-checks
  `http_client` against it, so the `httpx.Client` we handed it was a `TypeError`
  at construction — "llm setup failed" on every discovery call. 1.x also dropped
  `temperature` from `Messages.create()`, whose signature has no `**kwargs`;
  `classify_llm_error` read the resulting `TypeError` as transient, so discovery
  spent its whole retry budget on a call that could never succeed. Both legs now
  ask the SDK for its own `DefaultHttpxClient` rather than picking an httpx, and
  `temperature` travels in `extra_body` when the installed SDK will not name it.
  Verified in a scratch venv on anthropic 1.2.0 + httpx2 2.12.0: 4291 passed.
  openai 3.6.0 (what CI resolves) exports the same class and still names
  `temperature`, so its leg needed only the client change.
- **`test_ci_e2e_dispatch.py` is deleted.** It is KernelForge's contract test for
  *KernelForge's* `ci-e2e-dispatch.sh`, which P9 deliberately did not vendor.
  Hyperloom's same-named script predates this PR and is a different shape. It
  never showed up locally because its skipif wants `jq`, which this host lacks.
- **The mori vendor-playbook test** was the only one in its file reaching
  `_resolve_gpu_target()`, which ends in `rocminfo`. Green on a GPU box, red on a
  runner. It names `GPU_TARGET` now — the test is about the packaged bundle.
- **Gitleaks** flags `script_key="a8w8_bpreshuffle"` as a generic API key. Only on
  `pull_request`, which scans the whole directory; the `push` event scans the diff
  range, which is why every push run was green. Allowlisted, scoped to the tuner
  package and a snake_case literal.
- **CodeQL**: `shape_key()` returned `()` for a shape it could not parse, and all
  three callers unpack into three names — so it only moved the failure out a few
  frames and dropped the shape from the message. `"16x1536"` did not even fail
  there. It raises now; `cli.py` already treats that as "tier3 attempt failed;
  tuning continues".
- **CodeQL, second alert** — clear-text storage of sensitive data at
  `gemm_tune/cli.py:222`, the write of `tier3_outcome.json`. The SARIF taint path
  starts at `Tier3Outcome.trusted`: a **bool** saying whether an operator has
  signed a generated tuner script off. The query's name heuristic reads any field
  containing "trusted" as a secret. The field is `operator_signed` now, which is
  both outside the heuristic and a more accurate name -- what it records is the
  signature, not a trust level. `ledger.is_trusted()` keeps its name; a function
  is not a data node. Nothing reads the JSON key, so the rename is four lines.
- **py3.10.** `test_forge_codex_provider.py` imported `tomllib` bare, and
  `requires-python` is `>=3.10` where it is not stdlib -- a collection error that
  took out all four py3.10 shards and both coverage jobs. It uses the guard this
  repo already has in `test_packaging_lint.py`, with `tomli` from the `ci` extra.

## Linked issue(s)

None.

Both — roughly 250 test files move in, plus new guards for each silent-failure
class above (rename completeness, resource file-count floors, the
write-to-site-packages fixture).

```
ruff check . && ruff format --check .     # 1522 files, clean
pytest -q -n 8                            # 19453 passed, 18 failed
python scripts/check_wheel_contents.py …  # 1835 entries, declared assets present
reuse lint                                # compliant
python -m build --wheel                   # kernelforge/cli.py + data present, tests/ absent
bandit -c pyproject.toml -r src/… …       # advisory; ~27 medium+ in kernelforge, untriaged
```

Docs were repaired in the same pass: 14 sphinx `{doc}` cross-references in
`docs/kernelforge/**` had lost their `/kernelforge/` prefix in vendoring, and
two of them resolved *silently to the wrong page* because Hyperloom has its own
`docs/conceptual/optimization-loop.md`. Example paths now point at
`src/kernelforge/data/examples/`, with a note on reaching the same tree from an
installed wheel. `CHANGELOG.md` claimed `$FORGE_PATH` still worked — it is
removed — and now carries the three BREAKING entries below;
`$KERNELFORGE_PROJECT_ROOT` is documented in
`docs/reference/environment-variables.md`.

The 18 failures are **pre-existing and environment-dependent, not from this
branch**. Verified by running the same 7 files in a clean `origin/main`
worktree: identical 18 failures. They come from this host's `.env` credentials
leaking into tests, a vllm version mismatch, and no TraceLens patch for the
installed vllm.

End-to-end on MI355X / gfx950, Qwen3-14B-FP8: all three forge paths
(gemm-tuning, fusion, kernel-rewrite) run to completion.

## Breaking changes: yes/no

**Yes**, three — all with a clear failure mode rather than a silent one.

1. **`$FORGE_PATH` is deleted.** Nothing reads it. It *is* still forwarded —
   `FORGE_` is a dotenv prefix on env_safety's allowlist — so a stale value
   reaches the child and is then ignored rather than rejected; that is why
   `CHANGELOG.md` calls it out explicitly. The dev override that replaces it is
   `$KERNELFORGE_PROJECT_ROOT`, which is on the exact allowlist and documented
   in `docs/reference/environment-variables.md`.
2. **`forge-gemm-tune` is gone as both a console script and a distribution.**
   The tuner is now the `kernelforge.gemm_tune` subpackage of the same wheel,
   reached as `kernelforge gemm-tune`. There is no subtree left to
   `pip install` on its own.
3. **The `intellikit` kernel backend is gone.** Nothing in Hyperloom could
   reach it: `infer_kernel_backend` has no arm for it, the dispatch path only
   ever passes triton/flydsl/ck/aiter, and `grep -rn intellikit src/hyperloom/`
   is empty — only an explicit `--kernel-backend intellikit` selected it. Its
   author confirms it is no longer needed. Its `languages/asm/` knowledge tree
   (117 files: a vendored `ROCm/intellikit-asm-skills` copy with no LICENSE,
   plus CDNA4 ISA extracts) went with it, reachable from no other backend.
   Unrelated and untouched: `AMDResearch/intellikit`, the profiling toolkit
   Magpie depends on. Same name, different project.
4. **The pre-rename `fellow` vocabulary is no longer accepted** — not as a
   `-fellow` backend suffix, not as a campaign-config key, not as
   `FORGE_FELLOW`. A campaign config carrying the retired key now fails loudly
   at load instead of silently migrating.

## Known follow-ups (deliberate, not oversights)

- `[tool.mypy] files` stays `["src/hyperloom"]`. 90k lines of unannotated code
  in the advisory job is noise. **Bandit** does now cover `src/kernelforge` —
  it is the shell-out-heavy part of the tree, which is where bandit earns its
  keep — and its ~27 medium+ findings (incl. 2 HIGH `B602` in
  `fusion/validate.py`) are untriaged on an advisory job. **Pylint** is
  measured-and-declined, not skipped: adding the tree takes it 0 → ~58, all but
  2 being E1120 against click decorators, and both survivors a known
  `__dataclass_fields__` false positive. The measurement is recorded in
  `lint.yml` so the next person does not have to redo it.
- Two pre-existing bugs found while reading the code are **left alone** —
  they predate this migration and do not belong in a 1047-file PR:
  `tracelens_skill_runner.py:1526` (bogus `task_group.source_path` join), and
  whether `loop.py`'s early exit should collect multiple KEEPs for multi-patch
  fusion.
- **`deploy/` is not carried over.** Every file in the vendored tree targets
  the repository this PR retires — clone URL, runner registration, a cluster
  checkout path — and Hyperloom's own dispatcher only knows `kind:"hyperloom"`.
  Retargeting it is real work and belongs in its own change; the files are
  recoverable from the snapshot commit.
- **Third-party content is now named rather than blanket-claimed as AMD/MIT.**
  `reuse lint` was already green, which was the problem: the `**` annotation
  covers anything nobody listed. Five FlyDSL reference kernels (Apache-2.0),
  SGLang's `mxfp8_grouped_gemm.py` (Apache-2.0), and the SGLang serving patches
  (`Apache-2.0 AND MIT`) now carry overrides, and `THIRD_PARTY.md` records why
  each is in the tree — the question a release review actually asks, which
  `REUSE.toml` cannot answer. **Whether they should ship in the wheel at all is
  still a compliance call for a human before release.**
- `fail_under = 90` is enforced, and P9 fixed the `vars.*` mapping that had
  silently disabled the `COVERAGE_RELAX_FAIL_UNDER` escape hatch. If coverage
  dips after merge, the cause is almost certainly a mistranscribed `omit` glob
  (KernelForge had 24), not genuinely uncovered code. Fix the glob; do not set
  the variable.


ZhengGong-amd (approved):
Approved

Issue #1307, opened 2026-08-27
GEAK handoff uses a throughput reference that may not match the handed-off config


chao-xu-spec:
### Problem
The handoff does read configuration fields from current_best, but current_best is not guaranteed to be a self-consistent measured-config snapshot: its tput may be from a different or richer benchmark configuration than its persisted flags/env/stack.

`geak/handoff.json` sets `orchestrator_best_tput_same_config` from `state.current_best.tput`, while GEAK seeds its baseline from `accepted_flags`, `accepted_env`, and `baseline_env_spec`.

The field is intended to represent Hyperloom throughput for the *same configuration* GEAK launches. However, `current_best.tput` can refer to a richer or different configuration than the flags/env included in the handoff. This produces false GEAK baseline-alignment warnings and makes the cross-harness comparison misleading.

Bug site: `src/hyperloom/orchestrator/phases/kernel.py`

```python
"orchestrator_best_tput_same_config": float((state.current_best or {}).get("tput") or 0.0)
```

### Evidence

| Case | Handoff flags | Handoff reference | GEAK baseline | Reported divergence | Actual same-config result |
|---|---|---:|---:|---:|---|
| GLM-5.2-MXFP4 | DSA-only (`aiter` prefill + `tilelang` decode) | 1670.86 | 1402.73 | **−16.05%** | DSA-only Hyperloom stack_rebench = 1403.43; divergence **−0.05%** |
| Qwen3-0.6B | `--kv-cache-dtype fp8` | 20845.60 | 17971.52 | **−16.19%** | fp8-only Hyperloom stack_rebench = 17981.31; divergence **−0.06%** |

Data sources:
- `session-breakdown/102259`
- `session-breakdown/102313`
- 2026-08-25 to 2026-08-26

### Root Cause Analysis

1. **GLM:** `current_best.tput` is from a full stack containing speculative decoding / graph tuning, but the GEAK handoff contains only DSA flags. The throughput reference is therefore not for the handed-off configuration.

2. **Qwen3-0.6B:** `state.current_best` reports fp8-KV-only flags, but its `tput` is the 20845.60 result from a different qknormrope/compilation variant. The state contains a config-to-throughput mismatch.

GEAK correctly consumes this field as its same-config reference in `GEAK/interface/run_e2e.py`; the incorrect value originates in Hyperloom's handoff/state contract.

### Proposed Solutions

[...]

PR #1341, opened 2026-08-31
Fixes GEAK handoff provenance

chao-xu-spec:
## Description

Fixes current-best GEAK handoff provenance so Hyperloom only presents a same-config throughput reference when it is bound to evidence from the measurement that produced it.

### What and why

- Capture and persist launch evidence with each grid and baseline measurement: requested args/env, recipe digest, actual server-log provenance, warm-reuse metadata, and observed SGLang `ServerArgs`.
- Resolve warm-reused measurements to their own warmup server log only.
- Store an immutable `current_best_measurement` snapshot and distinguish `verified_observed`, `verified_declared_only`, and `unverified` identity states.
- Gate same-config throughput on observed identity verification.
- Preserve launch evidence through `integrate_patch` KEEP promotion and through baseline double-run hot measurements.
- Export the observed identity and measurement evidence in the GEAK handoff while retaining compatibility fields.

This prevents a declared recipe/config identity from being incorrectly treated as proof that the server configuration producing `current_best` throughput was observed.

## Linked issue(s)

- #1307 

### Added or updated

- Grid launch-evidence persistence and warmup-log attribution.
- SGLang `ServerArgs(...)` parsing and observed-vs-declared identity semantics.
- `integrate_patch` benchmark evidence propagation into `current_best_measurement`.
- Baseline evidence persistence and double-run warm-reuse provenance.
- Handoff exposure of structured observed identity and measurement evidence.

### Commands run

- Focused Hyperloom handoff, baseline lifecycle, and promotion regression suites: `175 passed`.
- Ruff checks: passed.
- Diff whitespace validation: passed.

### Offline log replay: 11 archived cases

- Replayed all 11 archived handoff/state artifacts.
- Every archive was classified as `legacy_unverifiable`: declared flags, environment, overlay, and throughput fields remained internally consistent, but the old artifacts lacked an immutable measurement identity and observed launch evidence.
- This verifies the new safety behavior: legacy data is no longer allowed to claim observed same-config verification merely from declared configuration fields.

### Online E2E validation: Qwen3-14B

- Hyperloom current-best rebench measured `1664.542 tok/s`; the GEAK Setup baseline measured `1656.670 tok/s`, a `-0.473%` difference.
- Hyperloom, GEAK Setup, and GEAK validation server logs produced identical observed SGLang identities, including TP, context length, effective memory fraction, FP8 KV cache, attention backends, and radix-cache settings.


src/hyperloom/orchestrator/loop/writeback.py
```diff
@@ -4319,6 +4342,238 @@ async def replay_for_resume(self) -> dict[str, Any]:
...
+        model_path = str(launch_evidence.get("model_path") or "").strip()
+        try:
+            from sglang.srt.server_args import ServerArgs
+
+            parser = argparse.ArgumentParser(add_help=False)
+            ServerArgs.add_cli_args(parser)
+            tokens = shlex.split(raw_args)
+            if model_path and not any(token == "--model-path" or token.startswith("--model-path=") for token in tokens):
+                tokens = ["--model-path", model_path, *tokens]
+            namespace, _unknown = parser.parse_known_args(tokens)
```
xiaofei-zheng, line 4552: **Blocking — `SystemExit` is not an `Exception`.**

`ServerArgs.add_cli_args` types the options, so `parse_known_args` on a malformed value (`--mem-fraction-static high`, `--tp-size abc`) calls `parser.error()` -> `sys.exit(2)` -> `SystemExit`, which derives from `BaseException` and slips past this `except Exception`. It then propagates out of `_resolved_sglang_server_config` and `_stamp_current_best_measurement` into `_lift_to_current_best` / `_promote_baseline`, which have already assigned `shared_state.current_best` but not yet returned — so the promotion is left half-applied with `current_best_measurement` unset.

Catching `BaseException` here (or parsing without argparse) would contain it.

Separately: in real sglang, `ServerArgs.__post_init__` reads the HF model config and probes the GPU. That is heavy, side-effecting work to run inside the promotion path — see the note on the critical section below.

src/hyperloom/orchestrator/loop/writeback.py
```diff
@@ -4319,6 +4342,238 @@ async def replay_for_resume(self) -> dict[str, Any]:
...
+        )
+        out: dict[str, Any] = {}
+        for field_name in fields:
+            value = getattr(resolved, field_name, None)
+            if isinstance(value, (str, int, float, bool)) or value is None:
+                out[field_name] = value
+        return out
+
+    @staticmethod
+    def _measurement_observed_server_identity(measurement: Mapping[str, Any]) -> dict[str, Any]:
```
xiaofei-zheng, line 4584: **The server-log re-parse below this is unreachable.**

`isinstance({}, Mapping)` is `True`, so an *empty* `observed_server_identity` returns here and never reaches the `server_log_path` / `actual_server_log_path` fallback. Both `_attach_grid_launch_evidence` and `_attach_baseline_launch_evidence` set `evidence['observed_server_identity']` unconditionally, defaulting to `{}` (log not yet flushed when evidence was attached, or `_launch_argv_from_log` returning empty) — so for every measurement this PR produces, the fallback is dead code and the measurement is classified `verified_declared_only` with the `ServerArgs` line sitting in the log on disk.

Guard on truthiness rather than on the isinstance check.

src/hyperloom/orchestrator/loop/writeback.py
```diff
@@ -4319,6 +4342,238 @@ async def replay_for_resume(self) -> dict[str, Any]:
...
+        if not marker:
+            return ""
+        paths = [str(measurement.get("server_log_path") or "").strip()]
+        if isinstance(evidence, Mapping):
+            paths.append(str(evidence.get("actual_server_log_path") or "").strip())
+        for key in ("benchmark_workspace", "workspace"):
+            workspace = str(measurement.get(key) or "").strip()
+            if workspace:
+                root = Path(workspace)
+                paths.extend((str(root / "server.log"), str(root.parent / "server.log")))
```
xiaofei-zheng: **This probe escapes the measurement's own directory.**

`Path(workspace).parent / "server.log"` re-introduces the cross-run attribution that `_measurement_server_log_path` explicitly promises never to do. For an `integrate_patch` lift, `measurement['benchmark_workspace']` is the executor workspace (e.g. `/benchmark`), so the parent probe resolves to `/server.log`. For a baseline whose workspace is the run root, `root.parent` is the shared `runs/` directory — a sibling run's `server.log`, a completely different launch's argv, gets scraped, published as this measurement's `resolved_server_launch_flags`, hashed into `launch_identity`, and shipped as `baseline_env_spec.config.server_launch_flags`.

src/hyperloom/orchestrator/loop/writeback.py
```diff
@@ -4319,6 +4342,238 @@ async def replay_for_resume(self) -> dict[str, Any]:
...
+            return "verified_declared_only"
+        return "unverified"
+
+    def _stamp_current_best_measurement(self, evidence: Mapping[str, Any] | None = None) -> None:
+        """Attach the exact promoted config identity to ``current_best``."""
+        cb = self.shared_state.current_best
+        if not isinstance(cb, dict):
+            return
+        evidence = evidence if isinstance(evidence, Mapping) else {}
+        launch_evidence = evidence.get("stack_rebench_launch_evidence") or evidence.get("launch_evidence") or {}
```
xiaofei-zheng: **These four keys have no producer.**

`grep -rn stack_rebench_launch_evidence src/` matches only these read sites. `stack_rebench_launch_evidence` / `_workspace` / `_server_log_path` / `_launch_evidence_path` are four dead `or` branches at the front of every lookup, implying a stack-rebench evidence contract that does not exist — the rebench path goes through `_promote_geak_from_candidate`, which supplies no evidence at all. Either wire the producer or drop the keys.

src/hyperloom/orchestrator/loop/writeback.py
```diff
@@ -4319,6 +4342,238 @@ async def replay_for_resume(self) -> dict[str, Any]:
...
+                or ""
+            ),
+            "server_log_path": str(
+                evidence.get("stack_rebench_server_log_path")
+                or evidence.get("server_log_path")
+                or launch_evidence.get("actual_server_log_path")
+                or ""
+            ),
+            "launch_evidence": launch_evidence,
+            "launch_evidence_path": str(
```
xiaofei-zheng: **Duplicated I/O inside the promotion critical section.**

This computes `_measurement_launch_flags` (opens and scans server logs), and then `build_env_spec()` below recomputes it on the same measurement dict, plus `_launch_recipe_digest` (full recipe read) and `_geak_overlay_digest` (reads `_overlay_manifest.json` and sha256s every referenced kernel body). On a large overlay that is seconds of blocking I/O added to the section `test_promote_shared_state_lock.py` exists to protect — and with the `ServerArgs.__post_init__` GPU probe noted above, more still.

Compute `server_launch_flags` once and pass it into `build_env_spec`.

src/hyperloom/orchestrator/loop/writeback.py
```diff
@@ -4357,16 +4612,20 @@ def build_env_spec(self) -> dict[str, Any]:
...
         """
         from ..source_snapshot import source_layer_overlay_dir, source_layer_reproducible
 
+        cb = self.shared_state.current_best if isinstance(self.shared_state.current_best, Mapping) else {}
         materialized = self._current_best_launch_config()
-        stack = [e for e in (getattr(self.shared_state, "optimization_stack", []) or []) if isinstance(e, dict)]
+        # current_best's embedded stack was promoted with its tput. Reading the
+        # mutable global stack here could attach an unrelated source patch after
+        # a resume or partial state write.
+        stack = [e for e in (cb.get("optimization_stack") or []) if isinstance(e, dict)]
```
xiaofei-zheng, line 4754: **Reading the stack from current_best silently drops source patches.**

`_promote_baseline` never sets `optimization_stack` on the current_best it builds, and neither do state files written before this PR. When `cb.get('optimization_stack')` is absent, `source_snapshots` is `[]` even though `shared_state.optimization_stack` holds `scope=source_patch` entries — so `baseline_env_spec` tells GEAK to reproduce the baseline from a stock framework tree.

The empty list is also what `_handoff_launch_identity` hashes on both the stamp and handoff side, so the mismatch reports as `verified` instead of being caught. The global-stack read that was removed here was the only thing surfacing those layers; consider falling back to it when the embedded stack is missing.

src/hyperloom/orchestrator/loop/writeback.py
```diff
@@ -4426,8 +4677,28 @@ def build_env_spec(self) -> dict[str, Any]:
             },
             "source_snapshots": source_snapshots,
             "overlay_pythonpath": materialized.get("final_overlay") or "",
+            "overlay_digest": _geak_overlay_digest(str(materialized.get("final_overlay") or "")),
+            "base_launch_recipe": str(getattr(self.shared_state, "baseline_config_path", "") or ""),
+            "base_launch_recipe_digest": self._launch_recipe_digest(
```
xiaofei-zheng, line 4810: **Hashing the recipe path invalidates a still-valid identity.**

`base_launch_recipe` is the materialized recipe *path*, and it is folded in alongside the content digest. A later baseline re-run (revalidation, warm replay) points `shared_state.baseline_config_path` at a new timestamped YAML but skips the current_best re-stamp when `optimization_stack` is non-empty — so at handoff this hashes to H2 != H1, `identity_matches` is `False`, and `orchestrator_best_tput_same_config` drops to 0.0 for a byte-identical config.

Hashing the digest alone would avoid the path half of this.

xiaofei-zheng (commented):
Reviewed the provenance rework. The direction is right — binding the same-config reference to evidence from the measurement that produced it is the correct fix for #1307. Three things, though, keep the PR from enforcing its own stated invariant, and one is a state-corruption risk rather than a metadata issue.

**Blocking**

1. **`SystemExit` escapes the promotion transaction** (`writeback.py:4428`). `_stamp_current_best_measurement` runs after `shared_state.current_best` is assigned but before `_lift_to_current_best` returns. `parse_known_args` raises `SystemExit` — a `BaseException` — on a malformed value, which `except Exception` does not catch, leaving `current_best` written, `current_best_measurement` unset and the promotion half-applied.

2. **The gate does not actually gate** (`kernel.py:837`). `verified_declared_only` still emits `same_config_reference_status: "verified"` and a non-zero throughput, and the declared-only test (`requested_server_args is not None`) is satisfied by every evidence dict the producers write, since both attach helpers set the key unconditionally (possibly to `""`). The PR's stated invariant — that a declared identity is not proof the server configuration was observed — is not enforced on either field.

3. **The GEAK promotion path carries no launch evidence.** `_promote_geak_from_candidate` (`kernel.py:1615`, and the same shape at `kernel.py:3318/4077/4324`, `prelude.py:2189`) lifts current_best with `{name, candidate_extra_server_args, extra_envs, final_overlay, workspace}` — no `launch_evidence`, `launch_evidence_path` or `server_log_path`. The stamp is empty, the status falls to `unverified`, and the next macro-cycle hands GEAK `orchestrator_best_tput_same_config = 0.0`, discarding the same-config reference in exactly the iterate-on-GEAK case this PR exists to serve.

Worth noting the failures run in both directions: several paths downgrade genuinely-observed measurements to declared-only (the dead log-reparse fallback, the 256-line cap), while the gate itself still lets declared-only through as verified. The Qwen3-14B E2E passes because it exercises the one path where both happen to line up.

**Also dead / unwired**

- `_scrape_resolved_launch_flags` (`coordinator_helpers.py:1892`) lost its only production caller in this PR; `grep -rn _scrape_resolved_launch_flags src/` now hits only `test_coordinator_helpers_launch_scrape_unit.py`. ~100 lines plus a test file describing behavior the orchestrator deliberately no longer performs.
- The four `stack_rebench_*` keys read in `_stamp_current_best_measurement` have no producer anywhere in the tree — see the inline note.

Remaining inline comments cover correctness details, duplication between the two evidence helpers, and I/O added to the promotion critical section.

src/hyperloom/orchestrator/phases/kernel.py
```diff
@@ -735,9 +735,63 @@ async def _run_geak_kernel_phase(self, *, from_phase: str) -> None:
...
+            launch_evidence.get("requested_server_args") is not None
+            or launch_evidence.get("requested_server_env") is not None
+            or launch_evidence.get("recipe_digest")
+        ):
+            reference_verification_status = "verified_declared_only"
+        else:
+            reference_verification_status = "unverified"
+        reference_verified = reference_verification_status != "unverified"
+        observed_identity = str(measurement.get("observed_launch_identity") or "")
+        if not observed_identity and identity_matches and observed_flags:
```
xiaofei-zheng: **This duplicates `_identity_verification_status` + `_observed_launch_identity`, and has already diverged from them.**

The observed hash is re-derived only when `observed_flags` is set, while `WritebackCollaborator._observed_launch_identity` also accepts an observed `ServerArgs` map. A measurement carrying only `observed_server_identity` — the warm-reuse / archived-ServerArgs case this PR adds, or a state file written by an older stamp — has empty `observed_flags`, so this branch is skipped and the handoff publishes `same_config_reference_verification_status: "verified_observed"` next to `same_config_reference_observed_identity: ""`. A consumer keying off the observed hash sees no proof for a record labelled observed.

Calling the two existing helpers instead of re-implementing them here would keep the two sites from drifting further.

src/hyperloom/orchestrator/phases/kernel.py
```diff
@@ -778,9 +833,22 @@ async def _run_geak_kernel_phase(self, *, from_phase: str) -> None:
             # Orchestrator throughput of the SAME config GEAK seeds its baseline
             # with, so run_e2e can compute a pure measurement divergence. 0.0 =>
             # no accepted config yet (falls back to raw baseline downstream).
-            "orchestrator_best_tput_same_config": float((state.current_best or {}).get("tput") or 0.0)
-            if isinstance(getattr(state, "current_best", None), dict)
-            else 0.0,
+            "orchestrator_best_tput_same_config": same_config_tput,
+            "same_config_reference_status": "verified" if reference_verified else "unverified",
```
xiaofei-zheng, line 858: **Blocking — `verified_declared_only` is still published as `verified` with a live throughput.**

Two problems compounding:

1. The declared-only test at line 773 (`launch_evidence.get('requested_server_args') is not None`) is true for every post-PR measurement — `_attach_grid_launch_evidence` writes `requested_server_args` unconditionally, possibly as `""`.
2. `verified_declared_only` sets `reference_verified = True`, so `orchestrator_best_tput_same_config = measurement['tput']`, and legacy consumers reading only `same_config_reference_status` see plain `"verified"`.

So a warm-reused measurement whose server log was never found is handed to GEAK as a verified same-config reference — which is the situation the PR description says it prevents. The observed/declared distinction needs to reach both `same_config_reference_status` and the throughput field, not just the new `*_verification_status` key.

src/hyperloom/orchestrator/loop/coordinator_helpers.py
```diff
@@ -1741,6 +1742,134 @@ def _launch_argv_from_log(path: str, marker: str) -> str:
     return ""
 
 
+_SGLANG_SERVER_ARGS_LOG_RE = re.compile(r"\bserver_args\s*=\s*ServerArgs\s*\(")
+_SGLANG_SERVER_ARGS_MAX_CHARS = 512 * 1024
+_SGLANG_SERVER_ARGS_MAX_LINES = 256
```
xiaofei-zheng: **Silent give-up truncates real logs.**

256 lines / 512KB is short for Hyperloom's `server.log`: it is the Magpie subprocess log, carrying the `+ python -m sglang.launch_server ...` trace, env dumps and torch/ROCm init noise before sglang's own `server_args=ServerArgs(...)` line. Once that line lands past the cap, this returns `{}`, `observed_server_identity` is empty, and a genuinely observed measurement is classified `verified_declared_only` — the exact distinction the PR is drawing.

Worth both raising the cap and logging on the give-up path; as written the downgrade is invisible.

src/hyperloom/orchestrator/actions/executors/_grid_runner.py
```diff
@@ -2717,6 +2718,7 @@ def _skip_rest_for_budget(idx: int, *, spent_on: str, rounds_left: int = variant
                 reported_success=measurement.get("reported_success"),
                 returncode=rc,
                 nonfatal_warnings=warnings,
+                server_log_path=_existing_log_path(server_log) or _existing_log_path(workspace / "server.log"),
```
xiaofei-zheng: **Recomputed and overwritten a few lines later.**

This evaluates `_existing_log_path(server_log) or _existing_log_path(workspace / 'server.log')` — two stat calls — and then `_attach_grid_launch_evidence` calls `_measurement_server_log_path(primary_log, workspace, slot=slot)`, which redoes exactly those two checks plus the warmup glob and assigns the result over this field.

src/hyperloom/orchestrator/actions/executors/_grid_runner.py
```diff
@@ -2789,6 +2797,139 @@ def _existing_log_path(path: Path) -> str | None:
...
+                    "warmup_round"
+                    if actual_log and str(actual_log).startswith(warmup_prefix + os.sep)
+                    else ("caller_ready_server" if caller_reused_ready_server else "fresh_or_unobserved")
+                ),
+                "source_server_log_path": actual_log or "",
+            },
+        }
+        result.launch_evidence = evidence
+        try:
+            slot.mkdir(parents=True, exist_ok=True)
```
xiaofei-zheng: **Writes evidence for variants that never launched.**

A budget stop (`_skip_rest_for_budget`) or a `capability_unsupported` fast-fail appends a `VariantResult` whose slot was never created. This loop still runs `slot.mkdir(parents=True, exist_ok=True)` and writes a `launch_evidence.json` whose `warm_reuse.reused_ready_server` is `caller_reused_ready_server` and whose provenance is `caller_ready_server` — asserting a warm server reuse for a measurement that does not exist, and filling `runs/` with `variant_NN_*/launch_evidence.json` directories for variants the grid explicitly refused to launch.

Skip results whose status is `skipped` or a pre-launch `failed`.

src/hyperloom/orchestrator/actions/executors/baseline.py
```diff
@@ -286,6 +286,100 @@ def _config_framework(config_path: Path | str) -> str:
     return str((cfg.get("benchmark") or {}).get("framework") or "").strip().lower()
 
 
+def _attach_baseline_launch_evidence(
```
xiaofei-zheng, line 322: **This is `_attach_grid_launch_evidence`'s body copied.**

The config YAML parse, the `server_args_env_name` split of requested args/env, the sha256 recipe digest, framework resolution, the `_launch_argv_from_log` / `_observed_sglang_server_identity_from_log` dispatch, the 12-key `evidence` literal and the persist-to-`launch_evidence.json` block are all duplicated verbatim from `_grid_runner.py:2832`.

The two have already diverged on `warm_reuse` — `'warmup_round' in Path(actual_log).parts` here vs. a `startswith(slot/warmup_round + os.sep)` prefix test there — so a schema change or a fix to one will silently leave the other behind. A shared `build_launch_evidence(config_path, actual_log, framework, slot)` would collapse both.

(Both helpers also `read_text()` the config for the YAML and then `read_bytes()` it for the sha256 — read the bytes once and decode.)

chao-xu-spec:
@xiaofei-zheng 
Addressed the provenance review findings.
Changes:
- Made current-best measurement stamping transactional: declared-config parsing now contains `SystemExit`, and `current_best` plus `current_best_measurement` are committed only after the measurement identity is fully built.
- Tightened the same-config gate so only `verified_observed` evidence publishes a verified same-config throughput to GEAK. Declared-only evidence remains explicitly classified and publishes zero same-config throughput.
- Propagated launch evidence, evidence sidecar paths, and server-log provenance through the GEAK 2a promotion path.
- Restored empty observed-identity fallback to the measurement’s own server log, removed parent-directory log probing, and raised bounded SGLang `ServerArgs` log scanning to cover real Magpie logs.
- Centralized baseline and grid launch-evidence construction, and prevented skipped or pre-launch-failed variants from materializing synthetic evidence.
- Removed obsolete launch-flag and stack-rebench evidence branches.

Validation completed successfully:
- Focused handoff, baseline lifecycle, promotion, grid, and GEAK sweep regression suites: 285 passed.
- Ruff, formatting, compilation, and diff whitespace checks: passed.
- Archived E2E provenance replay across 5 bundles: 373 real server logs parsed successfully; observed-vs-declared classification, empty-identity recovery, legacy safety, and evidence reconstruction all passed.
- Live GPU 2a validation: verified executable `final_launch.sh` selection, successful measurement, persisted `launch_evidence.json`, observed SGLang identity, and promotion evidence propagation.
- Live production baseline validation: Coordinator-dispatched SGLang baseline completed at 4000.917 tok/s; `current_best_measurement` carried persisted launch evidence, an existing server log, and `verified_observed` identity status.

The review invariants are now enforced by both implementation and regression coverage.

xiaofei-zheng (commented):
Re-reviewed at `4405183`. 11 of the 13 findings from the previous round are addressed — the `SystemExit` escape, the gate that published `verified_declared_only` as verified, the missing GEAK-promotion evidence, the unreachable log fallback, the `stack_rebench_*` keys, the dead `_scrape_resolved_launch_flags`, the 256-line cap, the parent-directory log probe, evidence for skipped variants, the duplicated evidence builders, and the duplicated I/O in the promotion path. The extraction into `_launch_evidence.py` / `launch_log_evidence.py` and the rewrite of the `ServerArgs` parser onto balanced-paren `ast.parse` both read better than what I suggested.

**One retraction.** My note on `build_env_spec` reading the embedded `optimization_stack` was wrong — details inline. Your reply was correct and no change is needed there; apologies for the noise if you had already started on it.

**Two small things left**, neither blocking: the recipe *path* is still hashed into the handoff identity, and the observed-identity hash is still gated on `observed_flags` alone while the status test now also accepts `observed_server_identity`. Both are a couple of lines, both inline.

(Posting as a review rather than as thread replies — EMU policy blocks the reply endpoint for me, so these are anchored to the lines instead of nested under the original threads.)

src/hyperloom/orchestrator/loop/writeback.py
```diff
@@ -4398,16 +4662,20 @@ def build_env_spec(self) -> dict[str, Any]:
...
         """
         from ..source_snapshot import source_layer_overlay_dir, source_layer_reproducible
 
+        cb = self.shared_state.current_best if isinstance(self.shared_state.current_best, Mapping) else {}
         materialized = self._current_best_launch_config()
-        stack = [e for e in (getattr(self.shared_state, "optimization_stack", []) or []) if isinstance(e, dict)]
+        # current_best's embedded stack was promoted with its tput. Reading the
+        # mutable global stack here could attach an unrelated source patch after
+        # a resume or partial state write.
+        stack = [e for e in (cb.get("optimization_stack") or []) if isinstance(e, dict)]
```
xiaofei-zheng, line 4754: **Retracting my earlier note on this line — I misread the guard.**

`_promote_baseline` only assigns `current_best` when the stack is empty:

```python
if anchor_accepted and not (getattr(self.shared_state, "optimization_stack", None) or []):
```

So the current_best it produces genuinely *was* measured on an empty stack, and an empty `source_snapshots` is the correct answer rather than a silently dropped layer. My claim that it "never sets `optimization_stack`" was wrong in the way that mattered. Reading the embedded stack rather than the mutable global one is the right call and the comment you added explains why — please disregard the suggestion to add a global-stack fallback.

The pre-PR-state-file half of the original note is technically still reachable (an old `current_best` carries no embedded stack while `shared_state.optimization_stack` is populated), but `legacy_unverifiable` already covers that conservatively, so there is nothing to do.

src/hyperloom/orchestrator/loop/writeback.py
```diff
@@ -4360,6 +4393,232 @@ async def replay_for_resume(self) -> dict[str, Any]:
             "verdicts_seen": len(verdicts),
         }
 
+    @staticmethod
+    def _handoff_launch_identity(env_spec: Mapping[str, Any]) -> str:
+        """Return a stable identity for the complete GEAK baseline launch."""
+        config = env_spec.get("config") if isinstance(env_spec.get("config"), Mapping) else {}
+        payload = {
+            "base_launch_recipe": str(env_spec.get("base_launch_recipe") or ""),
```
xiaofei-zheng: Still open from the last round: the recipe *path* is hashed alongside its digest.

```python
payload = {
    "base_launch_recipe": str(env_spec.get("base_launch_recipe") or ""),   # <- this line
    "base_launch_recipe_digest": ...,
```

A baseline re-run (revalidation, warm replay) repoints `baseline_config_path` at a new timestamped YAML but skips the current_best re-stamp when `optimization_stack` is non-empty, so a byte-identical config hashes to a different identity and `orchestrator_best_tput_same_config` drops to 0.0. Dropping this line leaves the digest, which already expresses everything the path stood in for.

The failure is conservative — a lost reference, never a wrong throughput — so this is not blocking. One consequence worth deciding on first: changing the payload invalidates every identity stamped under the old shape. Within a process that is invisible (stamp and handoff move together), but a state file resumed from an older build mismatches once and falls to `unverified` until the next promotion re-stamps it, including the 11 archives you replayed. That is consistent with how this PR already treats legacy data, so I would take the one-time churn — but a payload version field would avoid it if you would rather not.

src/hyperloom/orchestrator/phases/kernel.py
```diff
@@ -739,9 +739,63 @@ async def _run_geak_kernel_phase(self, *, from_phase: str) -> None:
...
+            str(launch_evidence.get("requested_server_args") or "").strip()
+            or bool(launch_evidence.get("requested_server_env"))
+            or str(launch_evidence.get("recipe_digest") or "").strip()
+        ):
+            reference_verification_status = "verified_declared_only"
+        else:
+            reference_verification_status = "unverified"
+        reference_verified = reference_verification_status == "verified_observed"
+        observed_identity = str(measurement.get("observed_launch_identity") or "")
+        if not observed_identity and identity_matches and observed_flags:
```
xiaofei-zheng: The classification is consistent now — both sites use the same `.strip()` / `recipe_digest` test, and `observed_server_identity` participates in the status decision. This branch did not follow:

```python
if identity_matches and (observed_flags or observed_server_identity):
    reference_verification_status = "verified_observed"
...
if not observed_identity and identity_matches and observed_flags:   # <- hash only when flags exist
```

A log carrying `server_args=ServerArgs(...)` but no `+ python -m sglang.launch_server` argv trace — the two come from different writers, so the second is not guaranteed — yields `verified_observed` next to `same_config_reference_observed_identity: ""`. The throughput is still published, so this is a missing proof field rather than a wrong value. Widening to `(observed_flags or observed_server_identity)` lines it up with the status test above.

chao-xu-spec:
@xiaofei-zheng Both remaining findings are addressed in `b6c3d90f`. The
`optimization_stack` retraction needed no change, so I left that as it was.

**Recipe path in the handoff identity.** Dropped `base_launch_recipe` from the
`_handoff_launch_identity` payload and kept the digest, taking the one-time churn
as you suggested. Reproduced first: stamp on `baseline.<ts1>.yaml`, re-point
`baseline_config_path` at a byte-identical `baseline.<ts2>.yaml`, and the identity
no longer matches the stamp, so `orchestrator_best_tput_same_config` falls to 0.0.
Now pinned by `test_measurement_identity_survives_recipe_repath_with_identical_content`,
next to the existing content-change test so both directions are fixed.

`env_spec` is unchanged — `base_launch_recipe` and the `launch_recipe` alias are
still exported — so this is a hash-input change only.

One edge this slightly widens: `_launch_recipe_digest` returns `""` on `OSError`,
so two unreadable recipes now hash alike where the path used to separate them. The
other payload fields still discriminate, so I left it rather than expand scope —
happy to make the identity return `""` on an empty digest if you prefer.

**Observed-identity hash gate.** Widened to
`(observed_flags or observed_server_identity)`, matching the status test above it
and `_observed_launch_identity`, which already used the wider condition; the
payload already carried `observed_server_identity`. Also reproduced first —
`verified_observed` next to an empty `same_config_reference_observed_identity` —
and now covered by `test_handoff_hashes_observed_identity_from_server_args_alone`.

Worth noting the live path fills `observed_launch_identity` via the correct helper,
so the narrow branch was mostly reached when the stamp was absent; the two tests
disagreeing was the real defect.

xiaofei-zheng (approved):

PR #1349, opened 2026-08-31
refactor(kb): three-column Recipe SDK with recorded apply roots for warm replay

meinali-566:
## Summary

Reworks the Recipe KB into three columns, each with a dedicated SDK facade, and makes the **recorded apply root the sole authority** for warm replay — replacing the env/allowlist root search that could apply a patch to a tree the gain was never measured on.

### 1. One facade per published column (`24053b815`)
- Consolidates the old five-column layout (`config`, `explore`, `framework`, `kernel`, `patch_timeline`) into three: `config`, `patch`, `kernel`.
- Each column has a single read+write owner: `ConfigKB`, `PatchKB`, `KernelAgentKB`.
- Deletes `patch_timeline`; replay order is now the lexicographic order of zero-padded overlay refs (`patch/overlays/<6-digit stack>/<2-digit member>-<name>.patch`), so there is no separate timeline to keep in sync.
- Adds a `host_origin` sanitizer exemption: the one subtree allowed to carry absolute host paths (secrets are still stripped there).

### 2. Each overlay applies into the checkout it was recorded against (`8e5c1ba19`)
- Write side records `provenance[].host_origin.apply_roots` (`{ref: absolute_root}`) per KEEP.
- Read side stamps each patch entry's `framework_root`.
- `_apply_warm_patches` places each overlay into its own recorded tree, in timeline order, snapshotting every touched tree before writing any, and rolling back **all** touched trees when a single overlay fails (fail-closed, never partial). Nested roots (`/sglang` and `/sglang/python/sglang`) collapse to the git work-tree root so `git apply` does not silently no-op.
- Promote validates every tree's `root == snapshot_manifest.repo_path` or rejects the whole replay.

### 3. Recorded root is the only authority; env/allowlist search deleted (`62c34f09c`)
- A record that cannot name its checkout is treated as broken: warm replay skips it whole rather than searching for a substitute tree.
- The `kernel` column now records `host_origin.apply_root` per fusion/rewrite item (from `kernel_repo` / `last_deploy_repo_root`, else derived by removing the diff's declared tail from the absolute target).
- Removes `resolve_warm_replay_framework_root`, `resolve_warm_replay_kernel_root`, `_resolve_warm_replay_patch_root`, the two allowlist root builders, `warm_replay_patch_sources`, and `_resolve_recipe_patch_target`. `resolve_source_file_allowlist` stays — it answers a write-side "may this be edited" question with many other callers.

### 4. Review fixes (`2f140278d`)
- **Adoption lineage.** `_adopt_replayed_prior` re-homes each replayed prior overlay's `apply_root` onto its new ref (a provenance row under the replay stack index). Previously an inherited overlay published rootless, and because replay is fail-closed on missing roots, the *next* generation skipped the whole Recipe — sinking the current generation's own well-rooted overlays as collateral. An ancestor that was already rootless stays rootless; nothing is invented.
- **Kernel publish degrades per item, not per session.** A fusion/rewrite KEEP whose checkout cannot be named is now dropped from the Recipe (its just-staged patch is discarded so it leaves no orphan) instead of raising and taking `config`, `patch`, and the still-rooted kernels down with it. Publishing it rootless would instead poison the combined replay, so dropping is the safe middle. A genuinely missing/unmaterializable patch still fails closed.
- **Realized-diff harvest gates on a real commit.** `integrate_patch` only harvests `HEAD^..HEAD` when `_git_commit_kept` actually advanced `HEAD` (empty note). A no-op commit (`nothing to commit`, or no patch-touched paths) leaves `HEAD` on the previous KEEP, so harvesting there published the *previous* KEEP's diff as this KEEP's realized change; it now falls back to the delivered patch (`realized=False`).

## Test plan

- [ ] `ruff check src/hyperloom/` (passing locally)
- [ ] `pytest src/hyperloom/inference_optimizer/tests/{test_column_kb,test_remote_recipe_v2,test_promote_shared_state_lock,test_framework_paths_units,test_warm_patch_apply,test_warm_replay,test_prelude_warm_kernel_kb,test_integrate_patch_helpers_unit}.py` (passing locally)
- [ ] Full `pytest src/hyperloom/inference_optimizer/tests/` regression — **not yet completed**, please run in CI
- [ ] Confirm on a real session that a stacked (2nd-generation) replay republishes with `apply_roots` covering both the inherited and the new overlay, and that an unrooted kernel KEEP degrades (item dropped, rest of the Recipe still published) rather than aborting.

## Follow-up (out of scope)

Bugbot separately flagged that `harvest_realized_diff` (new in `24053b815`, not pre-existing) can omit in-root artifact paths that were never committed, so a `realized=True` overlay may still drop side-effect files that live under the root but outside the committed pathspec. That is a distinct gap from the no-op-commit harvest fixed here and is left for follow-up.


ZhengGong-amd (approved):
Approved

PR #1331, opened 2026-08-28
Add SBD V6 startup and Framework Agent timeline

luochen-amd:
## Summary
- add the additive SBD V6 metadata and outcome projection while preserving V5 output
- persist ordered install and model-gate timeline events for fresh, resumed, and failed startup attempts
- add the merged Framework Agent timeline that replaces the former separate Explore stage
- preserve durable Framework discovery, authoring, integration, reauthor, and Critic evidence

## Scope
- V6 metadata and outcome
- install timeline
- model-gate timeline
- Framework Agent timeline
- top-level V6 close placeholder

## Framework Agent timeline
- projects configuration work into `config_arm`, including specialist suggestions, rounds, variants, measurements, and decisions
- projects source work into `source_arm`, including candidate discovery, authoring/reauthoring, patch integration, gates, and outcomes
- excludes Kernel, Enablement, and other non-Framework specialist activity
- records real discovery failures from production task/result shapes
- persists normalized Critic reviews outside the prunable scratch workdir and deduplicates legacy fallback evidence
- propagates `reauthor_attempt` through specialist, integration, progress, and exported timeline records
- aligns with the removal of stack rebench and maps legacy `KEEP_UNSTABLE` evidence to `REVERT`

## Review fixes
- do not create diagnostic sessions for expected preflight `SystemExit` or `KeyboardInterrupt` paths
- isolate unexpected resume-preflight failures in a new failed-attempt session instead of mutating the requested session
- store startup events only in the append-only `reports/sbd_v6/timeline/` ledger and remove read-side migration writes
- package the complete timeline history and durable `write_warnings.jsonl` ledger
- surface SBD V6 writer failures through `metadata.warnings`
- write fail-fast model-gate breakdowns once instead of exporting twice
- reuse shared collector helpers and Framework constants
- propagate structured `macro_cycle` context into durable Critic iterations instead of parsing prompt text
- document the four additive V6 top-level keys and add a changelog entry

## Compatibility
- keeps the existing V5 schema version and fields unchanged
- treats all V6 collection and persistence as best-effort telemetry
- does not change optimization decisions or runtime outcomes
- does not add token or monetary cost reporting

## Validation
- `46 passed` in the SBD V6 focused suite
- `17 passed, 4 deselected` in session-package tests; deselections are Windows-only symlink/FIFO cases
- `168 passed, 2 deselected` in Coordinator/Critic tests; deselections require the real Critic subprocess runtime or optional `claude-agent-sdk`
- Ruff passed
[...]

PR #1350, opened 2026-08-31
Add SBD V6 baseline/sweep/conc_sweep/kernel timeline and close

luochen-amd:
## What

PR #1331 landed 3 of the 14 SBD V6 timeline stages: the two durable ones (`install`, `model_gate`) and the `framework_agent` projection. The top-level `close` key was hardcoded to `{}`.

This adds five more units: **`baseline`, `sweep`, `conc_sweep`, `kernel`** (timeline) and **`close`** (top-level key).

## Why these need no durable event stream

`install` writes to disk because it runs before `session_dir` exists; `model_gate` because it fail-fasts before the state machine starts. All five units here run *inside* the Coordinator, so their data is already in `state.json` / recorder fragments / `reports/`. They are pure projections at export time:

- `_EVENT_TYPES` (`session/sbd_v6.py:24`) is untouched
- no V6 event writer is added
- under `orchestrator/`: the close-section refresh described below with the package rebuild it gates (best-effort, after the sequencer's last step), and a `fusion_run_id` stamped on the fusion result so its integrate verdict can be paired with it (see *Review round 3*)

V6 stays purely additive telemetry: collection runs on an isolated `v6_warnings` list surfaced only through `metadata.warnings`, so a V6 failure can never change a run's outcome or the V5 payload.

## Changes

| File | |
|---|---|
| `collectors/v6_stages.py` *(new)* | `project_baseline_event`, `project_sweep_event`, `project_conc_sweep_event`, `project_kernel_events` |
| `collectors/v6_close.py` *(new)* | `collect_v6_close` |
| `collectors/v6.py` | `collect_v6_timeline` gains optional keyword sections; `_phase_windows` (extracted last PR) reused for KERNEL_AGENT visits; the existing sort is unchanged |
| `collectors/_common.py` | `_KERNEL_PHASES` and `_operation_task_id` hoisted so both modules share them |
| `collectors/__init__.py` | re-exports |
| `exporter.py` | passes the already-computed V5 sections down, adds a V6-only `collect_geak` call, moves the `metadata["warnings"]` snapshot **below** the close collector, and adds `patch_breakdown_close` |
| `orchestrator/phases/close.py` | patches the `close` key as the sequencer's last act, and rebuilds the artifact package when that changed something |
| `recorder/instrument.py`, `schema.py` | fold the kernel-level `proposal` / `verification` / `best_artifact_path` onto the backend attempt that was adopted, so the verdict reaches the journey at all |
| `orchestrator/phases/kernel.py` | stamps `fusion_run_id` on the fusion result and on its integrate verdict, so the two can be told apart from a previous round's |

`kernel` emits one event per macro-cycle visit, carrying the five candidate lanes (`kernel_rewrites`, `fusion_runs`, `gemm_tuning_runs`, `collective_runs`, `geak_runs`) plus the unified final-rebench `attempts[]` ledger, each bucketed onto the visit that produced it.

## No-fabrication rules

A stage with no evidence produces **no event**, rather than an empty `skipped` shell. Fields the design names but that nothing persists are emitted as `null`, each with a comment saying why the nearest neighbour would be wrong — notably `sweep.plan.grid_source` and `sweep.input_anchor.*`, which `current_best` cannot supply honestly.

Ownership is never inferred from an action name: `integrate_patch` lands Kernel rewrites, Framework enablement and Explore config alike (`attribution.py:327`), so `_owns_kernel_attempt` decides on `agent`/`phase` and only falls back to a kind table for old rows carrying neither.

The same principle — ownership is never read off an attribute that several rows share — decides which of a kernel's backend attempts speaks for the kernel. Every attempt carries the same `kernel_id`, so the kernel-level `e2e` verdict, its best micro speedup, the integrated patch and its identity go to the adopted attempt alone — the one the recorder stamped `best_artifact_path` on, falling back to V5's own `(micro_speedup, ts)` ordering for sessions recorded before that stamp existed. A losing attempt keeps its own `micro_decision` and answers only to its own `rewrite_id`.

Rebench attempts are linked to lane candidates by identity (`kernel_id` / `run_id` / `task_id` / …). An attempt naming no subject is absorbed by a lane holding exactly one row, since there is nothing to be ambiguous with; one naming a kernel that matches no candidate is **never** absorbed, however few candidates are on offer. Where two rows of the same lane are indistinguishable, no edge is written and a warning is raised, because a wrong edge reads as a candidate having been validated when it was not. The conflict test is namespace-aware: `kernel_rewrites` / `collective_runs` are keyed on kernel identity, the other three on run ids, so a fusion rebench naming the kernel it patched is not a rival claim against a fusion run.

Producer vocabularies are normalized onto the closed enums rather than passed through — a backend attempt succeeds as `completed`, forge-fusion as `ok`, and `kernel_journey` rolls a kernel up as `adopted`, none of which are enum members. An unlisted spelling maps to the conservative member and raises a warning, so drift surfaces instead of being coerced.

One subtlety worth flagging: `collect_baseline` returns `0.0`, not `None`, for a session that never measured, so a *positive* throughput is the only proof a baseline exists. Without that guard every empty session grew a phantom `baseline` event.

## `close` is written twice, and the first pass is deliberately partial

`session_breakdown` is CLOSE step 2 (`orchestrator/phases/close.py:340`); `langfuse_flush` / `artifact_package` / `ndjson_drain` / `done` are recorded *after* it, and once `close_sequence_done` is true the CLI finally block skips `write_breakdown_json` entirely. The snapshot written at step 2 therefore describes the close-out only as far as itself, and honestly reports `degraded`.

The sequencer's last act now calls `patch_breakdown_close(session_dir)`, which recomputes **only** the `close` key against the finished `state.json` and splices it back in — the same read → recompute-one-key → atomic `mkstemp` + `os.replace` idiom as the existing `patch_breakdown_langfuse`. `_record_close_step` persists `state.json` on every step, so the sequence is complete on disk by that line. The call sits after `stop_reason` and `close_sequence_done` are settled, returns `bool`, and swallows everything; it cannot affect a run. A breakdown with no `close` key (a V5-only payload) is left untouched rather than grown a new one.

**Patching the session copy is not delivery.** `session_breakdown.json` is `PACKAGE_GLOBS[0]`, so the artifact package built at step 5 bundles it into the zip *and* the loose tree, and that package is what external sync ships. The patch therefore has to be followed by a rebuild — gated on the patch having changed something and on the first packaging having produced a path.

It is a rebuild rather than a reorder because the close section cannot be complete until `artifact_package` has an outcome to report: that is a fixed point, not an ordering bug, and the langfuse splice above documents the same invariant from the other side. Replacing the single member surgically would cost the same anyway (stdlib zip has no raw member copy, so the archive is rewritten regardless) and would leave `PACKAGE_MANIFEST.json` describing the old file. No close step is recorded for the rebuild: it rewrites the path `artifact_package` already names, and recording it would strand the bundled copy one step behind again.

So a healthy session now reports `succeeded` with all seven steps in every copy that ships, and `degraded` means what it should: the record of the close-out is incomplete — what a session killed mid-sequence leaves behind. Note that `langfuse_flush` only records a step *on failure*, so its silence is success; the module docstring says so, since absence of a step is not evidence against it.

`fact_finalize` is emitted by the sequencer but was missing from the field design's `step` enum. The gap is in the contract, not the producer, so the doc was corrected and the collector accepts it; genuinely unknown step names pass through unchanged with a warning.

## Review round 3: three lanes reported an outcome the session's own records contradict

**Backend rewrites were reading fields a real attempt does not have.** `run_attempt` returns `status` / `optimized_path` / `error`; the verdict is the result's top-level `proposal.decision` and the gates its top-level `verification`, and only `micro_speedup` was ever back-filled onto the adopted attempt. So on production data every rewrite reported `micro_decision: SKIPPED` with null gates, and `artifact_path` fell back to the attempt's own stdout log. `KernelJourneyEntry` carries no kernel-level verification, so the data was already gone before V6 saw it — the fix is in the recorder, which now folds `proposal` / `verification` / `best_artifact_path` onto the attempt named by `verification.best_attempt_id` and no other. An empty `best_attempt_id` means nothing was adopted, so nothing is stamped; a failed attempt with no verdict of its own now reports `FAILED` rather than `SKIPPED`. That stamp also repairs `_adopted_attempt`, whose `best_artifact_path` probe could never fire on real data before.

**A GEAK run's `status: ok` says the runner finished, not that the candidate was adopted.** A run the final rebench reverted was reported `KEEP` beside the attempt that rejected it. `outcome` now comes from the writeback's `revalidation_status` (`no_promote` / `no_material` / the failures — the field is negative-only; a promotion writes a `geak_e2e` stack entry and a rebench attempt instead, so its absence is not a KEEP) or from the linked final rebench. A run nobody adjudicated is `NEEDS_REVIEW`.

**`last_fusion` is overwritten every run; `last_fusion_integrate` only when integration runs** — which a fusion that was not kept never reaches, and one missing its patch or target returns before. A new run therefore inherited the previous round's verdict. Both now carry a `fusion_run_id` and V6 pairs on it, falling back to the patch the verdict names for sessions recorded before it; an unpaired verdict is dropped with a warning rather than reported. With no paired verdict, a micro `candidate` is `NEEDS_REVIEW` — KernelForge keeping a fusion is a microbenchmark result, and adoption is the e2e re-baseline's call. Collective campaigns get the same ladder: `kept` with no `integration_decision` is `NEEDS_REVIEW`.

Uniformly: a candidate produced but unadjudicated reports `NEEDS_REVIEW`, and `_settle_pending_outcomes` upgrades **only** such rows from a linked rebench that actually measured something. A candidate already carrying an explicit e2e verdict keeps it — two verdicts disagreeing is a fact worth seeing, not one to paper over by preferring whichever ran last.

Out of scope, flagged for follow-up: `knowledge/remote_recipe/values.py:586` `build_kernel_fusion_value` pairs `last_fusion` with `last_fusion_integrate` with the same blind fallback, so the KB write-back can attribute an old e2e gain to a new fusion. Pre-existing, and not V6.

## One deviation from the plan

The plan said the existing `test_framework_timeline_*` tests must stay green without modification. Five of them changed: they asserted over the *whole* timeline, and a state that walks the macro loop now legitimately produces `kernel` events too. A recorded KERNEL_AGENT phase transition **is** evidence the stage ran, and suppressing it to keep the assertions literal would fabricate absence — so the tests were narrowed with a new `_framework_events()` helper instead. No production behaviour was bent to fit a test.

## Testing

- `tests/test_sbd_v6_stages.py` *(new, 77 tests)* — happy paths for all five units, no-evidence cases, per-visit and per-lane kernel isolation, identity-based rebench linking with its ambiguity and named-conflict warnings, adopted-attempt ownership of a kernel's E2E verdict, agent/phase ownership of `integrate_patch`, the status ladders, the enum normalizers and the warnings they raise on drift, mixed-event ordering, a parametrized additivity test that monkeypatches each projector to raise and asserts the V5 payload plus top-level `warnings` are unchanged, six tests for `patch_breakdown_close`, and three that read the delivered `session_breakdown.json` back out of the zip and the loose copy (including the stale-bundle regression and the manifest).
- Round-3 regressions, in the same file: two drive the **real recorder → assembler → V6 chain** on a production-shaped forge result (attempts carrying only `status` / `optimized_path` / `error_type`, verdict and gates at the top level) and assert the winner gets the verdict with full gates and the rewrite as its artifact while the loser gets `FAILED` and its own stdout log, plus one pinning that an empty `best_attempt_id` stamps nothing. Then GEAK reverted-by-rebench / kept-by-rebench / each `revalidation_status` / unadjudicated, and fusion stale-verdict (must warn, must not be `KEEP`) / paired / pre-id patch-path pairing / micro-kept-without-integrate, and the collective pending-integration ladder.
- `tests/test_close_phase_sequencer.py` — one new test drives the real Coordinator through the CLOSE transition and asserts the packaged copies carry the post-step-2 sequence.
- `test_sbd_v6_initial.py` back to exactly its pre-existing failure set.
- Wide run (`-k "breakdown or v6 or geak or kernel or exporter or collector"`): before/after failure set byte-identical.
- `ruff format --check .` and `ruff check src/hyperloom/` clean.
- End-to-end on a synthetic session: the step-2 snapshot reports `degraded` with 4 steps; after the sequencer's refresh it reports `succeeded` with 7 steps and `close_sequence_done: true`, with every other key in the payload byte-identical.

Note for reviewers on Windows: 27 tests fail with `ModuleNotFoundError: No module named 'fcntl'` (POSIX-only, via `recipe_kb/local_store.py`). Pre-existing and unrelated; it also prevents `test_close_phase_sequencer.py` from being collected locally, so that file's coverage of the new call site relies on CI.

## Not in this batch

`warm_start`, `warm_replay`, `enablement`, `roofline`, `profile`, `discovery`, `kb_write_back` — deferred to a follow-up.


luochen-amd:
Thanks — all seven hold up. I reproduced each against source before touching anything, and every one was real. Pushed as `611110ea5`.

---

### [P1] #1 — CLOSE stuck on `degraded`

You're right that this is two defects, and the second one is the worse of the two.

**`sequencer_started` never settles.** It's recorded once as `running` at `close.py:228` and never revisited (`grep -c` on the file returns exactly 1). My `unsettled` check treated it as a step still in flight, which made `succeeded` unreachable *by construction* — no write-order fix would have helped. That's why your repro stayed `degraded` with full steps and `close_sequence_done=True`. It's now excluded via `_MARKER_STEPS`, with a comment saying why rather than just listing it.

**The write order.** Adding `patch_breakdown_close` (`exporter.py`), called as the sequencer's last act. It reads the breakdown, recomputes only the `close` key against the finished `state.json`, compares, and splices — the same idiom as the existing `patch_breakdown_langfuse` at `exporter.py:854`, down to the atomic `mkstemp` + `os.replace`. What makes this safe rather than a race: `_record_close_step` persists `state.json` on *every* step (`close.py:775`), so by the time `done` is recorded the whole sequence is already on disk.

The call site is 8 lines, after `stop_reason` and `close_sequence_done` are settled, returns `bool`, never raises. A payload with no `close` key — i.e. V5-only — returns `False` untouched, so this cannot grow a key on a surface that never carried it.

End-to-end on a synthetic session: step-2 snapshot `degraded` / 4 steps → after refresh `succeeded` / 7 steps / `close_sequence_done: true`, every other key in the file identical.

### [P1] #2 — `integrate_patch` collected as Kernel data

Correct, and the codebase already says so in as many words: `attribution.py:327` documents that `integrate_patch` lands Kernel rewrites, Framework enablement *and* Explore config, which is exactly why `instrument.py:224` routes it through `patch_author(result)` instead of reading the name. My filter did the thing the comment warns against.

Took your suggested shape: `_owns_kernel_attempt()` decides on `agent == "kernel_agent"` or `phase == "KERNEL_AGENT"` when either field is present, and only falls back to the kind table for old rows carrying neither. `_KERNEL_ATTEMPT_KINDS` is demoted from an ownership rule to a candidate filter. Note `optimizations.attempts[].agent` is already pre-resolved by `_attempt_agent` through recorded-value → kind-table → `patch_author`, so the good path is a direct read, not a re-derivation.

### [P2] #3 — `succeeded` with every attempt failed

Fixed. The ladder now separates *concluded* attempts from faulted ones; a rebench-only stage where every attempt is `status=failed` reports `failed`, not `succeeded`.

### [P2] #4 — `final_rebench_attempt_ids` full join

Right, and your note about the tests is the real finding — one lane row per `source_kind` in every fixture is exactly what hid it.

Linking is now by lane identity (`kernel_id`/`kernel_name`/`rewrite_id` for rewrites, `run_id`/`task_id`/`source_file` for fusion, and so on). Three cases: identity match wins; a single-row lane still absorbs leftovers, since there's nothing to be ambiguous *with*; and two indistinguishable rows in the same lane produce **no** edge plus a warning. I deliberately chose to withhold rather than guess there — a wrong edge reads as a candidate having been validated when it wasn't, which is worse than a missing one. Fixtures now carry multi-row lanes.

### [P2] #5 — sweep `skipped` alongside `stop_reason=sweep_failed`

Fixed. A hard failure with zero points is `failed`. Ladder now: some ok + any failure → `degraded`; ok only → `succeeded`; failures or a failing `stop_reason` with nothing kept → `failed`; genuinely nothing → `skipped`.

### [P2] #6 — `{"speedup": null, "error": "succeeded"}`

Fixed, and finding the fixture bug behind it was the useful part. `_conc_pair_error()` now names the failing arm and reads that arm's own point for the reason, e.g. `"optimized: server OOM at conc=16"`.

Worth recording why this was easy to miss: `conc_pair_comparison` (`common/gain_math.py:35`) is an outer join, so a missing arm yields `{}` and a `None` status. My fixture had a row with a baseline throughput but no baseline point and no `baseline_status`, which the real producer never emits. The fixture was rebuilt to match actual output — both arms stamped on every comparison row — and only then did the test exercise the case you described.

### [P2] #7 — `fact_finalize` outside the enum

The sequencer does emit it, so the gap is in the contract, not the producer. Added it to the field design's `step` enum with a note that the runtime is authoritative here, plus a note on `sequencer_started`'s status. The collector keeps a `_KNOWN_STEPS` set: recognized names pass silently, genuinely unknown ones pass through **unchanged** with a warning, so future drift surfaces instead of being coerced into the enum.

---

**Ruff.** My mistake — I'd run `ruff check` and CI gates on `ruff format --check .`, a different check. `ruff format` applied; both clean now.

**Tests** 31 → 50. Verified: `test_sbd_v6_stages.py` 50/50; `test_sbd_v6_initial.py` back to exactly its pre-existing 27 `fcntl` failures; wide-suite before/after diff contains removals only.

One gap I should flag rather than paper over: `test_close_phase_sequencer.py` can't be collected on Windows (`fcntl`, via `recipe_kb/local_store.py`), so I could not run it against the new call site locally. I read its assertions instead — it pins the exact 7-name `close_steps` list, and this change adds no step — and `py_compile`'d both changed runtime files. CI is the real check there.

luochen-amd:
All five hold up. I reproduced each against source before changing anything, and two of them turned out to be wider than reported. Pushed as `8df1e8cad`.

---

### [P1] #1 — the fix never reached the consumer

Correct, and this is the one that mattered: last round's patch landed on the copy nobody reads. `session_breakdown.json` is `PACKAGE_GLOBS[0]`, so it goes into the zip *and* the loose tree, and step 5 had already bundled the step-2 snapshot by the time step 6 patched the session dir.

The sequencer now rebuilds the bundle when the patch changed something. Three ways to fix it, and why this one:

- **Reorder** (patch before packaging) loses `artifact_package`'s own step and path from the delivered section — the close section cannot be complete until packaging has an outcome to report. That is a fixed point, not an ordering bug.
- **Surgical member replacement** costs the same as a repackage anyway (stdlib zip has no raw member copy, so the whole archive is rewritten) and additionally leaves `PACKAGE_MANIFEST.json` describing the old file.
- **Rebuild.** Gated on `patch_breakdown_close()`'s bool return *and* on the first packaging having produced a path, so an unchanged section or a failed package costs nothing. Deliberately records **no** close step for the rebuild: it rewrites the path `artifact_package` already names, and recording it would strand the bundled copy one step behind again — the same regress by a different route.

Tests, as asked — `test_sbd_v6_stages.py` now reads the delivered copies directly:

- `test_the_delivered_package_carries_the_finished_close_section` — asserts `close.status == "succeeded"` with all seven steps in **both** `zipfile.ZipFile(...).read("session_breakdown.json")` and the loose copy.
- `test_a_package_built_before_the_patch_ships_the_step_two_snapshot` — the regression guard, pinning exactly the state you found: session copy `succeeded`, both delivered copies `degraded` with no `artifact_package`.
- `test_the_delivered_manifest_describes_the_rebuilt_bundle` — the reason it is a repackage and not a member swap.

Plus `test_the_sequencer_delivers_the_finished_close_section_in_the_package` in `test_close_phase_sequencer.py`, which drives the real Coordinator through the transition and reads the zip. That file can't be collected on Windows (`fcntl`), so CI is its first real run — flagging that rather than implying I ran it.

### [P1] #2 — one kernel's backends sharing one verdict

Right, and V5's precedent is exactly where you pointed: `collectors/kernels.py:205-233` resolves `verification.best_attempt_id` → `best_backend` → highest micro speedup, and stamps the kernel decision onto that attempt alone.

One wrinkle: those hints are *consumed* inside the V5 collector and do not survive onto `KernelJourneyEntry`. What does survive is `best_artifact_path`, which is only ever written onto the adopted row. So `_adopted_attempt` reads that first and falls back to V5's own `(micro_speedup, ts)` ordering — same identity, reconstructed from what reaches the projector.

Only the adopted attempt now carries the E2E `outcome` and the kernel's identity when rebenches are linked; a losing attempt answers to its own `rewrite_id` alone and reports `outcome: SKIPPED` (or `FAILED` if it failed on its own terms) with `reason: "not the adopted backend attempt for this kernel"`. Its `micro_decision` is still its own — that part of the story is true of it.

Beyond the report: `verification.micro_speedup` had the same defect. It fell back to the kernel-level `micro_speedup`, which is the *best* attempt's number, so a backend that never compiled was reporting the winner's speedup. That fallback is now the adopted attempt's alone.

I chose to keep the losing attempts rather than emit only the best one — they are the evidence that other backends were tried, and `micro_decision` + `execution_status` still say what happened to each. What they no longer do is claim the outcome.

### [P2] #3 — a lone lane absorbing another kernel's rebench

Agreed, and the test pinning that behaviour was mine to remove. The fallback is now for attempts that name *no* subject; one that names a kernel matching no candidate stays unlinked and warns with the kernel it named.

One correction to the shape of the fix. My first attempt keyed the conflict test on "does this attempt carry a source identity", and it broke a legitimate link: `kernel_rewrites`/`collective_runs` are keyed on `kernel_id`, but `fusion_runs`/`gemm_tuning_runs`/`geak_runs` are keyed on run ids, so a fusion rebench that recorded the kernel it patched was being read as a rival claim against the visit's only fusion run. The check is now namespace-aware — only a lane keyed on kernel identity can be contradicted by an attempt naming a kernel — with `test_a_fusion_rebench_naming_its_kernel_is_not_a_conflict` guarding it.

### [P2] #4 — all-skipped reporting `succeeded`

Fixed. `concluded` was `status != "failed"`, which counts a skip as a conclusion. Only a `succeeded` attempt is a measurement now: all-skipped reports `degraded` when candidates exist and `skipped` when nothing was built either. `_final_rebench_attempt` already normalizes its status to `succeeded|skipped|failed`, so this is a predicate change, not a new normalization.

### [P2] #5 — production spellings against closed enums

Confirmed against the producers, not just the doc: backend attempts record `completed`, forge-fusion `ok`. Ordinary success paths, and a strict consumer would reject the payload over them.

`_lane_status` / `_lane_outcome` / `_micro_decision` / `_candidate_decision` now normalize every enum-typed field in the lanes, including the two `micro_decision` vocabularies the field design keeps separate (`KEEP|REVERT|PARTIAL|FAILED|SKIPPED` for `kernel_rewrites`, lowercase `candidate|no_improvement|failed|skipped` for the run-level lanes). `tuner_attempts.status` keeps its fourth member `empty`, which is the one lane whose enum has it.

An unlisted spelling maps to the conservative member **and** raises a warning — same idiom as `_KNOWN_STEPS` last round, so vocabulary drift surfaces instead of being silently coerced.

Beyond the report, `outcome` had the same problem from the other side: `kernel_journey`'s rollup vocabulary (`adopted` / `attempted` / `dispatched` / `discovered`) was being uppercased straight into a closed enum with no such members, so a perfectly healthy adopted kernel emitted `outcome: "ADOPTED"`. It is now translated through a table, and `PARTIAL` — legal as a `micro_decision`, not as an `outcome` — maps to `NEEDS_REVIEW`.

---

**Tests** 61 → 64 in `test_sbd_v6_stages.py`, plus 1 in `test_close_phase_sequencer.py`. Also fixed a fixture of my own that the new warnings caught: `_journey_kernel` gave a backend attempt `decision: "candidate"`, which is the *fusion* vocabulary — `KernelBackendAttempt.decision` is `KEEP|PARTIAL|REVERT|FAILED` (`instrument.py:2345-2380`). The same class of fixture error that hid #6 last round; the normalizer surfaced this one immediately, which is the argument for warning rather than coercing.

**Verified:** `test_sbd_v6_stages.py` 64/64; `test_sbd_v6_initial.py` back to exactly its pre-existing 27 `fcntl` failures; wide before/after diff (`-k "breakdown or v6 or geak or kernel or exporter or collector"`, 215 pre-existing failures/errors on Windows) byte-identical; `ruff format --check .` and `ruff check src/hyperloom/` clean.

luochen-amd:
All three issues were valid and have been fixed in `117e2351f`. Details follow.

---

### #1 — Backend rewrites did not handle the real recorder payload

Confirmed. Reproducing the path you described produced exactly `micro_decision: "SKIPPED"`, gates whose values were all null, and an `artifact_path` pointing to the stdout log.

The fix belongs in the recorder, as suggested. `KernelJourneyEntry` itself does not carry the kernel-level `verification` or `proposal`; that data had already been lost before reaching V6, so changing only `v6_stages.py` could not recover it.

`instrument.py` now resolves the kernel-level facts before iterating over attempts and applies them only to the winner named by `verification.best_attempt_id`:

- `proposal.decision` → that attempt's `decision`
- `verification.compile_passed` / `correctness_passed` / `correctness_source` → the corresponding fields on that attempt, without overwriting values already recorded by the attempt
- `verification.best_artifact_path` → the new `best_artifact_path` field; all remaining attempts receive `""`

Design choices:

- **Use `verification["best_attempt_id"]` instead of `_best_attempt_id()`.** The latter has a speedup fallback that could assign a kernel-level `REVERT` to an arbitrary row when compilation failed. An empty `best_attempt_id` means nothing was adopted, so no attempt should be decorated. It is safer to leave the fields empty than invent an association. `micro_speedup` still uses the existing helper, so that behavior is unchanged.
- **Add `best_artifact_path` instead of reusing `optimized_files`.** A real backend run's `optimized_path` is `optimized/{attempt_id}_stdout.log`, not the rewritten artifact. The two fields have different meanings, now documented in the schema docstring.
- This also fixes `_adopted_attempt`: it previously looked for `best_artifact_path`, but production attempts never contained that field, so adoption selection always depended on the speedup fallback.
- A failed attempt without a verdict now reports `FAILED` instead of `SKIPPED`.

The integration test uses the real path rather than constructing attempt rows by hand: `record_kernel_backend_result` → `assembler.assemble_parts` → `collect_v6_timeline`. The input matches production data, where the attempt carries only `status`, `optimized_path`, and `error_type`, while verdict and gates live at the top level. It asserts that the winner receives `KEEP`, complete gates, `correctness_source == "forge_rewrite_reference"`, and `artifact_path == "kernels/rmsnorm.py"`; the loser receives `FAILED`, null gates, and its own stdout log, with `warnings == []`. A separate test pins the rule that no fields are assigned when `best_attempt_id` is empty.

I also checked the blast radius: `assembler._compose_kernel_journey` passes payload rows through unchanged, `_kernel_outcome` and `_best_micro_speedup` do not read these fields, and `kernel_journey.backend_attempts` has no consumers other than V6 and tests.

---

### #2 — GEAK's final outcome could contradict the completed rebench

Confirmed. `status: ok` only means that the runner completed; it does not mean that the candidate was adopted. `outcome` now has two authoritative sources, neither of which is the runner status:

1. **`geak_result.revalidation_status`** — this field records only negative terminal outcomes: `failed`, `no_material`, `no_promote`, or `fallback_failed`. The positive path (`decision == "validated"` → `_promote_geak_from_candidate`) writes a `geak_e2e` optimization-stack entry and a final-rebench attempt, but does not set this field. Therefore, absence of `revalidation_status` must not be interpreted as KEEP. The field is also absent from `collect_geak`; V6 reads it directly from `state["geak_result"]`, leaving the V5 payload unchanged.
2. **The linked final-rebench attempt** — `_settle_pending_outcomes` runs after `_link_rebench_attempts` and settles the candidate using the rebench that actually measured it (`status == "succeeded"`).

When neither source exists, the result is `NEEDS_REVIEW`, never `KEEP`. The supplied reproduction now produces `REVERT`.

This "unadjudicated means pending" rule is consistent across all five lanes. `_settle_pending_outcomes` only updates rows currently marked `NEEDS_REVIEW`; an existing explicit E2E verdict is not overwritten. Conflicting verdicts remain visible instead of being hidden by a last-writer-wins rule.

---

### #3 — Fusion could pair a new run with an old integration result

Confirmed. `last_fusion` is overwritten for every run, while `last_fusion_integrate` is updated only when integration actually runs. A non-kept result never enters integration, and `kernel.py:4160` also returns early when the patch or target is missing.

The fix adds a shared identity as suggested: `_handle_fusion_result` stamps `fusion_run_id` onto the result before persisting state, and `_integrate_fusion` writes the same id into the integration verdict. V6 validates the pairing through `_fusion_paired`:

- both sides have an id → pair by id for new sessions
- neither side has an id → pair by the patch path named by the verdict for historical compatibility
- they do not match → discard the stale verdict and emit a warning; `KEEP` is then unreachable

Without a matching verdict, a micro-level `candidate` reports `NEEDS_REVIEW`: KernelForge's KEEP is a microbenchmark result, while adoption is decided by the E2E re-baseline. `no_improvement` maps to `REVERT`, and `failed` maps to `FAILED`.

Collective campaigns now use the same ladder: an explicit `integration_decision` wins; otherwise a micro-level `kept` result remains `NEEDS_REVIEW`. This lane is upserted by `collective_attempt_id`, so it does not have the cross-run pairing problem, but it can still produce the shape "micro KEEP without completed integration."

---

### Additional issue found outside this PR's scope

`build_kernel_fusion_value` in `orchestrator/knowledge/remote_recipe/values.py:586` uses the same unconditional fallback to pair `last_fusion` with `last_fusion_integrate`. As a result, KB write-back may attribute the previous round's E2E gain to a newer fusion run. This predates the PR and is outside V6 telemetry, so it was left for a follow-up.

---

### Verification

- `test_sbd_v6_stages.py`: 77 passed, including 13 new regression tests for these three issues
- `test_sbd_v6_initial.py` and `test_kernel_journey.py`: exactly the same pre-existing failure set as before the changes
- Broad run with `-k "breakdown or v6 or geak or kernel or exporter or collector"`: the before/after FAILED and ERROR lists were byte-for-byte identical
- `ruff format --check .` and `ruff check src/hyperloom/`: clean

On Windows, `test_shared_state_persistence.py`, `test_remote_recipe_v2.py`, and `test_coordinator_gemm_promote_units.py` cannot be collected because `fcntl` is unavailable. Those files cover the `last_fusion*` persistence path and therefore rely on CI for execution.

xiaofei-zheng (approved):

PR #1353, opened 2026-08-31
Publish session evidence to the Recipe KB and estimate uplift without running

rpoornac:
## Summary

Five commits: two building the feature, two fixing bugs the review found, one merging
current `main`.

**Publish side (`7e2cca6b`).** CLOSE recorded a session's validated gain but none of
the evidence behind it, so a later session could not tell how much roofline headroom
was left, what the host was tuned to, what had already failed, or what the run cost.
This publishes, all nested under `value{}` since `knowledge` forbids extra top-level
properties:

- `roofline` — compact baseline/optimized arms, host paths stripped, falling back to
  the analytic ceiling when no trace snapshots exist
- `platform` — CPU governor/SMT/NPS and GPU `gfx_arch`, no hostname
- `do_not_repeat` — capped fingerprinted skip list of rejected configs and kernel IDs
- `token_usage` + `elapsed_minutes` — folded from `llm_calls.jsonl`, because CLOSE runs
  before `session_breakdown.json` exists

None of these count as replay material; `has_replay_material` asserts that explicitly.
PRELUDE injects the skip list into `explore_search.rejected` and `rejected_kernel_ids`
so the dedup gate refuses known-failed configs mechanically, which lets the specialist
prompt drop `remaining_gaps` and the raw failure list rather than spending tokens
restating them. The sanitizer needed an allowlist entry so `token_usage` isn't mistaken
for a secret.

**Read side (`7198b911`).** A no-run estimator that mines the KB instead of running:
forecasts `min(H, P50(capture) × H)` where capture is the fraction of a prior run's
roofline gap it actually closed, and reports token/runtime priors from published spend.
New CLI: `python -m hyperloom.inference_optimizer.tools.estimate_no_run`.

**Two fixes (`95ec1ddb`, `c7605875`).** Both were silent failures in the publish side,
detailed under "What the review caught" below.

## Why the scoping matters

Pooling every session a search returns makes the median meaningless. Sessions are now
grouped by identity and by the full `tp/conc/isl/osl` replay scope, and `pool_warnings`
names it whenever a pool spans models, boards, frameworks, versions, or precisions. On
live MI355X/sglang data a 23.4% pooled median hides per-family medians ranging from
6.2% to 311%.

Parallelism gets the same treatment. TP is only compared within a fixed model,
precision, and ISL/OSL: ISL changes arithmetic intensity and model size dictates the TP
that fits at all, so without that grouping a 0.6B at TP1 against a 70B at TP8 reads as
3% scaling retention when nothing scaled. Across 103 live MI355X sessions every
workload family has exactly one TP, so cross-TP projection honestly reports "no
per-GPU baseline in this family" instead of a fabricated number.

The parallelism question the store *can* answer today is the layout inside a fixed GPU
count. `sharding_whatif` normalizes the vLLM and SGLang spellings of tp/dp/ep/pp and
ranks accepted layouts per replay scope. Live results:

| scope (tp = GPU count) | layout | n | p50 gain |
|---|---|---|---|
| deepseek-v4-pro/mxfp4/sglang/tp8 | `dp_attention=on` | 1 | 297.4% |
| deepseek-v4-pro/mxfp4/sglang/tp8 | framework-default | 1 | 0.0% |
| minimax-m3-mxfp4/mxfp4/sglang/tp8 | `ep=4 dp=8 dp_attention dp_lm_head` | 1 | 23.3% |
| minimax-m3-mxfp4/mxfp4/sglang/tp8 | framework-default | 2 | 12.5% |
| qwen3.8-2.4t-a95b/mxfp4/sglang/tp8 | `ep=8 moe_dense_tp=1` | 1 | 14.2% |
| qwen3.8-2.4t-a95b/mxfp4/sglang/tp8 | `moe_dense_tp=1` | 1 | 7.2% |
| qwen3.8-2.4t-a95b/mxfp4/sglang/tp8 | framework-default | 2 | 6.8% |

Read these as directional, not statistical: the arms are n=1 or n=2, and 40 of 43
scopes have only a single layout. Note also that the winning-by-gain minimax layout has
*lower* absolute throughput than its default arm (2675 vs 2835 tok/s), because gain is
measured against each session's own baseline — rank on whichever metric you care about.

Only winning layouts are published until `do_not_repeat` lands, so the report carries
`winners_only: true` — an absent layout means untried or unpublished, not worse.
Likewise `coverage` and `limitations` name the missing fields rather than reporting
zero savings, since no record in the store yet has roofline, spend, or elapsed data.

## What the review caught

Both bugs were in the publish side, and both failed silently rather than loudly, which
is why they're worth calling out to reviewers.

**`95ec1ddb` — one malformed rejected config destroyed the whole published recipe.**
`sanitize_publish_server_args` raises on args `shlex` cannot tokenize or that carry a
whitespace-bearing operand, and `_skip_explore_row` called it unguarded, so the
exception propagated through `build_remote_knowledge` and CLOSE published nothing at
all. Rejected configs are the likeliest place to find broken quoting, since they
already failed. A skip list is an optimization rather than replay material, so the row
is now dropped and the recipe survives. The same commit fixes a starvation bug: the cap
filled in strict ledger order and stopped at 32, so an explore phase with 32+ rejected
variants silently dropped every kernel skip — exactly the rows PRELUDE needs. Half the
cap is now reserved for kernel skips, with either kind reclaiming the other's unused
budget.

**`c7605875` — a single-snapshot roofline scored as zero capture.**
`build_publishable_roofline` fills the baseline arm from `snapshots[0]` and the
optimized arm from `snapshots[-1]`, the same object when a session captured one
profile, and then published both — asserting a post-optimization profile that never
ran. The miner read the equal arms as a zero delta and skipped the
`validated_e2e_gain` fallback, so a session with a real 30% gain contributed
`capture = 0.0` and pulled `P50(capture)` and the forecast toward zero. The publisher
now emits an optimized arm only once a second snapshot exists, and the miner prefers
the arm delta only when positive, falls back to the validated gain when the delta is
non-positive but the gain isn't, and still reports zero capture for a genuine
regression.

## Notes for reviewers

- The read side works against today's store and produced the table above with no new
  sessions. The publish side is inert until sessions flow through CLOSE, which is why
  the estimator's roofline/spend/elapsed coverage counters currently read zero.
- Store credentials are parameters (`--kb-store-token-file` preferred), and the report
  echoes the store URL but never the token.
- 11 files changed, 2524 insertions, 13 deletions.
- Current `main` is merged in (`24249abd`), so CI runs against the real target. This
  matters because `main` restructured `build_remote_knowledge` underneath the branch —
  staged section merging, required-owner validation, and section-routing-by-lever — and
  this branch attaches its new `value{}` sections inside that same function. The merge
  was textually clean and the combined KB suite passes, but passing tests show only that
  nothing broke, not that the routing is being used as intended, so a second opinion
  from @ZhengGong-amd on that interaction would be worth having.

## Test plan

- [x] `pytest test_no_run_estimate.py test_remote_recipe_v2.py test_warm_replay.py
      test_specialist_lessons_section.py test_config_patch_agent_kb.py
      test_promote_shared_state_lock.py` — 362 pass with `main` merged in, including
      main's own tests for the restructured `build_remote_knowledge`
- [x] Publish path: roofline/platform/`do_not_repeat`/`token_usage` built from state and
      ledgers, host paths stripped, and each asserted *not* to be replay material
- [x] Skip injection: `do_not_repeat` rows reach `explore_search.rejected` and
      `rejected_kernel_ids`; specialist section 5 keeps `do_not_repeat` and drops
      `remaining_gaps`/`what_failed`
- [x] Scoping regressions pinned: layouts are never compared across world sizes, and TP
      is never compared across models or across ISL
- [x] Robustness against silent publish failures: an unsanitizable rejected config is
      dropped rather than aborting the whole recipe publish, the 32-row skip cap
      reserves half its slots for kernel skips, and a single-snapshot roofline publishes
      no optimized arm so it cannot score as zero capture (`95ec1ddb`, `c7605875`)
- [x] Credential order (flag → file → env), missing URL exits 2 with no network call,
      and the token never appears in the report
- [x] Live end-to-end against the KB store on the merged tree for `--hardware mi355x`
      (103 sessions, 102 with gain, 14 carrying an explicit parallelism layout)
- [ ] Reviewer: confirm the `value{}` additions are acceptable to any downstream KB
      consumer that validates the envelope


meinali-566:
@rpoornac  Really appreciate your great work.
After reviewing this with our team, we decided to close this PR for the following reasons: 

- the Recipe KB should remain strictly focused on validated data required to select, reconstruct, and execute the next replay, including configuration, patches, kernel artifacts, and only the minimal metadata needed to apply them. 
- Fields such as roofline, platform, token_usage, and elapsed_minutes represent historical execution evidence or analytics inputs rather than replay data, so they should not be stored in the Recipe KB or added to its value{} schema. This information should instead be derived from archived session logs, SBD, or another session-evidence pipeline, and the no-run estimator should consume that source directly. Keeping this boundary avoids turning the Recipe KB into a general-purpose session data store.

/close

rpoornac:
Thanks @meinali-566, I've created PR#1391 as per your suggestion, please help to review.

PR #1359, opened 2026-09-01
fix(fusion): do not report an aborted forge-fusion as an optimization result

zengleixin-amd:
## Problem

forge-fusion can die before its loop runs a single attempt — the harness-authoring turn fails, or there is no git workspace to author in. `_normalize_manifest` did not read `termination_reason`, so that landed in the generic no-KEEP shape: `complete` / `no_improvement`, no `error_class`, empty `env_flags`.

That is wrong twice over. It records infrastructure failing as *"this model has no fusion opportunity"*, and `complete` satisfies the KERNEL-entry idempotency gate in `_fusion_required_before_kernel_opt`, so a single abort skips fusion for the whole remaining session.

## Impact on the fleet

Measured over the 8-25..8-28 window on `/shared_nfs/hyperloom-claw`:

- **8 sessions** lost fusion outright, every one with `termination_reason=harness_author_failed`, `attempts=0`, `best_env_flag=None`.
- Every one had **already located a recipe** — the flag name was sitting in `manifest.fusion.env_flag` (`DEEPSEEK_V4_FUSED_ATTN_REDUCE_INV_ROPE`, `GLM_MOE_DSA_FUSED_INDEXER_K_PREPARE_STORE`, …).
- 7 of the 8 `harness_author.log` files end with `end: subtype=success` immediately before the guard error: the agent had written the harness, validated it, and proved via `git status` that it touched no framework source.
- Across those 8 sessions KERNEL entry came round **15 times**. Fusion ran 8 times. **7 later entries could have retried and were turned away by the gate.**

The upstream trigger (a workspace guard reclaiming the agent's own declared target) was fixed in e91e543e0. This PR fixes the downstream normalization, which is what turned one aborted turn into a session-wide skip.

## Fix

Report an abort the way an LLM outage is already reported one branch above: `failed`/`REVERT` with an `error_class` — the established retryable shape.

Keyed on `termination_reason` rather than on the one failure that was observed, so the sibling abort path (`no_git_workspace`) is covered and a new one inherits the handling instead of silently regressing.

Guarded on `not kept` and `not compile_pass` so a measured result is never discarded. The attempt count is deliberately *not* load-bearing: KernelForge's abort handler builds a fresh `LoopResult` with an empty history, so `attempts` reads 0 even when earlier recipes ran full campaigns.

Retries are bounded. `no_git_workspace` does not heal mid-session, and each retry re-runs LLM discovery before failing in the same place. Capping is not the original bug returning — the record still reads `failed` with its `error_class`, so a run that gave up stays legible as infrastructure that gave up, not as a verdict on the kernel.

## Not changed

- The workspace guard, `PROTECTED_GLOBS`, the gate's status condition, the salvage path, KernelForge.
- `env_flags` is left empty on an abort. Its contract is "flags this run confirmed"; nothing measured the located recipe, so it rides in the error message instead.
- `serving_unconfirmed` is deliberately excluded from `INFRA_ABORT_REASONS`: it is set while *keeping* a micro KEEP, which is the opposite of "the loop never ran".

## Verification

- `test_forge_fusion.py`: **68 passed** (14 new, including parametrized case/whitespace and non-numeric `attempts`).
- New `test_fusion_retry_gate.py`: **11 passed**. Exercised against `KernelPhase` unbound on a stand-in `self` — building a `Coordinator` pulls in `recipe_kb` → `fcntl`, so the existing coordinator-based module cannot be collected off POSIX at all.
- **Replay against real fleet manifests**, same input into pre- and post-change code:

| case | before | after |
|---|---|---|
| real `8b372077` (`harness_author_failed`) | `complete`, gate blocks | `failed` + `error_class`, gate allows retry |
| `no_git_workspace` | same as above | same as above |
| real `bb534a39` (loop ran, `exhausted`) | `complete` + flags | **byte-identical** |
| `llm_unavailable` | `failed` + `error_class` | **byte-identical** |
| abort reason + `kept=True` | `ok` / KEEP | **byte-identical** |

- Full `agents/kernel/tests/` + `phases/tests/` regression: failure set identical to baseline, **zero new failures**.

## Follow-ups (not in this PR)

- The deployed runtime lagged the guard fix: e91e543e0 landed 08-28 05:28 UTC, and the last observed failure ran 08-28 13:12 UTC. Worth confirming what the fleet actually has installed.
- `llm_unavailable` retries remain unbounded (pre-existing). Left alone here rather than changing behaviour this PR did not introduce.

### DIFF
--- a/src/hyperloom/agents/kernel/tests/test_forge_fusion.py
+++ b/src/hyperloom/agents/kernel/tests/test_forge_fusion.py
@@ -807,9 +807,7 @@ def test_a_loop_that_ran_still_reports_no_improvement(tmp_path):
     """
     output_dir = tmp_path / "out"
     output_dir.mkdir()
-    manifest = _aborted_manifest(
-        "exhausted", attempts=1, best_env_flag="QWEN3_FUSED_QK_NORM_ROPE_KVCACHE"
-    )
+    manifest = _aborted_manifest("exhausted", attempts=1, best_env_flag="QWEN3_FUSED_QK_NORM_ROPE_KVCACHE")
     (output_dir / "fusion_manifest.json").write_text(json.dumps(manifest), encoding="utf-8")
 
     result = forge_fusion._normalize_manifest(str(output_dir), rc=0)
@@ -826,9 +824,7 @@ def test_an_abort_reason_is_matched_tolerantly(tmp_path, reason):
     no_improvement mapping, i.e. straight back into the bug this prevents."""
     output_dir = tmp_path / "out"
     output_dir.mkdir()
-    (output_dir / "fusion_manifest.json").write_text(
-        json.dumps(_aborted_manifest(reason)), encoding="utf-8"
-    )
+    (output_dir / "fusion_manifest.json").write_text(json.dumps(_aborted_manifest(reason)), encoding="utf-8")
 
     result = forge_fusion._normalize_manifest(str(output_dir), rc=0)
 
--- a/src/hyperloom/agents/kernel/tests/test_forge_fusion.py
+++ b/src/hyperloom/agents/kernel/tests/test_forge_fusion.py
@@ -675,6 +675,170 @@ def test_an_llm_outage_verdict_is_matched_tolerantly(tmp_path):
     assert result["status"] == "failed"
 
 
+def _aborted_manifest(reason, **loop_extra):
+    """A manifest for a run that located a recipe, then died before attempting it.
+
+    This is exactly the shape observed on the fleet: discovery succeeded, so
+    ``verdict`` is ``candidate`` and ``fusion.env_flag`` names a real flag, while
+    ``fusion_loop`` reports zero attempts and no promoted flag.
+    """
+    loop = {"termination_reason": reason, "attempts": 0, "best": None, "best_env_flag": None}
+    loop.update(loop_extra)
+    return {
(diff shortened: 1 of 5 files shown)


zengleixin-amd:
Addressed in 215aa4ea7.

The cap was resettable. The tally lived inside `last_fusion`, which every run replaces, so any outcome without the abort marker — a subprocess timeout, a handler crash — carried no count forward and handed the cap a clean slate. `abort → timeout → abort` retried indefinitely.

Moved it to its own `SharedState.fusion_infra_aborts` field: it is a property of the session, not of any one run. Keeping it off the record also stops a per-run result from reporting a number that is not about that run.

Monotonic, with nothing clearing it. Every outcome that would justify a reset — `ok`, `complete`, `kept` — already stops the gate on its own, so a reset would buy nothing and only reopen the hole. A new field defaults to `0`, so a `state.json` written before this needs no migration (verified: round-trip, and `from_dict` on a blob with the key removed).

The interleaved path now has the regression test it was missing (`test_an_unrelated_failure_between_aborts_does_not_reset_the_cap`). It fails against f068e61bf, along with 5 others that were sensitive to where the count lived.

`test_fusion_retry_gate.py`: 13 passed. Full `agents/kernel/tests/` + `phases/tests/` regression unchanged from baseline.

xiaofei-zheng (approved):

PR #1363, opened 2026-09-01
fix(kernel): name the reason when kernel_opt is never dispatched

BaoYunkai:
## Problem

A session that reached `KERNEL_AGENT` without a candidate table dispatched no `kernel_opt` at all and left nothing on record saying so.

The summary's six `unattempted_reason_breakdown` buckets each mean "the candidate table listed this kernel and nobody tried it", so an absent table counts zero in every one of them. The zero-attempt takeaway then offered two guesses — `kernel_opt` disabled, or no candidate qualified — and neither is what happened: nothing was ever asked for, because `untried_hot_reusable_kernels()` reads `last_trace_analyze` and that snapshot was empty.

Seven forge sessions (2026-08-29..31) spent ~28 GPU-hours in the phase with `attempted=0`, `top_candidates=0` and all six buckets at 0, and read as workloads with no headroom. Telling "the lane died for want of input" apart from "candidates arrived and were all filtered" required reading `state.json` by hand.

## Fix

The field that should carry this already exists: `record_kernel_opt` writes `last_kernel_opt_dispatch_skip` for an empty dispatch queue and the summary surfaces it as `dispatch_skip_reason`. Both wholesale-skip paths return before `run_optimization` is called, so that writer never runs for them.

- `phases/kernel.py`: KERNEL entry records the breadcrumb on both skip paths, separating the states `_kernel_opt_work_remains()` collapses into a single `False` — `auto_kernel_opt_disabled`, `no_candidate_table`, `no_untried_hot_kernels` — plus `no_candidates_path` for a snapshot that carries no candidates artifact. The reason is derived from the field the gate itself admits on (`hot_kernels_top15` / `hot_kernels`), not from the snapshot dict being non-empty: a trace_analyze that ran and crashed leaves `{"status": "failed", ...}` behind, and calling that "the kernels were all tried" would assert the very conclusion this breadcrumb exists to prevent. The evidence fields (`trace_analyze_empty`, `candidates_path`, `profile_trace`, `profile_status`, `roofline_failure_streak`) answer "why was the table empty" without a `state.json` dig, and are persisted where they are written — the run this serves is the one that spends hours in the phase and is then killed.
- `phases/kernel.py`: the entry batch clears the field before dispatching. It names no `kernel_id`, so `record_kernel_opt` — which retires the breadcrumb only for a result that names one — never cleared it, and a later entry that dispatched and had its candidates filtered by the handler's own floor still reported "never dispatched" as fact.
- `kernel/attempt_summary.py`: with nothing attempted, `_build_top_takeaways` states the recorded reason instead of guessing; the original wording is kept when no breadcrumb exists. Worded around the skip rather than around "never dispatched", because the field has two writers with different meanings — the phase records declining to ask, while `no_eligible_kernels` records a dispatch that did happen and found nothing eligible.

No schema change: both the report key and its state field predate this, and `unattempted_reason_breakdown` / `totals` are untouched.

## Tests

Seven tests, each verified red on the tree it targets:

- `test_prelude_roofline.py`: an empty snapshot records `no_candidate_table` with its evidence; a failed trace_analyze records `no_candidate_table` rather than `no_untried_hot_kernels`; a spent table (a reusable hot kernel above the floor with an attempt recorded against it, asserted to leave the gate empty) records `no_untried_hot_kernels`; a snapshot without a candidates artifact records `no_candidates_path`; a `batch_mode` dispatch retires an earlier breadcrumb.
- `test_kernel_attempt_summary.py`: the takeaway names the recorded reason; it does not claim a dispatch never happened for `no_eligible_kernels`; it keeps the generic line when nothing was recorded.

### DIFF
--- a/src/hyperloom/inference_optimizer/tests/test_kernel_attempt_summary.py
+++ b/src/hyperloom/inference_optimizer/tests/test_kernel_attempt_summary.py
@@ -435,6 +435,38 @@ def test_dispatch_skip_reason_empty_by_default(tmp_path: Path) -> None:
     assert out["dispatch_skip_reason"] == {}
 
 
+def test_takeaway_names_the_dispatch_skip_instead_of_guessing(tmp_path: Path) -> None:
+    """With nothing attempted, the takeaway must state the recorded reason.
+
+    The old sentence offered two guesses -- disabled, or nothing qualified --
+    and so could not express the case that actually happened: the candidate
+    table was never produced, so kernel_opt was never asked for. A reader
+    holding six zero buckets and that sentence concluded the workload had no
+    headroom.
+    """
+    state = _make_state(top15=[])
+    state.last_kernel_opt_dispatch_skip = {
+        "reason": "no_candidate_table",
+        "trace_analyze_empty": True,
+        "roofline_failure_streak": 3,
+        "ts": "2026-09-01T00:00:00+00:00",
+    }
+    out = build_kernel_optimization_summary(state, tmp_path)
+    joined = " ".join(out["top_takeaways"])
+    assert "no_candidate_table" in joined
+    assert "never dispatched" in joined
+    assert "no candidates qualified" not in joined
+
+
+def test_takeaway_keeps_the_generic_line_without_a_recorded_reason(tmp_path: Path) -> None:
+    """No breadcrumb => the original wording stays (nothing to name)."""
+    state = _make_state(top15=[])
+    out = build_kernel_optimization_summary(state, tmp_path)
+    joined = " ".join(out["top_takeaways"])
+    assert "No kernels were attempted" in joined
+    assert "no candidates qualified" in joined
+
+
 def test_zero_attempts_session_does_not_crash(tmp_path: Path) -> None:
     state = _make_state(top15=[])
     out = build_kernel_optimization_summary(state, tmp_path)
--- a/src/hyperloom/inference_optimizer/tests/test_kernel_attempt_summary.py
+++ b/src/hyperloom/inference_optimizer/tests/test_kernel_attempt_summary.py
@@ -454,8 +454,28 @@ def test_takeaway_names_the_dispatch_skip_instead_of_guessing(tmp_path: Path) ->
     out = build_kernel_optimization_summary(state, tmp_path)
     joined = " ".join(out["top_takeaways"])
     assert "no_candidate_table" in joined
-    assert "never dispatched" in joined
     assert "no candidates qualified" not in joined
+
+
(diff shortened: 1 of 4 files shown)


BaoYunkai:
All four confirmed and fixed in `ea9d64e6b`.

**1. A non-empty snapshot without a hot-kernel table (must-fix)**

Correct, and worse than a mislabel: the reason asserted the conclusion the breadcrumb exists to prevent. `_kernel_opt_dispatch_skip_reason` tested the dict for emptiness while the gate reads `hot_kernels_top15` / `hot_kernels`, so `{"status": "failed", "error": "roofline crashed"}` came back as `no_untried_hot_kernels`. Both now read the same field, as suggested.

The fixture was wrong the way you describe — `kernel_roofline_top15` is what `attempt_summary` reads for its top15, not what the gate admits on, so the test exercised the no-table state its docstring denied. It now carries a reusable hot kernel at 40% GPU with an attempt recorded against it in `kernel_opt_task_attempts`, and asserts `untried_hot_reusable_kernels() == []` up front so the fixture cannot silently decay back into the no-table case. A separate test pins the failed-trace_analyze state to `no_candidate_table`; it fails on the previous commit.

**2. The breadcrumb outliving a real dispatch (must-fix)**

Confirmed: the entry batch names no `kernel_id` by design, `record_kernel_opt` clears the field only for a result that names one, and the handler returns `batch_mode`. Cleared in `_run_kernel_opt_entry_batch` right before dispatching, per your suggestion. Test asserts the field is empty after a `batch_mode` result and fails on the previous commit.

Agreed on the severity ordering: a confident false statement is worse than the hedge it replaced.

**3. Skip paths not persisted**

Saved inside `_record_kernel_opt_dispatch_skip`, so both call sites are covered by one line and any future one is too. Wrapped so a save failure can never fail the phase for a breadcrumb.

**4. Wrong verb for `no_eligible_kernels`**

Took the neutral-verb option: `No kernels were attempted in this session (kernel_opt dispatch skip: <reason>)`. Branching would have meant either a reverse import from `phases/` into `kernel/attempt_summary.py` or duplicating the reason literals across two modules, and duplicated literals drift. A test asserts the sentence does not claim a dispatch never happened when the reason is `no_eligible_kernels`.

**Verification**

Ruff 0.16.2 `check` + `format --check` clean (the earlier red was a collapsible line in the takeaway, fixed in `85d7db21c`). Full `inference_optimizer` suite: 11419 passed, 15 failed — the same 15 that fail on an unmodified tree (real subprocess / real ROCm build / real LLM backend / optional-package). `ci-e2e/run` times out identically on #1358, #1359, #1360 and #1361, so it is not this branch.

xiaofei-zheng (approved):

PR #1365, opened 2026-09-01
Publish the AgentX workload identity in the GEAK handoff

amd-ethany:
### Description

**What.** In AgentX mode Hyperloom benchmarks a replay of real agentic traces, but
the GEAK handoff only ever carried the CLI's synthetic `isl`/`osl`/`conc` triple.
This PR adds an explicit `benchmark.workload_spec` to the handoff that names the
workload, carries the shape actually observed on the baseline, and pins the
concurrency the recipe really serves.

**Why.** GEAK had no way to tell "these numbers describe the load I must
reproduce" from "these numbers are a placeholder the bench client ignores", so it
measured a ~1k-token synthetic sweep and divided the result into an agentic
denominator. On Kimi-K3 that produced:

| Measurement | Value |
| --- | --- |
| Hyperloom agentic baseline | 168.998 tok/s |
| GEAK synthetic client at ISL/OSL = 1024 | 465.676 tok/s |
| Ratio an ungated GEAK would publish | **2.76x "win" with no kernel changed** |

The real workload is nothing like 1024/1024 — the corpus averages **111,874**
input and **797** output tokens per request.

### Changes

**1. `workload_spec` production — `_workload_envs.py`**

`build_agentx_workload_spec()` emits the workload identity that travels to GEAK:
`kind` (`agentx_trace_replay`), concurrency, duration, dataset, scenario, warmup
depth/grace, and the failed-request threshold. `apply_agentx_switch()` attaches it
to the recipe. Non-AgentX materializations carry no `workload_spec`, so every
existing fixed-ISL/OSL path behaves exactly as before.

**2. Observed replay shape — `kernel.py`**

`_observed_replay_shape()` derives the real workload shape from the baseline's own
`inferencex_result.json` (`total_input_tokens / completed`, likewise for output)
and injects `observed_isl`, `observed_osl`, `observed_requests`, `observed_source`
and `observed_tput_match_err` into the spec, so kernel and GEMM shape synthesis
targets the shape the workload actually has instead of the CLI default.
`_resolve_workload_spec()` reads the spec back off the materialized recipe, and
`_resolve_launch_server_script()` resolves the server launcher for the handoff.

**3. Concurrency invariant — `_workload_envs.py`**

`workload_spec.concurrency` is the load GEAK replays with, so it MUST equal the
`CONC` the recipe serves. It did not. `apply_agentx_switch()` builds the spec early
in `materialize_config_with_envs()`, before the resolved `CONC` is projected into
`envs`, so a base config carrying the parser default (64) shipped that default to
GEAK while the recipe served the operator's value (8):

```
apply_agentx_switch()   -> workload_spec.concurrency = 64   (pre-override envs)
CONC projected to envs  -> envs.CONC = 8                    (operator value)
```

No downstream gate caught it, because `workload_spec.kind` still read
`agentx_trace_replay` on both sides. The fix re-pins the spec to the served value
once `CONC` is resolved, and warns when they disagree so a producer bug is visible
rather than silent.

Measured consequence of the old behaviour, replaying at 64 against a baseline
served at 8:

| Metric | Baseline (conc 8) | Replay at conc 64 |
| --- | --- | --- |
| Output throughput | 168.09 tok/s | 41.46 tok/s |
| Mean TTFT | 3,361 ms | 298,556 ms |

That is KV-cache thrashing, not a kernel regression — but it is indistinguishable
from one in the results, and the leg was marked `usable_for_acceptance: true`.

**4. `AGENTX_CLIENT_SCRIPT` — `agentx/deploy.py`**

Names the `aiperf_client.sh` constant beside the deployment that publishes it, so
consumers distinguishing "drives a client" from "launches a server" match on one
definition instead of re-spelling the filename.

## Linked issue(s)

_None filed; this was found while running a 24h Kimi-K3 AgentX campaign._

## Tests

**Added:**
- `test_agentx_workload_spec_concurrency_tracks_served_conc` in
  `test_workload_envs_branches_unit.py` — drives the full
  `materialize_config_with_envs` path with a base config at `CONC: 64` and an
  operator `CONC: 8`, asserting the spec follows the served value.
- `test_geak_handoff_launch_server_script.py` (new, 186 lines)
- `test_geak_observed_replay_shape.py` (new, 146 lines)
- `test_geak_workload_comparability.py` (new, 165 lines)

**Updated:** `test_agentx_switch.py` (+29), `test_geak_resume_recovery.py` (+80).

**Commands run:**

```bash
# Targeted: workload envs, AgentX switch/budget/guards, GEAK resume/handoff suites
python3 -m pytest \
  src/hyperloom/inference_optimizer/tests/test_workload_envs_golden_lock.py \
  src/hyperloom/inference_optimizer/tests/test_workload_envs_branches_unit.py \
  src/hyperloom/inference_optimizer/tests/test_agentx_budget_and_guards.py \
  src/hyperloom/inference_optimizer/tests/test_agentx_switch.py \
  src/hyperloom/inference_optimizer/tests/test_geak_resume_recovery.py \
  src/hyperloom/inference_optimizer/tests/test_cli_workload_envs.py \
  src/hyperloom/inference_optimizer/tests/test_geak_handoff_launch_server_script.py \
  src/hyperloom/inference_optimizer/tests/test_geak_observed_replay_shape.py \
  src/hyperloom/inference_optimizer/tests/test_geak_workload_comparability.py -q
# -> 160 passed

# Full suite
python3 -m pytest src/hyperloom/inference_optimizer/tests -q
# -> 11586 passed, 27 failed, 17 skipped, 1 xfailed
```

The 27 failures are pre-existing and environmental, not from this change: they are
the credential-dependent suites (`test_preflight_auth_override`,
`test_critic_agent_backend`, `test_proposal_scorer_coverage_unit`,
`test_common_utils`, `test_coverage_gap_units`) which need LLM-gateway credentials
absent on the test host. Confirmed by running the same suite in a pristine
worktree of the base commit and diffing the failure sets — identical distribution,
no failures unique to this branch.

The new concurrency test was also verified to genuinely catch the defect: with the
guard disabled it fails `assert 64 == 8`, the exact symptom seen in the campaign.

## Breaking changes

**No.** Purely additive:

- Handoffs without `workload_spec` keep today's behaviour; the field is only
  produced for AgentX materializations.
- The concurrency guard only changes a value that was already self-contradictory
  (spec disagreeing with the recipe's own served `CONC`), and logs a warning when
  it fires.
- Requires the matching GEAK change to *consume* the spec. Until that lands GEAK
  ignores the extra field, so the two can merge in either order.


ZhengGong-amd:
## Changes required before merge

1. **Rebase this PR onto the latest `main`**, which now contains PR #1362. The branch is based on an older mainline revision and has a real conflict in:

   ```text
   src/hyperloom/orchestrator/actions/executors/_workload_envs.py
   ```

2. **Resolve the `_workload_envs.py` conflict in the correct semantic order.** Keep both changes, with PR #1362's warmup adjustment first and PR #1365's workload publication second:

   ```python
   # First: derive and assign the CONC-scaled value from PR #1362.
   envs["AGENTX_WARMUP_GRACE_PERIOD"] = scaled_value

   # Then: publish the effective workload used by AgentX.
   bench["workload_spec"] = build_agentx_workload_spec(
       bench,
       envs,
       model_path=model_path,
   )
   ```

   `build_agentx_workload_spec()` reads `AGENTX_WARMUP_GRACE_PERIOD` and publishes it as `warmup_grace_period_s`. Calling it before PR #1362's override would silently record the raw/stale value in the GEAK handoff while the aiperf client executes with the scaled value. Do not resolve this conflict with a simple `ours`/`theirs` selection or concatenate the blocks in the opposite order.

3. **Fix the current Ruff formatting failure.** `ruff format --check` is red and reports changes in:

   ```text
   src/hyperloom/orchestrator/actions/executors/_workload_envs.py
   src/hyperloom/orchestrator/phases/kernel.py
   src/hyperloom/inference_optimizer/tests/test_geak_workload_comparability.py
   ```

   Run the repository's Ruff formatter, commit the formatted output, and confirm the required Ruff check passes.

4. **Rerun integration coverage after the rebase.** PRs #1362 and #1365 both extend the AgentX-to-GEAK handoff contract. At minimum, rerun the tests covering:

   - AgentX switch and workload environment materialization;
   - GEAK handoff construction and resume recovery;
   - launch-server-script selection;
   - observed replay shape;
   - workload comparability;
   - PR #1362's warmup-grace and throughput behavior.

   Verify explicitly that the published `workload_spec.warmup_grace_period_s` equals the final CONC-scaled value used by the aiperf client. The isolated integration merge passed 164 relevant tests; its two subprocess import failures were caused by the review environment not installing `hyperloom`, not by either PR.

5. **Resolve or rerun the failing E2E check.** Its current failure is caused by missing CI configuration:

   ```text
   E2E_API_BASE is required
   ```

   This does not appear to be a code regression, but the CI environment or dispatch configuration must be corrected if this is a required check.

## Recommended cleanup

6. **Load the Recipe YAML once.** `_resolve_workload_spec()` and `_resolve_launch_server_script()` currently open and parse the same `baseline_config_path` independently, duplicate the `benchmark` shape validation, and maintain separate broad fallback paths. Parse and validate the Recipe once, then pass the resulting benchmark mapping to both resolvers.

7. **Use one source of truth for the concurrency default.** `build_agentx_workload_spec()` currently falls back to `8`, while workload materialization uses `parser.DEFAULT_CONC` (`64`). Remove the conflicting literal default and derive both paths from the same resolved value or shared constant.

8. **Consider constructing the workload spec after environment projection.** `apply_agentx_switch()` currently runs before the effective `CONC`, `ISL`, `OSL`, and related values are projected into `envs`, requiring a later warning-and-repin block to repair `workload_spec.concurrency`. Moving workload-spec construction after environment projection would record the correct value on the first pass and remove redundant corrective logic.

9. **Reduce model-family/corpus rule duplication.** The AgentX model-family whitelist and default corpus selection are duplicated across `_workload_envs.py`, `aiperf_client.sh`, and `agentx/preflight.py`, with slightly different normalization rules. Centralize the mapping or add an explicit cross-implementation consistency test so future model additions cannot silently change the workload reported to GEAK.

10. **Make replay-result filtering path-aware.** `_observed_replay_shape()` currently excludes results using substring checks such as `"geak" in path`. Prefer checking path components or known result directories so an unrelated parent directory containing that substring does not silently exclude a valid baseline result.

xiaofei-zheng (approved):
## Review: what this PR does

In AgentX (trace-replay) mode the GEAK handoff previously carried only the CLI's synthetic `isl`/`osl`/`conc` triple and a fixed output axis, so GEAK searched on the wrong load and the wrong throughput axis. This PR publishes the AgentX workload identity so GEAK can reproduce the real load and stay comparable:

- New leaf module `hyperloom/common/workload_defaults.py` centralizes `DEFAULT_ISL/OSL/CONC/TP/EP/PRECISION`, breaking the orchestrator↔`cli.parser` import cycle; both sides read the same constants.
- `_workload_envs.py` adds `build_agentx_workload_spec`, written into the recipe at the END of `apply_agentx_switch` (after all derived values land in `envs`), plus `geak_metric_axis` (output/total axis by grader) and the corpus/family helpers that mirror `aiperf_client.sh`.
- `phases/kernel.py` attaches `workload_spec` (overlaying observed ISL/OSL from the round's own `inferencex_result.json`), resolves the launch-server script, aligns `E2E_METRIC` to the guarded axis, and refuses to compute `self_reported_gain_pct` when GEAK reports the workloads are not comparable.
- `_geak_sweep.py` reads the metric-neutral `throughput_tok_s_median` with a legacy fallback.

## Verification depth

- **Concurrency invariant (the headline fix):** confirmed correct *by construction*. `agentx_env_for_conc(conc)` returns an env copy with the rung's `CONC`, and `build_agentx_workload_spec(..., env=_agentx_env)` resolves `CONC` via `served_knob` (process-env first), so `workload_spec.concurrency` equals the served `CONC` rather than the stale parser default. Verified `served_knob` reads `_agentx_env` before `envs`.
- **Non-AgentX path byte-identical:** `apply_agentx_switch` returns early when AgentX is off or the framework is scriptable, so `workload_spec` is never set. The relocated fallbacks (`DEFAULT_ISL/OSL/CONC` = 1024/1024/64) equal the previous inline literals, so `materialize_config_with_envs` is unchanged for fixed-ISL/OSL runs.
- Read the `_record_geak_candidate` comparability gate, the warmup-grace scaling, and the corpus/family normalization; the shell↔Python corpus parity is locked by `test_agentx_corpus_rules_consistency` exercising the real shell function.

## Blocking issues

None. The change is complete, call sites are migrated, failures are not silently swallowed, and docstrings/comments track the code.

LGTM.

PR #1367, opened 2026-09-01
ci: key E2E workflow_dispatch concurrency on head_sha

jiagaoxiang:
## Summary

`workflow_dispatch` has no `pull_request.number` / `issue.number`, so CI E2E concurrency fell through to `github.ref` (`refs/heads/main` for every dispatch). `cancel-in-progress: true` then killed the previous GPU smoke — #1325's dispatch cancelled #1322's run 33463143636 after ~58 min on the GPU.

This keys the group on `inputs.head_sha` (then `inputs.head_ref`) before `github.ref`, so two fork-PR dispatches can coexist. Same idea on Forge E2E job concurrency (`head_sha` before `head_ref`).

`forge-kernel-bench.yml` already uses a single `forge-kernel-bench` group with `cancel-in-progress: false`, so dispatches serialize instead of killing each other. Left it alone.

## Test plan

- [ ] Dispatch CI E2E for two different `head_sha`s from `main` and confirm the first run is **not** cancelled
- [ ] PR / `/retest` / `retest` label path still cancels the same-PR in-flight run

### DIFF
--- a/.github/workflows/ci-e2e.yml
+++ b/.github/workflows/ci-e2e.yml
@@ -39,12 +39,17 @@ on:
 # comments get a group of their own (keyed by comment id) so they cancel nothing. The
 # `/retest` predicate is kept identical to the one in `resolve.if`, so exactly the
 # comments that can start a run are the comments that can cancel one.
+#
+# workflow_dispatch has no pull_request.number / issue.number, so the group used
+# to fall through to github.ref (refs/heads/main for every dispatch) and
+# cancel-in-progress killed the previous GPU smoke. Key on inputs.head_sha
+# (then head_ref) so two fork-PR dispatches can run without cancelling each other.
 concurrency:
   group: >-
     ci-e2e-${{
       (github.event_name == 'issue_comment' && !contains(github.event.comment.body, '/retest'))
       && format('noop-comment-{0}', github.event.comment.id)
-      || (github.event.pull_request.number || github.event.issue.number || github.ref)
+      || (github.event.pull_request.number || github.event.issue.number || inputs.head_sha || inputs.head_ref || github.ref)
     }}
   cancel-in-progress: true
 
--- a/.github/workflows/forge-e2e.yml
+++ b/.github/workflows/forge-e2e.yml
@@ -190,7 +190,7 @@ jobs:
     # Concurrency starts only after resolve has authenticated and parsed the
     # event. A no-op comment therefore cannot cancel an in-flight GPU job.
     concurrency:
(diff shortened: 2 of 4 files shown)


zoroyihan7 (approved):
LGTM — approving. This is the fix for a bug that was actively burning GPU hours, and the diagnosis and the key are both right.

For context on why this matters: I hit this twice from both sides. A dispatch I made for #1322 ran ~58 minutes and was cancelled 18 seconds after a dispatch for #1325 was created; later my dispatch for #1322 killed one of yours that had been queued for 2h51m. Every `workflow_dispatch` collapsed to `ci-e2e-refs/heads/main`, and since dispatch is the only way to get a real e2e on a fork PR (no secrets on `pull_request`), two fork PRs could never both be validated — and the loser vanished silently.

### What I verified

- **`inputs.head_sha` is safe on non-dispatch events.** This was my main worry, since the workflow-level `concurrency` expression is evaluated before any job's `if`, so a bad reference would break run creation for every event. Empirically fine: the `pull_request`-triggered `CI E2E` and `Forge E2E` runs on this very branch both completed successfully with the new expression.
- **`resolve` really does export `head_sha`** in both workflows, so the forge-e2e side isn't silently falling through to `head_ref`.
- **The keys behave correctly at the boundaries.** Two dispatches for *different* SHAs land in different groups (the fix); two for the *same* SHA still supersede each other (desirable). The `noop-comment` guard is untouched and still first in the chain.
- **Merged `main` locally** — this branch is 121 commits behind and #1376 touched both `forge-e2e.yml` and `test_forge_e2e_gate.py`. Merges clean, fix intact, 15 passed, and `scripts/tests/` plus the guard file are 195 passed.
- **Mutation-tested it.** Reverting the ci-e2e group, reverting both forge-e2e groups, and moving `head_sha` *after* `github.ref` each turn a test red. So the tests genuinely pin the change rather than just describing it.

### One hole, non-blocking

Deleting `head_sha: ${{ steps.decide.outputs.head_sha }}` from `resolve.outputs` in `forge-e2e.yml` leaves all 15 tests green. The group expression then silently falls through to `head_ref`, and the bug is back with CI still passing — which is the same silent-fallthrough shape this PR exists to fix. The tests string-match the *consumer* (the group expression) but never assert the *producer* it depends on.

One line in `test_forge_e2e_gate.py` closes it:

```python
assert "head_sha: ${{ steps.decide.outputs.head_sha }}" in workflow
```

Worth adding, but I don't want to hold up a fix that's costing real GPU time. Take it here or as a follow-up.

PR #1375, opened 2026-09-02
Stop reporting a GEAK run that found nothing as a fault

luochen-amd:
`no_gain` is the **third most common GEAK status in production** — 212 sessions against 115 `ok` over the 2404 that carry a geak block, roughly **9% of every session that engaged GEAK** — and the V6 kernel lane was reporting all of them as `status: failed` / `outcome: FAILED`.

## Why it is not a failure

It is a completed run that found nothing. The runtime treats it as terminal beside `ok` in three places:

- `orchestrator/phases/machine_state.py` lists it among the statuses that end the kernel phase
- `breakdown/collectors/geak.py` pairs it with `ok` when backfilling accepted kernels — `status in ("ok", "no_gain")`
- `kernel_work_pending` reads it as "nothing left to try"

Falling through to `_lane_status` mapped it to `failed`, which is the right default for a genuinely unrecognized word and the wrong answer for one the runtime relies on. It is also the exact conflation this lane was built to remove, one status further in.

## What it maps to now

`succeeded`, and no further. `status` says the runner finished, which is a fact about the runner; whether the candidate is adopted stays with the orchestrator's own rebench, so the outcome ladder leaves a `no_gain` run at `NEEDS_REVIEW` unless a writeback or a linked rebench settles it. GEAK's self-report is the claim, not the verdict.

## How it was found

Not by inspection. The normalizer added in #1350 warns on an unrecognized status, and

```
v6.timeline.kernel: unrecognized geak_runs status 'no_gain'; reported as failed
```

surfaced in `metadata.warnings` on the first two V6 sessions that ran GEAK. The drift channel worked as designed; the vocabulary was incomplete.

## Tests

The new test pins the whole vocabulary the runtime writes — `ok`, `no_gain`, `missing`, `no_result_recovered_from_disk`, `timeout`, `error`, `skipped` — so the next spelling fails there instead of in a report. Verified it fails on the pre-fix tree for `no_gain` alone and passes for the other six.

100 → 107 in `test_sbd_v6_stages.py`. Ruff format + check pass.

### DIFF
--- a/src/hyperloom/inference_optimizer/breakdown/collectors/v6_stages.py
+++ b/src/hyperloom/inference_optimizer/breakdown/collectors/v6_stages.py
@@ -132,10 +132,17 @@ _GEAK_REVALIDATION_OUTCOMES = {
 # the orchestrator's own rebench, so the outcome ladder below leaves a
 # ``no_gain`` run at ``NEEDS_REVIEW`` unless a writeback or a linked rebench
 # settles it. GEAK's self-report is the claim, not the verdict.
+# ``baseline_reproduction_failed`` is the runner reporting that its baseline
+# reference did not reproduce the orchestrator's, which makes the whole run's
+# gain non-comparable -- ``phases/kernel.py`` reads it as a guard and refuses to
+# promote a phantom-baseline gain. ``failed`` is where the unknown-status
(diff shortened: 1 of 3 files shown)


ZhengGong-amd (approved):
Approved

PR #1389, opened 2026-09-02
testintel: add missing test for SharedState.record_collective_integration (created by dougljia)

jiagaoxiang:
This PR was created by dougljia via Test Gap Resolver.

- Description: No test forces SharedState.save to fail during record_collective_integration, so the atomic rollback of last_collective/collective_attempts on save exception (lines 2228-2237) is unverified; a regression here would let in-memory state diverge from the persisted session file after a failed write.
  This PR adds a focused unit test for `SharedState.record_collective_integration`; it does not change production code.
- Linked issue(s): None
- Tests: Adds a unit test for `SharedState.record_collective_integration`. In `src/hyperloom/inference_optimizer/tests/test_collective_lane.py` (`pytest src/hyperloom/inference_optimizer/tests/test_collective_lane.py`). Sits beside `test_collective_replay_preserves_completed_integration`. Not executed by Test Gap Resolver; CI on this draft reports results.
- Breaking changes: No

### DIFF
--- a/src/hyperloom/inference_optimizer/tests/test_collective_lane.py
+++ b/src/hyperloom/inference_optimizer/tests/test_collective_lane.py
@@ -322,6 +322,191 @@ def test_collective_replay_preserves_completed_integration(tmp_path):
     assert state.last_collective["integration_decision"] == "KEEP"
 
 
+# --- record_collective_integration ---------------------------------------------
+
+
+def _seeded_collective_state(tmp_path):
+    """A state whose last_collective/collective_attempts hold one open KEEP campaign."""
+    state = SharedState.load_or_init(tmp_path)
+    campaign = {
+        "collective_attempt_id": "attempt-1",
+        "integration_id": "integration-1",
+        "status": "ok",
+        "decision": "KEEP",
+        "engine": "forge_collective",
+        "kept": True,
+        "requires_e2e_validation": True,
+    }
+    state.record_collective(campaign, tmp_path)
+    return state
+
+
+def test_record_collective_integration_updates_last_collective_and_history(tmp_path):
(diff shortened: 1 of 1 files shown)


jiagaoxiang:
**TestIntel PR Steward:** Poll found CI red at `02c9f0a7d` (`e2e` and `ruff` failing). Pushed `de93837c` to fix what's ours and to flag what isn't:

- **Lint (`ruff format --check .`) — fixed.** The added `next(...)` generator expression in `test_record_collective_integration_updates_last_collective_and_history` spanned 4 lines; `ruff format` wants it on one line. Collapsed it.
- **Duplicate test name — fixed.** `github-advanced-security` (CodeQL) flagged the same root cause independently: this PR's new `test_record_collective_integration_rolls_back_when_save_fails` had the exact same name as a pre-existing test later in the file. In Python, the second `def` silently shadows the first at module load, so the new rollback-on-save-failure test — the one covering the exact atomic-rollback risk this PR describes — was never actually collected or run. Renamed the new one to `test_record_collective_integration_rolls_back_when_save_fails_with_legacy_field` (it exercises the legacy `integration_status` field name, distinct from the existing test's `patch_cleanup_status` path), so both tests now run.
- **`e2e` — not touched, looks unrelated to this PR.** The job fails immediately in `.github/scripts/ci-e2e-dispatch.sh` with `E2E_API_BASE: E2E_API_BASE is required` before any of this PR's code runs — a missing CI env var/secret for the e2e dispatcher, not something a test-only PR against `SharedState` can cause. Flagging for a human/infra owner; will not attempt to patch CI config from this PR.

Will keep following CI, review threads (including bot findings), and will merge only after a human approval plus green required checks.

PR #1391, opened 2026-09-02
Estimate uplift and compare parallelism layouts without running

rpoornac:
## Why

Deciding whether a target is worth a session meant running one. This mines the
Recipe KB instead and reports what prior sessions already settled for a scope:
the distribution of validated gain, and which parallelism layout won inside a
fixed GPU count.

## Scope

This is the read half of #1353, re-opened on its own after that PR was closed.
The publish half is not here and is not coming back: roofline arms, host
platform, token spend and elapsed time are execution evidence rather than replay
data, so they do not belong in the Recipe KB or its `value{}` schema.

This reads only fields the KB already owns as replay material —
`validated_e2e_gain`, `workload_shape`, and the accepted
`value.config.extra_server_args` that a layout is read out of. Nothing is added
to the schema, and no publish path is touched. A test pins the boundary rather
than leaving it to review, because the cheap way to add a headroom forecast
later is to start reading roofline arms back out of a replay record.

Worth being explicit that this costs nothing: every number in the table below,
across 103 live MI355X sessions, came from those three fields. The forecast and
spend priors in #1353 were the parts reading the contested fields, and their
coverage counters were zero because no record in the store carries them.

## Why the scoping matters

Pooling every session a search returns makes the median meaningless. Sessions
are grouped by identity and by the full `tp/conc/isl/osl` replay scope, and
`pool_warnings` names it whenever a pool spans models, boards, frameworks,
versions, or precisions. On live MI355X/sglang data a 23.4% pooled median hides
per-family medians ranging from 6.2% to 311%.

Parallelism gets the same treatment. TP is only compared within a fixed model,
precision, and ISL/OSL: ISL changes arithmetic intensity and model size dictates
the TP that fits at all, so without that grouping a 0.6B at TP1 against a 70B at
TP8 reads as 3% scaling retention when nothing scaled. Across 103 live MI355X
sessions every workload family has exactly one TP, so cross-TP projection
honestly reports "no per-GPU baseline in this family" instead of a fabricated
number.

The parallelism question the store *can* answer today is the layout inside a
fixed GPU count. `sharding_whatif` normalizes the vLLM and SGLang spellings of
tp/dp/ep/pp and ranks accepted layouts per replay scope. Live results:

| scope (tp = GPU count)             | layout                               | n | p50 gain |
| ---------------------------------- | ------------------------------------ | - | -------- |
| deepseek-v4-pro/mxfp4/sglang/tp8   | dp_attention=on                      | 1 | 297.4%   |
| deepseek-v4-pro/mxfp4/sglang/tp8   | framework-default                    | 1 | 0.0%     |
| minimax-m3-mxfp4/mxfp4/sglang/tp8  | ep=4 dp=8 dp_attention dp_lm_head    | 1 | 23.3%    |
| minimax-m3-mxfp4/mxfp4/sglang/tp8  | framework-default                    | 2 | 12.5%    |
| qwen3.8-2.4t-a95b/mxfp4/sglang/tp8 | ep=8 moe_dense_tp=1                  | 1 | 14.2%    |
| qwen3.8-2.4t-a95b/mxfp4/sglang/tp8 | moe_dense_tp=1                       | 1 | 7.2%     |
| qwen3.8-2.4t-a95b/mxfp4/sglang/tp8 | framework-default                    | 2 | 6.8%     |

Read these as directional, not statistical: the arms are n=1 or n=2, and 40 of
43 scopes have only a single layout. Note also that the winning-by-gain minimax
layout has *lower* absolute throughput than its default arm (2675 vs 2835
tok/s), because gain is measured against each session's own baseline — rank on
whichever metric you care about.

A record keeps only the layout its session settled on, so the report carries
`winners_only: true` and says in `limitations` that an absent layout is untried
or unpublished rather than beaten.

## Follow-ups, not in this PR

Archived `session_breakdown.json` is a strict superset of what this reads: its
`workload` block carries the identity dimensions and the full replay scope, and
its `final` block carries `cumulative_gain_pct_validated`,
`throughput_tok_s_per_gpu` and `extra_server_args`. A second projector plus a
`--sessions-root` flag would let the estimator mine archived sessions directly,
and since SBD already carries `roofline`, `token_usage` and elapsed time, it
would restore the headroom forecast without a single field going into the KB.
Left out here to keep this reviewable on its own.

## Test plan

* `pytest test_no_run_estimate.py` — 20 pass
* `pytest test_remote_recipe_v2.py test_warm_replay.py test_specialist_lessons_section.py test_no_run_estimate.py` — 249 pass
* Scoping regressions pinned: layouts are never compared across world sizes, and
  TP is never compared across models or across ISL
* Boundary pinned: the report carries no `forecast` or `savings_prior` section
  and `coverage` is exactly `{with_gain}`
* Credential order (flag → file → env), missing URL exits 2 with no network
  call, and the token never appears in the report
* `ruff check` and `ruff format --check` clean

### DIFF
--- a/docs/reference/operator-scripts.md
+++ b/docs/reference/operator-scripts.md
@@ -1,15 +1,16 @@
 ---
 myst:
     html_meta:
-        "description": "Reference for Hyperloom operator scripts: dump_session_breakdown, dump_session_report, and event_counts. Use these utilities to inspect, export, and report on session data."
-        "keywords": "Hyperloom, operator scripts, session breakdown, session report, event counts, LLM inference, AMD GPU, ROCm, debugging, observability, operator tools"
+        "description": "Reference for Hyperloom operator scripts: dump_session_breakdown, dump_session_report, event_counts, and estimate_no_run. Use these utilities to inspect, export, and report on session data, and to estimate uplift before running."
+        "keywords": "Hyperloom, operator scripts, session breakdown, session report, event counts, uplift estimate, Recipe KB, LLM inference, AMD GPU, ROCm, debugging, observability, operator tools"
 ---
 # Hyperloom operator scripts
 
 A short reference for the operator-facing scripts under
 `src/hyperloom/inference_optimizer/tools/`. These are not part of the agent loop —
-they are utilities you run by hand against a finished or in-progress
-session directory.
+they are utilities you run by hand, most of them against a finished or
+in-progress session directory. The exception is `estimate_no_run.py`, which
+reads the Recipe KB rather than a session.
 
 When no explicit `--session-dir` is given, scripts resolve the active session in
 two steps: `INFERENCE_OPTIMIZER_CURRENT_SESSION_DIR` when set, otherwise the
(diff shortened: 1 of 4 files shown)


rpoornac:
Context for this PR, following the close of #1353.

Taking the boundary as given: session evidence — roofline, platform, token spend,
elapsed time — is not replay data and does not belong in the Recipe KB or its
`value{}` schema. The publish side of #1353 is dropped entirely, including the
sanitizer allowlist entry, and nothing here changes the schema.

What's left reads three fields the KB already owns as replay material:
`validated_e2e_gain`, `workload_shape`, and the accepted
`value.config.extra_server_args`. That is enough for the scoped gain distribution
and the layout comparison in the description — the 103-session MI355X table came
from those three. The parts that read the contested fields were the headroom
forecast and the spend priors, and both are gone. A test pins the boundary so it
does not creep back.

The second half of your note — have the estimator consume SBD directly — is a
follow-up rather than this PR. It turns out to be a clean fit: SBD's `workload`
block carries the identity dimensions and the full tp/conc/isl/osl scope, and its
`final` block carries the validated gain, throughput and accepted server args, so
it is a strict superset of what I read from the KB. Since SBD also carries the
roofline and token sections, the forecast can come back from the evidence pipeline
without a field going into the KB. I kept it separate so this one stays reviewable.

One open question there. There is no cross-session index over SBDs today —
`breakdown/SKILL.md` says one file per session, with fleet views left to a
downstream job. The cheapest thing I see is globbing `**/session_breakdown.json`
under an archived session root, building no new store. If there is an intended
direction, or one already in flight, I would rather write against it.

meinali-566:
@rpoornac  as we discussed offline,  this feature shouldn't be done in Hyperloom workflow, so I close this PR. thanks.

PR #1397, opened 2026-09-03
[CHORE] Drop the deterministic trace-analysis route

fengshaoyi-amd:
## Why

`deterministic` was a no-LLM TraceLens route: it shelled out to the Python
toolchain (perf report, `orchestrator_prepare`, the per-category scripts,
`generate_priority_data`) and then re-extracted hot kernels out of
`priority_data.json`.

`bypass` already answers the same question — hot kernels with no LLM call — by
reading the profiler trace directly. So `deterministic` was a second extraction
pipeline to keep correct for no additional capability. Only `agent` and `bypass`
remain.

## Behaviour change

A request still naming `analysis_route=deterministic` is **rejected before
dispatch**: the handler returns `status=failed` with
`error_class=invalid_analysis_route` without resolving TraceLens or launching a
subprocess, and the error names `bypass` as the no-LLM replacement. Falling back
to `agent` would spend an LLM session — the opposite of what a caller asking for
a no-LLM route wants — so an explicit unrecognized route fails loudly instead.
Only an omitted or blank route defaults to `agent`.

The same rule covers `HYPERLOOM_TRACE_ANALYSIS_ROUTE`. An operator whose `.env`
still carries the line this repo's own template used to suggest verbatim
(`HYPERLOOM_TRACE_ANALYSIS_ROUTE=deterministic`) now sees the roofline fail with
a message that names the fix, rather than silently paying for an agent session.
`roofline.py` continues in degraded mode rather than retrying, so the failure is
loud without being fatal.

`tracelens_analysis.py` loses `--analysis-route` entirely rather than keeping a
flag with one legal value.

## One non-obvious cleanup

`record_kernel_discovery` takes `source` and `tool` separately, and
`deterministic` was the only reason that split was reachable from
`trace_analyze_handler`: it recorded `source="bypass"` (a no-LLM route, grouped
with bypass on the dashboard) over `tool="tracelens"` (what actually ran, so the
version provenance stayed under the TraceLens toolchain). With it gone the two
are equal for this caller, so `tool=` is dropped and defaults to `source`.

The split itself stays — the GEAK backend still uses it from the kernel phase
(`source="bypass"` with `tool="geak"`).

Two consequences fall out of dropping `tool=`:

- A bypass run now keys its own `versions["bypass"]` entry, so `"bypass"` needs
  a `_TOOL_PROVENANCE` recipe. The reader ships inside this distribution, so it
  resolves Hyperloom's version the way `forge` already does; without the entry
  the session recorded an all-empty
  `{"tool":"bypass","root_dir":"","commit":"","version":""}`.
- `assert "bypass" not in out.get("versions", {})` is retired. It only held in
  the `source=bypass` + `tool=tracelens` combination; for a real bypass run
  `versions["bypass"]` is now expected, and a new case pins its version.

Separately, `scan["analysis_route"]` now records the route rather than the
toolchain. It read `_disc_source`, so an agent run wrote `"tracelens"` — a value
no caller can pass, and outside the `agent` / `bypass` vocabulary this PR
declares in the CHANGELOG. The test that pinned `"tracelens"` there was pinning
the bug.

`_analysis_md.py` is kept; `bypass` is now its only caller.

## Verification

Windows has pre-existing failures in these files (path separators, missing
`python3`), so this was checked against a throwaway worktree at the merge base
rather than by reading the numbers alone. Scope: the seven test files this PR
touches, run under a local `fcntl` stub so `test_profile_and_kernel_handlers.py`
collects at all.

| | failed | passed |
|---|---|---|
| merge base (`da54496de`) | 24 | 894 |
| this branch | 23 | 866 |

The two failure sets are identical line for line except one: the failure that
disappears is `test_deterministic_steps_quote_priority_output_dir`, a deleted
test. The passed count drops because the `deterministic` tests went with the
route.

`test_profile_and_kernel_handlers.py` cannot be collected on Windows (`fcntl` is
Unix-only) and was only run under that stub, so **Linux CI is the real gate for
that file**.

`ruff check` and `ruff format --check` are clean across `src/`.


tsrikris:
Pls refer to https://github.com/AMD-AGI/Hyperloom/pull/1368 for comprehensive PR that was being worked upon for cleanup. Feel free to combine changes if necessary (though I'd looked at your PR for confirmation). I'd suggest focussing on that PR since it updates the TraceLens SHA, the post-processing layer as well.

xiaofei-zheng (approved):

Issue #431, opened 2026-06-04
CUDA-graph capture folds kernels into hipGraphLaunch → trace_analyze hot_kernels=0 (#210)


iraj465:
# [Hyperloom] ISSUE (#210) — CUDA-graph capture folds kernels into hipGraphLaunch wrappers → trace_analyze hot_kernels=0

- **Type:** ISSUE (bug / analysis gap)
- **Target repo:** AMD-AGI/Hyperloom  (`inference_optimizer` profile/analysis; relates to existing #210)
- **Status:** PREPARED ON DISK — not filed (gh not authenticated).

## Problem
When the server runs with CUDA/HIP graphs enabled (default), per-kernel device activity is folded into `hipGraphLaunch` wrappers and the captured `pt.trace.json.gz` carries **no `user_annotation` events**. `trace_analyze` then cannot attribute per-kernel time and emits **0 hot kernels** (empty candidate set). The optimization pipeline has nothing to work on.

## Evidence (hard, on disk)
- `/home/sapmajum/PROJECTS/runs/gptoss120b_20260602_194008/kernel_e2e3.log:33` and `:98`:
  ```
  trace structure check: [3] main trace ...pt.trace.json.gz has no user_annotation events —
  InferenceX per-step annotations didn't fire. Verify roofline_annotations reached the framework
  (PROFILE_EXTRA_BODY consumed; see #210).
  ```
  These runs used cuda-graph (`--max-cudagraph-capture-size 2048`).
- Eager vs cuda-graph contrast (session measurement): eager run → Compute **39.92%**, **10** hot_kernels; cuda-graph run → Compute **14.54%**, **0** hot_kernels.
- Corroborating: the eager analysis at `.../20260603T170856Z/.../parallel_e2e_summary.json` reports `num_hot_kernels: 10`.

## Workaround (validated)
Run the profiling pass with `--enforce-eager` (vLLM `enforce_eager=True`). Per-kernel attribution returns and the analysis yields ~10 candidates.

## Proper fix (recommended)
Fold the CUDA-graph capture traces (`capture_traces/*.pt.trace.json.gz`) into the analysis so per-kernel attribution survives graph capture, instead of relying solely on the live (graph-folded) trace. Then `trace_analyze` can recover hot kernels without forcing eager mode (which perturbs the very workload being optimized).

chennyiiis:
Fixed in #487 (merged).

[...]

PR #1399, opened 2026-09-03
fix(profile): stop the per-batch-size graph-capture exporter taking the server down

fengshaoyi-amd:
## Why

`profile_sglang.yaml` selects SGLang's per-batch-size graph-capture exporter.
That exporter cannot survive a DSA dual-graph model: it names each trace after a
batch size, tracked by one index advanced once per profiler callback.

```python
self._profile_bs_list = list(reversed(self.capture_bs))
self._profile_bs_idx = 0

def on_trace_ready(prof):
    bs = self._profile_bs_list[self._profile_bs_idx]
```

But `capture()` fires one callback per `(bs, lora, dsa)` combination. A model
that captures a dense and a sparse graph for each batch size therefore raises
the index past the end of the list, and the server goes down during graph
capture — before it is ready to serve a single request.

Switching to the single-trace exporter takes that path out of play. Upstream
gates the per-bs branch on the single-trace flag being unset
(`decode_cuda_graph_runner.py:765-766`), so setting it is enough; the buggy
indexing is never reached.

## Behaviour change and blast radius

**Breaking changes: no.** This is a profile-time capture setting; nothing in the
serving or baseline path reads it.

**Graph-capture trace layout changes.** One combined trace instead of one file
per `(batch size, rank)`. `_capture_shapes.py` classifies capture sidecars by
shape rather than by an exact name, and matches `graph_capture` anywhere in a
filename, so the combined trace is still recognised as a capture rather than
mistaken for the workload trace.

## What this does not fix

Stated so the PR is not read as closing more than it does. A profile round can
still come back with an empty hot-kernel table, for reasons this change does not
touch:

- the TraceLens 0.5.18 patch set being short three annotation patches — a
  TraceLens issue, tracked there;
- per-kernel device time folding into `hipGraphLaunch` under cuda-graph capture
  (#431), which is what actually empties the steady-state window in the captures
  looked at here;
- the AgentX `_AGENTX_PROFILE_MAX_ITERS = 8` capture cap, which lives in
  `_workload_envs.py` and carries its own host-RAM rationale.

## Dropped from the first revision of this PR

The first revision also set `SGLANG_PROFILE_WITH_STACK` to `false`, on the
argument that callstacks crowd GPU kernels out of the steady-state window. That
argument does not hold and the change has been reverted.

`python_function` events carry no correlation id — 0 of 641782 in the capture
that was quoted — and the splitter counts a chunk's GPU events solely by
collecting correlation ids from the CPU events inside a step window and looking
up the GPU events they map to. Callstacks never enter that count; removing them
changes it by zero.

A closer control says the same. Qwen3-0.6B `20260826T072156Z-edacec52` profiled
at **97.33%** `python_function`, *higher* than the 94.62% capture quoted as
evidence, under the same concurrency, the same two profiler env vars and the
same 128 steps — and did not hit `steady_state_chunk_empty`. Both captures
collapse the same way instead: 128 `hipGraphLaunch`, exactly one of which
carries GPU events, leaving 127 of 128 steps with zero. Which chunk the splitter
lands on decides whether a run survives.

The blast-radius check in that revision was also wrong to conclude nothing reads
a callstack. It searched for the env var name, which does appear once. But
`_trace_launcher_resolver` reads the `python_function` chain rather than the env
var, and runs unconditionally from `tracelens_analysis.py:5064` to recover a
Python launcher frame for the kernels TraceLens files as `(Synthetic Op)` with
`launcher_path = "Not found"` — hand-written Triton and AITER kernels, the ones
kernel optimisation is most likely to want. Without callstacks it falls back to
name-grep, which yields no line number and can match a test file or a CPU
implementation. TraceLens' own `mla_decode_pseudo_ops` matches `python_function`
names too.

## Verification

The four suites that read this config: **381 passed**.

```
pytest src/hyperloom/inference_optimizer/tests/test_workload_envs.py \
       src/hyperloom/inference_optimizer/tests/test_workload_envs_branches_unit.py \
       src/hyperloom/inference_optimizer/tests/test_profile_and_kernel_handlers.py \
       src/hyperloom/inference_optimizer/tests/test_aiperf_client_sh.py
```

The config parses and materialises the intended values, and no reference to the
replaced env var remains anywhere in the tree.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,34 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ### Changed
 
+- **Graph-capture traces go through the single-trace exporter, and the shape
+  manifest says when that costs it something.**
+  `SGLANG_ENABLE_CUDA_GRAPH_CAPTURE_TRACE` replaces
+  `SGLANG_GRAPH_BATCH_CAPTURE`, which selected a per-batch-size exporter that
+  cannot survive a model capturing more than one variant per batch size. It
+  indexes a list built from `capture_bs` once per profiler callback, but
+  `capture()` fires one callback per `(bs, lora, dsa)` combination, so a DSA
+  dual-graph model — dense and sparse for every batch size — walks off the end
+  of the list and takes the server down during graph capture, before it is
+  ready to serve. Upstream gives the single-trace flag precedence when both are
+  set (`decode_cuda_graph_runner.py::_graph_batch_capture_active`, from
+  sgl-project/sglang#24370), so an inherited `SGLANG_GRAPH_BATCH_CAPTURE=1`
+  cannot re-select the crashing path.<br/>
+  **The layout it writes is not a drop-in.** One combined file per rank holds
+  every batch size at once and carries no variant identity in its name, so the
+  shape-manifest classifier rejects it — deliberately, since admitting it would
+  mint one shape-identical variant per rank and defeat the dedup behind
+  `variant_count`. The identity is not recoverable from inside the file either:
+  the capture-phase `record_function` spans are not emitted on this path
+  (measured on a GLM-5.2 MI355X capture — 15166 `user_annotation` events, not
+  one of them a capture span). The manifest therefore falls back to eager and
+  Forge loses the per-batch-size dense shapes.<br/>
+  That fallback is also what an eager profile produces legitimately, so until
+  now the two were indistinguishable downstream. The manifest now emits
+  `shape_manifest_capture_not_indexable` when capture ran and only its layout
+  was unreadable, naming the files. A server that starts takes precedence over
+  a manifest that indexes, but the trade is recorded rather than silent.
+
 - **AgentX installs its own benchmark client instead of letting an agent guess
   at it.** `HYPERLOOM_AGENTX` declares aiperf as a required, version-pinned
   dependency and `install.sh` already owned that install, but it was gated on a
--- a/src/hyperloom/agents/kernel/tests/test_bypass_trace_analysis.py
+++ b/src/hyperloom/agents/kernel/tests/test_bypass_trace_analysis.py
@@ -895,7 +895,7 @@ def test_discover_capture_shards_dedups_tp_ranks(tmp_path):
 
 
 def test_discover_capture_shards_dedups_a_runner_prefixed_batch(tmp_path):
-    # An SGLang without the profiler patch prefixes the runner name:
+    # The upstream per-bs exporter prefixes the runner name:
     # DecodeCudaGraphRunner_bs_104_rank0. The batch token is still what
     # identifies the variant, so the ranks must collapse the same way — an
     # anchored read finds nothing here and falls back to the per-rank stem,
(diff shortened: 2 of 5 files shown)


devalshahamd:
@fengshaoyi-amd Can you please tell what exact issue are you facing and for which workload. We intentionally store one capture trace per batch size since that is required for downstream analysis flow and this change might break that. 
@mohbasit

fengshaoyi-amd:
@mohbasit Thanks — and you are right that this change breaks the per-batch-size
layout your downstream flow depends on. Let me give you the exact failure first,
then what I think the real fix is, because I don't think we actually disagree.

## Workload and exact failure

GLM-5.2 (MXFP4) on MI355X, TP=8, DSA dual-graph capture enabled. The server
never reaches ready — it dies during startup graph capture.

`decode_cuda_graph_runner.py` (v0.5.18) names each capture trace from a list
indexed once per profiler callback:

```python
# _init_profile_context_and_memory_record, lines 781-791
self._profile_bs_list = list(reversed(self.capture_bs))
self._profile_bs_idx = 0

def on_trace_ready(prof):
    bs = self._profile_bs_list[self._profile_bs_idx]      # <-- unguarded index
    trace_file = os.path.join(trace_dir, f"{runner_name}_bs_{bs}_rank{rank}.json.gz")
    prof.export_chrome_trace(trace_file)
    self._profile_bs_idx += 1
```

But `capture()` is a triple loop (lines 1094 / 1105 / 1107), so the callback
fires once per `(bs, lora, dsa)` combination, not once per bs:

```python
for bs in capture_range:
    for variant_label, _variant_has_lora in lora_variants:
        for dsa_variant in dsa_variants:
            self.capture_one_shape(bs, ...)
```

With 8 capture_bs and DSA dual-graph (dense + sparse per bs) that is 16
callbacks against a list of 8, so callback #9 raises IndexError and takes the
server down before it can serve a request. Any model that captures more than
one variant per batch size hits this: DSA, LoRA, or both.

## The part I think matters more to you

The crash is not the worst of it. The first 8 callbacks **succeed**, and 7 of
them write a trace under the wrong batch size:

```
DecodeCudaGraphRunner_bs_7_rank0.json.gz   actually holds bs=8
DecodeCudaGraphRunner_bs_6_rank0.json.gz   actually holds bs=7
```

So on a multi-variant model the per-bs layout does not deliver the guarantee
it is there to provide. Downstream does not get a missing shape it can detect
— it gets a confidently mislabelled one, and roofline / arithmetic intensity
are computed against the wrong batch. A model with fewer buckets than variants
would never crash at all and would mislabel silently, end to end.

## On the combined trace — your concern is correct

We measured it, and it is worse than "might break": the combined layout carries
no variant identity in the filename, the capture-phase `record_function` spans
are not emitted on that path (15166 `user_annotation` events in our capture, not
one of them a capture span), and the sibling `execution_details.json` records a
single `batch_size` for a file holding every batch size (256 for all 8 decode
ranks). So the shape manifest degrades to eager and per-bs dense shapes are lost.

We switched only because a degraded manifest is recoverable and a server that
never starts is not. The PR now emits an explicit warning instead of degrading
silently, but that is damage control, not a fix.

## What I would rather do

Name the trace after the shape actually being captured, instead of indexing a
list by callback count. I have a patch that does this and I would rather land it
than keep the combined-trace workaround:

```python
self._set_profile_capture_tag(bs, variant_label, dsa_variant)   # before capture_one_shape
# -> DecodeCudaGraphRunner_bs_8_dense_rank0.json.gz
#    DecodeCudaGraphRunner_bs_8_sparse_rank0.json.gz
```

Replayed against four capture shapes:

| model shape | callbacks | current | with patch |
|---|---|---|---|
| Qwen3-8B (52 bs, no variants) | 52 | 52 traces, 0 mislabelled | **identical filenames** |
| LoRA (8 bs x 2) | 16 | crash at #9, 7/8 mislabelled | 16 traces, 0 mislabelled |
| DSA (8 bs x 2) | 16 | crash at #9, 7/8 mislabelled | 16 traces, 0 mislabelled |
| DSA + LoRA (8 bs x 2 x 2) | 32 | crash at #9, 7/8 mislabelled | 32 traces, 0 mislabelled |

Single-variant models produce byte-identical filenames, so nothing in your
current flow changes. Multi-variant models get the per-bs layout for the first
time, with the variant in the name so dense and sparse no longer collide.

Applies cleanly to v0.5.18 and to upstream main (the bug is still there on main).
Happy to send it upstream, or to route it through the TraceLens patch set if you
would rather carry it there — your call, since your flow is the consumer. Would
that address the downstream requirement you have in mind?

fengshaoyi-amd:
Closing this one.

It was parked waiting for TraceLens to confirm the downstream impact. After the
crash mechanism and the capture-shape replay were posted on Sep 4 there has been
no reply in three weeks, but the silence is not the reason to close it â€” the
merits are.

As argued in the thread above, this change is damage control rather than a fix.
It sidesteps the crash by selecting the combined-trace exporter, and the price is
precisely what @devalshahamd objected to: the per-batch-size dense shapes the
downstream flow needs. Trading a capability away to route around a defect is the
wrong shape of change, however well the trade is documented.

The root cause is in SGLang's capture-trace naming. `_profile_bs_idx` advances
once per profiler callback against a list built from `capture_bs`, while
`capture()` fires one callback per `(bs, lora, dsa)` combination. The fix is to
name each trace after the shape actually being captured, which keeps the per-bs
layout intact and additionally corrects the mislabelled traces the current
exporter writes before it crashes. That belongs upstream, not in a Hyperloom
config switch, so it will not arrive as a revision of this PR.

Worth stating plainly so closing this does not read as the problem being solved:
the bug is unfixed on `main`. `profile_sglang.yaml` still sets
`SGLANG_GRAPH_BATCH_CAPTURE: 1` and the crash path is unchanged, so any model
capturing more than one graph variant per batch size â€” DSA, LoRA, or both â€” still
dies during startup graph capture. A model with more batch-size buckets than
variants does not crash at all and instead writes traces under the wrong batch
size. Tracked separately from this PR.

PR #1403, opened 2026-09-04
Score the rewrite on per-case speedups, not a ratio of aggregates

jiaqiang-dot-liu:
## Summary

`rewrite_by_flydsl` graded the rewrite on `source_ms / flydsl_best_ms`, a **ratio of aggregate timings**. forge-loop decides KEEP/REVERT on the **equal-weight mean of per-case speedups**, and it already carries `aggregate_regression_detail` precisely because those two statistics can disagree. So the pipeline searched one objective and published a verdict on another.

On a suite whose cases span orders of magnitude the gap inverts the verdict. Measured against the aiter `a16w16` baseline at n=k=6144 on MI355X, whose 13 cases run 0.0128 ms to 0.2290 ms:

| candidate | mean over cases | ratio of aggregates |
|---|---|---|
| 2x on the 8 cheap cases, 1x on the 5 expensive | 1.615x | 1.107x |
| 3x cheap, 0.9x expensive | 2.192x | 1.041x |
| **5x cheap, 0.8x expensive** | **3.385x** | **0.955x** |
| 10x on the single largest case only | 1.692x | 1.492x |

The third row is not a corner case for this work — it is the shape a FlyDSL port takes when it beats aiter's own small-M kernels but not hipBLASLt at large M. The old number called it a regression.

## Where the score comes from

The rewrite needs the **finished kernel against the original source**. forge-loop's own `mean_case_speedup` cannot answer that: its anchor is the port, so it measures what OPTIMIZE *added*. Both halves are already measured, and no extra driver run is needed:

- **source side** — the per-case timings from the `--ref-bench-mode` preflight
- **candidate side** — forge-loop's `_promote_best` records the kept candidate's per-case medians; `cli.py` now reports them next to `best_ms`, together with the loop's own unscored list

When OPTIMIZE produces nothing the port is the final kernel, so the preflight candidate measurement already describes it. The per-case timings and `best_ms` therefore always describe the same kernel, and the medians mean a single noisy reading cannot move the verdict.

## Threaded through the paths that decide

Publishing the statistic is not enough — every gate reads it from somewhere else:

- **`applyback`** writes `speedup` as the mean when per-case timings exist, and carries `mean_case_speedup`, `aggregate_speedup`, `speedup_basis`, both sets of case times and the unscored list in the manifest. The consumer gates on the manifest.
- **`forge_submit`** grades the publication gate and the published micro gain on `_rewrite_micro_speedup`, which **recomputes** the declared mean from the manifest's own per-case timings and grades on the recomputed value. A producer is a separate process, and a declared score its evidence does not reproduce is not a number to grade against. Booleans are rejected explicitly (`isinstance(True, int)` is true).
- **`kb`** applies the same preference in its improvement gate and records `speedup_basis` alongside `speedup`, because warm-start ranking compares records against each other and two numbers from different statistics are not comparable.
- **`improved`** is driven by the mean and is *not* withdrawn when the aggregate disagrees. This diverges from forge-loop deliberately, and the reason is stated where it is implemented: the loop publishes an aggregate wall time as its headline, while this layer answers "is the rewrite worth adopting?" on the statistic the search optimized. Vetoing on the aggregate would reinstate the gate this work removes, and every consumer reading `improved is True` — `forge_collective` among them — would reject exactly the rewrites this route looks for. The contradiction is still reported by name in `aggregate_regression`.

## Unscored cases

A case may be marked `unscored`: measured and guarded, but kept out of the score because its run-to-run spread swamps a real change. The exclusion belongs to the comparison, not to the reading, so **both sides pass their full case set and name the exclusions**. Filtering one side makes the two sets unequal and the comparison is refused outright; filtering both without declaring it lets an excluded case's noise carry equal weight in the mean. The union of both sides' exclusions is what stays out.

## Parsing

One shared `parse_case_timings` in `bench.py`, used by the bench tool and the rewrite contract, so the ids a coverage check accepts and the times a score is built from cannot come from two regexes that agree today. Duplicate case timings and unparseable case times are contract failures — matching bench.py, which refuses the same measurement — and are fatal on both driver paths. An unparseable case still declares coverage, so two paths reporting the same malformed case cannot pass the coverage check on the subset that parsed.

## When no mean is available

`speedup_basis` names which statistic produced the number. Two outcomes are kept apart because they call for opposite handling:

- **no per-case timings** — the driver reported only an aggregate, so its ratio is the best available answer and is labelled as such
- **case coverage mismatch** — the two paths timed different work, so no comparison is published at all and `speedup_unavailable_reason` says why

## Test plan

- [x] `pytest src/kernelforge/tests/test_rewrite_by_flydsl.py` — 54 passed
- [x] `test_rewrite_by_flydsl_applyback.py` + `test_rewrite_by_flydsl_pipeline.py` + `test_rewrite_by_flydsl.py` — 135 passed
- [x] `test_rewrite_driver_contract.py` + `test_rewrite_by_flydsl_kb.py` + `test_flydsl_rewrite_driver_preparation.py` + `test_rewrite_agent_kb.py` — pass
- [x] hyperloom gate suites (`test_forge_best_result_authority`, `test_forge_collective`, `test_forge_export_restore`) — 234 passed
- [x] `ruff check` clean on every changed file
- [x] New coverage: the 3.385x vs 0.955x divergence on measured timings; the submission gate and the KB gate graded on the mean; the manifest hand-off; the declared-mean cross-check and the boolean rejection; an unscored case neither blocking the score nor entering the mean; pre-filtering one side being what refuses to score; malformed timings fatal on both paths; both parsers agreeing
- [x] `test_loop_runner.py` has 15 pre-existing failures in this environment (missing `pytest-asyncio` config); verified identical with the branch stashed
- [ ] End-to-end `forge-rewrite-by-flydsl` run on gfx950

## Notes for review

The end-to-end run is the one item not exercised. What the design buys is that there is no new driver invocation to validate: the candidate-side timings come from the measurement forge-loop already takes, so the added paths are the plumbing that carries them and the gates that read them, all of which are unit-tested.

`_publish_patch`'s new parameters are required rather than defaulted. Defaults would let a future caller silently omit the mean and fall back to the aggregate at the submission gate, which is the exact failure this PR is about.


jiaqiang-dot-liu:
Thanks — the first blocking point was the real one: I published the statistic and left every gate dividing aggregates, so on my own example (mean 3.385x / aggregate 0.955x) the artifact was still rejected and still recorded as a 0.955x regression. Pushed e051e6b.

## Blocking

**1. Decision paths now read the mean.** The gap was that the mean stopped at the result JSON while the consumer gates on the *manifest*.

- `applyback._publish_patch` writes `speedup` as the mean when per-case timings exist, and carries `mean_case_speedup`, `aggregate_speedup` and both sets of case times.
- `forge_submit` grades the publication gate **and** the published micro gain on a new `_rewrite_micro_speedup(applyback)`, which reads the producer's mean first and falls back to the aggregate pair only for a driver that reported no cases. The old `best_ms >= baseline_ms` comparison and the separate `baseline_ms / best_ms` publish are both gone.
- `kb.write_flydsl_kb_solution` takes `mean_case_speedup` and applies the same preference in its improvement gate.

**2. `improved` is no longer withdrawn.** You're right that this made the whole change inert for `forge_collective.py:846`. For this layer the mean is authoritative — letting the aggregate veto it restores exactly the gate the PR removes. `aggregate_regression` still reports the contradiction, as a record rather than a verdict. My test now asserts `improved is True` on the 3.385x case.

**3, 4, 5 plus both minor runner items: the re-measurement is deleted.** All five came from it, and it was the wrong mechanism — forge-loop had already measured what I needed. `loop.runner._promote_best` records the kept candidate's per-case medians; `cli.py` now reports them next to `best_ms`. So:

- no single reading replaces the loop's medians (`opt["best_ms"]` is never overwritten)
- no extra driver run, so nothing draws on `deadline_unix` or the apply-back reserve
- no fatal-failure asymmetry, because there is no second check to downgrade
- the case times and `best_ms` cannot describe different kernels: either both come from the loop's best, or OPTIMIZE produced nothing and both come from the port that is then the final kernel

`best_kernel_restored is False` now falls back to the port too, instead of attributing a timing to a file that is no longer in the worktree.

## Minor

- **`driver_contract.py:56`** — one shared `parse_case_timings` in `bench.py`, used by both. The anchored pattern was the worse half of that bug: a collectives driver's `[rank0] ` prefix read as *no cases at all* and degraded silently. Unscored cases are now excluded from the mean via `DriverReading.scored_case_times`, while coverage checking still sees every id. Tests cover the prefix, the marker, and that both callers agree.
- **`driver_contract.py:157`** — duplicate case timings now fail the contract with `DUPLICATE_CASE_TIMINGS`, matching bench.py.
- **`report.py:141`** — `cross_language_mean_case_speedup` returns `(speedup, reason)`. A coverage mismatch publishes **no** speedup (`speedup_basis == ""`, `speedup_unavailable_reason == "case_coverage_mismatch"`); only "driver reported no per-case timings" falls back to the aggregate. The test that claimed "refuses to score" now actually asserts a refusal.
- **`report.py:23`** — `field` removed; `ruff check` is clean on every touched file.

One deliberate choice worth flagging: `_publish_patch`'s three new parameters are required rather than defaulted. Defaults would let a future caller silently omit the mean and fall back to the aggregate at the submission gate, which is the exact failure this PR is about. That cost one existing test an update.

## Tests

`test_rewrite_by_flydsl.py` 50 passed, `test_rewrite_by_flydsl_applyback.py` + `test_rewrite_driver_contract.py` + `test_rewrite_by_flydsl_kb.py` 139 passed together, and the hyperloom gate suites (`test_forge_best_result_authority`, `test_forge_collective`, `test_forge_export_restore`) 234 passed. New coverage pins the submission gate and the KB gate to the mean, the manifest hand-off, and the four parsing behaviours above.

Still not exercised end to end on hardware — that item stays unchecked in the description. What the deletion of the re-measurement buys is that the remaining path is the one forge-loop already runs, so there is no new driver invocation left to validate.

Issue #1320, opened 2026-08-28
is_multigpu_kernel substring-matches collective tags: 'smallreduce' reserves a second GPU


zoroyihan7:
## Problem

`is_multigpu_kernel` decides whether a kernel is a multi-GPU collective using a bare substring test with no word boundaries:

https://github.com/AMD-AGI/Hyperloom/blob/main/src/hyperloom/agents/kernel/tools/tracelens_analysis.py#L3543-L3573

```python
blob = f"{name} {source_file}".lower()
return any(
    tag in blob
    for tag in ("all_reduce", "allreduce", "all_gather", "allgather",
                "reduce_scatter", "broadcast", "p2p", ...)
)
```

So ordinary single-GPU kernels get classified as collectives:

| kernel name | matches tag | result |
|---|---|---|
| `smallreduce` | `allreduce` | ✗ false positive |
| `broadcastable_check` | `broadcast` | ✗ false positive |
| `broadcast_add` | `broadcast` | ✗ false positive |

A source path containing `dist/` (e.g. `.../redist/...`) trips the `"dist/"` tag the same way.

## Impact

This is not a cosmetic label. The classification drives GPU reservation:

- `item["num_gpus_recommended"] = 2 if item["is_multigpu"] else 1` — `tracelens_analysis.py:4736`
- `num_gpus = max(1, args.num_gpus or candidate["num_gpus_recommended"])` — `kernel_optimization.py:1578` / `:2101`
- → `forge.submit(num_gpus=num_gpus, ...)` — `kernel_optimization.py:2129`, which reaches the Ray runtime `--num-gpus` reservation.

So a false positive **reserves a second GPU** for a kernel that only needs one. This is the mirror image of #559 (which fixed under-allocation on the same `num_gpus_recommended` path).

## There is already a correct implementation

`_collective_names.kernel_name_implies_multigpu` does word-bounded matching over a normalized name (camel-split, non-word chars collapsed, Itanium digit-prefix handling):

```python
re.compile(r"(?:^|_)all_?reduce(?:_|$)")
re.compile(r"(?:^|_)broadcast(?:_|$)")
...
```

`is_multigpu_kernel` predates it and was never migrated.

## This needs a decision, not just a patch

Worth flagging before someone "fixes" it: `_collective_names.py` documents the **opposite** intent in its module docstring —

> *"Name-pattern fallback for multi-GPU collectives TraceLens missed; false positives are cheap so we bias toward them."*

[...]

PR #1405, opened 2026-09-04
fix(kernel): stop the name heuristic unresolving lane-identified collectives

zoroyihan7:
Refs #1320. **Please read the correction below — I filed #1320 and its stated impact was wrong.**

## The claim in #1320 that does not hold

#1320 says a false positive "reserves a second GPU" via `forge.submit(num_gpus=...)` → Ray `--num-gpus`. I traced the call chain when I filed it but never checked whether the far end consumed the argument. It didn't: #1381 removed the parameter with the message *"submit() accepted num_gpus and prefer_ray and **read neither**"*. So the reservation cost was never real, and today `num_gpus_recommended` only reaches prompt text (`kernel_optimization.py:1663-1668`, `:1831-1833`) and `_invocation_spec.py:872`.

Investigating properly also inverted the priority. The false positives #1320 leads with — `smallreduce`, `broadcastable_check` — appear **only as probe names in `test_collective_names.py`**. Sweeping every kernel-name-shaped literal in the repo turned up no false positive in trace-derived data, and no non-collective source path matches either (`dist-packages` ≠ `dist/`; `torch/distributed/` doesn't contain `dist/`; `communicator` doesn't match `communication.py`).

The **false negatives are the observed ones**, and this PR fixes those.

## The defect

`_stamp_candidate_metadata` recomputed `is_multigpu` from the name, overwriting what `_nccl_summary_candidates.py:266` had established from TraceLens' `nccl_summary` table plus a resolved device symbol:

| row | lane | after the stamp |
|---|---|---|
| `small_collective` | True | **False** |
| `EpDispatchIntraNodeKernel_bf16` | True | **False** |
| `ncclDevKernel_Generic_1` | True | **False** |
| `all_to_all_dispatch` | True | **False** |
| `cross_device_reduce_2stage` | True | True |

The flip is self-defeating. `is_collective_candidate` (`_kernel_decisions.py`) gates the collective lane on `candidate_source == "nccl_summary"` **and** `is_multigpu` — so the lane refuses rows it produced itself, and an 8-rank collective gets told it has one GPU and should write a rank-slice surrogate.

`small_collective` isn't hypothetical: it's a real `__global__` in the repo's own fixture, resolved by device-symbol lookup and then un-resolved by a substring guess.

## The change

A positive `nccl_summary` verdict now outranks the name. The heuristic still runs for every other row, **and** still runs for an `nccl_summary` row that did not claim multi-GPU, so it can still raise a collective the lane missed.

Also fills three vocabulary gaps. The list carried both spellings of `all_reduce` and `all_gather` but only the delimited `reduce_scatter`, and no all-to-all form at all — so `ReduceScatterFusion`, `all_to_all_dispatch` and `rccl_AllToAll` read as single-GPU while `ncclDevKernel_AllGather` did not. The word-bounded sibling `kernel_name_implies_multigpu` already agrees with all three.

## Mutation results

| mutation | result |
|---|---|
| delete the guard (restore the clobber) | 🔴 |
| guard always on (never re-derive) | 🔴 |
| drop the `candidate_source` qualifier | 🔴 |
| drop the `is_multigpu` qualifier | 🔴 |
| remove `reducescatter` / `all_to_all` / `alltoall` | 🔴 ×3 |

Tests drive the real `_stamp_candidate_metadata` and assert `num_gpus_recommended`, not the bare boolean. Kernel suite: 2049 passed vs 2046 on `main`, same 10 pre-existing failures.

## Deliberately not in this PR

**Word-bounding the tags.** Delegating to `kernel_name_implies_multigpu` is a **regression** — it returns False for `cross_device_reduce_2stage`, `cross_device_reduce_1stage`, `send_recv_kernel` and `p2p_copy`, all of which `_enrich_kernel_contract` enumerates as an all-reduce family. It also fixes only 2 of the 3 FPs #1320 names (`broadcast_add` still matches). Doing it safely needs the sibling's normalisation machinery plus restored `cross_device`/`p2p`/`send_recv` patterns, and it would flip one real trace row (`EpCombineIntraNodeKernel_bf16_nop2p`, which currently matches on the *negation* `nop2p`). That is a real decision with a real cost, and with the GPU-reservation impact gone the remaining cost is prompt wording — so it should be weighed on its own, not smuggled in here.

I'll update #1320 with the corrected analysis so the FP question isn't ranked off a chain that no longer exists.

### DIFF
--- a/src/hyperloom/agents/kernel/tests/test_collective_candidate_routing.py
+++ b/src/hyperloom/agents/kernel/tests/test_collective_candidate_routing.py
@@ -128,3 +128,70 @@ def test_nccl_marker_check_precedes_source_resolution():
     item = _candidate("ncclDevKernel_AllReduce", _AITER_SRC)
     reusable, _reason = tl.classify_patchability(item)
     assert reusable is False
+
+
+# --- A resolved lane verdict outranks the name heuristic ---------------------
+
+
+def _nccl_summary_row(name: str, source_file: str) -> dict:
+    """A row shaped like the one _nccl_summary_candidates.py:266-267 emits."""
+    return _candidate(name, source_file, candidate_source="nccl_summary")
+
+
+def test_lane_resolved_collective_is_not_unresolved_by_its_name():
+    """A name the heuristic reads as single-GPU must not clobber the lane's verdict.
+
+    _nccl_summary_candidates sets is_multigpu=True from TraceLens' nccl_summary
+    table plus a resolved device symbol. Re-deriving from the name flips these
+    rows to False, and is_collective_candidate then refuses them because it
+    gates on candidate_source == "nccl_summary" AND is_multigpu -- the lane
+    rejecting its own output.
+    """
+    for name in ("small_collective", "EpDispatchIntraNodeKernel_bf16", "ncclDevKernel_Generic_1"):
+        assert tl.is_multigpu_kernel(name, "") is False, f"{name} should be a name-heuristic miss"
+        item = _nccl_summary_row(name, "/sgl-workspace/aiter/csrc/include/other.cuh")
+        tl._stamp_candidate_metadata(item, None)
+        assert item["is_multigpu"] is True, name
+        assert item["num_gpus_recommended"] == 2, name
+
+
+def test_the_name_heuristic_still_runs_without_a_lane_verdict():
+    """The override is scoped: only a positive nccl_summary verdict is preserved."""
+    # No candidate_source at all -> derived from the name.
+    item = _candidate("small_collective", "/w/other.cuh")
+    tl._stamp_candidate_metadata(item, None)
+    assert item["is_multigpu"] is False
+    assert item["num_gpus_recommended"] == 1
+
+    # A different lane does not get the same authority.
+    item = _candidate("small_collective", "/w/other.cuh", candidate_source="other_bucket_fallback")
+    tl._stamp_candidate_metadata(item, None)
+    assert item["is_multigpu"] is False
+
+    # nccl_summary that did NOT claim multi-GPU is still derived, not forced True.
+    item = _candidate("plain_gemm_kernel", "/w/gemm.cu", candidate_source="nccl_summary")
+    item["is_multigpu"] = False
+    tl._stamp_candidate_metadata(item, None)
+    assert item["is_multigpu"] is False
+    assert item["num_gpus_recommended"] == 1
+
+    # ...and the derivation still RUNS for such a row rather than being skipped:
+    # the lane left is_multigpu False, but the name is an unmistakable collective,
+    # so the heuristic must be allowed to raise it. Without the is_multigpu
+    # qualifier on `authoritative` this row would stay False.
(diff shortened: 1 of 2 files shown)


zoroyihan7:
## Integration check on the real chain

The unit tests in this PR call `_stamp_candidate_metadata` with hand-built rows. To confirm the defect and the fix on the path that actually runs, I drove the **real producer** end to end on both trees — a genuine TraceLens metrics file (`category_data/multi_kernel_metrics.json` with `nccl_summary.top_ops`) plus a device-source tree, so `extract_collective_candidates` resolves each symbol by real lookup rather than being handed a dict:

```
extract_collective_candidates   (real producer, real symbol resolution)
  -> _stamp_candidate_metadata  (the line this PR changes)
  -> _enrich_kernel_contract
  -> is_collective_candidate    (the collective lane's admission gate)
```

### Baseline — `origin/main` @ 2950de599

| kernel | lane says | after stamp | `num_gpus_recommended` | lane admits |
|---|---|---|---|---|
| `small_collective` | True | **False** | 1 | ❌ |
| `EpDispatchIntraNodeKernel_bf16` | True | **False** | 1 | ❌ |
| `all_to_all_dispatch` | True | **False** | 1 | ❌ |
| `cross_device_reduce_2stage` | True | True | 2 | ✅ |

### With this PR — da2f3d9f1

| kernel | lane says | after stamp | `num_gpus_recommended` | lane admits |
|---|---|---|---|---|
| `small_collective` | True | **True** | **2** | ✅ |
| `EpDispatchIntraNodeKernel_bf16` | True | **True** | **2** | ✅ |
| `all_to_all_dispatch` | True | **True** | **2** | ❌ (see below) |
| `cross_device_reduce_2stage` | True | True | 2 | ✅ |

Three of four rows the lane resolved were being rejected by the lane itself: `_nccl_summary_candidates` identifies them by device-symbol lookup, the name heuristic overwrites that verdict, and `is_collective_candidate` then refuses them at `_kernel_decisions.py:1190` because `is_multigpu` is no longer True.

### One row is still refused, and it is not this PR

`all_to_all_dispatch` now carries `is_multigpu=True` and `num_gpus_recommended=2` correctly, but the lane still declines it at a **different gate**:

```python
SUPPORTED_COLLECTIVE_OPS = frozenset({"all_reduce", "reduce_scatter", "all_gather"})  # :50
```

`_enrich_kernel_contract` gives it `collective_op="all_to_all"`, which isn't in that set, so `_kernel_decisions.py:1186` rejects it. That is the lane's current capability boundary, unrelated to this change.

### Also worth flagging, also not this PR

`small_collective` and `EpDispatchIntraNodeKernel_bf16` both come out of `_enrich_kernel_contract` with `collective_op="all_reduce"`. For an EP dispatch kernel that is probably wrong — it's an all-to-all. The contract's op is inferred separately from `is_multigpu` and I haven't touched it here, but someone who owns that path may want to look.

### What this does not show

This exercises the candidate-selection chain only. It says nothing about the optimization behaviour of a genuine multi-GPU collective — that needs a real multi-rank e2e, and the box I ran this on has a single MI325X.

chaojhou (approved):

PR #1409, opened 2026-09-04
[Refactor]  Rework the enablement / bring-up round lifecycle

ZhengGong-amd:
49 sessions were reported stuck in PRELUDE (68 ended `enablement_stalled` across the wider
1256-session corpus). The cause is one latch. `enablement_close_guard_active()` suppressed
`skip_to_close` while a run was not yet enabled, and every declared exit depended on something
that latch blocked:

- `baseline_tput > 0` — `baseline` is denied by the latch, so throughput never becomes positive
  (40/49 hit `baseline:enablement_round_in_flight`, longest streak 380).
- `prelude_baseline_failed` — a denied action never *fails*, so the counter never advances.
- `enablement_stalled` — the rearm that advances the stall streak never ran; all 49 end with a
  non-empty `inflight_task_id` and a streak of ≤ 4.
- wall clock — reachable but late; the deadline is tested at the top of a tick whose body can
  block for `max_turns × per_turn_max_seconds`.
- `skip_to_close` — dropped by the guard; 18/49 emitted it, one 56 times, never routed.

The latch was a string field on in-process state: set before the work it guarded existed, cleared
on paths that could be skipped, read by the gate as authority, invisible outside the loop, and
reconciled against nothing.

## What changed

**The round is a durable, owned row.** Acquired by compare-and-acquire under `BEGIN IMMEDIATE`
with a fencing token and an owning task id, settled through an append-only outbox. Exclusion is
bounded by the lease, so a round nobody settles frees itself and a dead holder is visible to
someone else. Ray's custom resources remain the physical GPU mutex; these rows are the accounting
view (`state/round_store.py`).

**Reconciliation runs unconditionally, at the top of every tick.** Previously the self-heal sat
behind a cadence guard, a positive-tick guard and a modulus, inside the pump it was meant to heal.
It is now an independent pass with per-rule isolation — expired round, dead holder, unanswered
review, stranded revalidation — so one failing rule cannot skip the others.

**PolicyGate is a pure function again.** It reads an advisory projection built once per tick from
the round rows and the live task set, instead of a mutable latch it also probed the environment
during. The close guard is bounded and its count persisted, so a resume cannot reset it and latch
again.

**Preflight refuses an unlaunchable round before it costs a launch.** Server argv is sealed and
digested by one finalizer, then validated against the framework's own parser in the interpreter
that will serve; an unknown flag is a named terminal, with one repair allowed. Environment checks
run cheapest-first. A missing checkpoint is terminal; a framework that does not import yet, an
unbuildable extension and a held port are *unavailable* verdicts, not faults — installing and
building is what a round does.

**Boot outcomes are classified once and persisted.** `BootObservation` / `LadderStage` carry a
stable `failure_digest`; every baseline attempt persists one, and both halves of a comparison read
the artifact by path rather than re-deriving a verdict. A session-start tree identity makes digests
comparable across rounds.

**Delivery is verified.** A pre-round baseline records what each declared target held, for git and
non-git trees alike (pip-installed frameworks have no git to diff). The non-git apply drops
`patch`'s fuzz and rejects an offset-only match, so a tree that merely resembles the post-state no
longer reads as already-applied. Backups go to an fsynced ledger, so a process that dies mid-apply
still leaves a possible revert.

**Reap and deadlines.** A reap collects the descendant set before signalling (a child that left
its group with `setsid` is still reachable) and reports what its outcome actually proves, never
that the tree is gone. `Deadline` is an absolute monotonic instant that crosses the subprocess
bridge, so a budget cannot be re-anchored by the crossing.

**Out-of-band supervision.** A separate process reads a tick stamp — written off the loop thread,
since it fsyncs onto a possible network mount — and signals the coordinator when the tick stops
advancing. It never opens the session database (its journal mode is unsafe with a second writer
on NFS), never transitions round state while the coordinator is alive, and runs with control-plane
credentials scrubbed.

**Testability.** A rehearsal harness scripts the launch boundary and the specialist and installs a
virtual clock at the `time` module, which is what makes a multi-tick round-lifecycle test
writable. Excluded from the wheel; nothing shipped imports it.

## Protocol change

Session breakdown goes to **v6.0**. The round ledger is additive within it: a section under
`enablement`, which is already carved out of the stability guarantee. The stop-reason vocabulary
now has one definition instead of two drifted copies, and the CLI exit code reads the same set the
report grades against. `enablement_stalled` keeps its slot for archived sessions but no longer has
a producer — the monotone attempt cap (`enablement_attempts_exhausted`) is what bounds a session
that cannot boot. Field dispositions are in `docs/reference/session-breakdown.md`.

## Testing

- `pytest src/hyperloom/orchestrator src/hyperloom/common`: 755 passed.
- `ruff check` / `ruff format --check`: clean.
- The `inference_optimizer` suite has pre-existing sandbox failures unrelated to the branch
  (ambient `ANTHROPIC_API_KEY`, a missing TraceLens asset, a wall-clock-sensitive budget test);
  they fail identically at base.

Run tests with an **absolute** `PYTHONPATH`; a relative `PYTHONPATH=src` breaks every test that
spawns a subprocess from a different cwd and manufactures phantom failures.

## Size

169 files, +17,348 / −2,979. Large for the problem. `audit.md` lists every commit with its purpose
and where new code was written instead of reusing an existing path — read it before the diff.

## Not done

- Forensics on the original 49 sessions: no `sacct`, `dmesg` or journal access here, so the
  external cause of the first stuck holder is inferred, not proven.
- Resume is deliberately unhandled; the round store is new state and old sessions do not carry it.


xiaofei-zheng (approved):
Reviewed at head `db4b42e7` against merge base `f705855a` (`git merge-base origin/main refactor/zgong/enablement-rework`), non-test surface only: 110 files, +10,851 / −2,006.

**No blocking issues found.** Approving.

### What I read

Core new machinery, in full:

- `orchestrator/state/round_store.py` — CAS acquire under `BEGIN IMMEDIATE`, fencing token, lease-bounded `_LIVE_EXCLUSION`, `settle` outcome validation + replay detection, `consecutive_stalled` ordering, `round_events` outbox.
- `orchestrator/bringup/reconcile.py` — per-rule isolation, `_UNDECIDED_PROPOSALS_SQL` NULL handling, `_TIMEOUT_DENY_SQL` insert-if-absent so a real verdict wins the race, `_resolve_open_rounds` / `_advance_or_expire` / `_expire`, `_rebuild_projection`.
- `orchestrator/bus/resource_lock.py` — `BRINGUP_ROUND_LANE` deliberately absent from `KNOWN_LANES`, `ROUND_LEASE_PID = 0` skipping `reap_dead_holders`, idempotent `hold_round_lane`.
- `orchestrator/policy/projection.py` + `policy/gate.py` — the gate is now pure/synchronous; `_validate_baseline_not_mid_round` is unconditional and admits the round's own holder; `RULE_LEASE_NOT_LIVE`; `CORE_STATE_FIELDS` swap.
- `orchestrator/enablement/lane.py` (789-line diff) and `enablement/params.py` — gate order, `_open_authoring_round`, renew/handoff/settle, `_pump_enablement_safely`.
- `orchestrator/state/shared_state.py` — leg-based budget (`elapsed_charged_sec` / `leg_anchor_unix`), `from_dict` back-fill, `MAX_SKIP_TO_CLOSE_SUPPRESSIONS`.
- `common/deadline.py`, `loop/offload.py`, `loop/signals.py`, `loop/coordinator.py`, `phases/close.py`.
- `bus/storage/schema.py` + `connection.py`, `common/proctree.py`, `actions/executors/_nogit_patch.py`, `bringup/argv_preflight.py`, `supervisor/watch.py`, `breakdown/stop_reasons.py`.

### Things I specifically chased and cleared

- **The original wedge is genuinely fixed and bounded.** `enablement_close_guard_active()` (`shared_state.py:1505`) short-circuits at `skip_to_close_suppressions >= 5`, and `intent_router.py:1045` increments on every suppression, so the latch can no longer be the sole authority denying a session its last exit.
- **`_lease_iso()` vs `now_iso()`** produce byte-identical shapes (`...T...:...:....000000+00:00`), so the string comparison the lease TTL sweep uses is sound.
- **`deadline_unix` is fully removed** as a `SharedState` field; the remaining repo hits are unrelated `forge_submit.py` / `forge_collective.py` parameters, so the `CORE_STATE_FIELDS` swap leaves no forgeable budget field behind.
- **`params["enablement_candidate_refs"]` / `["enablement_failure_kind"]`** are always present when params is non-`None`, so the lane's direct indexing cannot `KeyError`.
- **Every `_maybe_rearm_enablement` / `_settle_enablement_round` / `_handoff_enablement_round` call site is awaited** after the sync→async conversion (`build.py:390,414`, `intent_router.py:447`, `phases/framework.py:362`, `writeback.py:954,986,3208,3219`, `proposals.py:531`).
- **A round leaked by an early return** (e.g. `_maybe_rearm_enablement` returning on a non-enablement result) self-heals: the reconciler treats the lease table as the round's clock and expires it after `TERMINAL_HOLDER_CAP_SEC`. This is the point of the durable design, and it holds.
- **`ResourceFacts.read = False` denies nothing** (`round_excludes` is `False` on a default projection), so a gate with no repair pass behind it cannot refuse a dispatch on facts nobody looked up. The one-tick staleness window between opening a round and the next `_rebuild_projection` is advisory only — the store's CAS is the actual authority.
- **`SignalDrain`'s `set_wakeup_fd`** cannot be tripped by other in-process Python handlers: only SIGINT/SIGTERM are installed here (`robustness/main.py` is a separate process), and on 3.12 `ThreadedChildWatcher` installs no SIGCHLD handler.
- **`RoundStore.renew` does not test lease expiry**, but an expired round is settled by the reconciler at the top of every tick before anything can acquire, and a settled round fails the `state = 'open'` CAS. No path reaches a renew of a round someone else has taken.
- **Schema v4 → v5 is purely additive** (`bringup_rounds`, `round_events`, both `IF NOT EXISTS`, both registered in `_MANAGED_TABLES`); an existing v4 database just gains the tables. No destructive migration.
- **Journal mode**: the default is still WAL, so no behaviour change for existing deployments; the new read-back check only converts SQLite's silent refusal into an explicit `JournalModeError`.
- **Stop-reason vocabulary stays backward compatible**: `enablement_stalled` is retired as a *producer* but retained in `machine_state.py:253` and the `report.py` description map, so historical sessions still read. `enablement_attempts_exhausted` is registered in all three places and maps to `failed`. No import cycle from `stop_reasons.py` → `supervisor/watch.py` (verified by importing it).
- **`proctree.proc_identity` / `group_members` field offsets** are correct after the `rfind(')')` split (ppid at `[1]`, pgrp at `[2]`, starttime at `[19]`), and the comm-with-parens case is handled.
- **`argv_preflight`** only reaches the terminal `INVALID` verdict when the parse probe succeeded *and* the probe interpreter's identity exactly matches the serving one; every indeterminate reading returns non-terminal `UNAVAILABLE`. That is the right bias for a verdict that stops a run.

### Verification I ran

Minimal, scoped to the code above — not the full suite (CI covers that). `PYTHONPATH=<abs>/src`:

- Direct-call repro against `RoundStore` on a temp DB: second concurrent `open()` refused with `excluded`; `excluding()` reports the live round; settle → replay returns `ok=True, duplicate=True`; acquire succeeds after settle and after lease expiry; `not_owner` / `stale_fence` CAS refusals; `ValueError` on an unknown outcome; `consecutive_stalled` counts FAILED rounds and resets to 0 across a `booted`.
- Direct import + `outcome_status()` check for `enablement_attempts_exhausted` / `enablement_stalled` / `supervisor_tick_stalled`.
- `pytest test_round_store.py test_reconcile.py test_deadlines.py test_sqlite_journal_mode.py test_supervisor.py` → **74 passed**.
- `pytest test_bringup_round_scenario.py test_argv_refusal_round.py test_environment_fault_round.py test_stop_reason_outcomes.py test_exit_code_stop_reason.py` → **43 passed**.
- `py_compile` over every changed `.py`, and `ruff check src/hyperloom/orchestrator src/hyperloom/common` → clean.

### Non-blocking, recorded only so it is not lost

Nothing here should hold the merge:

- `_apply_patch_no_git` passes `--reject-file=-` (discard) but `_collect_rej_files` then scans for `*.rej`, so `ApplyFeedback.rejected_hunks` will always come back empty on the nogit channel — degraded diagnostics, not wrong behaviour.
- `Supervisor._end` never escalates past the single SIGTERM: a coordinator that survives the ask is left running and reported, which the code documents deliberately, but the docstring on `_escalate_wedged` reads as though a kill follows.
- `offload()` in `lane.py` is called without `default=`, so a 45s timeout and a genuine "nothing to author against" both collapse to `None` and route to `_maybe_record_enablement_human_review`. In practice `_discover_enablement_candidate_refs` self-bounds at the same deadline and returns partial results, so the timeout branch is unlikely to fire; the next tick retries either way.

Resume being unhandled is fine as stated — the round store is new state and pre-existing sessions carry none of it.

PR #1410, opened 2026-09-04
[Stepping] Stateless orchestration: one full projection per turn

ZhengGong-amd:
Replaces the SEED/DELTA conversation model with stateless turns, tightens bus
delivery and liveness, completes the tested-variant ledger, and removes the
machinery all of that made dead.

`8 commits · 121 files · +1049 / -4521`

## What changes

**Bus (`26513b0`)** — `message_bus` enforces per-role topic subscriptions and
excludes self-echo. `replay_for` now default-denies unknown agents instead of
falling back to delivering every message.

**Liveness (`5233534`)** — heartbeat envelopes and event-scan inference are gone.
The coordinator stamps timestamps directly; the critic emits an *idle* intent on
the `observation` topic instead of a heartbeat.

**Stateless turns (`17f1e2f`)** — deletes SEED/DELTA gating, the provider
conversation thread (session id, resume, `reset_conversation`, `needs_seed_for`,
`conversational`), and the whole checkpoint subsystem. Every orchestration turn
now carries one unconditional full-fidelity projection. `orchestration_memory` is
produced from exactly one place: a dedicated LLM call at the macro-cycle boundary
feeding `next_cycle_directive` into the system-prompt rebuild, with
`_cycle_directive_fallback` as the deterministic path.

**Ledger, findings, reverse paths, guards (`dde1720`)**
- *Ledger*: render every tested variant with decision-bearing fields only, no
  truncation; surface re-proposal warnings in `last_action_failures`.
- *Findings*: specialist findings are domain-agnostic; `specialist_done` unpacks
  into inbox events; the `empty` field is gone (`proposal_set` emptiness plus a
  `summary` rationale is the authoritative signal).
- *Reverse paths*: discarded-escalation advisory, cancelled queued tasks surfaced
  via observation + `record_action_failure`, and an explicit `mode`
  (`patch`/`env_only`) replacing integrate shape-inference.
- *Guards*: `NO_PROGRESS_TICKS` and `no_levers_min_ticks` deleted; maintenance
  cadence and alert cooldown moved to wall-clock seconds; new
  `gain_gated_action_count` windows `gain_plateau` on real measurements.

**Cleanup sweep (`507e0ca`, `921bc4c`, `a7debcf`)** — drops the now-dead
`conversation_progress` signal end to end, `_enrich_with_tested_gain()`,
`is_degenerate_checkpoint()`, and stale coordinator delegations; renames
checkpoint/compaction to macro-cycle memory handoff (`MEMORY_REQUEST_PROMPT`,
`parse_memory_reply`, `last_capture_*`, `capture_count`), `gain_window_ticks` to
`gain_window_actions`, `cooldown_ticks` to `cooldown_sec`; aligns
`orchestration.md`, `optimization-loop.md`, `session-breakdown.md` and the four
agent SKILL/reference files with the runtime.

## Tests (`d238ad4`)

- **Deleted**: `test_codex_conversation_continuity.py`,
  `test_orchestration_memory_unit.py`, and three `test_inbox_parser*` files —
  they assert SEED/DELTA, checkpoint, and inbox-header behaviour that no longer
  exists.
- **Added**: `test_bus_delivery.py` (self-echo exclusion, subscription denial),
  `test_gain_gated_counter.py`.
- **Updated**: renamed symbols across the suite; dropped assertions for
  `compactions_per_tick`, `seed_prompts`, `delta_ratio`.

## Notes for reviewers

Breaking for anything reading the removed config fields (`cooldown_ticks`,
`gain_window_ticks`, `no_levers_min_ticks`) or the `empty` field on specialist
envelopes. Prompts and docs are updated in the same branch, so agent-facing
protocol text and runtime should not drift.


xiaofei-zheng:
## What this PR does

Moves orchestration from "persistent conversation + SEED/DELTA incremental pushes + checkpoint compaction" to a stateless, full-state-projection-per-turn model, and cleans up the surrounding machinery:

- **Conversational state removed**: `ClaudeBackend.conversational` / SDK `resume` sessions, `CodexBackend._thread_seeded` / `needs_seed_for`, `reset_conversation`, `_orchestration_needs_seed` are all gone; `ClaudeBackend` now selects its reasoning-effort tier via an explicit `effort_role`. The whole checkpoint/compaction subsystem (`CheckpointPolicy`, `CheckpointTracker`, `MODEL_CONTEXT_WINDOWS`, `render_memory_for_seed`) is deleted.
- **Macro-cycle handoff instead**: new `ExplorePhase._capture_cycle_memory()` asks orchestration for a JSON working-memory record at the cycle boundary; `next_cycle_directive` is injected into the next cycle's system prompt (falling back to the deterministic `_cycle_directive_fallback`).
- **Bus tightened**: `TOPIC_ALLOWLIST` reduced to 12 topics, new `ROLE_SUBSCRIPTIONS`, `replay_for` filters by subscription and excludes self-echo; the `heartbeat` topic is retired in favour of coordinator-stamped `agent_last_active` timestamps, rendered as `agent=Ns ago` and re-parsed by the robustness stall signal.
- **Guards de-tick-ified**: `MAINTENANCE_EVERY_TICKS=50` -> wall-clock `MAINTENANCE_INTERVAL_SEC=1800`; robustness action-ladder `cooldown_ticks` -> `cooldown_sec`; the plateau window now keys on the new `gain_gated_action_count` (rounds that actually produced a throughput measurement).
- **Protocols made explicit**: `integrate` takes an explicit `mode: patch|env_only` instead of inferring from payload shape, validated in `policy/gate.py`; the specialist envelope drops the `empty` field in favour of `proposal_set` emptiness + `summary`; `_harvest_research_scout` generalises to `_harvest_specialist_findings` for any domain, triggered on `new_findings`.
- **GPU-leak teardown**: `PR_SET_PDEATHSIG` SIGKILL -> SIGTERM, `GpuSpecialistLease.close()` calls `stop()` before `ray.kill`, wider server cmdline markers (`sglang.srt`, `EngineCore`, `VLLM::Worker`, ...), and orphan reaping also runs at shutdown.
- **Rendering**: `explore_search.tested` is now enumerated in full (KEEPs first, untruncated), replacing the previous last-5-accepted / last-15-rejected view.

## Blocking comments

### 1. `integrate` defaulting to `mode="patch"` will silently measure an unrelated patch for LLM-authored env-only requests

`src/hyperloom/orchestrator/kernel/request_handlers.py:6605`

On the merge base, env-only was decided by payload shape (no `patch_path`/`target_file`/`source_file`, but `extra_envs`/`extra_server_args` present). The deleted comment stated the reason explicitly: it "keeps such a request away from `_resolve_integrate_payload`, which would otherwise back-fill patch_path/target_file from the last kernel optimization and silently measure an unrelated patch."

The mode is now caller-supplied and defaults to `patch`. All three in-tree producers were updated (`phases/kernel.py:2663`, `phases/kernel.py:3716`, `phases/kernel_stack.py:650`), but nothing enforces it for an `integrate` the orchestration LLM writes itself. If the LLM emits `integrate{kernel_id, extra_envs, base_tput}` and omits `mode`, the request falls into the `else` branch, `_resolve_integrate_payload` back-fills `patch_path` and `source_file` from `last_kernel_opt` / the per-kernel ledger (`request_handlers.py:1334-1343`), and the patch is actually applied and benchmarked — with the result attributed to the env change. On a KEEP this can also re-promote a patch that is already on `optimization_stack`. There is no warning anywhere on this path; the output is wrong attribution data.

`policy/gate.py:1109` only validates the value when present, so it does not catch this. Suggest keeping a shape-based fallback: when `mode` is absent and the payload names no artifact but carries env fields, still treat it as `env_only` (or reject and require an explicit `mode`).

### 2. Only `next_cycle_directive` is consumed from the macro-cycle memory; the other fields are dead state, and the prompt tells the agent otherwise

`src/hyperloom/orchestrator/state/orchestration_memory.py`, `src/hyperloom/orchestrator/prompts/orchestration.md:32`

`MEMORY_REQUEST_PROMPT` asks the agent for six fields (`current_plan`, `hypotheses`, `tried_and_why`, `pending`, `learnings`, `next_cycle_directive`), and `build_memory_record` does non-empty-wins merging, learnings dedup and a cap of 50 before persisting. But this PR deletes `render_memory_for_seed`, and across the tree (tests excluded) `orchestration_memory` is read in exactly three places — `phases/explore.py:334`, `phases/machine.py:471`, `cli/__init__.py:2791` — all of which take only `next_cycle_directive`. (The `tried_and_why` / `learnings` in `phases/framework.py` belong to framework's own separate memory.) The other five fields never reach any prompt again.

Meanwhile `orchestration.md:32`, added by this PR, tells the agent: "The Coordinator persists it and pastes it back as `=== Your working memory (recovered) ===` on later turns, so put intent and rationale there". That block no longer exists — its only producer was the deleted `render_memory_for_seed`, and nothing else emits it. So the agent is instructed to write cross-cycle intent and rationale into something that is never fed back, costing one LLM call per cycle boundary for discarded output, and is given a description of the system that contradicts its actual behaviour. Either wire the record back into the prompt, or drop that promise from the prompt along with the fields that are no longer consumed.

### 3. `explore_search.tested` is rendered in full into every agent's prompt on every turn

`src/hyperloom/orchestrator/state/_shared_state/render.py:930`

`_format_search_state` changed from "last 5 accepted + last 15 rejected" to enumerating every `tested` entry with explicit no truncation; the only bound is `_EXPLORE_TESTED_CAP = 5000`. A rendered row with a typical `extra_server_args` plus two envs measures **182 chars**:

```
1000 tested entries -> 182,063 chars (~45k tokens)
5000 (the cap)      -> ~910KB       (~230k tokens)
```

`_format_search_state` hangs off `to_prompt_summary()`, which `_compose_prompt()` calls for **every agent on every tick**, so this is a per-turn fixed cost that grows monotonically with run length. Each `tested` entry is one actually-benchmarked variant (`executors/explore.py:1953` = winners + losers), so at ~4 variants per round over a multi-day run, several hundred to a thousand entries is within the intended envelope — `_cap_tested_ledger`'s own docstring says "Bound the explore_search negative ledger for multi-day runs." Compared with the previous steady ~20 lines (~4KB), this is tens of thousands of tokens of fixed per-turn amplification, and this same PR removes the context-token-watermark compaction that used to bound prompt growth on long runs. The dedup-correctness motivation is understandable, but the renderer needs an upper bound of its own — e.g. render all KEEPs plus the most recent N rejects, or bucket by `macro_cycle`/bottleneck and render only the current bucket.

ZhengGong-amd:
Thanks — all three are accurate on the facts. Fixed 1 and 2; declining 3, reasoning below.

## 1. `integrate` defaulting to `mode="patch"` — fixed, and it was worse than reported

Confirmed the regression end to end, including that the back-fill source is still live after #1408: the fusion lane still writes `kernel_opt_task_attempts` / `pending_kernel_integrations` via `enqueue_nominated_patch` (`phases/kernel.py:4954`), so `_resolve_integrate_payload` can and does resolve a real patch for an env-only payload.

One thing the review missed, which made it worse: the gate check read `payload.get("mode")` on the REQUEST **envelope**, but a request carries its handler payload under `params` — `intent_router.py:911` does `{**intent.payload, **params}` and only then dispatches. So the check never saw an LLM-authored mode at all. The value validation was unreachable dead code, and *every* agent request fell through to the `patch` default, not just the ones omitting `mode`.

Fixed in 27525c313: read `params` the way the router does, and **require** `mode` rather than defaulting it. Took the "reject and require an explicit mode" half of your suggestion rather than the shape fallback — the two modes differ in whether a patch is applied at all, so a missing one is a question rather than a value to guess, and inferring it again would walk back the contract this PR introduced. Internal callers reach `integrate_handler` directly and are unaffected; this validates the one boundary an LLM crosses, which is also what AGENTS.md asks for. The denial carries a hint naming both modes so the agent can correct on the next tick.

Seven existing tests emitted an integrate request with no mode and all failed first — that they failed is what confirms the check is now reachable. Added two tests: one for the missing-mode denial, one pinning that `mode` is read from `params` and not the envelope.

Two notes for the record:
- `apply_patch` is an alias of `integrate` (`action_surfaces.py:53`) and the gate matches on the literal kind, so it skips this check. Left alone deliberately: the action name *is* the request to apply a patch, so the `patch` path is what it asks for — no wrong attribution.
- Back-fill that resolves nothing returns `missing_integration_inputs`, a clean rejection. So the silent-wrong-data case needed a `kernel_id` that actually carries an artifact — which is exactly what the prompt tells the agent to pass, so the window was genuinely open.

## 2. Dead memory fields + a prompt promising a block that no longer exists — fixed

Confirmed: `=== Your working memory (recovered) ===` appears in exactly one place in the tree, `orchestration.md:32`, with zero producers since `render_memory_for_seed` was deleted. And the three readers do only take `next_cycle_directive`. I also checked `envelope.py:178` and `gate.py:633`, which mention `orchestration_memory` — both are write-lock lists (fields an LLM `update_state` must not touch), not prompt projections, so they don't rescue the other five fields.

You missed one instance of the same claim: the `orchestration_memory` field comment in `shared_state.py` also still said "pasted back into the projection and used on crash-recovery rebuild". Both corrected in c2fadc7bd to describe what actually happens — `next_cycle_directive` becomes the next cycle's `## CYCLE DIRECTIVE` section, the rest is run-report evidence and is not replayed.

Took the "drop the promise" half rather than wiring the record back in. Kept the five fields: the handoff call has to happen anyway for `next_cycle_directive`, so their marginal cost is a few output tokens once per macro-cycle, and they are useful post-hoc evidence in the run artifact. Trimming them touches `MEMORY_REQUEST_PROMPT`, `build_memory_record`'s merge/dedup and the history ring, which is a bigger change than this thread warrants — happy to do it separately if you'd rather they go.

## 3. Full `tested` ledger render — declining, this is the intended design

The measurements are right, and if anything understated: I measured a realistic row (name + `extra_server_args` + two envs + reason) at **250 chars**, not 182. `to_prompt_summary()` is called unconditionally at `conversation.py:386` for every agent every tick, and yes, this PR also removed the context-token watermark that used to bound prompt growth.

Keeping it anyway, deliberately. Rendering the ledger in full is the point of the change, not a side effect: a variant missing from the render reads as untried and gets re-proposed, which costs a real benchmark — minutes of GPU time — every time it happens. `fix(ledger): stop spending the 30-slot failure ring on non-failures` in this same PR depends on exactly this: it removed the per-re-proposal `record_action_failure` calls *because* the tested ledger is rendered in full, so those rows no longer had to defend against a variant reading as untried. Truncating the render silently undoes that commit's premise, and a bounded render without an explicit "N omitted" marker actively misleads the agent — worse than the token cost, because the failure mode is wasted hardware rather than wasted context.

On magnitude: `_EXPLORE_TESTED_CAP = 5000` is not a reachable operating point. Each entry is one actually-benchmarked variant, so 5000 is roughly 17 days of nothing but benchmarking; the ~910KB / 230k-token figure is a theoretical ceiling rather than a run we can have. The realistic multi-day shape is several hundred rows, and we are choosing to spend that context on dedup correctness.

If prompt size does become the binding constraint on a long run, the right lever is lowering `_EXPLORE_TESTED_CAP` — that shrinks the ledger and the render together and keeps them consistent — not truncating the render while the ledger keeps growing behind it.

xiaofei-zheng:
Re-reviewed at `a4eaf0c10`.

Findings 1 and 2 are fixed and verified:

- `27525c313` — `gate.py` now merges `{**payload, **(payload.get("params") or {})}` before reading `mode`, which matches `intent_router.py:910-911`, and denies with `missing_integrate_mode` / `invalid_integrate_mode`. Both new tests cover the envelope and the `params` placement. `prompt_builder.py:769` states the no-default contract.
- `c2fadc7bd` — `orchestration.md:28-33` and `shared_state.py:1198` now describe what is actually read back.

Finding 3: your rationale holds — full render is the point of the change, `7fd8282cb` depends on that guarantee to stop writing re-proposals into the 30-slot failure ring, and 5000 fingerprints is not a reachable operating point. Withdrawn; the cap is the right lever if it ever needs one.

One residual blocking item from the `mode` fix:

**`orchestration.md:272` still documents `mode` as optional, in the same prompt that now says it is required.**

```
`integrate` request schema: `{kernel_id, patch_path, target_file, base_tput,
extra_server_args, config_path, mode}`. `mode` (str): `"patch"` (default) to
apply a kernel source patch, or `"env_only"` to validate env/config changes
without applying a patch (used by GEMM tuning E2E validation).
```

That block sits under `<!-- phase: KERNEL_AGENT -->` (orchestration.md:260), and `_KERNEL_OPT_PIPELINE_BODY` is appended under `_KERNEL_REQUEST_PHASES = frozenset({"KERNEL_AGENT"})` (`prompt_builder.py:59`, appended at `:1059`). So the KERNEL_AGENT system prompt now carries both texts: one says `patch` is the default, the other says omitting `mode` is denied rather than guessed.

Trigger: an agent following the schema line omits `mode` on the mandatory "Drain pending KEEPs first" path (orchestration.md:266-271) and is denied by the gate on every pending KEEP — a wasted tick per attempt, on the path the prompt itself marks as do-this-before-anything-else. Dropping the `(default)` clause from that line resolves it.

ZhengGong-amd:
Thanks for the re-review, and for finding the contradiction — but I'm not fixing it by editing either sentence. Stepping back from the finding: the mechanism it describes should not exist.

## `mode` was never an agent-facing field

`mode` has exactly two producers in the tree, both in `phases/kernel.py` — the GEMM paired confirmation (`:2665`) and the GEMM tuning E2E validation (`:3716`). Both are Coordinator-owned lanes calling `integrate_handler` **directly**, never crossing the bus or the gate. There has never been an agent producer.

I checked the merge base: it documented no `mode` and no `env_only` to the agent at all. Both texts — the `prompt_builder` schema entry and the `orchestration.md:272` block you quoted — arrived in this branch, with the explicit-mode contract. The `(used by GEMM tuning E2E validation)` parenthetical in my own line was the tell, and I missed it.

So I handed the agent a field it has no use for. An agent `integrate` means one thing: land the KEEP'd patch for this `kernel_id`. Env / serve-flag changes are `explore`'s lever, where variants are fingerprint-deduped through `explore_search`. Publishing `env_only` invented the very request your first finding correctly identified as dangerous — and then each round added another layer around it: a validator, a required-field contract, seven updated payloads, a per-kind params fixture, and finally two prompts disagreeing about the default. That is accretion in response to a surface that shouldn't have been published.

## What I did instead (b8fa0cd03)

Removed the surface, not the symptom:

- **Deleted** the PolicyGate `integrate` mode validation outright — including the `missing_integrate_mode` / `invalid_integrate_mode` rules and both tests I added last round.
- **Deleted** `mode` from both prompt schemas, and deleted the duplicate schema block this branch added to `orchestration.md`. `prompt_builder` already owns that schema; carrying it twice in one KERNEL_AGENT system prompt is precisely how the two descriptions drifted apart, so removing the copy fixes the contradiction at its cause rather than resyncing two texts that will drift again.
- **Reverted** the seven test payloads and the routing fixture — they only existed to satisfy the gate check.
- **Kept** `mode` as a handler parameter defaulting to `patch`, now documented at `request_handlers.py` as Coordinator-internal, with the two GEMM lanes still setting it explicitly.

Net **−85 lines**.

The trigger you described is gone by construction: an agent cannot omit a field it is never offered, and there is no "Drain pending KEEPs first" denial path left. And your original finding #1 dissolves with it — `integrate{kernel_id, extra_envs, base_tput}` now means what it says (land `kernel_id`'s patch, with those envs as serving config), because `env_only` is no longer something an agent can intend. A `kernel_id` with no artifact still gets a clean `missing_integration_inputs` rejection, as before.

Worth noting the merge-base shape inference was hazardous in the mirror direction on this same path: an agent `integrate{kernel_id, extra_envs}` inferred env-only and silently *skipped* the patch it was asked to land. The explicit `mode` is the right fix for the internal callers; publishing it was the mistake.

Rebased onto current main (`921b45d70`) while I was in here. `ruff check` / `format --check` clean; 1442 tests across `orchestrator` + `agents` + the gate/prompt/routing files pass locally, and 20546 collect clean.

xiaofei-zheng (approved):
Re-reviewed at `b8fa0cd03`. **No blocking issues left.** Agreed with the direction — removing the surface is the right call, and it dissolves the original finding rather than papering over it.

What I checked:

- **`mode` producers.** `git grep` over `src/` (tests excluded) finds exactly three writers: `phases/kernel.py:2665` and `:3716` set `env_only` and call `integrate_handler` directly, and `phases/kernel_stack.py:650` emits `mode: "patch"` on the Coordinator's own auto-integrate bus request — explicit and equal to the handler default, so it's a no-op either way. No agent-authored producer exists, and the merge base (`0a8edb157`) documented neither `mode` nor `env_only` to any agent. The Coordinator-internal claim holds.
- **The path that motivated finding #1.** An agent `integrate{kernel_id, extra_envs, base_tput}` now falls to `mode="patch"` at `request_handlers.py:6609`, goes through `_resolve_integrate_payload`, and lands the KEEP'd patch — which is what the request means. A `kernel_id` with no artifact still gets `missing_integration_inputs`. The merge-base shape inference that silently skipped the patch is gone, and no gate rule is needed to hold that in place.
- **The contradiction.** Deleting the duplicate `orchestration.md` block rather than resyncing it removes the drift source; `prompt_builder.py:764-772` is now the single schema and no longer mentions `mode`. No `mode` / `env_only` text reaches any agent prompt.
- **Gate deletion is scoped.** Only the `integrate` mode branch went; `_validate_gemm_tuning_action`, the `source_file` trusted-scope checks, and the tool-whitelist collision rule are untouched. The reverted test payloads are the same ones the previous round added.
- **Tests.** `test_policy_gate.py`, `test_agent_roles_and_policy.py`, `test_session_layout.py`, `test_lifecycle_wiring.py`, `test_kernel_integrate_and_report.py`, `test_profile_and_kernel_handlers.py` — 365 passed on `b8fa0cd03`.

Findings 1, 2 and 3 are all settled: 1 and 3 dissolved or withdrawn on their merits, 2 fixed in `c2fadc7bd`.

One non-review note: `main` has moved three commits since your rebase (`d5b62a3fc refactor: compact Python comments and docstrings (#1421)` being the wide one), and GitHub currently reports the PR as CONFLICTING. Mechanical only, but it needs another rebase before this can go in.

PR #1411, opened 2026-09-05
[FEAT] KernelForge: trace-driven nominator behind KERNELFORGE_NOMINATOR=trace

npoulad1:

`kernelforge/nomination` ships a documented stub seam. It only accepts rows that are already source-resolved, ranks them on whatever `gpu_pct` the candidate list was authored with, and splits the campaign budget evenly across whatever survives. That is fine as a placeholder and wrong as a nominator: the thing that knows what is hot is the profiler, not the candidate list.

This adds a nominator driven by real trace evidence, with three capabilities:

1. **Trace evidence overrides the row.** Per-kernel GPU shares are read from the profiler output, so the trace decides what is hot rather than the authored `gpu_pct`.
2. **Unresolved rows get rescued.** A row with no `source_file` is searched across the configured roots by *definition site* — a regex over `__global__`, `def`, and templated/static declarations — instead of by substring. Substring matching is actively wrong here: it resolves a kernel named `blind` to any file that happens to contain those five letters in a comment. There is a test for exactly that false positive.
3. **Budget follows share.** The campaign budget is split in proportion to each target's measured GPU share, with a floor, so a 9% kernel is not handed the same wall clock as a 0.5% one.

**Backward compatibility is the design constraint here, not an afterthought.** 48 shell tests encode the stub's semantics as a contract, including "skip unresolved rows" — which is the exact behaviour capability 2 inverts. So the new nominator is:

- opt-in behind `KERNELFORGE_NOMINATOR=trace`,
- defaulting to `stub`,
- and falling back to the stub when it nominates nothing, because an empty nomination ends the lane and falling through to the already-resolved rows is strictly better than returning nothing.

None. Raised from a KernelForge campaign on DeepSeek-V4-Pro-0813 decode; happy to file a tracking issue if you would prefer one.

9 new tests in `src/kernelforge/tests/test_nomination_trace.py`, covering: trace override of the row's `gpu_pct`, share-weighted budget split, the budget floor, rejection handling, definition-site resolution, the substring false positive, mangled-symbol reduction, and the `KERNELFORGE_NOMINATOR` env gate.

```
pytest src/kernelforge/tests -m "not critic_agent_e2e and not robustness_agent_e2e and not targeted_build_e2e"
ruff check src/kernelforge
ruff format --check src/kernelforge
```

Result against `main` (9ae79d6) on the same machine and interpreter:

| | passed | failed |
|---|---|---|
| `main` | 3594 | 28 |
| this branch | 3603 | 28 |

`+9 passed, +0 failed` — the 9 new tests, no regressions. The 28 failures are pre-existing on `main` and environmental: they are `test_forge_loop_resume.py` cases that need an LLM provider extra (`AgentProviderUnavailableError: no Agent provider is available; install the 'claude' or 'codex' extra`), which is not installed in this sandbox. They fail identically with and without the patch.

`ruff check` and `ruff format --check` both clean.

No. Default behaviour is byte-for-byte the stub. The new path is only reachable with `KERNELFORGE_NOMINATOR=trace` set, and even then it falls back to the stub if it nominates nothing.

### DIFF
--- a/src/kernelforge/nomination/__init__.py
+++ b/src/kernelforge/nomination/__init__.py
@@ -1,0 +1,397 @@
+# SPDX-FileCopyrightText: 2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Consumer half of the nomination contract.
+
+Mirrors ``hyperloom.orchestrator.kernel.nomination_request`` field for field.
+The two halves are deliberately separate modules rather than a shared import:
+forge stays independently importable, and ``PROTOCOL_VERSION`` is what keeps
+them honest. Bump both together.
+
+The seam a real implementation replaces is :func:`nominate`; everything else
+here is shell -- read, validate, delegate, assemble.
+"""
+
+from __future__ import annotations
+
+import json
+from dataclasses import asdict, dataclass, field
+from pathlib import Path
+from typing import Any
+
+#: Must equal the producer's value. A mismatch stops the run rather than
+#: risking a partially understood request.
+PROTOCOL_VERSION = 1
+
+#: Must equal the producer's ``candidate_manifest.MANIFEST_VERSION``; an absent
+#: version is a skew too. No compatibility path, so both halves move together.
+MANIFEST_VERSION = 2
+
+#: Every field a candidate row must carry. A row missing one comes from a
+#: producer this build does not understand, so the gap is refused, not guessed.
+_REQUIRED_CANDIDATE_FIELDS = ("kernel_name", "gpu_pct", "source_file", "reason_class", "attempts", "rejected")
+
+#: Mirrors ``kernel_source_contract.KNOWN_REASON_CLASSES`` as a literal so forge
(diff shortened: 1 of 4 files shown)


jiaqiang-dot-liu:
Thanks. But I'd suggest closing this PR.

Selection no longer lives in `nomination`. #1408 deleted
`src/kernelforge/nomination/` outright and replaced it with
`kernel_rewrite_controller/`. Nothing on current `main` imports
`kernelforge.nomination`, and the controller already covers all three
capabilities this PR adds.

The rebase brought the deleted files back. The diff is now `+897 / -0`, and two
of those files — `nomination/__init__.py` and `stub.py` — are byte-identical to
the versions #1408 removed. The `-0` is the tell: a PR that modifies an existing
seam can't have zero deletions. I don't think this was intentional — `git
rebase` resolves "deleted by them" by keeping the branch side, so the conflict
disappeared cleanly while also reverting part of #1408.

So merging this would leave us with a package that has no callers, plus an
undeclared partial revert.

PR #1412, opened 2026-09-05
[FIX] KernelForge: exempt the loop's own aiter cache from the workspace guard

npoulad1:

**The bug.** `configure_aiter_cache_isolation` roots its cache at `<experiments-dir>/aiter_cache`, and the shipped AITER example places `--experiments-dir` *inside the workspace*. When a candidate triggers a JIT build, the cache writes new untracked files **during** the turn. The workspace guard sees new untracked paths and rejects the entire turn — for files the framework itself just wrote.

`allow_dirty_baseline=True` already forgives cache shards that exist when a turn *begins*. The gap is only the files created mid-turn.

**What it cost.** This is not hypothetical. Iteration 5 of a DeepSeek-V4-Pro-0813 campaign produced a candidate that the in-session gate explicitly passed:

```
[in-session-gate] ALLOW (correct + faster: mean case speedup mean score=1.026621x
  >= 1.015225x from scores=[1.025718, 1.026704, 1.027443]) edit=21
```

Three independent measurements, tightly clustered, all better than the candidate that was ultimately kept (1.0138x). It was then destroyed by:

```
[agent] ERROR: new non-ignored files are unsupported:
  forge_experiments/aiter_cache/sources/d7287ba5e00fb3a3529d082a/.forge_cache_owner.json,
  forge_experiments/aiter_cache/sources/d7287ba5e00fb3a3529d082a/flydsl_cache/launch_...
[REVERT] Protected integrity violation; canonical validation skipped
```

Every one of those paths was written by KernelForge. Because the revert happens *before* the candidate is committed, the source was not recoverable from git — no dangling object, no stash, and `candidates/` has archives for iterations 1–4 but none for 5. The campaign reported 1.38% when it had measured 2.66%.

**The fix.** The mechanism already existed. `spec.ignored_untracked_globs` was added for precisely this class of bug when rocprofv3 droppings failed a session; the loop's own cache was simply never added to the list. This patch:

- lifts the glob list into a module constant, `TOOL_OWNED_UNTRACKED_GLOBS`, so tests assert against the list that actually ships rather than a copy of it;
- adds `aiter_cache/*` and `*/aiter_cache/*`. The directory name is fixed by `configure_aiter_cache_isolation` even though `--experiments-dir` is not, so the two forms together cover every placement of the cache inside a workspace.

It stays a **per-path declaration, not `allow_untracked`** — an undeclared stray file is still a violation, and there is a test asserting that.

**One thing worth considering alongside this patch** (not changed here, happy to fold in if you want it): the loop prints

```
[git] 2 new file(s) outside commit_new_paths, not removed: forge_mhc_driver.py, graph_harness.py
```

on **every** iteration. Those are the protected driver and harness, which the shipped AITER example deliberately keeps untracked. The message reads like a fault report for the documented pattern.

None filed. This is a bug report and its fix in one; glad to split out an issue if that is preferred for tracking.

2 new tests in `src/kernelforge/tests/test_aiter_cache.py`: one reproducing the exact untracked paths that lost the candidate (`.forge_cache_owner.json` and `flydsl_cache/launch_*` under `aiter_cache/sources/<hash>/`), and one asserting that a file smuggled in next to the cache still fails the guard.

```
pytest src/kernelforge/tests -m "not critic_agent_e2e and not robustness_agent_e2e and not targeted_build_e2e"
ruff check src/kernelforge
ruff format --check src/kernelforge
```

Result against `main` (9ae79d6) on the same machine and interpreter:

| | passed | failed |
|---|---|---|
| `main` | 3594 | 28 |
| this branch | 3596 | 28 |

`+2 passed, +0 failed`. The 28 failures are pre-existing on `main` and environmental (`test_forge_loop_resume.py` needs an LLM provider extra that is not installed in this sandbox); they fail identically with and without the patch.

`ruff check` and `ruff format --check` both clean.

No. This widens an existing exemption list by two globs and refactors it into a named constant. A turn that passed the guard before still passes; the only turns whose outcome changes are the ones that were being rejected for the framework's own cache files.

### DIFF
--- a/src/kernelforge/orchestrator/agent.py
+++ b/src/kernelforge/orchestrator/agent.py
@@ -45,6 +45,40 @@ _REPO_EXTRA_PROTECTED_GLOBS = [
     "conftest.py",
 ]
 
+# Untracked paths written by the tooling rather than by the agent, declared so a
+# turn is not rejected for them. Deliberately per-path: an undeclared stray file
+# is still a violation, which is why this is not ``allow_untracked``.
+#
+# Exported as a constant so the guard's behaviour can be tested against the list
+# that actually ships rather than against a copy of it.
+TOOL_OWNED_UNTRACKED_GLOBS = [
+    # The profiler writes where it is run. Both forms must reach any depth: it
+    # runs in ``run_cwd``, the kernel file's parent, while the guard reports
+    # paths relative to the git toplevel. ``fnmatch`` crosses "/", so
+    # ``*_results.db`` already does; ``.rocprofv3/`` has to be spelled twice.
+    ".rocprofv3/*",
+    "*/.rocprofv3/*",
+    "*_results.db",
+    # The loop's own aiter cache, for the same reason and with the same remedy.
+    # ``allow_dirty_baseline`` forgives the shards present when a turn begins,
+    # but a candidate that triggers a JIT build creates new ones *during* the
+    # turn -- ``<experiments-dir>/aiter_cache/sources/<hash>/...``,
+    # ``.forge_cache_owner.json``, ``flydsl_cache/launch_*`` -- which land among
+    # the untracked and get the turn rejected for files the framework wrote.
+    #
+    # This is not hypothetical: a candidate measured at 1.027x, which the
+    # in-session gate had already ALLOWed as correct and faster, was reverted as
+    # a "Protected integrity violation" and -- never having been committed --
+    # could not be recovered afterwards.
+    #
+    # ``configure_aiter_cache_isolation`` roots the tree at
+    # ``experiments_dir / "aiter_cache"``, so the directory name is fixed even
+    # though ``--experiments-dir`` is not; matching on it covers every placement
+    # inside the workspace.
+    "aiter_cache/*",
+    "*/aiter_cache/*",
+]
+
 # task_type values that mean "a full source tree, not a self-contained snippet".
 _REPO_TASK_TYPES = {"repository", "image_kernel"}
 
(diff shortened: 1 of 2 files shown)


jiaqiang-dot-liu:
LGTM

xiaofei-zheng (approved):
**No blocking issues — LGTM.**

Verified the following rather than taking the description at face value:

- **Root cause matches the code.** `configure_aiter_cache_isolation` roots the tree at `experiments_dir / "aiter_cache"` (`src/kernelforge/loop/aiter_cache.py:71`), so the directory name is indeed fixed while `--experiments-dir` is not. The guard's `verify()` raises on any untracked path that survives `_drop_ignored_untracked` (`src/kernelforge/agent_backends/workspace_guard.py:850`), and `allow_dirty_baseline` only snapshots what exists at `prepare()` time — mid-turn JIT writes genuinely fall through the gap.

- **Both globs are necessary, neither is redundant.** `_drop_ignored_untracked` uses `fnmatch` (`workspace_guard.py:315`), which anchors both ends. Checked directly:
  - `fnmatch("forge_experiments/aiter_cache/sources/abc/flydsl_cache/launch_1", "aiter_cache/*")` → `False`
  - same path against `"*/aiter_cache/*"` → `True`
  - `fnmatch("aiter_cache/x/y", "aiter_cache/*")` → `True`

  So the nested form covers `--experiments-dir` anywhere inside the workspace and the bare form covers it at the git toplevel. Since `*` crosses `/`, arbitrary depth under the cache root is covered by both.

- **Single production call site.** `ignored_untracked_globs` is consumed only by `workspace_guard.py:312` and set only at `src/kernelforge/orchestrator/agent.py:716`, so lifting it to `TOOL_OWNED_UNTRACKED_GLOBS` and widening it there is the complete end-to-end fix; no other spec construction needs the same treatment.

- **Blast radius is confined to untracked paths.** The filter is applied in `_current_changes()` only. The `staged`, `protected_changes`, `protected_untracked`, `changed_snapshots` and `index_changed` buckets in `verify()` are unaffected, so this cannot weaken the protected-path or index guarantees.

- **Tests, scoped to the change** (against merge-base `9ae79d6`): `pytest src/kernelforge/tests/test_aiter_cache.py` → 18 passed; `pytest src/kernelforge/tests -k "workspace_guard or guard or agent_spec"` → 81 passed. Left the full suite to CI.

One non-blocking observation, no action needed: exempted paths are also skipped by `rollback()` / `_restore_baseline()`, since they read the same filtered untracked list — so the cache now survives a revert. That matches the existing `.rocprofv3/*` behavior and is what `prune_aiter_cache_shards` / `cleanup_owned_aiter_cache` are there for, so it reads as intended rather than as a leak.

PR #1414, opened 2026-09-07
enablement: make the recipe's replay sufficiency fail closed on what a consumer actually receives

zoroyihan7:
Continues `feature/r1-recipe-impl`. The branch already carried R1a's `recipe_steps`
projection and R1b's `replay_sufficiency` contract; this run closes the findings the R1b
design was approved into implementation still carrying, and the acceptance clauses an
audit of both designs found named but unreached.

## Carried findings

**B42 — the design document exceeds a 900-line cap.** Informational; the designs are frozen
and no code follows from it. Not addressed.

**B43 — a recipe could be replay-sufficient while its referenced payloads were not
deliverable.** `evaluate_replay_sufficiency` already took a `delivered_paths` set, but no
producer passed one, so every session took the branch that judges a recipe on its content
alone.

- The recipe now names the bytes behind its manifests and digests
  (`referenced_payloads`), and `session_package.deliverable()` answers per candidate:
  matched by the curated selection, held as a regular file inside the session, and small
  enough that a bundle could carry it. Absent for any reason is absent.
- The enablement KEEP's snapshot overlay matched no package glob, so the byte-exact
  content of every patch and artifact target the recipe declares was never shipped.
  `optimization_stack/enablement/**/files/**` is now selected — the captured bytes only,
  since each capture's own manifest records the absolute framework root it was taken
  under, and the portable projection of it already travels in the breakdown.
- The capture directory is keyed by a digest of the root path, so every KEEP round wrote
  into one directory and the mechanism never cleared the files beside the manifest it
  overwrote. Each capture now starts from an empty overlay, so what ships is the accepted
  stack rather than the union of every round.

Patch steps are judged through their root's snapshot rather than their own `path`, which
names the authoring workspace no bundle ships.

**B44 — `base_sha` recorded the tree after the KEEP commit.** Already closed on the branch:
`_git_head_sha` is read before the stash and before any apply, and
`test_a_kept_patch_applies_exactly_once_to_the_recorded_base` applies a real unified diff
to the recorded base and is refused a second time. This run closes the per-root half —
`is_git`/`base_sha` were read off `framework_root` and stamped onto every record, so a
second contributing checkout was asserted non-git with no base commit — and refuses a git
root whose commit the record does not name, which a failed `rev-parse` used to leave
standing unjudged.

**B45 — successful setup steps could depend on inputs neither captured nor rejected.**
Capture and refusal were already on the branch. The missing half was delivery: the identity
stopped at durable state, so the emitted step handed a consumer a command string and
nothing else. The `setup` element now carries `input_identity` and `unresolved_inputs` and
the projection states the replay rule; the identity names the file where a delivery would
carry it, and a payload the bundle does not carry is refused rather than advertised.

## Also in this change

Five rules were reporting more than the evidence supported: an absent `launch_evidence`
judged only when a configuration had also projected; `setup_ledger_truncated` firing for a
command the accepted round ran and failed; `closure_scope_incomplete` gated on an execution
succeeding; `dependency_closure_status` reading `"verified"` when the ledger that decides
scope was absent or capped; and `present_at_final_launch` never taken back from an earlier
KEEP round, which let a stale row answer for a command the validated launch never ran and
made truncation unable to fire at all.

Test additions cover the audit's unreached clauses: the build element's key set pinned
closed, declarativeness walked to the leaves of a nested `build_inputs`, the published
recipe key set actually inspected, multi-root rounds in both the git and non-git direction,
the KEEP probe's fail-closed half paired with the bypass backend that can resolve an
interpreter, a build inheriting a credentialed index, a runtime with neither rebuild
source, and the env values that must not travel beside their digest.

## Deviations from the frozen designs, stated rather than hidden

- `setup_inputs_incomplete` is not in R1b §3.1's closed code table, and the file read that
  produces its evidence is not in §2.1's authorized-collection list. B45 asks for exactly
  this — an immutable identity for file and VCS inputs, or an explicit replay-blocking
  reason — and neither is reachable without them. Both designs are frozen, so the extension
  is recorded here rather than in the design.
- The delivered set is a prediction, made at collect time, of what packaging will include.
  It is the only point at which the decision can be made, since the breakdown is itself one
  of the packaged files.
- The size test is per payload rather than over the whole selection: a bundle-wide cap
  cannot be evaluated without enumerating the session tree, and that walk sits on the path
  the coordinator drives at shutdown.

## Verification

- `pytest` full suite: **20712 passed, 54 failed, 40 skipped, 8 xfailed** (23m). All 54
  failures are environmental and pre-existing: 47 reproduce byte-identically on a clean
  `origin/main` worktree in the same environment (credential/preflight/kernelforge suites
  that read ambient provider keys), and the other 7 were an artefact of my own run having
  unset `OPENAI_API_KEY`. None is in any module this branch touches. Two suites
  additionally need the package importable in a child interpreter (`pip install -e .`).
- `ruff check .` and `ruff format --check .` — clean.
- `mypy src/hyperloom` is advisory and reports 1084 pre-existing errors across 188 files,
  none in any module this branch touches.

### DIFF
--- a/src/hyperloom/common/launch_log_evidence.py
+++ b/src/hyperloom/common/launch_log_evidence.py
@@ -6,6 +6,7 @@
 from __future__ import annotations
 
 import ast
+import hashlib
 import logging
 import re
 import shlex
@@ -19,6 +20,10 @@ log = logging.getLogger(__name__)
 _RUN_SPECIFIC_LAUNCH_FLAGS: frozenset[str] = frozenset(
     {
         "--model-path",
+        # vLLM's own spelling. Without it the projected argv keeps a run-local
+        # model operand the SGLang spelling has always had stripped.
+        "--model",
+        "--tokenizer",
         "--tokenizer-path",
         "--served-model-name",
         "--host",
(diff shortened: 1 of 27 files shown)


chaojhou (approved):

ZhengGong-amd:
Scoped to redundancy and dead code. I'm not raising the fallback/silent-failure paths — the
section is written to disk for an external consumer, so degrading to a refusal there is the
right shape.

## 1. `deliverable()` runs the packer twice, and the prediction cannot be exact

`session_package.deliverable()` re-runs `_select()` + `_pack()` at breakdown-record time to
predict what the later real packaging will include. But `session_breakdown.json` is itself one
of the packaged files, so the simulated tree is missing the file being written and the budget
accounting diverges from the real run by construction.

Invert it: let the packer report what it actually shipped, and backfill the missing references
from that. `_pack()` already returns `packed / truncated / dropped`. One fact instead of two
runs of the same selection logic.

## 2. Duplicated infrastructure

| new | already in the repo |
|---|---|
| `integrate_patch.py:660` `_git_head_sha()` | `_patch_source_pr.py:28` `_git_head_sha()` — same package, same name, same `rev-parse`, same 30s timeout; the new one drops the stderr the existing one returns |
| `_patch_snapshot.py:297` `_git()` | `_git.py:16` `_run_git()` / `_run_git_cp()` — `_patch_snapshot.py` already imports and uses `_run_git_cp`. The isolated-env need is real, but it's a keyword on the existing runner, not a third parallel one |
| `_patch_snapshot.py:218` `_trust_config()` (40 lines, sha256-named config file + two `git config --file` subprocesses) | `common/git_safety.py:28` `safe_directory_args()` |
| `integrate_patch.py:256` `_merge_established_server_args()` (58 lines to merge two strings; `shlex.split` falls back to `.split()` twice) | `_grid_server_args.py`: `merge_server_args`, `tokenize_server_args_preserving_json`, `_shell_safe_dedupe` |

`_patch_snapshot.py` now holds five answers to "which files does this patch touch":
`_patch_touched_paths_split`, `patch_declared_ops`, `_step_inventory`,
`overlay_inventory_without_base`, `_patch_touched_paths_from_text`. The new docstrings argue
against the old ones (`patch_declared_ops` says `_patch_touched_paths_split` "asks the tree
instead, which is the wrong question twice over"; `_step_inventory` says `patch_file_targets`
"cannot tell a file header from a hunk body") — but none of them were removed. Please converge
on one and delete the rest.

## 3. One rule, two modules

`_established_enablement_config`'s docstring claims ownership:

> Every dispatcher that opens an enablement integrate_patch crosses this executor, so the
> inheritance lives here rather than in each emitter.

The same PR then adds the same inheritance to an emitter at `phases/explore.py:1604-1627`. The
40-line fallback in the executor exists to survive the resulting double merge. Pick one owner
and the fallback goes away with it.

## 4. Dead code

**a. The same condition twice; the second is unreachable** — `integrate_patch.py:3771-3780`:

```python
            if replayed is None and root_base:
                log.warning(...)
                continue
            if replayed is None and root_base:
                continue
```

**b. Guard on top of guard, with a stale comment** — `enablement_section.py:239-251`. The
`_tri` loop already handles `build_extensions_not_carried` through the `_ABSENT` sentinel; the
`_carry` block below repeats it. Its comment ("the loop above, which drops anything falsy")
refers to the earlier `for key, value in (...)` loop — the one the `_tri` loop was added to fix
in `43b964fb0`. The `_carry` block is the previous iteration's patch, left behind.

**c.** `_patch_snapshot.py:131` `_GIT_DIFF_BLOCK_RE` — zero references anywhere including
tests, and it is the only user of the module's `import re`.

**d.** `patch_declared_ops` (54 lines, 32 of them docstring) is exported in `__all__` but has no
production caller; its only consumers are 10 assertions in `test_enablement_keep_records.py`.
Production uses `replayed_stack_ops` / `overlay_inventory_without_base`.

**e.** `declared_targets(..., upserted=(), deleted=(), ...)` — the only production call site
passes both empty, which makes the two loops consuming them at `recipe/keep_records.py:141-144`
unreachable.

PR #1415, opened 2026-09-07
fix(sbd-v6): name the copies an enablement round left in the archive

lunarwhite:
## What

**An enablement attempt row names files a reader cannot fetch.**
`patches_applied` and `accepted_config_path` are the round's own paths —
workspace originals under `runs/`. `session_package.py`'s `PACKAGE_GLOBS` takes
all of `reports/enablement/**` and only a narrow allow-list under `runs/`,
which patches and materialized configs are not on. Running that matcher
directly confirms it: `runs/specialist/<tid>/patches/001_fix.patch` and
`runs/integrate_patch/<tid>/integrate_patch.with_envs.yaml` are both **not
packaged**.

**`accepted_config_path` is the dangling pointer
[#1394](https://github.com/AMD-AGI/Hyperloom/pull/1394) fixed**, re-recorded a
layer down. That fix holds where it was applied — `lane.py` overrides
`state.enablement.accepted_config_path` with the archived copy, so
`ext.result.accepted_config_path` resolves — but the per-attempt row takes
`res["enablement_accepted_config_path"]`, which is
`bench_result["materialized_config"]`, the value from before the archive ran.

## How

`snapshot_round` already returns a `RoundArchive` that records only copies that
landed. A new `record_archive` writes what it holds onto the row the dispatch
opened — `files` with its roles, and `accepted_config_path` read out of that
same manifest so the two cannot disagree.

It is its own entry point rather than a field of `record_round`. That call is
the verdict projection, built from the round's own `res`; which copies exist is
a fact only the coordinator has and only once the snapshot has run, so it is
recorded where it comes into being, on the same tick, keyed by the
`(task_id, attempt)` the dispatch already filed. The recorder is not put inside
`snapshot_round` itself: that helper is copy and IO, has no attempt key, and
should not reach for the SBD. `lane.py` still reads
`archive.path_for(ROLE_LAUNCH_CONFIG)` to re-point
`state.enablement.accepted_config_path`, which is control plane rather than
observability.

The call sits inside the archive block's `try`, so a snapshot that raised
leaves both fields **absent** rather than asserting an empty manifest — it did
not establish that nothing landed, and `null`/absent is what this layer
reserves for a fact that was not produced. The round is still ruled either way.

## Notes

**The roles are what the round's own result cannot supply.** A patch the round
applied and one it refused are archived apart — `ROLE_PATCH` against
`ROLE_PATCH_EVIDENCE` — and nothing in `res` distinguishes them, so without the
manifest a consumer cannot tell which patch to replay and cannot enumerate the
refused ones at all, their names appearing nowhere else on the event. The
archive's 2 MiB ceiling is the other half: it silently declines a copy it
cannot take, which is what makes "the round reported this patch" and "the
archive holds it" two different facts.

**A list of `{path, role}` rather than the keyed `artifacts` dict** the other
event types use (`conc_sweep_event`, `framework_event`, `close_out`). Those
hold one file per name; four of the archive's roles repeat — `ROLE_PATCH` by
design — so a keyed map cannot express this. `files` rather than `artifacts`
because `artifacts_applied` already sits on the same row meaning something
else: whole files installed into the framework tree.

**The workspace paths stay**, as identity rather than as a way to fetch. A
patch name joins a round to its `files` entry, and an artifact's target is the
position a replay installs to. The schema docstring says which of the two each
field is, and flags the deliberate asymmetry with
`ext.result.accepted_config_path`, which stays absolute because the
revalidation baseline opens that one directly.

**`files` names copies the archive took, which is weaker than "fetchable".**
`reports/enablement/**` is packaged outside `RESERVED_PATHS`, so a session that
hit the file or byte cap ships a truncated package and a consumer has to
tolerate a miss. The schema docstring says so rather than promising more than
the packer does. Moving these under the reserved partition would be a separate
call, and a bigger one than this PR should make.

**One distinction is given up.** A config that existed but whose copy the
archive declined now reads, at row level, like no config at all.
`ext.result.accepted_config_path` still separates the two, keeping the `runs/`
original in that case.

**`accepted_config_path` changes meaning** from an absolute `runs/` path to a
session-relative archived one.
[#1455](https://github.com/AMD-AGI/Hyperloom/pull/1455) merged this week, so no
production payload carries the old form.

**The first commit is independent** and fixes a pre-existing leak:
`Coordinator.__init__` binds the recording session and drops the token, so any
test constructing one leaves the rest of the worker with a session bound. Only
*sync* construction leaks — `pytest.mark.asyncio` runs the coroutine in a Task
that gets a copy of the context — which is why it has stayed latent under
alphabetical collection order. Happy to split it out.

## Test plan

- [x] `pytest test_sbd_v6_enablement_timeline.py test_sbd_v6_enablement_wiring.py
      test_enablement_artifacts.py` — 88 passed
- [x] `pytest` over `-k "enablement or sbd_v6 or breakdown or recorder or
      coordinator or integrate_patch"` — 1848 passed
- [x] Full `inference_optimizer` + `phases` suites against `main`: identical
      failure set, all pre-existing
- [x] Each fix verified to fail its test when reverted, including deleting the
      `record_archive` call in `lane.py`, which two wiring tests catch
- [x] Leak reproduced on `main` before the fixture and green after
- [x] `ruff check` / `ruff format --check` clean; `mypy` reports nothing new on
      the changed modules

New coverage: the archive and the verdict are two calls on one tick and land on
one row, not two; a copy the archive declined is named nowhere rather than named
and dangling; a snapshot that raised leaves the row silent rather than claiming
an empty manifest, and the round is still ruled; the round's workspace originals
survive as identity; and in `test_sbd_v6_enablement_wiring.py`, which exists so
that a call site quietly not making a call fails there, the real rearm records
the archive it took.

_Made with Cursor_

### DIFF
--- a/src/hyperloom/orchestrator/enablement/lane.py
+++ b/src/hyperloom/orchestrator/enablement/lane.py
@@ -627,10 +627,21 @@ class EnablementLane(CoordinatorCollaborator):
         res_fw_root = str(res.get("framework_root") or "").strip()
         if res_fw_root:
             state.enablement.framework_root = res_fw_root
+        # The ordinal the dispatch filed this round under, read before the
+        # settle scores it, so every row of this round meets on the same number.
+        attempt = await self.rounds.consecutive_stalled() + 1
         setting_script = ""
         archive = RoundArchive(self.session_dir)
         try:
             archive = snapshot_round(self.session_dir, res)
+            # Where the copies come into being, and from the archive rather
+            # than from ``res``. Inside the try so that a snapshot which raised
+            # leaves the row silent instead of claiming nothing landed.
+            enablement_event.record_archive(
+                task_id=_round_task_id(state, res),
+                attempt=attempt,
+                files=archive.to_list(),
+            )
             if status in ("kept", "advanced"):
                 setting_script = write_setting_script(
                     self.session_dir,
@@ -650,9 +661,6 @@ class EnablementLane(CoordinatorCollaborator):
         archived_config = archive.path_for(ROLE_LAUNCH_CONFIG)
         if status == "kept" and archived_config:
             state.enablement.accepted_config_path = str(Path(self.session_dir) / archived_config)
-        # The ordinal the dispatch filed this round under, read before the
-        # settle scores it, so the two rows meet on the same number.
-        attempt = await self.rounds.consecutive_stalled() + 1
         # A rearm always ends the round; only a KEEP booted and was graded, and
         # an advance is the ledger's record that the cap must not charge it.
         is_advanced = status == "advanced" or bool(res.get("advanced"))
@@ -705,6 +713,15 @@ class EnablementLane(CoordinatorCollaborator):
                 log.exception("ENABLEMENT %s (%s) failed", pump.__name__, caller)
 
 
+def _round_task_id(state: Any, res: dict[str, Any]) -> str:
+    """The specialist task id every row of one round is keyed by.
+
+    A phase-synthesised rearm carries no id of its own, so the in-flight one
+    the dispatch filed stands in.
+    """
+    return str(res.get("specialist_task_id") or "").strip() or str(state.enablement.last_specialist_task_id or "")
+
+
 def _record_enablement_round(
     state: Any,
     res: dict[str, Any],
@@ -723,10 +740,9 @@ def _record_enablement_round(
     dispatch filed it under.
     """
     lane = state.enablement
-    round_tid = str(res.get("specialist_task_id") or "").strip() or str(lane.last_specialist_task_id or "")
     succeeded = bool(lane.succeeded)
     enablement_event.record_round(
-        task_id=round_tid,
+        task_id=_round_task_id(state, res),
         attempt=int(attempt or 0),
         result=res,
         stall_streak=int(stall_streak or 0),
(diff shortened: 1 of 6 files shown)


src/hyperloom/orchestrator/enablement/lane.py
```diff
@@ -732,6 +735,8 @@ def _record_enablement_round(
         stall_streak=int(stall_streak or 0),
         succeeded=succeeded,
         validation_pending=bool(lane.validation_pending),
+        files=archive.to_list(),
+        accepted_config_path=archive.path_for(ROLE_LAUNCH_CONFIG),
```
haishuok0525: Taking the live `RoundArchive` instead of `res` is the right source — those copies exist only in the coordinator, and `patches_applied` cannot say whether the 2 MiB ceiling actually kept them.

Ferrying the manifest through `record_round` is the part that sits oddly against the recorder's design. That entry is the settle projection (`status`, stall streak, the round's own `res`). `files` / the archived `accepted_config_path` are produced by `snapshot_round`, a few dozen lines above, on the same tick. The enablement recorder's contract is that each producer records its own facts when they exist (`record_dispatch` at dispatch, `record_build` in the executor), not that the lane later stuffs someone else's object into the settle call.

Prefer a dedicated `record_*` that upserts onto this same attempt row immediately after `snapshot_round` returns, using the in-flight `(task_id, attempt)` already known from dispatch. Do not put the recorder inside `snapshot_round` itself: that helper is copy/IO, does not have the attempt key, and should not import SBD. The lane can still consume `archive.path_for(ROLE_LAUNCH_CONFIG)` to rewrite `state.enablement.accepted_config_path` — that is control-plane, not an observability courier.

lunarwhite: Added a dedicated `record_archive(task_id, attempt, files)`

haishuok0525 (approved):

PR #1416, opened 2026-09-07
feat(atom): support Atom end to end — reap its recovery workers, default its kernel backend to forge, and publish it as supported

zengleixin-amd:
This branch makes Atom a framework Hyperloom actually supports, and says so. It
carries three concerns that share a session's failure story but not a code path;
they should be reviewed separately.

| | Concern | Production diff | Framework scope |
|---|---|---|---|
| **A** | ATOM recovery worker reaping | `recover.py` +103/-2 | ATOM-specific |
| **B** | Eval-failure attribution | `baseline.py` +90/-43 | Framework-agnostic |
| **C** | Kernel backend default + publishing Atom as supported | `cli/__init__.py` +33/-4, plus examples/docs | ATOM-specific |

Diff composition: production 274 lines (19%), tests 676 (48%), examples and
documentation 415 (29%), CHANGELOG 53.

**C is why A and B ship together with the docs rather than after them.** Merging
the code without it would ship an optimizer that handles Atom alongside a
compatibility page that still lists only SGLang and vLLM.

---

# A. Reap recorded ATOM recovery workers

## Scope

Fix two ATOM recovery gaps in Hyperloom, without changing the normal vLLM/SGLang lifecycle or any third-party project/dependency pin.

1. `_server_lifecycle` and `recover` did not recognize `atom.entrypoints.openai_server`, so a recorded ATOM server failed the owner check.
2. After recovery sent group TERM, the ATOM leader could exit while anonymous workers ignored TERM. The subsequent cmdline-based check no longer recognized an owner, so those workers did not receive KILL.

**Note after merging `main`:** `main` has since added `atom.entrypoints` to `_SERVER_CMDLINE_MARKERS` itself (alongside `EngineCore` and `VLLM::Worker` for its dead-leader-live-group branch). Gap 1 is therefore already closed upstream, and `_server_lifecycle.py` **no longer appears in this diff at all**. What remains here is gap 2, in `recover.py`.

## Implementation

- Add `atom.entrypoints` to the existing owner list in `recover.py`. Select ATOM recovery from the `-m atom.entrypoints.openai_server` module entrypoint, not the first framework substring found in model arguments.
- For a recognized, live ATOM leader from this session's pidfiles, snapshot group members before TERM. Record PID, PGID and process start time, then revalidate the leader. If the leader or its identity cannot be confirmed, skip this candidate before sending any signal; do not fall back to the stale discovery record.
- After the existing grace period, revalidate each recorded survivor before sending individual KILL signals. A reused PID, changed process group or unreadable identity is not sufficient authority to signal it.
- Record the worker PIDs actually signalled; do not report a worker's KILL as a KILL sent to its already-exited leader.
- Retain the pidfile while the recorded process/group is still alive. Sending KILL is not proof of exit.

The vLLM/SGLang recovery branches and warmup/measure scheduling remain unchanged.

## Revision from the earlier PR version

The earlier generic `_descendant_pgids` expansion has been withdrawn. It changed teardown for every framework but did not recover lineage after a wrapper/leader had already exited. `_subprocess_kill.py` has **no diff from the base commit**. A real-process regression verifies that a completed warmup wrapper leaves its persistent server available for the next round.

Earlier wording claiming that `recover` unconditionally deleted live groups' pidfiles was incorrect. Its existing deletion helper checks liveness; the explanation and changelog have been corrected. Historical GPU-leak experiments motivated the investigation, but are not being claimed as end-to-end validation of this revised patch.

## Verification (A)

Run on Linux/Python 3.12.13 in a dedicated checkout created from the complete PR snapshot, not the earlier mixed experimental workspace. No GPU/model serving was needed for these process-protocol checks.

- Real process regression: a recognized ATOM leader exits on TERM, its anonymous worker ignores TERM, recovery must terminate the worker and leave an unrelated process alive. **Failed before the recovery change; passed after.**
- Safety cases: changed leader identity, reused/moved/missing worker PID, old group gone with reused leader PID, new owner marker, unrecorded member, unreadable/malformed `/proc` data, zombie process and non-POSIX behavior.
- Recovery audit records the actual worker KILL and retains a still-live group's pidfile.
- vLLM/SGLang retain their existing group-cleanup behavior without entering the ATOM snapshot branch.
- Baseline/grid/bypass warmup and measured-round tests now exercise ATOM, vLLM and SGLang. These are orchestration tests with mocked servers, not GPU E2E.
- Real `run_with_session_kill` process tests confirm successful warmup preserves a detached persistent server.

Re-confirmed at `dbe38bb40`, after the `main` merge and both part B commits, on Linux/Python 3.10.12 (kernel 6.8.0-107), source archive sha256 `48b5fde2f54601132b753185ab8bab127e1c4edbb263182072355e36272cf342`:

```
six lifecycle/recovery/runner test files  370 passed in 129.35s
```

(The earlier 332-passed figure was taken before the merge and is superseded. This run predates the `e16a12c22` correction below; see [Verification against the current head](#verification-against-the-current-head) for the figures that apply now.)

## Live reproduction on 8×MI355X

The gap this part closes was hit four separate times while driving a real ATOM ×
GLM-5.2-FP8 optimisation session on an 8×MI355X node: every time the owning
process went away, the server survived with `PPID=1` in its own process group,
still holding 88–93% of VRAM on all eight cards, and the next round could not
allocate until it was cleared by hand. That is the failure this patch exists for,
observed in the wild rather than constructed.

It was then reproduced deliberately and handed to this branch's executor. The
server was started the way Magpie starts it (`setsid python3 -m
atom.entrypoints.openai_server --model <704 GB GLM-5.2-FP8> -tp 8`), the model
was loaded so the cards were genuinely occupied, and the parent was allowed to
exit:

```
BEFORE  leader alive: True   VRAM: 88 89 89 89 88 88 88 88

_kill_stale_owners returned 11 entries in 5.0s
    pid 44184  pattern=atom.entrypoints  signal=KILL     <- leader, identified by entrypoint
    pid 44588  pgid=44184                signal=KILL     <- resource_tracker
    pid 44589  pgid=44184  cmd=''        signal=KILL     <- anonymous: empty cmdline
    pid 44758  ATOM::TP0               signal=KILL
    pid 44759  ATOM::TP1               signal=KILL
    ...        ATOM::TP2/3/4           signal=KILL

AFTER   leader alive: False  VRAM: 0 0 0 0 0 0 0 0
```

Eleven processes reaped in five seconds; all eight cards back to zero.

What this confirms, point by point:

- **Entrypoint-based identification** — the leader matched `atom.entrypoints`
  through `-m atom.entrypoints.openai_server`, not through a framework substring
  that happens to appear in the model path.
- **The member snapshot is what makes the anonymous workers reachable** — pid
  44589 has an *empty* cmdline. That is precisely the case the description calls
  out: once the leader exits, a cmdline check no longer recognises an owner, so
  without the pre-TERM snapshot this process is skipped and its VRAM stays held.
- **Identity is checked before signalling** — `(pgid=44184, starttime=14318133)`
  was captured for each of the 19 members seen at snapshot time.
- **The audit records the workers actually signalled**, one entry per pid.

The pidfile is still present afterwards. That is the documented conservative
behaviour — the group is retained until it is observed gone, because sending
KILL is not proof of exit — not a leak.

One correction worth recording: the first run of this experiment reported *no*
reap. That was a defect in the harness, which set `session_dir` where
`_discover_stale_pids` reads `_active_session_dir`, so discovery returned an
empty candidate list and none of this code ran. Re-run with the right attribute,
it passed first time.

## Focused review follow-up

- Invalidated ATOM leader identity now blocks TERM as well as worker KILL. The new missing/changed/replaced-identity and path-marker cases failed before the fix (10 failures, 2 unchanged-framework controls passed).
- The real-process recovery test launches a temporary ATOM module, rather than merely including an ATOM string in `python -c` arguments.
- The persistent-server test publishes its PID file atomically from the wrapper holding the process handle; failed publication cleans up that process.
- The first test process is protected by `finally` before creating the second. An injected second-launch failure first reproduced the leaked leader, then passed after the cleanup fix.
- No pidfd framework, cross-package helper dependency, test-efficiency rewrite, or third-party change was introduced.

---

# B. Stop reading an unreachable server as a capability gap

## The failure this comes from

On a real 24-hour session, a baseline's throughput pass had **already produced a number** when its server was torn down mid-eval. The client's next request was refused:

```
aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host 0.0.0.0:41099
  ssl:default [Connect call failed ('0.0.0.0', 41099)]
ERROR: run_eval failed with exit code 1
```

InferenceX reports a vanished server and a model that scored badly the same way — a non-zero `run_eval`. So the eval-rooted branch in `baseline.py` stamped this as an eval-failure contract and handed it to the enablement lane, which then spent **five specialist rounds** hunting a capability gap the evidence never supported. The run ended on the stall cap with a terminal reason naming enablement rather than the server, and the measured throughput was discarded with `baseline_tput = 0.0`.

## Precedent

This is the same shape as the AgentX-preflight case already documented in `writeback.py:1195-1215`:

> Measured: routed as an ordinary launch failure, this opened an enablement round instead. [...] the PolicyGate's `enablement_round_in_flight` rule then blocked the baseline for the rest of the run — 24h of budget spent retrying a problem one operator action fixes.

The remedy there was to refuse to open a round that no amount of authoring can close. This change applies the same remedy to a second instance.

## Implementation

- Add `_EVAL_SERVER_UNREACHABLE_MARKERS` (`ClientConnectorError`, `Cannot connect to host`, `Connect call failed`) and `_is_server_unreachable_eval_failure`. It reads only this result's own error and warnings, deliberately **not** through `_failure_carries_markers`: that helper climbs out of a round directory and scans every log under the shared task root, so a warmup round whose server crashed would decide the classification of a measure round that did reach a verdict.
- At **both** `_stamp_eval_failure_contract` call sites -- `_run_retrying` and `_maybe_stop_on_missing_baseline_accuracy` -- an eval-rooted failure carrying refused-connection evidence no longer enters the eval-failure branch. It is recorded as an ordinary baseline failure with a `eval_failed_server_unreachable` warning, so the existing `_BASELINE_MAX_TOTAL_FAILURES` backstop ends the run on the cause it actually had.

Deliberately **not** changed: the accuracy gate is untouched, no throughput is salvaged, `RUN_EVAL` is not forced off, and no "retry only the eval" mechanism is introduced. A baseline that fails this way still fails.

## Verification (B)

```
test_baseline_eval_fallback.py + test_baseline_eval_rooted_failure.py   69 passed in 68.46s
combined with the six part-A files                                     439 passed in 196.96s
ruff check (whole src tree) / ruff format --check                      passed
```

Same Linux host and source hash as part A. All eight CI test shards (py3.10 and py3.11) pass at this head.

### The markers were re-derived from a live failure

The three marker strings were copied out of a session log, and the unit tests
assert with those same hand-copied strings -- so no test could notice if
aiohttp reworded the message. Provoked for real against a closed port
(aiohttp 3.13.5):

```
ClientConnectorError: Cannot connect to host 0.0.0.0:39871 ssl:default [Connect call failed ('0.0.0.0', 39871)]
```

All three markers match it and `_is_server_unreachable_eval_failure` returns
`True` on that exact text. This pins the predicate to the library's current
wording rather than to a transcription of it.

### The routing test earned its keep

The routing test could not run on the Windows workstation this was authored on (`test_baseline_eval_fallback.py` imports `fcntl`; the module's pre-existing tests fail identically there). The first commit was pushed with the predicate covered by executed tests but the wiring reviewed statically only — and **CI immediately caught a real defect in it**:

```
FAILED test_eval_failure_from_an_unreachable_server_is_not_an_enablement_gap
  AssertionError: assert True is not True        (py3.10 and py3.11, shard 3/4)
```

`_stamp_eval_failure_contract` has **two** call sites. Gating only `_run_retrying` left `_maybe_stop_on_missing_baseline_accuracy` still stamping the contract: its guard admits any eval-rooted failure, and a vanished server is genuinely eval-rooted, so the run reached the accuracy branch, found no accuracy, and recorded `baseline_eval_failed` after all. Both routes now return early on the same evidence — an eval that never reached a verdict produced no accuracy signal, so neither the accuracy stop nor the enablement routing has anything to act on.

Worth stating plainly: the predicate unit tests were all green while the feature was still broken. Only the end-to-end routing test found it.

---

---

# C. Default the kernel backend on atom, and publish Atom as supported

## The gap

`_apply_atom_auto_tighten` enforced the `--nodes>=2` guard and then printed that
kernel-agent was "wired for atom". The phase does run -- but the backend it got,
GEAK, is on weaker ground there. Its extractor's own rule for a quantized,
non-vLLM backend forbids *guessing* a rewrite seam (a wrong fp8/attn seam is
just another dead rebind) and requires grepping the live server for the real
quant-apply or backend forward instead. That is a harder path, not a closed one,
but it is unproven on atom -- and the default phase split hands that phase half
the session, which is a poor place to find out. Only a reader of the demo skill
would have known to set `KERNEL_OPT_BACKEND_ORDER=forge` by hand.

## Implementation

- On atom an unset or blank `KERNEL_OPT_BACKEND_ORDER` is filled in with `forge`
  and the choice is reported at launch. The fill-in lands at CLI line ~1932,
  before `_seed_shared_state` (~2074) reads the env to record
  `kernel_optimizer`, so a session cannot record `geak` while running forge.
- A value the operator set is **kept**, with a warning that GEAK's seam
  resolution is unproven there. Running GEAK on atom deliberately, to measure
  exactly that, stays possible.
- `--no-kernel` skips the defaulting; the phase it applies to does not run.
- Scoped to the `framework == "atom"` branch. SGLang and vLLM keep GEAK.

The function had no tests. It has twelve now, including one that asserts a named
backend is not rewritten, one that pins the ordering above, and one that reads
the call site out of the source to keep the `framework == "atom"` guard from
being widened -- every other test calls the function directly, so none of them
would notice SGLang and vLLM starting to get the default too.

## Publishing

- `examples/hyperloom-qwen3-14b-fp8-12h-atom` -- the existing 12h demo moved onto
  a different serving framework, workload and budget copied verbatim so the
  variants stay comparable.
- `docs/compatibility.rst` -- Atom joins the summary row, the framework table and
  the image table. ROCm 7.2.4 and the image tag were read out of the image. The
  GPU column records what was verified (MI355X) rather than implying the
  MI300X/MI325X coverage the other rows carry.
- The setup skill stops asking which kernel backend to use when `FRAMEWORK` is
  `atom`; only one answer works there and the CLI already applies it.

Server arguments in the example were measured, not copied from Atom's published
`Qwen3-8B-FP8` recipe: on MI355X the engine serves Qwen3-14B-FP8 at `-tp 1` with
nothing beyond `--max-model-len`, and 64 concurrent 1024/1024 requests all
completed. That recipe's `--level 0 --block-size 64 --kv_cache_dtype bf16`
belongs to a small-VRAM gfx1201 consumer card.

## Explicit boundaries

- No changes to Magpie, InferenceX, ATOM, aiter or dependency pins.
- No host-wide `pkill`, no killing by `spawn_main` alone, no use of host-namespace KFD PIDs as container-local PIDs.
- No claim to recover ownership of anonymous workers whose leader exited before recovery took its snapshot. A member that changed groups is conservatively skipped; members created *after* the snapshot are covered only by the closing group KILL, which is aimed at the group as a whole and proves nothing about any individual rank. The existing GPU probe can still report `needs_review`.
- PID/starttime validation narrows reuse hazards but is not an atomic pidfd guarantee.
- The generic subprocess/profile teardown and third-party cleanup limitations remain outside this patch; this is not a claim that every ATOM exit path is now fixed.
- Part B does not make a stalled session succeed. It converts a long, mislabelled failure into a prompt, correctly attributed one. The boot failure that started that session is in `aiter` (its `init_process_group` passes no `timeout=`, unlike vLLM's and SGLang's) and is out of scope here.
- Part C's example was verified as far as the engine booting Qwen3-14B-FP8 at `-tp 1` and serving 64 concurrent 1024/1024 requests. That combination has **not** been driven through a complete 12h optimize session; the end-to-end Hyperloom-on-Atom evidence is from GLM-5.2-FP8 on 8xMI355X (baseline 758 tok/s, framework-agent variants measured and graded), on an earlier working copy.
- A full `optimize --framework atom` session was attempted against this branch and
  did not reach the baseline action: the orchestrator re-dispatches
  `target_analysis` and never `baseline`, with each LLM call timing out after
  ~125s. That is **not** this branch's doing -- unmodified `origin/main`, same
  container, same credentials, reproduces it identically, and the three files
  that decide actions (`phases/machine_state.py`, `roles/critic_agent.py`,
  `roles/robustness_agent.py`) are byte-identical to main here. Four
  configurations were crossed (sonnet/opus x advisory on/off x 2h/12h budget x
  Magpie absent/present) with no change. Reported separately; what this branch
  changes was verified by other means, listed above.
- A framework-agnostic port defect is **not** fixed here and still bites atom hardest: `_pick_free_port` binds port 0, so the reserved port comes from the kernel's ephemeral range, and the server only binds it after the model has loaded. On a large checkpoint that window is long enough -- 22 minutes, measured on GLM-5.2-FP8 -- for the kernel to hand the port to an outbound connection, and the launch then fails with `address already in use`. It needs its own change.
- The hosted E2E for the earlier head `c735bffb4` timed out with 217 of 220 polls reporting `Queued` and repeated `GPU quota exceeded`; it provided no acceptance evidence and is superseded by the single-GPU teardown measurement below. That measurement covers the teardown path only -- it is not a full optimize-session E2E.

---

## Correction: the identity gate was lost, then restored (`e16a12c22`)

`dbe38bb40` set out to stop the ATOM branch short-circuiting the KILL phase. That
part was right -- a rank forked after the member snapshot is in neither the
recorded set nor any cmdline that still reads as an owner, so without a closing
group KILL it survives holding its cards. But the same commit also dropped the
`_process_identity(pid) != (pgid, members[pid])` check and turned *skip when
ownership is unconfirmed* into *signal anyway*, contradicting the behaviour this
description, the CHANGELOG entry and the branch's own tests all specified.

A recorded pgid is stale data. Once the kernel recycles it, the generic teardown
aims TERM at whoever now holds it. Reproduced with real processes: a leader that
still reads as an ATOM server, a record naming a pgid that meanwhile belongs to an
unrelated group.

| | signals sent | unrelated group |
|---|---|---|
| before `e16a12c22` | `group <other> TERM`, `pid <leader> KILL` | killed |
| after | none | survives |

The two goals do not conflict. The transient-`/proc` case `dbe38bb40` cited never
reaches the fall-through: if `/proc` cannot be read for the leader it cannot be
read for the group either, so `pid not in members` skips first. The closing group
KILL is kept, on the confirmed path only, and no longer force-removes the pidfile
-- it targets members this pass never enumerated, so it is not proof of exit.

## Verification against the current head

Linux, in a container on one MI355X node.

- `482 passed` across every test file this branch touches, the new
  kernel-backend defaulting included. A wider sweep (lifecycle / server /
  teardown / reaper) is `554 passed, 2 failed`; both failures reproduce
  identically on unmodified `origin/main`, where that test stubs `builtins.open`
  while `is_multi_node()` still wants to read a state file.
- Real-process reproductions: late-forked members reaped; recycled pgid left
  untouched; and the before/after table above.
- **Real serving ATOM server**, Qwen3-14B-FP8 at `-tp 1` -- the earlier
  process-protocol tests used a stand-in, this one serves the model:

  | scenario | before | `_kill_stale_owners` | after |
  |---|---|---|---|
  | mid-boot | 2 engine procs | 1 × TERM | 0 procs, 0% VRAM |
  | fully serving | 1 leader, **87% VRAM** | **4 × KILL** (leader + 3 anonymous workers) | 0 procs, **0% VRAM** |

- CI on `79e15dbc9`: 30 of 33 checks pass, 2 skipped (main-only jobs). One
  `test_session_time_budget.py` failure did not reproduce on re-run and passes
  111/111 three times in a row locally; that file is untouched by this branch.


xiaofei-zheng (approved):
Reviewed against `merge-base origin/main` (`97eadf5e1`). **No blocking issues — LGTM.**

Paths I actually walked, so you know the depth:

**A — `recover.py`**
- `_process_identity` field offsets verified against `proc(5)`: after `rsplit(")", 1)`, `fields[0]=state`, `fields[2]=pgrp` (field 5), `fields[19]=starttime` (field 22). Correct. Zombie (`Z`) returning `None` means such a pid can never satisfy the identity equality, so it is skipped rather than signalled.
- `_is_atom_server` exercised over launcher-prefix, vLLM-with-`atom`-in-model-path, truncated `-m`, and trailing-`-m` inputs. `args[i+1:][:1]` cannot raise `IndexError`, and the vLLM-serving-a-path-named-`atom.entrypoints.openai_server` case correctly returns `False`, so the ATOM snapshot branch stays off for vLLM/SGLang.
- The TERM gate (`pid not in members or not _is_atom_server(...) or _process_identity(pid) != (pgid, members[pid])`) is the same predicate the KILL phase uses, so a stale/recycled pgid cannot be signalled from either side. Confirmed the `e16a12c22` restoration is present in the merged tree, not just described.
- `atom_members[pid]` is populated before the TERM attempt, but the KILL loop iterates `killed`, which only receives entries where `sent` was true — so a failed TERM leaves an orphaned snapshot entry that is never acted on. Harmless.
- The `continue` after the closing group KILL skips `_remove_finished_pidfile`, so an ATOM pidfile survives the pass. `_discover_stale_pids` tolerates that: a dead pid yields an empty cmdline, `_process_group_owner_cmd` finds no marker, and the entry is skipped with a warning rather than signalled. If the pid is later recycled by a new ATOM leader, the recorded pgid no longer matches and the identity gate skips it. No path where the retained pidfile causes a wrong signal.
- Adding `atom.entrypoints` to `_OWNER_PATTERNS` also widens `_process_group_owner_cmd`, which the vLLM/SGLang `still_owned` check uses. Confirmed that only adds ATOM recognition; it cannot change the verdict for a vLLM/SGLang group.

**B — `baseline.py`**
- The central claim — that skipping `_stamp_eval_failure_contract` lets the ordinary backstop fire — checks out end to end: the stamp is what sets `BASELINE_EVAL_FAILED_KEY`, `writeback.py:1359` reads exactly that key into `eval_failed`, and `eval_pending_suppress` (`:1383`) is therefore `False`, so `baseline_total_failures >= _BASELINE_MAX_TOTAL_FAILURES` (`:1436`) is reached. The run ends on the real cause instead of stalling.
- The second call site is genuinely covered. In `_maybe_stop_on_missing_baseline_accuracy` the early return is correctly ordered *after* `_salvage_sibling_baseline_accuracy` and gated on `salvaged is None`, so a sibling attempt that did reach a verdict still anchors and a measured under-floor accuracy still routes. Returning early also skips `request_baseline_accuracy_stop` on the non-enablement path, which is the intended handoff to the backstop above rather than a dropped stop.
- `_is_server_unreachable_eval_failure` deliberately does not climb to the task root the way `_failure_carries_markers` does. Worth stating as a known limit rather than a defect: on a double-run layout the eval-rooted verdict can come from the sibling warmup round while the refusal evidence stays there too, in which case the new branch does not fire and the old routing applies. That is the conservative direction, and the docstring says why.

**C — `cli/__init__.py`**
- Ordering claim verified in the merged tree: the fill-in is at `:1939`, `_seed_shared_state` at `:2081`, so a session cannot record `geak` while running forge.
- The resume call at `:1695` sits in the `--resume-from` branch and the fresh call at `:1939` in the other, so `_apply_atom_auto_tighten` runs once per invocation. `getattr(args, "no_kernel", False)` and `getattr(args, "nodes", 1)` keep the resume path safe if the namespace is thinner there.
- `forge_explicitly_enabled()` is an exact match on `forge`, so a composite order such as `forge,geak` takes the warning branch. The message is slightly off for that input, but the value is kept and behaviour is correct.

Tests: not run locally — this workstation is Windows and `test_baseline_eval_fallback.py` imports `fcntl`, which fails identically on `origin/main`. CI covers it: all 12 shards green on py3.10/3.11 plus ruff, mypy, pylint and the sphinx build.

PR #1417, opened 2026-09-07
docs(examples): fix 3h demo phase budget and align skill docs with the CLI

lishuoshuo-amd:
The 3h demo asked for `--max-minutes-framework-pct 0.90`, which silently produced *less* budget than intended. Under `--no-kernel`, KERNEL_AGENT's 0.50 share is not dropped -- `redistribute_budget_pct` adds it on top of the requested value, weighted by base pct. 0.90 redistributes to 1.3945, and `normalize_budget_pct` then rejects anything above 1.0 as out-of-range and reverts that phase to its 0.40 default, logging a single WARNING. So the run asked for 90% and got 40%: worse than passing no flag at all.

0.50 is the effective ceiling -- it redistributes to 0.9902, which is what the requirement text beside it already described as "~99% of wall clock". The flag and its own prose had drifted apart; this makes the flag match the prose. The 12h demo is unaffected: it keeps KERNEL_AGENT enabled, so 0.43/0.42 pass through untouched.

Also aligned three smaller doc/code mismatches:

- README's setup question order listed Secrets third; the setup skill asks it second, with the model asked alongside the other non-secret values. The run-mode option lost its `(Recommended)` label on `docker`, contradicting the Docker recommendation the same README leads with.
- Dropped `local_dir_use_symlinks=False` from all three `snapshot_download` calls. It was removed in huggingface_hub 1.0, and the skills install with `pip install -U`. It is still accepted and ignored with a UserWarning rather than raising, so this is noise removal, not a break.
- Noted in the advanced skill that global flags such as `--verbose` must precede the `optimize` subcommand. The template already places it correctly; the note keeps a reader from appending one to OPT_FLAGS, where argparse fails with `unrecognized arguments`.

Verified against the CLI rather than by reading: replayed the documented flags through `normalize_budget_pct`/`redistribute_budget_pct` (0.9902, survives re-normalization), and re-ran the full flag diff against `optimize --help` before and after -- identical, no invented flags.


ZhengGong-amd (approved):
Approved, the author promised further revision

PR #1418, opened 2026-09-07
[Stepping] Make the source scope advisory, not a permission set

ZhengGong-amd:
Specialists run under `--permission-mode bypassPermissions` with the framework
source trees passed as writable `--add-dir`. The allowlist that predates that
model still decided what specialists were *told* to look at, and still rejected
their work afterwards.

Evidence from 51 dispatches across three sessions under
`Hyperloom-Sessions/vllm/`:

- All 51 prompts render the same eleven roots, session-independent: the whole of
  site-packages, `/opt/rocm`, a diffusion framework, and sglang on vLLM runs.
  Two do not exist on the Kimi-K3 host — specialists hit `ls: cannot access`.
- Of 626 source-path references, 80% land in `dist-packages/vllm` and 12% in
  `aiter`. Four advertised roots got zero.
- Section 7 called the trees read-only, `--add-dir` made them writable, iron
  rule 2 forbade writing them. Three statements, three answers.
- One freeform mandate told a specialist that `/sgl-workspace/vllm` has no git
  tree and must never be patched, while Section 7 of the same prompt offered it
  as a source root.
- `enablement_ops` printed `ALLOWED SOURCE ROOTS (code edits outside these are
  rejected)`. `allowed_root_hints` had no reader anywhere.

## Scope is derived

`paths.py` already held the shape. `resolve_framework_tree` names the tree a
session optimises, keyed by the framework's own name; the search roots exist
because *"correct for a 'may this file be edited' containment test and wrong for
a recursive grep"* was already written down.

Section 7 now leads with the session tree, annotates each root as a git checkout
or an installed package with no git tree, drops what is absent from the host,
and resolves focus directories against the tree instead of leaving a relative
path to be joined onto eleven candidates. Classification keeps its own
unfiltered prefixes: a kernel path arrives from a trace produced on a serving
pod, so filtering by what exists locally turned every FlyDSL and multi-node path
into unknown source.

## Caps deleted, signals kept

`resolve_source_file_allowlist` and every cap it fed are gone — PolicyGate's
`source_file_outside_trusted_scope` and `rocm_runtime_write_denied`,
`integrate_patch`'s root refusal and its artifact and localization gates, the
controller publication check, the multi-node target assertion, and
`allow_unknown_target`, which every production caller already passed `True`.
Resolution survives where it answered a real question: `integrate_patch` takes
its candidate roots from the session tree, and the test is whether the patch's
target exists there rather than whether a path is on a list.

Three sites had no signal that survived a tick. Results settling `apply_failed`
or `reverted` are promoted, so `last_action_failures` never learned why a patch
did not land; specialist notes reached only the one-shot inbox line, and
`specialist_rounds` renders proposals but not notes; the controller's skipped
publications rode a `response` message rendered as a raw dump. All three now
record through `record_action_failure`, which SEED renders every turn.
`patch_safety` stops dropping a patch it cannot ground and reports it, so a real
`git apply` says which tree was tried.

Focus directories open to every domain and become a documented dispatch param,
alongside `scope`, `task_description`, `mode` and `bench` — all live, none in
the table. Their precision now comes from `model_info["quantization"]` rather
than `$PRECISION`, which read `fp8` for an MXFP8 checkpoint: static-recon was
handed the fp8 directories, reported the seeded items did not apply, and found
its one result in the two mxfp8 directories that selection had withheld.

## Three bugs caught in review

- The focus-directory join repeated the package name
  (`dist-packages/vllm/vllm/model_executor/...`).
- Existence-filtering the classifier roots made every FlyDSL and multi-node path
  unrecognisable.
- Removing `allow_unknown_target` made snapshot mode's fail-fast unreachable: an
  unrecognised layout would have written the optimized bytes to a fabricated
  path beside the target and reported `ok`.

## Template

- **Linked issues:** none.
- **Tests:** `orchestrator/` and `agents/` trees plus every touched
  `inference_optimizer` file — 3670 passed. New coverage for Section 7 ordering
  and annotation, absent roots, focus-directory joins, the precision resolver,
  the two failure-record paths, and the search-roots/classifier-prefixes split.
  Remaining failures (credentials, model-name assertions) reproduce on `main`.
  `ruff check` and `ruff format --check` clean.
- **Breaking changes:** yes, internal. `resolve_source_file_allowlist`,
  `resolve_patch_target_roots`, `is_rocm_hip_writable_path`,
  `assert_target_path_allowed`, `assert_revert_paths_allowed` and the
  `allow_unknown_target` parameter are removed; `patches_dropped_by_grounding`
  is now `patches_ungrounded`. No serving or CLI surface changes.
- **Single concern:** yes — the source scope stops being a permission set. The
  feedback wiring is in scope because deleting a cap must not delete its signal.
- **Upstream root cause:** no.


xiaofei-zheng:
## What this PR does

Turns the framework source allowlist from a permission set into advisory navigation. The single `resolve_source_file_allowlist()` used to answer three unrelated questions at once (may this be edited / where does a patch land / how do I classify this path); it is replaced by three resolvers with distinct semantics:

- `resolve_framework_tree()` — the one tree this session optimises;
- `resolve_kernel_search_roots()` — existence-filtered trees worth grepping;
- `resolve_known_source_prefixes()` — unfiltered string prefixes for classifying trace/manifest paths.

The enforcement machinery goes away with it: `assert_target_path_allowed` / `assert_revert_paths_allowed` / `allow_unknown_target`, the PolicyGate rules `source_file_outside_trusted_scope` and `rocm_runtime_write_denied`, `is_rocm_hip_writable_path` and all `/opt/rocm` root handling, and `_discover_installed_package_roots()` (which returned bare site/dist-packages parents). `_STATIC_PATCH_FALLBACK_ROOTS` becomes `_STATIC_SOURCE_LAYOUTS` and now only feeds prefix classification.

The rest is renames and feedback wiring: `patches_dropped_by_grounding` → `patches_ungrounded` (`PatchSafetyReport.dropped` → `.ungrounded`); `patch_targets_missing` gains `_P_STRIP_LEVELS = (1, 0, 2, …, 8)` so grounding is immune to the root-definition change; specialist prompt section 7 is rewritten from "read-only boundary" to "starting points, not a boundary", with a new `_source_root_row()` annotating each root as checkout vs installed package and `_resolve_focus_dir()` joining onto the parent for pip trees (`__init__.py` present) and directly for checkouts; three failure sites (promoted `integrate_patch` results carrying an `error_class`, specialist `patches_ungrounded`, skipped controller publications) now flow through `SharedState.record_action_failure` and are rendered every SEED turn.

## Blocking

**`_resolve_artifact_target` silently installs artifacts to a doubled path on pip-installed hosts** — `src/hyperloom/orchestrator/actions/executors/integrate_patch.py:1329-1348`

The root source changed from `resolve_source_file_allowlist()` (which contained the site-packages **parent**) to `resolve_kernel_search_roots()` (which contains the **package directory** itself, e.g. `.../dist-packages/vllm`), but the join logic did not change with it, and it did not get the `__init__.py` rule that `_resolve_focus_dir()` received in this same PR. A package-prefixed framework-relative target is therefore joined one level too deep. Measured on the same synthetic pip layout, merge base `0a8edb1` vs this branch:

```
target: vllm/model_executor/layers/fused_moe/configs/E=8.json
base   -> .../dist-packages/vllm/model_executor/layers/fused_moe/configs/E=8.json        # correct, file exists
branch -> .../dist-packages/vllm/vllm/model_executor/layers/fused_moe/configs/E=8.json   # does not exist
```

Neither fallback loop rescues it: the first requires `cand.parent.is_dir()` and falls through, the second only checks containment — and `vllm/vllm/...` *is* inside the root, so it is returned. The install site (`integrate_patch.py:4311-4317`) does `spec.target.parent.mkdir(parents=True, exist_ok=True)` followed by `shutil.copy2`, so the bogus directory is created, the file is written, and the action reports success — while the vLLM runtime never reads that path. The tuned fused_moe config simply has no effect, with no error signal anywhere.

The trigger is realistic rather than constructed: this PR's own prompt change (`_resolve_focus_dir` joining onto the parent for pip trees) is what leads the specialist to author package-prefixed targets like `vllm/...`. The function's docstring ("so a ``vllm/...`` config lands under the vllm root") is still written against the old root semantics.

Fix: reuse the `_resolve_focus_dir()` join rule — when the root itself carries `__init__.py`, also try `root.parent / rel` and prefer the candidate that exists.

Nothing else came up as blocking. Two things I checked and ruled out: patch grounding is not broken by the root narrowing (`patch_targets_missing` retries strip levels 0–8, so a `vllm/...` pre-image still matches a package root); and `test_agentx_switch.py::test_agentx_package_not_imported_on_off_path` fails on the merge base too, so it is not introduced here.

xiaofei-zheng (approved):
LGTM — the blocking issue I raised is fixed by 2af32a4f, and nothing else came up.

Re-verified on 2af32a4f, same synthetic pip layout as before:

```
target: vllm/model_executor/layers/fused_moe/configs/E=8.json
  -> .../dist-packages/vllm/model_executor/layers/fused_moe/configs/E=8.json   (correct)
target: model_executor/layers/fused_moe/configs/E=8.json
  -> .../dist-packages/vllm/model_executor/layers/fused_moe/configs/E=8.json   (correct)
```

The `_artifact_candidates` split is the right shape: the parent-join is only offered when the root itself carries `__init__.py`, and the retained `_is_within(cand, root)` check means it can only ever win for a target that repeats the package name — a `model_executor/...` target falls back to the direct join instead of escaping into `dist-packages/`. `relative_to(root)` stays valid for both candidates, so the KEEP snapshot key is unaffected. `test_integrate_patch_artifacts_unit.py` passes on the fix commit.

Paths I checked beyond that one, none blocking:

- `_resolve_framework_root` / `resolved_explicit_root` — an explicit or recorded root is now any existing directory rather than an allowlist member. Strictly more permissive, and `resolve_patch_apply_root` still requires a single unambiguous tree, so a patch set spanning two trees is not silently placed.
- Patch grounding across the root-definition change — `patch_targets_missing` retries strip levels 0-8, so a `vllm/...` pre-image still matches a package-directory root. No regression.
- `vet_patches` now appending to both `ungrounded` and `kept` for `missing_target` / `ambiguous_root` — intentional under the new model: those are reported as feedback rather than dropped, and `is_garbage` is correctly narrowed to the two cases that really are garbage (`GROUND_NOT_DIFF`, `GROUND_PATH_ESCAPE`).
- `/opt/rocm` removal vs the `_ROCM_HIP_ROOT_HINT` still advertised in `enablement_ops` — the hint is prompt prose, and with `resolved_explicit_root` accepting any existing directory a recorded ROCm apply root still resolves, so I could not turn this into a real failure.
- `subprocess_.py` dropping framework source roots from `--add-dir` — reads are unaffected; only the write sandbox narrows.

CI is green across all shards on this commit. `test_agentx_switch.py::test_agentx_package_not_imported_on_off_path` fails on the merge base too (`ModuleNotFoundError: No module named 'hyperloom'` in the subprocess probe), so it is not introduced here.

PR #1419, opened 2026-09-07
fix(teardown): reach the server our own wrapper setsid-ed away

zengleixin-amd:
## What

`kill_my_spawned_server` signals one process group: the wrapper's. Magpie's wrappers `setsid` the serving process, which is exactly what moves it out of that group. The wrapper then exits, the server reparents to init, and every rank it spawned keeps its GPUs.

This is the teardown used wherever `server_lifecycle` does not apply — which includes **every profile round**, since reuse is ineligible once `torch_profiler` is on.

## Evidence

Measured on MI355X with the real Magpie wrapper and a real ATOM server:

```
before teardown : 10 processes, 647,204 MiB
killpg(wrapper_pgid, SIGTERM) → SIGKILL
after teardown  : 10/10 alive, 647,206 MiB unchanged
                  leader ppid 2502 → 1, whole tree in pgid 2519
```

And inside a live optimisation run, on the profile round:

```
pid=298494  ppid=1  pgid=298317      8 workers, 2,188,381 MiB held, leader gone
```

Signalling the escaped group freed all of it.

## Approach

Escaped groups are enumerated **before** anything is signalled. The link that attributes them to this run — the wrapper still being their parent — disappears the moment the wrapper dies, so reading it afterwards is too late.

Attribution is **by descent from our own wrapper, not by cmdline**. Two reasons:

- A framework this repository has not heard of is still ours if our wrapper spawned it.
- The per-rank workers of the ATOM build measured here carry no identifying argv at all — they are `multiprocessing.spawn` children, indistinguishable by name from any other Python process. Ten processes in that tree, and only the leader mentions `atom`.

Nothing outside the tree we launched is ever signalled; our own process group is excluded explicitly, and the walk is depth-bounded.

## Tests

`test_kill_my_spawned_server_reaps_a_server_the_child_setsid_ed` — a wrapper `setsid`s a server into its own group; after teardown the server must be gone. Asserts the escape actually happened (`getpgid(server) != getpgid(wrapper)`) so the test cannot pass vacuously.

On the parent commit: `Failed: DID NOT RAISE ProcessLookupError` — the escaped server survives.

Verified on MI355X (Linux):

```
test_kill_spawned_server.py                          52 passed
every test file importing _subprocess_kill (10)      608 passed, 1 failed
ruff check / ruff format --check                     clean
```

The one failure, `test_grid_runner.py::TestSessionBudgetWarmupRounds::test_the_budget_is_re_checked_after_the_uncapped_server_restart`, is **pre-existing**: it fails identically with the changed file reverted to `main`.

## Relationship to #1416

Independent and non-overlapping. #1416 teaches the *lifecycle* teardown and `recover` to recognise an ATOM server by name; this teaches the *non-lifecycle* teardown to follow its own wrapper's descendants regardless of name. Different files, different paths, either can merge first.

## Deliberately not in scope

- The duplicated marker lists in `_server_lifecycle` and `recover` (adding a framework means editing both).
- The "pid reuse" wording both refusal messages use for what is really an unrecognised framework.
- A leaked group whose leader has already exited: `_process_group_looks_like_server` still cannot attribute it, because nothing left in the group carries a matching cmdline.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -196,6 +196,24 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ### Fixed
 
+- **Teardown reaches the server its own wrapper `setsid`-ed away.**
+  `kill_my_spawned_server` signals one process group: the wrapper's. Magpie's
+  wrappers `setsid` the serving process, which is exactly what moves it out of
+  that group — the wrapper then exits, the server reparents to init, and every
+  rank it spawned keeps its GPUs. This is the teardown used wherever
+  server_lifecycle does not apply, which includes every profile round, since
+  reuse is ineligible once `torch_profiler` is on. Measured on MI355X: a
+  torn-down ATOM server left eight per-rank workers alive holding 2,188,381 MiB;
+  signalling the escaped group freed all of it.<br/>
+  Escaped groups are now enumerated *before* anything is signalled, because the
+  link that attributes them to this run — the wrapper still being their parent —
+  disappears the moment the wrapper dies. Attribution is by descent from our own
+  wrapper rather than by cmdline: a framework this repository has not heard of is
+  still ours, and the per-rank workers of the ATOM build measured here carry no
+  identifying argv at all (`multiprocessing.spawn` children, indistinguishable
+  from any other Python process by name). Nothing outside the tree we launched is
+  ever signalled.
+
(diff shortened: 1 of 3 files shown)


zengleixin-amd:
Folded into #1416 — the two fixes are the same defect on two teardown paths and review better together. The commit is preserved there verbatim.

PR #1420, opened 2026-09-07
fix(kernel): make the rewrite lane able to reach an operator, measure it, and land it

jiaqiang-dot-liu:
A GLM-5.2-MXFP4 run gave the rewrite controller 754 minutes and got nothing: two staged tasks, both refused, no forge-loop started. Tracing that back turned up a chain of separate defects — one per stage of the lane — where every stage after the first was also broken, just never reached. This fixes the chain.

Session under investigation: `20260905T122810Z-b5a88594` (MI355X, sglang 0.5.16).

## The operator could not be named

Both tasks were refused for `operator_name does not normalize to identity.kernel_name`. The agent had named the operator from the TraceLens pseudo-op in one field and from the aiter entry point in the other, and the contract required them to agree.

`kernel_name` is the address form of `operator_name`, not a dimension the agent picks, so the parser derives it instead of checking it. forge-loop resolves its knowledge-base page from `--operator-name` alone; deriving is what keeps that page and the controller's canonical id at one address, where a rule the agent could break silently split them into two half-filled histories. A draft that still carries a `kernel_name` has the value replaced, not refused.

Replaying the session's two rejected drafts through the new publisher: both publish.

## The agent never learned it had failed

Validation runs out of process on a timer, so the refusal reached only the log. The session ended reporting both tasks as published, with 34 of 60 analysis minutes unspent — enough for several revisions of a one-field mistake.

Each refusal is now written to `rejection.json` inside the draft that earned it, and the `Stop` hook — a backend capability this agent already required and never used — holds the session open while one stands, capped at three so a draft the agent cannot fix does not consume the rest of the budget. The receiving half already worked: the publisher re-validates a draft whose mtime changed.

## Four rules cost an operator and protected nothing

`schema_version` described version skew that cannot exist, since `task.json` is written and read inside one controller run. An unknown field was refused outright although every field the run acts on is required and the optional ones are read by name. `identity.backend` was checked against the registered set even though forge-loop answers an unregistered backend with no prompt layer rather than an error. `shape_cases` was type-checked although no code reads a case.

All four are gone; unknown fields are logged and dropped. The rules that remain have teeth, and the two an agent actually trips now name the remedy rather than the rule — a `repo_root` below the top level reports the top level (already resolved, previously discarded), and a source path missing at the base commit names the repository searched and says a cross-repository reference belongs in `evidence`. That last one is the common case: a GLM-5.2 rewrite reaches aiter through sglang, and against the wrong repo it reads as an ordinary typo.

## Every campaign would have measured the unmodified original

The task repositories are editable installs, so `import` answers from a `sys.meta_path` finder pinned to the live directory and PYTHONPATH cannot outrank it. The controller handed each campaign a private git worktree, which is therefore edited and never loaded.

Finding that out mid-campaign, the forge-loop agent did the only thing that could produce a measurement and wrote the live tree instead. Nothing owned that tree, so two rejected candidates from earlier sessions were still sitting in `/sgl-workspace/aiter` days later — 386 lines whose `_FORGE_GATE_MODE` default broke MXFP4 MoE numerics, cost that session's PRELUDE ten hours and six failed patch attempts, and made the one patch the lane ever published unappliable.

A task whose repository is an editable install now borrows the live repository under a shared lock, on a campaign branch cut from the base commit, and hands it back in a `finally`. Handing back is safe precisely because it is late: a best commit that cleared correctness and the microbenchmark is already an exported patch, so the tree it was built in holds nothing not saved elsewhere. Restoring is scoped — tracked files return to the base commit one by one, only producer-owned untracked paths are deleted, HEAD moves back by ref. No `reset --hard`, no `git clean`; the repository also carries runtime caches and whatever its owner left there.

The verdict and the lock moved to `kernelforge.loop.editable_repo`. A second opinion on either is a defect: two verdicts would have one lane copy while another edits in place, and two lock implementations would serialize nothing.

## A campaign had no base commit it could name

Hyperloom dirties the framework trees by design — instrumentation patched in place, fusion's module written beside the model it edits, every lane's KEEP uncommitted until the session ends. So the code being served and the code at HEAD are different things, and an in-place borrow refuses a tree that is not the base commit it claims.

The serving tree is now sealed into a commit before the handoff is written, at the last moment the trees stand still. That makes three drifting things one object id: the base a task pins, the diff's starting point, and the state a borrowed repository is handed back at. Tracked changes only — tuned GEMM tables and JIT caches belong in no patch and so in no commit. Sealing commits without touching file contents, so a running server is unaffected.

Sealing has no inverse: the repository is left on `hyperloom/<session>-cN`, with the
session's own instrumentation and every other lane's KEEP in its history. That is
deliberate -- the base has to be the code the server is running -- but it is a durable
change to a tree the operator owns, and it means a patch from this lane has to be
rebased onto a clean upstream commit before it can go anywhere else. The seal names the
branch it left each repository on in the log.

Which repositories to seal is read from the interpreter, not from configuration: `vllm`, `sglang` and `aiter` are probed by name and resolved to their git top level. In the session all three configured sources were empty and the handoff reported no source repository at all, so a configured-only lookup would have sealed nothing and skipped every operator. A wheel-installed framework resolves to no repository, which is correct — it has no source to rewrite.

## A killed controller lost the patches checkpointing had saved

A hard timeout kills the process tree, so the controller's restore never runs and a borrowed repository is left on its campaign branch. Integration reads HEAD, finds it is not the base commit its publications name, and refuses all of them — discarding exactly what per-second checkpoint publication exists to save from the kill. Hyperloom now takes those repositories back before integration reads anything.

## Two things nobody could see

`gpu_pct` carries the operator's share of end-to-end GPU time as a field instead of a sentence in the task's prose, and the operator report renders it beside the speedup. Never read to decide anything and never type-checked: refusing a task over a number's formatting would trade an operator for a cast.

The opportunity analysis now files what it spent. It runs in the controller's own process and the ledger only read per-operator forge-loop rows, so a run that published no operator reported no spend — the sessions that spent most while buying least looked free. This one drove Opus 5 for twenty-six minutes through two context compactions and filed zero.

## Also

`editable_roots()` is memoized. It opens every editable marker under every site-packages on the path, and the checkpoint probe reached it once a second for the length of a campaign: roughly 5400 full scans per ninety-minute loop, for an install layout that cannot change inside one run.

## Test plan

- [ ] CI green
- [ ] A `KERNEL_OPT_BACKEND_ORDER=forge` session on GLM-5.2-MXFP4 reaches a forge-loop: tasks publish, the borrow succeeds, and the restored tree returns to the sealed base commit
- [ ] Deliberately malformed draft: `rejection.json` appears, the agent revises it within the same session, and the task publishes
- [ ] Kill the controller mid-campaign and confirm the published patches still integrate

## Known gaps, not addressed here

- The fusion lane edits the live tree through a shadow `GIT_DIR` and takes no repository lock. Within one KERNEL entry it finishes before the controller starts, so they are serial in fact but not by construction. Its own PR: it touches the lane that currently works.
- Private worktrees are still never reclaimed; 21 stale registrations were observed in aiter. Editable repositories no longer create any, so this now only affects non-editable patch targets, and the closing recovery sweep still reads results out of them.
- `reclaim_campaign_repositories` takes the pins the seal returned, so a seal that fails outright leaves it with nothing to reclaim.
- `release_repo_lock` is not idempotent — a double release raises `ValueError`. Unreachable today (one call site, in a `finally`), but this is now shared code.


ZhengGong-amd:
## Review: things worth changing before merge

The diagnosis chain is convincing and each fix maps to a concrete failure in the session under investigation. Pulling the editable-install verdict and the repo lock into a single `kernelforge/loop/editable_repo.py` — rather than growing a second copy in the controller — is the right call, and the docstrings/prompt were kept in sync with the contract changes. Ran the touched suites locally: `test_campaign_baseline.py`, `test_editable_repo_cache.py`, and `src/kernelforge/tests/kernel_rewrite_controller` → 219 passed.

Below are the items I would fix first.

---

### 1. [blocking] The prompt tells the agent to delete a draft it has no tool to delete

`opportunity_agent.py` — both the new system-prompt paragraph and the `_on_stop` block reason say:

> Delete the draft instead if the operator turned out not to be publishable.

But the analysis session runs with `AgentToolPolicy(read=True, search=True, write=True, shell=False, ...)`. There is no `Bash`, no `rm` — Write/Edit cannot remove a directory. So for a draft the agent genuinely cannot fix (an operator that should never have been proposed), the only escape it is told about is unexecutable: it burns all three Stop denials plus the turns in between, which is the same budget-loss shape this path exists to prevent.

Either give it an exit it can actually perform — e.g. instruct it to overwrite `task.json` with a withdrawal marker, and have `pending_rejections` treat that marker as clearing the refusal — or drop the sentence and say the draft will simply be reported as rejected.

### 2. [blocking] The in-place restore runs before the controller's closing recovery sweep and destroys what that sweep reads

`dispatcher.py` now ends in `finally: release_operator_worktree(worktree)`, which for an in-place task calls `_remove_producer_untracked()` and deletes `forge_experiments/` from the live repository.

`controller.py:357` then runs `recover_all_task_results(layout)` over every task, and `recovery.py:147` resolves the workspace with `operator_workspace(task, layout)` — which, for an in-place task, is exactly that just-emptied live repository. `_trusted_manifest(workspace)` finds nothing, leaving only the task-directory sidecar.

The ordering inside `dispatch_single_task` is right (recovery runs before the `finally`), so this only bites when the in-dispatch recovery also failed but the workspace did hold a good manifest — which is precisely the case the closing sweep exists for. `release_operator_worktree`'s docstring argues the tree "carries nothing that is not saved elsewhere"; that holds for the exported patch, not for the best manifest.

Suggested fix: copy `forge_experiments/best_result.json` (or the directory) into `layout.workspace_dir(operator_id)` or the task directory before `_remove_producer_untracked` deletes it.

### 3. Sealing has no un-seal, and that is a persistent side effect on live repositories

`_seal_one` does `checkout -B hyperloom/<session>-cN` and commits, and nothing ever moves the repository back: `reclaim_campaign_repositories` only recognises the `forge/controller/` prefix. After a session, the operator's `vllm`/`sglang`/`aiter` checkout is permanently parked on a Hyperloom branch. That is self-consistent for integration, but it is a durable mutation of a tree the user owns and the PR description does not state it. Worth calling out explicitly in the description and in a closing log line ("sealed at X, left on branch Y").

### 4. The handoff re-shells for a commit the seal already returned

`phases/kernel.py` seals and receives `baseline_pins` — a `{repo: commit}` map — and then `build_serving_context_md` immediately spawns one `git rev-parse HEAD` per repository through the new `_head_commit()` to recover the same values. That is both a redundant subprocess per repo and a TOCTOU window between the seal and the read. Passing `baseline_pins` into the handoff builder removes `_head_commit` entirely and makes the document provably name the sealed commits.

### 5. `source_repository_roots` is a pure forwarder

`forge_handoff.source_repository_roots(state)` only does `return campaign_repositories(state)`. It keeps the diff small but adds an indirection with no behaviour; importing `campaign_repositories` at the call site is more honest.

### 6. Two implementations of "reclaim a campaign branch"

`worktree._reclaim_abandoned_campaign` (checkout --force + branch -D) and `campaign_baseline.reclaim_campaign_repositories` (checkout --force + branch -D + rmtree `forge_experiments`) do the same thing at two trigger points. By the principle this PR states in `editable_repo.py`'s own module docstring — a second opinion is a defect — these should converge on one shared helper. Same for the subprocess wrapper: `campaign_baseline._git` and `forge_handoff._head_commit` are two spellings of one thing.

### 7. Smaller items

- **`gpu_pct` is interpolated into Markdown unvalidated.** `publisher.py` renders `` f"- **Operator share of E2E GPU time:** `{task.gpu_pct}`" `` while the field is `Any` and deliberately unchecked. A value containing a backtick or newline breaks the report. `{task.gpu_pct!r}` — or formatting it as a number when it is one — costs nothing and keeps the "never refuse over it" property.
- **`release_repo_lock` is not idempotent.** A second release raises `ValueError` from `fileno()` on a closed file; the function only catches `OSError`. Acknowledged in the description, but this is shared code now — a `_released` flag is a one-line fix.
- **`shape_cases` scalar-wrapping.** `task.py` wraps a non-list into `[value]`. If nothing reads a case, wrapping decides nothing either; passing it through verbatim is simpler and matches the stated rationale.
- **Seal failure surfaces far from its cause.** A repository that fails to seal produces one `log.warning`, and then *every* in-place operator for that repository later fails at `_require_tree_at`. In the spirit of this PR's other messages (which now name the remedy), that error should mention that the baseline seal for the repository may have failed.

---

Nothing above changes the shape of the design — 1 and 2 are the two that would cost real budget in the next session.

xiaofei-zheng:
## What this PR does

Fixes a chain of independent defects along the whole kernel-rewrite lane — from staging a task to landing a patch — found by tracing the GLM-5.2-MXFP4 session that spent 754 minutes and produced nothing:

1. **Task contract.** `identity.kernel_name` is now derived from `operator_name` instead of being cross-checked against it. Four rules that could cost a whole operator without protecting anything are gone (`schema_version`, unknown-field refusal, the `backend` registry check, `shape_cases` type checking); the two rules an agent actually trips now name the remedy (report the Git top level, say that a cross-repository reference belongs in `evidence`).
2. **Refusal feedback loop.** Every refusal is written to `rejection.json` inside the draft, and a new `Stop` hook holds the session open while a refusal stands, capped at three denials.
3. **Editable-install correctness.** `needs_inplace` and the repo lock move out of `forge_submit.py` into a shared `kernelforge/loop/editable_repo.py`. A task whose repository is an editable install now borrows the live repository under that lock on a campaign branch and hands it back in a `finally` (`_borrow_live_repository` / `release_operator_worktree`) instead of editing a private checkout nothing imports.
4. **A nameable base commit.** New `campaign_baseline.py` seals each serving repository's tracked changes into a commit at KERNEL entry, with the repository set probed from the interpreter (vllm, sglang, aiter) rather than read from configuration. After a hard kill, Hyperloom reclaims those repositories before integration reads any HEAD.
5. **Accounting.** `gpu_pct` is carried and rendered in the operator report; the opportunity analysis files its own LLM ledger row via a `UsageAccumulator`; `editable_roots()` is memoized.

All 229 tests in the touched suites pass locally.

## Blocking

**`release_operator_worktree` leaves campaign-created files behind in the live serving tree** — `src/kernelforge/kernel_rewrite_controller/worktree.py:229-259`

The restore order is: `git checkout <base> -- <path>` for every path in `git diff --name-only <base>`, then `_remove_producer_untracked`, then move HEAD, then `git reset HEAD -- .`.

For a file the campaign **created and committed** on its branch, `git checkout <base> -- <path>` fails with `pathspec did not match any file(s) known to git` — swallowed by `check=False` — so the file stays in the working tree. The later `reset` turns it into an untracked file, but `_remove_producer_untracked` has already run, and `is_producer_owned_path` only covers `forge_experiments`, `.forge_rewrite` and `.forge_driver_*`, so reordering alone would not catch it either.

Reproduced against this branch by calling `release_operator_worktree` directly, with a campaign commit that edits one existing file and adds one new file:

```
HEAD branch: main                  # HEAD restored correctly
a.py: a                            # tracked edit reverted correctly
status: ?? new_kernel_helper.py    # campaign-created file left behind
files: ['a.py', 'new_kernel_helper.py']
```

The consequence is exactly the residue class this PR set out to eliminate: unowned files accumulating in the live serving repository (the `/sgl-workspace/aiter` leftovers described in the PR body). Because the seal uses `add --update`, such a file never enters any commit, and `_require_tree_at` does not see it either — so it survives every subsequent campaign.

The in-place path already working in this repo, `_restore_inplace` in `src/hyperloom/agents/kernel/tools/backends/forge_submit.py`, does this correctly: it records a `baseline_untracked` inventory at prepare time and deletes untracked paths that were not in it. The new path needs the same inventory (or at minimum, move `_remove_producer_untracked` after the index reset and decide ownership against an untracked snapshot taken before the borrow). `reclaim_campaign_repositories`, which restores via `checkout --force`, has the same gap.

No other blocking issues found. The gaps already disclosed in the PR body (`release_repo_lock` is not idempotent, private worktrees are never reclaimed, a failed seal leaves nothing to reclaim) check out as described and are not on a path this PR makes reachable.

jiaqiang-dot-liu:
@xiaofei-zheng Thank you — the reproduction was the useful part, and it holds. I ran your case against the branch before touching anything and got the same `?? new_kernel_helper.py`.

Fixed in 5486ffdf3.

You had the mechanism right on both counts, and the second one is why reordering alone would not have done it. `git checkout <base> -- <path>` cannot restore a path the base does not carry, the failure was swallowed by `check=False`, and the file only reads as untracked once HEAD and the index have moved — which happened after the removal pass. And `is_producer_owned_path` was never going to catch it: it knows forge bookkeeping, not a header or a generated kernel a rewrite adds beside the file it was pointed at. By name those are indistinguishable from something the repository's owner keeps.

So it now works the way `_restore_inplace` does. `untracked_paths()` is recorded before the branch is cut, `remove_foreign_untracked()` deletes what is untracked and absent from that inventory, and it runs last — after `git reset <base> -- .`, for exactly the reason you gave.

`reclaim_campaign_repositories` had the same gap and takes the same inventory now: the seal records it per repository alongside the commit (`RepoBaseline`), and both entry points go through one `reclaim_campaign_branch()` helper rather than two spellings of it.

Your case, re-run after the fix:

```
HEAD:        master
a.py:        a                        # tracked edit reverted
git status:  ?? operator-notes.txt    # the operator's own file, untouched
repo:        ['a.py', 'operator-notes.txt']
```

The committed file, a directory the campaign created, and an uncommitted scratch file all go; the pre-existing untracked file stays. Covered by `test_release_removes_what_the_campaign_created_and_keeps_what_it_found` and `test_reclaiming_a_branch_also_removes_what_that_campaign_created`.

One thing your review changed for me beyond the fix: an existing test of mine wrote its "operator's own file" *after* the borrow, so it was asserting that a campaign-created file survives. It passed for the wrong reason, and under the inventory it correctly fails. Moved the write before the borrow.

`release_repo_lock` is idempotent now too (`_released` flag) — you were right that "unreachable today" was not worth relying on in code three lanes call.

jiaqiang-dot-liu:
@ZhengGong-amd Thank you — both blocking items were real and I had missed the second one entirely. All of it is addressed in 5486ffdf3, except where noted.

**1. The unexecutable exit — fixed.** You are right, and the failure shape you describe is the one this path exists to prevent: `shell=False`, `Bash|Shell|Task.*|Agent` denied by hook, Write cannot remove a directory. So "delete the draft" would have cost every Stop denial plus the turns discovering it.

Took your first suggestion. `WITHDRAWN_KEY` — an agent rewrites `task.json` as `{"withdrawn": "<why>"}`, `pending_rejections` stops counting that draft, and the publish scan stops offering it. Both the system prompt and the `_on_stop` reason now name that as the exit and say why it is the only one available. The draft stays on disk as the record of a candidate examined and rejected.

**2. The restore destroying what the sweep reads — fixed, and it went deeper than you wrote.** Chasing it, the sweep turns out to be *structurally* unable to recover an in-place task, not just unlucky: it runs after every task's release, and by then the campaign branch is deleted, so the best commit `export_patch_from_base` needs is unreachable no matter what the manifest says. Preserving the bundle alone would not have helped.

So recovery gets one more attempt inside the dispatch's `finally`, before anything is destroyed — the last moment the tree and the branch both exist. Publication is idempotent, so it is free when the first attempt already succeeded, and it is the only place a recovery that failed for a passing reason can be retried.

The bundle is preserved as well, moved to the task's workspace directory rather than deleted, so a run that published nothing can still be read. It cannot stay in the repository: every in-place task there is handed the same `--experiments-dir`, so leaving one campaign's state would have the next resume it — which I think is a bug the old unconditional delete was accidentally preventing.

**3. Sealing has no inverse — disclosed.** Added to the description: the repository is left on `hyperloom/<session>-cN` carrying the session's instrumentation and other lanes' KEEPs, that is deliberate because the base must be the served code, and the consequence is that a patch from this lane needs a rebase onto a clean upstream commit before it goes anywhere else. The seal's log line now names the branch it left each repository on. Accumulation across sessions is listed under known gaps; I have not added a sweep.

**4. The redundant `rev-parse` — fixed.** `write_forge_handoff` and `build_serving_context_md` take the baselines, `_head_commit` is gone. You were right that the TOCTOU window mattered more than the subprocess: the document now provably names the commits the seal produced rather than whatever HEAD happens to be when it is written.

**5. The forwarder — removed.** `campaign_repositories` is imported at the call site.

**6. Two reclaims — converged.** One `reclaim_campaign_branch()` in `worktree.py`, called by the next borrow and by Hyperloom after a kill. This also fixed a bug: @xiaofei-zheng found that the restore left campaign-created files behind, and `reclaim_campaign_repositories` had the same gap — having one implementation means the inventory fix landed in both at once, which is your argument for converging them, demonstrated.

I left the two subprocess wrappers alone. `campaign_baseline._git` and `forge_handoff._head_commit` were the pair you named and `_head_commit` no longer exists, so there is only one now.

**7.**
- `gpu_pct` → `{task.gpu_pct!r}`. Agreed, and it keeps the never-refuse property intact.
- `release_repo_lock` idempotent via a `_released` flag.
- `shape_cases` no longer reshaped — your reasoning is exactly right, if nothing reads a case then wrapping decides nothing either.
- Seal failure now says the repository's in-place operators will be skipped for want of a base, and `_require_tree_at` says a still-dirty repository is one whose seal did not take. That is the second half of the same message and I had only written the first.

492 tests pass in the touched suites.

xiaofei-zheng:
Re-checked against `5486ffdf3`.

## The original finding is genuinely fixed

Re-ran my case by calling `release_operator_worktree` directly:

```
HEAD: main
a.py: a                        # tracked edit reverted
status: ?? operator-notes.txt  # untracked file that predates the borrow, kept
files: ['a.py', 'operator-notes.txt']   # the campaign's committed file is gone
```

The shape of the fix is right: the inventory is taken before the branch is cut, `remove_foreign_untracked()` runs after `git reset <base> -- .`, `reclaim_campaign_branch()` is one implementation reached from both directions, and `RepoBaseline` carries the inventory alongside the commit. `release_repo_lock` being idempotent is a good call too.

## Blocking: the borrow-side reclaim passes no inventory, which means "delete every untracked file"

`src/kernelforge/kernel_rewrite_controller/worktree.py:271`

```python
reclaim_campaign_branch(repo_root, task.base_commit)   # baseline_untracked defaults to frozenset()
```

`remove_foreign_untracked(repo, frozenset())` reads as "everything untracked is this campaign's". With an empty inventory that is every untracked, non-ignored path in the live repository. The function's own docstring says the opposite:

> Without one, nothing here can tell those files from the operator's own and **they are left alone**.

Reproduced by calling `reclaim_campaign_branch` exactly the way `_borrow_live_repository` calls it, on a repository left on a campaign branch by a killed run:

```
before:  fusion_generated_module.py, operator-notes.txt (untracked)   debug.log (gitignored)
after:   ['.gitignore', 'a.py', 'debug.log']
```

Both untracked files the repository's owner kept are deleted; only the gitignored one survives, via `--exclude-standard`.

These files exist in a real serving tree. The seal uses `add --update`, so a new file another lane wrote — fusion's module beside the model it edits is the example from your own PR body — never becomes tracked and stays untracked for the whole session.

The path is reached when a controller is hard-killed **and** Hyperloom's `reclaim_campaign_repositories` does not run either — Hyperloom killed too, the reclaim raising, or the case in your known gaps where a seal that fails outright leaves nothing to reclaim — so the abandoned branch survives to the next borrow.

Taking the inventory before the reclaim instead of after would be strictly better, but only covers half of it — a matrix over whether the campaign commits with `add -A`:

| campaign uses `add -A` | inventory taken before reclaim | owner's untracked file |
|---|---|---|
| no | no (current) | **deleted** |
| no | yes | kept |
| yes | no (current) | **deleted** |
| yes | yes | **deleted** |

When the campaign committed with `add -A`, the owner's files were tracked on the campaign branch at the moment the inventory would be taken, so they are absent from it and read as foreign after the switch.

Getting this fully right needs the real pre-campaign inventory on this path, which means persisting `RepoBaseline` where the next borrow can read it (the controller layout) rather than only in the memory of the process that got killed. If that is more than you want in this PR, passing `untracked_paths(repo_root)` taken before the reclaim is a clear improvement over the empty set, and the docstring should then say what it actually does.

jiaqiang-dot-liu:
@xiaofei-zheng Caught. That is my regression — the removal did not exist in `_reclaim_abandoned_campaign` before I converged the two, and I added it without noticing the borrow side passes nothing. Reproduced your case first and got the same result: the fusion module and the operator's notes both deleted, only the gitignored file surviving via `--exclude-standard`.

Fixed in cc511b18c, and your matrix was the useful part — the first fix I wrote covered only half of it, exactly as you said it would.

**Unknown is now a distinct answer.** `baseline_untracked` defaults to `None`, and `None` deletes nothing and logs that it is not guessing. `frozenset()` still means "known to have been empty", so a genuinely clean repository is still cleaned. That restores what the docstring claimed.

**The inventory outlives the process.** A borrow writes it to `.git/forge_campaign_baseline.json` — beside the lock, same lifetime, removed once the repository is actually handed back — and the next borrow reads it on finding an abandoned branch. Your point that it needed persisting rather than living in the memory of the process that got killed is the whole fix; taking it at reclaim time would have been a patch over the symptom.

**Your `add -A` row turned out to be worse than a bad inventory.** Chasing it: at that point the owner's files are tracked on the campaign branch, so `git checkout --force <base>` deletes them itself, before any inventory is consulted. So both restores now go through one sequence that never uses `--force` or `reset --hard` — base content over the tree, then HEAD and the index, and only then removal. A file the campaign committed reads as untracked by that point, which is what lets the inventory judge it instead of Git judging by reachability.

The matrix, re-run:

```
                    owner's untracked file    campaign's file
no inventory / -u   kept                      kept   (nothing guessed)
no inventory / -A   kept                      kept   (nothing guessed)
inventory    / -u   kept                      deleted
inventory    / -A   kept                      deleted
```

Covered by `test_reclaiming_keeps_untracked_files_the_campaign_did_not_write` (parametrized over both staging modes), `test_reclaiming_without_an_inventory_deletes_nothing`, `test_a_borrow_records_the_inventory_for_whoever_has_to_hand_it_back`, and `test_a_later_borrow_reclaims_a_killed_campaign_from_the_record`, which walks the real path: a borrow the host killed, then the next one.

497 tests pass in the touched suites.

Worth saying plainly: converging the two reclaims is what let a single missing argument reach a path I had not tested, and it is also what made one fix cover both. I would still converge them, but I should have looked at every caller when I did.

xiaofei-zheng:
Re-reviewed at `cc511b18c`.

**The round-2 finding is fixed, and I confirmed the matrix.** Ran the four cases through the real `reclaim_campaign_branch` on a repository left on an abandoned campaign branch (owner's `operator-notes.txt` and a gitignored `debug.log` present before the campaign, campaign writes `fusion_generated_module.py` and mutates `a.py`, committed once with `add --update` and once with `add -A`, then the process dies without a release):

```
add -u, record present : ['.gitignore', 'a.py', 'debug.log', 'operator-notes.txt']  HEAD=main
add -A, record present : ['.gitignore', 'a.py', 'debug.log', 'operator-notes.txt']  HEAD=main
add -u, record missing : campaign file kept, owner's file kept, warning logged
add -A, record missing : campaign file kept, owner's file kept, warning logged
```

The `None`/`frozenset()` distinction and the ordering change in `_restore_tree_to_base` both do what you describe, and `record_campaign_baseline` restoring `origin_ref` means the repository comes back on its own branch instead of detached, which the old `checkout --force` did not. `test_worktree.py` passes (26).

## Blocking

**`seal_campaign_baseline` runs before anything reclaims, so a whole-process kill turns a dead campaign's rewrite into the next session's serving baseline.** `src/hyperloom/orchestrator/phases/kernel.py:4034` seals at KERNEL entry; the only reclaim is at `:4144`, after the controller has already exited *in the same process*. When the hard timeout kills only the controller, that ordering is fine. When the orchestrator itself dies — OOM, preemption, container restart, Ctrl-C — nothing reclaims, and the next session's KERNEL entry seals a repository that is still sitting on `forge/controller/…` with the campaign's commit as HEAD.

Both shapes end with the campaign's unvalidated code as the baseline, and neither is recoverable afterwards:

```
# session 1: borrow, campaign rewrites kernel.py, commits, whole process is killed
# session 2: KERNEL entry -> _seal_one()

campaign committed everything (tree clean):
  sealed commit  = the campaign's own commit          kernel.py = REWRITTEN_BY_DEAD_CAMPAIGN
  HEAD           = forge/controller/dead-1
  later borrow   -> reclaim_campaign_branch(repo, task.base_commit) fires, but
                    task.base_commit *is* the campaign commit, so it restores to it
                    and then forget_campaign_baseline() drops the only record of
                    the real pre-campaign commit

campaign left the tree dirty:
  _seal_one does `checkout -B hyperloom/s2-c0` from the campaign branch and commits,
  sealed commit  = campaign content + its dirt        kernel.py = REWRITTEN_BY_DEAD_CAMPAIGN = 2
  HEAD           = hyperloom/s2-c0
  later borrow   -> CAMPAIGN_BRANCH_PREFIX no longer matches, reclaim returns ""
                    and the record is left on disk unread
```

The information needed to avoid this is already written down by this commit and never read: `record_campaign_baseline` persists `base_commit`, but `read_campaign_baseline` returns only `(origin_ref, untracked)` and drops it. That field is exactly the pre-campaign commit `_seal_one` would need to reclaim *before* it seals.

So the fix looks like reclaiming at KERNEL entry rather than only at exit: for each repository in `campaign_repositories(state)`, if `.git/forge_campaign_baseline.json` exists, restore to its recorded `base_commit` and inventory first, then seal. That also makes the record's `base_commit` field load-bearing rather than decorative.

jiaqiang-dot-liu:
@xiaofei-zheng Right on both counts, and the second one is the whole fix: `record_campaign_baseline` persisted `base_commit` and `read_campaign_baseline` returned everything except it. Fixed in 88a4502.

**The reclaim moved to the seal, because the seal is the only place it can go.** Both existing reclaims run *after* a campaign, in the process that held it — so neither can cover a process that died holding it. The next session's KERNEL entry is the other end of that window, and it is the last moment the tree is still answerable. `_seal_repository` now checks whether HEAD is on a campaign branch and returns the repository before `_seal_one` commits anything.

Your two shapes collapse into one once the reclaim runs first: in both, HEAD is still on `forge/controller/…` at that point. The `hyperloom/s2-c0` case — where `CAMPAIGN_BRANCH_PREFIX` no longer matches and the record is left on disk unread — only existed because the seal got there first.

**The record is now the authority, not a fallback.** All three answers a restore needs come from it — commit, ref, inventory — because the borrowing process is the only one that saw the repository before it changed. Arguments stand in when no record survived. That also fixes the borrow-side call you traced: `reclaim_campaign_branch(repo, task.base_commit)` no longer restores to `task.base_commit` when a record disagrees, which was your first shape's second half.

**Where I went further than your suggestion.** A repository on a campaign branch with *no* record is not sealed at all, and says so. Restoring to a commit nothing vouches for is the same defect from the other side, and I could not find an honest way to derive the pre-campaign commit from the branch alone. So its operators are skipped — which is what a missing baseline already means — and the log names the branch for whoever wants to recover it by hand.

**The lock, for this and only this.** An ordinary tree still seals without it. A tree on a campaign branch may be one a live campaign is writing, and sealing mid-rewrite is the same poison by a different route. `test_an_ordinary_tree_is_sealed_without_the_repository_lock` is the guard against my having gated all sealing on a lock file.

Your round-2 matrix still holds — I checked before assuming it: `test_reclaiming_keeps_untracked_files_the_campaign_did_not_write` (both staging modes) and `test_reclaiming_without_an_inventory_deletes_nothing` pass unchanged. `frozenset()` still means "known to have been empty"; `None` still deletes nothing.

New tests, each of which I ran against the branch *before* the fix — the three seal-side ones fail there, and `test_an_ordinary_tree_is_sealed_without_the_repository_lock` passes there as it should:

- `test_a_dead_campaigns_rewrite_is_not_what_the_next_session_seals`
- `test_a_campaign_branch_nothing_recorded_is_refused_rather_than_sealed`
- `test_a_repository_a_live_campaign_is_still_writing_is_not_sealed`
- `test_reclaiming_with_neither_a_record_nor_a_base_commit_changes_nothing`
- `test_a_recorded_base_commit_outranks_the_one_the_caller_offers` — fails before the fix exactly as you would predict: it restores to the caller's older commit

469 tests pass in the touched suites. The branch is also merged up to `main`, which had landed a repo-wide docstring compaction across the files this PR touches.

One thing worth saying: three rounds of this were all the same defect wearing different clothes — something that knew the pre-campaign state failed to write it down, or wrote it down and nobody read it. It is now written once, read by every path that restores, and dropped only when the repository is actually back.

jiaqiang-dot-liu:
@ZhengGong-amd Coming back to two of your items. One I reported as done and had not actually done; the other is still open by choice, and I want to be explicit about that rather than leave it implied.

**Item 7, `shape_cases` — you were right, and my last reply was wrong.** I said it was no longer reshaped. What 5486ffd actually did was move the wrap inline: `shape_cases = [shape_cases]` became `tuple(...) if isinstance(shape_cases, list) else (copy.deepcopy(shape_cases),)`. A non-list was still turned into a one-element sequence — identical behaviour — and I had written a comment beside it claiming nothing was reshaped, so the code and the comment now disagreed. Worse, the test asserting it was already named `test_shape_cases_are_carried_verbatim` while pinning the conversion.

Fixed properly in 88a4502. The field is typed as written (`shape_cases: Any`) and the parser does `copy.deepcopy(shape_cases)` and nothing else. Your argument is what decided the type: nothing reads a case, so reshaping one decides as little as refusing it would, and it costs the reader the answer the agent wrote. I checked that nothing reads `task.shape_cases` and that `KernelRewriteTask` is never hashed, so a frozen dataclass carrying a non-tuple here is safe. Covered by `test_shape_cases_are_carried_verbatim` (now asserting the list it was given) and `test_shape_cases_that_are_not_a_list_are_not_reshaped`; both fail before the fix.

**Item 3, sealing has no inverse — still not fixed, and I am not proposing to fix it here.** What you asked for concretely is in: the description states it, and `_seal_one` logs `sealed <repo> at <commit> (was <head>) and left it on <branch>`. What remains is the underlying thing — nothing moves the repository back off `hyperloom/<session>-cN`, and those branches accumulate across sessions. That is deliberate for the reason in the description (the base has to be the code the server is running), but "deliberate" does not make it free, and I did not want to write a branch GC into this PR: it would delete history, and I cannot tell a branch nobody wants from one an operator is still holding something on. It stays under known gaps as a decision for a separate change.

Worth connecting: item 3 is where @xiaofei-zheng's third round came from. A repository parked on a branch with no inverse is exactly the state a killed session leaves, and the next session's seal was committing on top of it — turning a dead campaign's unvalidated rewrite into the baseline. That is now fixed at the other end (88a4502): the seal reclaims an abandoned campaign before it commits, or refuses to seal that repository if nothing recorded what it held beforehand. So the consequence you were pointing at is closed even though the accumulation is not.

Your item 6 also paid off again here. Because there is one `reclaim_campaign_branch`, making the record the authority on the base commit fixed the borrow path and the post-kill path at once, and gave the new seal-time path something correct to call rather than a third implementation.

Everything else from your review is unchanged and verified against the current tree: the withdrawal exit (`WITHDRAWN_KEY`, named in both the system prompt and the `_on_stop` reason), recovery inside the dispatch `finally` with the bookkeeping archived before anything is deleted, `_head_commit` and `source_repository_roots` both gone, one reclaim and one `_git` wrapper, `gpu_pct` rendered with `!r`, `release_repo_lock` idempotent, and the seal failure naming the operators it will cost. 469 tests pass in the touched suites.

xiaofei-zheng (approved):
Re-reviewed at `88a4502`. **No blocking issues left — approving.**

## What I verified

**The seal-time reclaim closes the round-3 finding, both shapes.** Ran the four cases through the real `_seal_repository` on a repository a killed session left on `forge/controller/dead-1` (owner's `notes.txt` present before the campaign, campaign rewrote `kernel.py` and added `gen.py`, committed with `add -A`, second variant leaving the tree dirty on top):

```
campaign clean / record present : baseline=b3a96712  HEAD=main  kernel.py='SERVED = 1'  files=[kernel.py, notes.txt]
campaign dirty / record present : baseline=b3a96712  HEAD=main  kernel.py='SERVED = 1'  files=[kernel.py, notes.txt]
campaign clean / no record      : baseline=None      HEAD=forge/controller/dead-1  (refused, branch named in the log)
campaign dirty / no record      : baseline=None      HEAD=forge/controller/dead-1  (refused, branch named in the log)
```

The dirty shape no longer diverges, which is the point: with the reclaim ahead of `_seal_one`, `CAMPAIGN_BRANCH_PREFIX` still matches in both. `gen.py` is gone and `notes.txt` survives, so the inventory is doing the deciding rather than reachability. The record now outranking the caller's `base_commit` also fixes the half of the first shape where `reclaim_campaign_branch(repo, task.base_commit)` would have restored *to* the campaign's own commit.

**The refusal path is safe rather than merely honest.** A repository skipped for having no record is absent from `sealed` in `build_serving_context_md`, and if it were reached anyway `_require_tree_at` refuses the borrow — a tree still carrying the campaign's rewrite cannot pass `git diff --quiet <base>`. So "skipped" ends at a refused operator, not at a campaign measured against a rewrite. Taking the lock only for a repository on a campaign branch is the right narrowing, and `test_an_ordinary_tree_is_sealed_without_the_repository_lock` is the guard I would have asked for.

**`shape_cases` is now genuinely verbatim.** `copy.deepcopy(shape_cases)` and nothing else, with the field typed `Any`. I checked the two things that would make that unsafe: nothing in the controller reads `task.shape_cases` (the only consumers are the Hyperloom-side `task_group_shape_cases` helpers, which work on the raw group dict, not on the parsed task), and `KernelRewriteTask` is never hashed or put in a set, so a frozen dataclass carrying a list here is fine.

**Round 2 still holds** — re-ran the untracked-ownership matrix at this commit, unchanged.

Tests: `test_campaign_baseline.py`, `test_worktree.py`, `test_task_parser.py` — 92 passed. CI green on the shards that have reported.

Agreed on leaving the `hyperloom/<session>-cN` accumulation out of this PR. The consequence that made it dangerous is what got fixed; a branch GC that cannot tell an abandoned branch from one an operator is holding is a separate decision, and guessing wrong there is the same class of defect as guessing a base commit.

PR #1422, opened 2026-09-07
fix(server_patcher): stop crashing on a namespace-package sglang install

BaoYunkai:
## Problem

`_discover_sglang_plan` / `_discover_sglang_ck_plan` in `_server_patcher.py` (the TraceLens roofline patcher and the KernelForge fp8 block-scale CK patcher) resolve the installed SGLang package root via:

```python
sglang_module = Path(sglang.__file__).resolve()
```

On a PEP 420 namespace-package SGLang install, `sglang.__file__` is `None`, so this raises `TypeError: argument should be a str or an os.PathLike object ... not 'NoneType'` instead of failing soft. This breaks the documented contract of `ensure_sglang_patched_for_tracelens()` / `ensure_sglang_patched_for_ck_blockscale()` ('any failure returns False') and crashes the profile run that calls them.

## Fix

Add `_sglang_package_dir()` to resolve the package directory from `__file__` when available, falling back to `__path__[0]` (always present on an importable package, namespace or regular, and already the package directory itself). `_resolve_sglang_apply_root()` now takes the resolved package directory instead of a file path, so it always returns a plan (removing a redundant `None` branch). Both discover functions now fail soft with a warning instead of raising when neither `__file__` nor `__path__` resolves anything.

Added `test_server_patcher_sglang_namespace_package.py`: reproduces the original crash, verifies the fail-soft path, and guards the common (non-namespace) editable/wheel install path is unchanged.

### DIFF
--- a/src/hyperloom/inference_optimizer/tests/test_server_patcher_sglang_namespace_package.py
+++ b/src/hyperloom/inference_optimizer/tests/test_server_patcher_sglang_namespace_package.py
@@ -1,0 +1,132 @@
+# SPDX-FileCopyrightText: 2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""SGLang plan discovery must fail-soft, not crash, on a namespace-package install.
+
+A PEP 420 namespace-package ``sglang`` install (``sglang.__file__ is None``,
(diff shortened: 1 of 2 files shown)


BaoYunkai:
Abandoning: this was a bug fix (namespace-package sglang crash), not the code simplification task it was meant to address.

PR #1423, opened 2026-09-07
refactor(server_patcher): dedupe the ensure_* plan-to-patched tail

BaoYunkai:
## Before

`ensure_vllm_patched_for_tracelens`, `ensure_sglang_patched_for_tracelens`, and `ensure_sglang_patched_for_ck_blockscale` each repeated the same three-line tail:

```python
plan = _discover_*_plan(...)
if plan is None:
    return False
return _ensure_patched(plan)
```

## After

Extracted `_ensure_patched_from(plan)` so there is one place that answers 'is this optional plan patched', instead of three copies. No behavior change; all 42 existing `_server_patcher` tests pass unchanged.


BaoYunkai:
Abandoning: not a meaningful enough simplification to keep.

PR #1424, opened 2026-09-07
refactor(kernel): merge the collective lane into the rewrite controller

jiaqiang-dot-liu:
Communication operators and dense operators are the same rewrite job. The only thing that ever separated them was the measurement harness — N ranks, a process group, cross-rank reduction — and that belongs to a task's driver, not to a lane of its own. Everything above the driver was a second copy of what the kernel rewrite controller already does: pick a target, prepare a workspace, run forge-loop, integrate the patch.

**97 files, +2,186 / −24,241.**

## What replaces the lane

A collective is now an ordinary controller task carrying one extra field.

| Stage | How a collective is handled |
|---|---|
| Selection | Opportunity analysis, from the same `kernel_candidates.json` every operator is chosen from. The `nccl_summary` rows that resolve a mangled comms symbol to editable source are kept and still injected. |
| Contract | `world_size` on `KernelRewriteTask`, optional and defaulting to 1. Deliberately not part of the identity six-tuple, which is the experience KB's primary key; parallelism is carried in `operator_name`. |
| Workspace | The live repository is borrowed under a lock when it answers `import` from an editable install, because a private checkout would be edited and never loaded. Handed back at the base commit it was taken at. |
| Driver | Copied into the workspace under the producer-owned `.forge_driver_` prefix, so it stays untracked, never reaches the exported patch, and is removed with the rest of the campaign's leavings. |
| Measurement | `--nproc-per-node` reaches forge-loop, and preflight holds the run to the distributed contract. |
| Integration | `integrate_controller_patches`, the same end-to-end gate every controller patch passes. |

Every collective-specific mechanism is gated on `world_size > 1`. A task that never asks for ranks takes none of these paths; the single-rank probe is byte-identical to before.

## What is enforced, and what is not

This distinction is the part worth reviewing. The retired lane locked its measurement invariants into a driver it generated from a template. The driver is now authored per task, so each invariant is either observed at runtime or stated in the contract the author reads — and the two are not interchangeable.

**Observed, and rejected when violated.** The graph probe already writes one shard per rank; it additionally records the device each rank bound, the reduce ops the benchmark used, and whether the process group was still standing at exit. That rejects a single-process run for a multi-rank task, a rank count that disagrees with the task, two ranks sharing a device, an averaging reduction that hides the laggard, and a leaked process group. None of these can be read off the source, which is why an earlier attempt at this — matching the driver text against idioms — was replaced.

**Stated in the driver contract, measured by nothing.** Barrier placement inside the timed region, rank-distinct inputs, a correctness reference that is itself distributed, and parity validated over two back-to-back calls rather than one. The contract says which of its requirements are enforced and which are not, and separates a slip that costs a wasted validation from one that lets a candidate quietly dropping work report a real speedup — the end-to-end gate downstream can adopt on throughput without scoring accuracy, so nothing after the driver is guaranteed to catch the second kind.

**Refused before the campaign starts.** A rank count on an operator whose name does not read as multi-rank, and a rank count larger than the visible GPUs. The name test is fuzzy and deliberately permissive: kernel names arrive camel-cased, mangled, abbreviated (`custom_ar`, `ag_gemm`) or named for a role rather than a collective (`EpDispatchCombineOp`), and refusing a real collective costs the operator the whole lane while letting a doubtful one through costs a check the run makes anyway. The GPU test skips rather than fails — the task is sound, the machine is not — and is silent when the count cannot be established.

## Also removed

`forge_submit.py` and its dependencies. `submit()` lost its last caller when #1408 removed `run_optimization`, and `forge_collective` was the only remaining runtime importer, so the backend went with the lane along with `_flydsl_rewrite.py`, `_collective_names.py`, and the stdout parsers for markers nothing emits any more. The vendor-operator-playbook registry and its `classify_patchability` tests describe live behaviour and stay; only the tests that drove `submit()` are gone.

## Breaking change

The `collective` section of `session_breakdown.json` is gone. Communication results now appear as ordinary controller rows, which is the point of the merge, but a dashboard reading that section will find it missing. `docs/reference/session-breakdown.md` is updated.

Two deterministic entry gates are also gone: TP > 1, and the comm-share floors. Target selection is KernelForge's job per #1408; if analysis is later found to spend budget on low-communication workloads, the threshold belongs in its prompt rather than back in Hyperloom.

## Verification

On 2× and 8× MI355X against the live aiter checkout:

- A full campaign through `dispatch_single_task`: in-place borrow, `--nproc-per-node` forwarded, distributed preflight passed, `change.patch` + `publication.json` produced. 1.0707x on `custom_all_reduce.cuh`, correctness SNR 49.37 dB. The repository came back at its base commit with no leftover branch.
- Each distributed rejection, by mutating a working driver: averaging reduction, leaked process group, a rank count disagreeing with the task, and two ranks on one device. The conforming driver is not rejected.
- The preparation agent turning a single-process driver into a conforming 2-rank one from the contract alone: `ranks_ok: True`, 112 graph replays.
- Concurrent borrows of one repository are refused by the lock.
- `pytest src/hyperloom src/kernelforge`: 20,097 + 4,615 passing.

## Not verified

A full TP > 1 session has not been run, so opportunity analysis publishing a communication task **on its own** is the one link no measurement covers. Everything downstream of a published task is exercised above. Five framings that would have argued an analyst out of publishing a collective — a vendor-library kernel reading as non-editable, a per-row share understating a split collective, absent shapes reading as weak evidence — have been removed from its prompt, but removing the objections is not the same as observing the decision.

Preparation costs roughly 40 minutes of a task's 90-minute budget when the driver needs distributing, which is always for a collective. That fits, but not with much room.


ZhengGong-amd:
Reviewed. The direction is right and this is a real subtraction — the retired lane was a second copy of the controller, and deleting it rather than relocating it is the correct call. The `PROBE_FAILED` / `PROBE_DISTRIBUTED_VIOLATION` split (a verdict on the driver must not be reported as "the tool broke"), collapsing "where does this task's workspace live" into `operator_workspace()`, and reusing the existing `path_ownership` registration for `.forge_driver_*` are all done well. The driver contract note is the best-written prose in the PR: every requirement carries its mechanism, and the closing paragraph separating "costs a wasted validation" from "lets a candidate that drops work report a real speedup" is exactly the distinction that matters.

Requesting changes on two items; the rest are small.

## Must fix before merge

**1. The orchestration prompt still teaches a lane that no longer exists.**

`src/hyperloom/orchestrator/prompts/orchestration.md` still carries `run_collective` at lines 296–299 and 398–408 (`run_collective_done`, `collective_integrate_done`, `collective:forge_collective`), and `src/hyperloom/inference_optimizer/SKILL.md:29` still describes `run_collective` as a Coordinator-owned lane sharing the handler table. The handler, the `COORDINATOR_OWNED_KERNEL_REQUEST_KINDS` entry and every producer of those responses are gone, so the orchestration agent reads a contract naming a request kind and an inbox response that cannot occur.

Worth noting how this got through: the PR updated the **docstring** of `test_prompt_kernel_lane_ownership.py` to drop `run_collective` from the owned-lane list. That test asserts no part of the prompt advertises those kinds as requestable — removing `run_collective` from its list removed the only thing constraining the leftover prompt text. The test was routed around, not satisfied. Please also add the inverse assertion: a retired request kind must not appear in the assembled prompt at all, otherwise the next lane removal repeats this.

**2. The closing recovery sweep is inert for exactly the in-place case.**

The new `finally` in `dispatch_single_task` calls `release_operator_worktree`, which for an in-place task runs `_remove_producer_untracked` → `shutil.rmtree(repo_root / FORGE_LOOP_OUTPUT_DIRNAME)`. The closing sweep at `controller.py:328` (`recover_all_task_results` → `recover_task_result` → `operator_workspace()` → `_trusted_manifest(workspace)`) reads the best manifest out of that same `forge_experiments` directory. For every in-place task it is now guaranteed to return `"no trusted forge-loop best result"`.

The in-dispatch `recover_task_result` (line 204, inside the `try`) covers the happy path, so this is not a regression of a normal run — what is lost is the safety net, in the case that needs it most (host kill, timeout), and it is lost silently. The comment on the `finally` asserts the opposite:

> A private checkout is left standing instead, because the controller's closing sweep still reads results out of it.

Either copy the best manifest out before release (same best-effort pattern as `_keep_preparation_audit`, which is already right next to it), or have in-place tasks opt out of the sweep explicitly so it does not appear to have looked. Either way the comment needs correcting.

## Strongly suggested, same round

**3. Rule 7 is prose where a machine check is nearly free.** `world_size` is deliberately outside the identity six-tuple, so whether a TP-2 and a TP-8 all-reduce collide in the experience KB depends entirely on the analyst obeying "operator_name must encode the parallelism". `parse_task_payload` already validates the name and `_PARALLEL_SUFFIX` already exists; it is just not required. Requiring the suffix when `world_size > 1` — or appending `_tp{n}` to `operator_id` automatically — turns a convention into a guarantee and lets rule 7 come out of the prompt.

**4. Three comments that contradict their code.**

- `_visible_gpu_count`: the comment says "An empty mask means no GPU at all, which is a real answer", but it returns `0`, and the caller's `if visible and ...` treats `0` as falsy. An explicit zero-GPU mask and "could not read" are indistinguishable downstream, and the real answer the comment claims is discarded. Return a distinct unknown sentinel, or change the comment.
- `_read_graph_probe_shards` docstring says "the run must reduce across ranks". It does not: `_distributed_contract_violation` returns clean when `observed_ops` is empty, deliberately — `test_a_run_that_reduced_nothing_is_not_second_guessed` documents that choice. The choice is right; the docstring needs to stop claiming otherwise.
- `needs_inplace`: "(or contains/sits under)" understates the consequence. `repo.startswith(r + os.sep)` means one editable subpackage inside a monorepo causes the whole monorepo to be borrowed in place and force-checked-out. `_require_tree_at` guards it, but the docstring should say what the match implies.

**5. Drop the fuzzy-matching layer in `_collective_names.py`.** The module's own docstring argues it is deliberately permissive and that the real verdict comes from watching the run, so letting a doubtful name through costs little. That is a good argument — and it also argues against `SequenceMatcher` over three window widths for every phrase at every offset, plus `_FUZZY_RATIO` and `_FUZZY_MIN_LEN` to tune. `_PHRASES` (substring), `_ABBREVIATIONS` (whole word) and `_PARALLEL_SUFFIX` already carry the load; the fuzzy layer only adds truncated/misspelled spellings, whose cost of escaping is ~0 by the module's own reasoning. Removing it drops ~40 lines, two magic numbers and an O(n·m·k) scan.

## Next round is fine

**6. `editable_repo.py` has no direct test.** It is a lift-and-shift from the deleted `forge_submit.py`, which is the right instinct, but it is also the highest-blast-radius code here — it decides whether the user's live repository gets `checkout --force` / `branch -D` / `rmtree`. The `test_forge_submit_*` files that exercised it were deleted with it, so net coverage on this logic goes down. Its three `except Exception: pass` / `except OSError: continue` sites are the specific concern: a failed probe makes `needs_inplace` return `False`, which yields a private checkout, which is the exact failure mode the module's opening docstring describes — the campaign edits a copy that is never imported and reports no improvement. Silent, and indistinguishable from a genuine null result. The style is legacy too (`d` / `n` / `m` / `_fh2`), which is fine for a move but worth a naming pass now that it is a shared kernelforge module.

**7. Reference reductions and metric reductions are indistinguishable to the probe.** The probe hooks `dist.all_reduce` globally, while the driver contract requires the correctness reference to be the matching `torch.distributed` collective. A conforming driver therefore records the reference's `SUM`. A driver that reduces no metric but prints from rank 0 lands on `observed_ops = {SUM}`, `gathers = 0` and is reported as an averaging reduction — a false positive on a driver that did nothing wrong. Consider giving the metric reduction a recognizable entry point in the contract (e.g. a harness-provided `reduce_metric()`) so the two can be told apart.

**8. Small cleanups.** `if task is not None` in the new `finally` is unreachable-false — `task` is assigned before the `try` and the `parsed.task is None` branch already returned. And `_all_reduce`'s `_sentinel` + dual forwarding can be the `*a, **k` passthrough that `_gathers`' own `_wrap` already uses two blocks below; two styles for the same job in one file.

## On the unverified items

Agreed with the PR's own assessment, and the honesty is appreciated — "removing the objections is not the same as observing the decision" is the right framing for analysis publishing a collective on its own. Two things worth watching immediately after merge: that link, and the preparation cost (~40 of 90 minutes when the driver needs distributing, which is always for a collective). The second is the one with no headroom.

ZhengGong-amd:
Re-checked against the current head (`a9f46ac4f`). The only movement since my review is a merge from `main` — `git diff --stat` between the two heads is empty, and every item above is still present verbatim. One correction to my own earlier comment: `_insufficient_gpus` does still have its docstring and `world_size <= 1` early return; I misread a truncated diff. That point is withdrawn.

More importantly: **that list should not be the fix.** Patching all five items one by one leaves the code longer than it is now. Two root changes make most of them disappear instead.

**1. Ship a fixed `dist_harness.py`, the way `graph_harness.py` is already shipped.** `task_preparer.py:2145` already pre-places a known-good, capture-guarded harness in the workspace and tells the agent "Do NOT rewrite it". The distributed case should reuse that pattern: torchrun bootstrap, `set_device(LOCAL_RANK)`, process-group init/destroy, `WORLD_SIZE` assertion, per-rank seeding, barrier outside the timed region, `MAX` reduction, rank-0 printing, two-call parity — all KernelForge code, with the agent authoring only `build_inputs` / `call_candidate` / `reference`.

Today those invariants are generated text, then observed by a probe, then restated as prose, then guarded by a name heuristic — four layers, and the four properties that matter most (barrier placement, per-rank distinct inputs, distributed correctness reference, two-call parity) are measured by none of them. A shipped harness makes them structurally unwritable errors and deletes `_distributed_contract_violation` (~60), the probe's `dist` interposer (~50), most of the contract note (~30), and `_insufficient_gpus` / `_visible_gpu_count` (~35).

**2. Remove the `needs_inplace()` runtime fork.** "Where does the code live" should have one answer by construction — either the workspace *is* the import root (per-campaign editable install / clean import env), or every task runs in place serialized. Not decided at runtime. The interest on that fork is `operator_workspace` + flock + `_require_tree_at` + `release` + `_reclaim_abandoned_campaign` + `_remove_producer_untracked` + `_keep_preparation_audit` + the closing sweep that item 2 above shows is already broken — ~370 lines, and items 2, 6 and 8 all hang off it.

Net ≈ −545 lines rather than +50 of new defence. `_collective_names.py` (123 lines) also goes: with a shipped harness the gate can be one evidence check — the candidate row came from an `nccl_summary` — instead of fuzzy-matching operator names.

Happy to be wrong on either; but I'd rather settle the shape now than review the next round of patches on top of this one.

jiaqiang-dot-liu:
Pushed; head is now `70c71bb8d`. Your re-check read `a9f46ac4f`, which was four commits behind my working tree — that is on me for not pushing, and it means part of your evidence was stale. Specifically, item 2's closing sweep was already fixed in `94a3cbd56` before you cited it as still-broken support for root change 2.

## The list, before the shape argument

All eight are done, but not all as patches.

Items 1–5 and 8 landed in `8a7acdf58`, `94a3cbd56`, `11cd7aa5a`, `19b43be0a`. Two things from that round are worth flagging. You were right that the test was routed around, and the inverse assertion is now there: `_RETIRED_VOCABULARY` names the retired terms directly and `test_no_retired_request_kind_survives_in_the_prompt` parametrizes over them, because a retired lane leaves no table entry for anything derived to notice. Separately, `19b43be0a` made the probe watch `dist.reduce` and not only `dist.all_reduce` — reducing MAX to rank 0 is the natural shape when only rank 0 prints, and watching `all_reduce` alone had been refusing such drivers for the collective they chose rather than for what they measured. That was a live false rejection, not only wording.

Item 2 took your second option: in-place tasks opt out of the sweep explicitly, so `_nothing_to_recover_reason` says the borrowed repository was handed back rather than reporting a verdict on a bundle that is gone. To be clear about what that does and does not buy — the comment no longer lies, but the safety net is not restored. A host kill during an in-place campaign still loses the result, and the sweep now says so instead of implying it looked.

Item 6 is `c6eb4466f`. Item 4 had one instance more than you found: `tracelens_analysis.py` explained the nccl-summary override by what `is_collective_candidate` would do with a re-derived `is_multigpu`, and that function left with the lane. `74b5f7574` restates the reason against the invocation contract, which is what gates on the flag now.

## Root change 1 is done

`dist_harness.py` owns the torchrun re-exec, device binding, process-group lifetime, per-rank seeding, warmup, graph capture, barrier placement, the cross-rank reduction, rank-0 printing and teardown. The driver supplies `build_inputs` / `call_candidate` / `reference`.

Two of the four properties you named are now unwritable rather than checked. Inputs cannot be identical across ranks or across the two parity calls, because the harness reseeds instead of trusting `build_inputs` to vary. No barrier can land inside the timed region, because the driver never sees the timed region. Two-call parity is structural for the same reason: both candidate calls are issued before either is compared, so a second call overwriting the first's registered scratch buffer is visible.

That collapsed the observation side to one question. The probe no longer interposes `torch.distributed` at all — it records whether the rank measured inside the harness, and `_distributed_contract_violation` asks only that. This is also the real answer to item 7: the reduction-op inference could not tell a metric reduction from the correctness reference's, so it is gone rather than made more precise.

## Two deliberate deviations

`_insufficient_gpus` stays. You listed it among what the harness deletes, and the harness does refuse the same shortage, but the two say different things. Asked before a campaign opens, too few GPUs is a skip that leaves a sound task for a host that can run it. Left to the harness it is a driver exiting non-zero, which is a failed campaign against a task that deserved none, after preparation has already spent most of its budget. Deleting it converts a machine problem into a task verdict.

`examples/mori_ep_dispatch_combine` is not ported. It spawns its own ranks via `mp.spawn` with a fixed master port, which preflight now rejects, and its measurements were taken on real 8-GPU hardware in that shape. I marked it read-for-the-operator-not-the-launch rather than rewrite 520 lines of mori-specific code with no hardware to re-validate on. It needs a real port, and that port should not be a blind one.

## The line count does not land where you predicted

Root change 1 on its own is a net **add** of roughly 300 production lines: the harness is 368, `task_preparer` nets −69 (+128/−197). Your −545 assumed both changes, and the deletion you costed mostly belongs to root change 2. What change 1 buys is not size — it is four properties moving out of "stated in the contract, measured by nothing" and into code that cannot express the error.

## Verification, plainly

The harness's capture, timing, parity and scoring paths run for real on one GPU with the cross-rank collectives stood in for: both candidate calls issued before either comparison, the two parity calls never receiving equal inputs, exactly one barrier before the timed region, each sample reduced with MAX, and a graph replay per timed iteration.

Not verified: anything that needs more than one device — the binding itself, and any real NCCL/RCCL behaviour. This needs a 2×/8× run before merge and I am not claiming otherwise.

## Root change 2 I would like to push back on

Removing the `needs_inplace` fork is ~370 lines and I agree the fork is the interest payment you describe. But both replacements carry costs the estimate does not. A per-campaign editable install means a real build of the target repository per campaign; for aiter with its HIP kernels that is minutes at best, and a plausible new failure mode on a path that currently has none. "Every task runs in place, serialized" gives up same-repository concurrency outright, and preparation budget is already the tight resource here by your own reading (~40 of 90 minutes). I would want the replacement named concretely, with its build cost, before trading the fork for it.

Happy to be wrong on that — but I would rather settle it the way you asked for the shape: before the next round of patches.

xiaofei-zheng:
Re-reviewed at head `70c71bb8d`. The `dist_harness` turn is the right one — moving the launch, device binding, seeding, barrier placement and the cross-rank reduction out of generated driver text and into shipped KernelForge code replaces four weak inference layers with one fact the probe can actually establish (`MEASURED[0] > 0`). I read the whole of `src/kernelforge/loop/dist_harness.py`, the rewritten probe/`_distributed_contract_violation`, the harness placement in `prepare_task`, `carries_parallelism_suffix` in `task.py`, and the `_nothing_to_recover_reason` / `finally` change in the dispatcher, and found no correctness defect in the production code.

But the branch is red, and it was green at the merge base, so this is the PR's own regression, not the environment's.

## Blocking

**1. `src/kernelforge/tests/test_distributed_contract_note.py` — 6 tests assert prose that `e0b9ea08f` deleted.**

`_distributed_contract_note` was rewritten from the hand-written-launch contract to the `dist_harness` contract, and this file (added earlier in this same PR) still asserts the old strings. Reproduced locally at `70c71bb8d` with `PYTHONPATH=src pytest src/kernelforge/tests/test_distributed_contract_note.py`:

```
FAILED test_parity_is_validated_over_two_calls_not_one
FAILED test_the_correctness_reference_is_itself_distributed
FAILED test_inputs_are_seeded_per_rank
FAILED test_the_timed_region_is_kept_free_of_synchronisation
FAILED test_the_note_says_which_requirements_are_actually_enforced
FAILED test_a_correctness_slip_is_not_described_as_a_wasted_validation
6 failed, 17 passed
```

e.g. `test_distributed_contract_note.py:39` wants `"two calls back to back, never one at a time"`, and `:73` wants `"without ever scoring accuracy"`. Both properties are still true of the new note — the harness enforces the first and the note still warns about the second — they are just worded differently now. The assertions need to move to the new wording, or to `dist_harness` itself, which is where the guarantees now live.

**2. `test_dist_harness.py` and `test_task_preparer_harness_placement.py` fail collection on every shard: `ModuleNotFoundError: No module named 'torch'`.**

Both do `from kernelforge.loop import dist_harness`, and `dist_harness.py` imports `torch` / `torch.distributed` at module scope. The CI test image has no torch, so these two modules error at collection in all 8 `test (py3.1x shard n/4)` jobs and both `coverage` jobs — which is why every shard is red rather than only the one carrying the stale note tests.

Worth noting the production side already avoids this correctly: `_dist_harness_text()` reads the file as text rather than importing it, so nothing in `prepare_task` needs torch. Only the tests import it. A `pytest.importorskip("torch")` at the top of both files, or asserting against the source text the way `_placement_expression` already does for `task_preparer`, restores the invariant that the suite runs without torch.

**3. `test_measurement_fidelity.py::test_wide_suite_excludes_the_noisy_cases_from_scoring` — `AttributeError: module 'torch' has no attribute 'cuda'` at `sitecustomize.py:40`.**

py3.10/py3.11 shard 1 only. Line 40 of the generated probe is `orig = torch.cuda.CUDAGraph.replay` in `_GRAPH_PROBE_SITECUSTOMIZE`, which sits outside the `try:` that guards the `import torch`, so the stub torch `_load_driver_module()` installs (it has no `.cuda`) turns the probe's lazy install into an uncaught `AttributeError`. I could not pin down what puts that probe's import hook into the pytest process at head — it passes standalone, and I have torch installed locally so I cannot reproduce the torch-less ordering faithfully. Flagging it as part of the same red run rather than as a diagnosis. The merge base `c49cf3326` is green on all 8 shards, so it is attributable to this branch either way.

## What I checked and found clean

- **`dist_harness` stdout/flag contract against the consumers.** `--warmup` / `--iters` / `--bench-mode` match what `bench_wallclock` sends (`bench.py:310-318`), `--bench-case` matches `SWEEP_CASE_FLAG`, `--profile-run` matches `PROFILE_RUN_FLAG`. `wall_ms:` lines alone satisfy `bench.py:348` (the aggregate is derived; `median_ms:` is optional), and `case_ms:` is printed per case, so the `expected_case_ids` check in `_preflight_async` is satisfiable. `--repeat` and `--local-rank=` are absorbed by `parse_known_args` rather than failing the run.
- **The torchrun re-exec.** `--standalone` resolves to `rdzv_endpoint=localhost:0` in torch 2.11, so concurrent campaigns cannot collide on a port. `config_from_args` builds a real `python -u <driver.py>` subprocess (not `runpy`) unless `--run-path` is given, so `sys.path[0]` is the driver's directory and `import dist_harness` resolves to the copy `prepare_task` placed beside it. No `start_new_session`, so a timeout kill still reaps the ranks.
- **The launcher process does not poison the contract check.** The un-ranked parent (and a torchrun agent that inherited `RANK=0`) writes a shard with `local_rank: None`, which `_read_graph_probe_shards` routes to `unranked_replays` before `_distributed_contract_violation` sees it — so the new "every rank must report `harness: True`" rule does not reject a conforming run on its own launcher.
- **`dist_harness.py` placement and lifetime.** Written to `driver_access_dir` (beside the driver) on both the initial scaffold and `_reset_scaffold`, added to the protected/read-only set only when `nproc_per_node > 1`, and — unlike `graph_harness` — never dropped by the unused-harness sweep in `_finish_success`. Inside the KRC it lands in the `.forge_driver_*` stage, which carries `.gitignore *`, so it cannot reach the exported patch.
- **`carries_parallelism_suffix`.** `normalise_kernel_name` does not split digits off `tp8`, so `custom_all_reduce_tp8` / `EpDispatchCombineOp_ep8` pass and a bare `custom_all_reduce` with `world_size: 8` is now refused at parse time. Dropping the suffix from `looks_like_multi_rank_operator` is right — otherwise the collective test became vacuous once the suffix was mandatory, and `fused_moe_tp8` would have qualified.
- **The dispatcher `finally`.** Dropping `if task is not None` is safe: `task` is bound before the `try`, and the `parsed.task is None` branch has already returned.
- **Tests I ran at `70c71bb8d`** (`PYTHONPATH=src`): `test_dist_harness.py`, `test_distributed_measurement_contract.py`, `test_editable_repo.py`, `kernel_rewrite_controller/test_multi_rank_constraints.py`, `kernel_rewrite_controller/test_driver_staging.py`, `test_task_preparer_harness_placement.py` — 112 passed.

## Not blocking, but worth a line before merge

Your own commit message on `e0b9ea08f` says the multi-device path is unverified and needs a 2x/8x run. I agree, and I want to name the one thing static reading cannot settle: `_check_case` computes `reference` *after* both `call_candidate` calls on the same input objects. For an out-of-place candidate that is exactly right; for a candidate that reduces in place, the reference then reads already-reduced buffers and parity fails loudly. It is a false failure rather than a false pass, so it is not a correctness hazard, but it is the kind of thing a real 8-rank run will surface immediately and a single-GPU stub never will.

jiaqiang-dot-liu:
Merged `main` again at `f0dfd89f7`, and this one is worth a note rather than a line in the log: #1420 landed while this branch was in review and built the same in-place borrow.

Its own summary makes the call for us — "The verdict and the lock moved to `kernelforge.loop.editable_repo`. A second opinion on either is a defect: two verdicts would have one lane copy while another edits in place, and two lock implementations would serialize nothing." There were two. So I kept main's and dropped this branch's rather than reconciling them hunk by hunk:

- **`editable_repo`** — main's. It memoizes the scan the controller runs once a second and makes lock release idempotent for the three lanes that now take it, neither of which this branch had. This branch's naming pass and the docstring stating what the containment half decides are applied on top, and its tests now exercise main's implementation, clearing the new cache between cases.
- **`worktree`** — main's, which is a superset: a persisted campaign baseline, a restore scoped by inventory rather than by name, and an archive that keeps a campaign's bookkeeping when the repository is handed back. Only `stage_operator_driver` came across, with the driver stage named in the restore the way the experiments directory is — both carry their own `.gitignore`, so the untracked listing sees neither.
- **`dispatcher`** — main's last-chance recovery before release, which is the option you offered for the closing sweep and a better answer than the one this branch took. `_keep_preparation_audit` is kept beside it.

One thing the merge exposed that neither side had on its own. main archives the forge-loop bundle to the operator's directory on release, but `operator_workspace` answers `repo_root` for an in-place task — so the sweep and the archive were looking in different places and a host-killed run still recovered nothing. The sweep now reads the archive when the workspace holds nothing. That is item 2 actually closed, rather than made honest about being open.

**This changes root change 2, and not in my favour or yours exactly.** My earlier objection was about what replacing the `needs_inplace` fork would cost. That argument now applies to code that has just landed on `main` through #1420 and is carrying a persisted baseline, a scoped restore, a campaign-branch reclaim and an archive — so removing the fork is no longer a question about this PR at all. If you still want it gone, it is worth raising against `main` rather than here, where it would mean this branch deleting a mechanism it no longer owns.

jiaqiang-dot-liu:
All three fixed, at `c644881de`. Thank you for the cross-checks in the second half — the stdout/flag contract against `bench.py`, `--standalone` resolving to `rdzv_endpoint=localhost:0`, and the launcher shard routing to `unranked_replays` are all things I had reasoned about and not verified, and one of them (the launcher poisoning the new every-rank rule) would have been a silent false rejection.

**1 — stale contract-note assertions.** Already fixed; you read `70c71bb8d` and the merge at `f0dfd89f7` rewrote that file. You were right that the properties are still true and only reworded, and that is roughly where they went: the parity, seeding and barrier claims are now asserted against `dist_harness` in `test_dist_harness.py`, because that is where they became guarantees rather than instructions, and `test_distributed_contract_note.py` keeps only what the note still has to say — the handover, and that a correctness slip is not described as a wasted validation. The assertions also normalise whitespace now, since the note is wrapped and the claims are not.

**2 — collection without torch.** Fixed, and you identified the right asymmetry: production never imports the harness, `_dist_harness_text()` ships it as text. The tests now match. `test_dist_harness.py` takes `pytest.importorskip("torch", exc_type=ImportError)`; `test_task_preparer_harness_placement.py` no longer imports the module at all, and the entry-point test parses the shipped source instead of executing it. Verified by putting a `torch` that raises `ImportError` first on `PYTHONPATH`: the whole `src/kernelforge` suite collects, 4722 tests, and the harness file skips cleanly.

**3 — `AttributeError` at `sitecustomize.py:40`.** Your read of the mechanism was right even without a reproduction. `_install()` caught the `import torch` and then read `torch.cuda.CUDAGraph.replay` outside the guard, so a torch that exists but carries no `.cuda` raised out of an import hook that runs in every process on the path — which is why the damage showed up in a fidelity test that has nothing to do with this work. The whole install is inside the guard now, and `test_the_probe_survives_a_torch_that_carries_no_cuda` pins both the module-scope install and the hook against exactly that shape.

**The non-blocking one, which I think was the most valuable finding.** You are right, and it was worth more than a line before merge — `_check_case` handed `reference` the same input objects `call_candidate` had already been given. For an in-place collective the reference would have been computing from the candidate's own output, and parity would fail on an operator that did nothing wrong. It is a false failure rather than a false pass, but a false failure that costs a campaign is still a defect, and reducing in place is ordinary for this class of kernel.

The reference's inputs are now rebuilt from their seeds rather than reused. That is sound only because the seed makes the rebuild identical to what the candidate was passed, so there is a test for that too, alongside one that runs an explicitly in-place candidate and expects it to pass.

That one did not need 8 ranks to show, which is a fair hit — I had written the single-GPU stand-in to exercise capture, timing and scoring, and it was capable of catching this. The multi-device gaps I named in `e0b9ea08f` still stand and still need the 2×/8× run.

xiaofei-zheng (approved):
Re-reviewed at `c644881de`. No blocking issues — LGTM.

All three items from the previous round are fixed, and CI is green on the new head (all 8 `test (py3.1x shard n/4)` jobs plus both `coverage` jobs, against the red board at `70c71bb8d`).

What I checked this round:

- `task_preparer.py:160` `_install()` — the whole install is inside the `try`, so a `torch` module carrying no `cuda` returns `False` instead of raising out of the import hook. `test_the_probe_survives_a_torch_that_carries_no_cuda` pins the shape by exec-ing `_GRAPH_PROBE_SITECUSTOMIZE` against a stub module and asserting `_graph_ready is False`.
- I separately extracted `_GRAPH_PROBE_SITECUSTOMIZE` to a real `sitecustomize.py`, ran a genuine CUDA graph with 5 replays under `GRAPH_PROBE_OUT`, and the shard read exactly `{"replays": 5, "rank": "0", "local_rank": null, "world_size": "1"}` — the unconditional `_graph_ready = _install()` in the hook does not double-wrap `CUDAGraph.replay` through nested `torch.*` imports, so the `replays >= iters` gate is not weakened.
- Collectability without torch: `test_dist_harness.py` now gates on `pytest.importorskip("torch", exc_type=ImportError)` before importing the harness, and `test_task_preparer_harness_placement.py` dropped the `dist_harness` import entirely and compares `_dist_harness_text()` against the file on disk read as text — the same reason production ships it as text rather than importing it.
- `dist_harness._check_case` — `first_want`/`second_want` are now computed from `_seeded_inputs(case, context, seed)` and `_seeded_inputs(..., seed + _PARITY_SEED_STRIDE)` rather than the buffers the candidate was already called on, so an in-place collective is no longer scored against its own output. Same seeds, so the rebuild is bit-identical to what the candidate received; `test_the_two_parity_calls_cannot_receive_the_same_inputs` now asserts four builds with the expected equality pattern, and `test_an_in_place_candidate_is_not_failed_for_overwriting_its_input` covers the regression directly.
- `test_distributed_contract_note.py` — confirmed the rewrite is substantive, not gutted: the assertions track the new handover text (`from dist_harness import Case, run`, `world_size=8`, `do NOT launch ranks yourself`, `SLOWEST rank's time`, `the one requirement the harness cannot hold for you`), with whitespace normalized in the fixture.
- Re-walked the production paths that carry the harness: placement beside the driver in `driver_access_dir`, protection only when `nproc_per_node > 1`, rewrite on `_reset_scaffold`, and the `uses_harness` sweep in `_finish_success` still dropping only `graph_harness` — `dist_harness.py` stays inside the `.gitignore`-covered `.forge_driver_*` stage and cannot reach the exported patch.
- Rank routing: shards with `rank is None` or a null `local_rank` still land in `unranked_replays`, so the launcher/torchrun-agent process cannot trip the new "every rank reports `harness: True`" rule. Reproduced locally that this box presets `RANK=0`/`WORLD_SIZE=1` with no `LOCAL_RANK`, which is the shape the code comments describe.

Tests run locally in a clean worktree at `c644881de` (`PYTHONPATH=src pytest`): `test_dist_harness.py`, `test_task_preparer_harness_placement.py`, `test_distributed_contract_note.py`, `test_distributed_measurement_contract.py`, `test_measurement_fidelity.py`, `test_task_preparer_probe_contract.py` — 75 passed.

PR #1426, opened 2026-09-07
feat(forge): optimize compiler-emitted AMDGPU assembly

chennyiiis:
Adds `forge-loop --kernel-backend assembly` to optimize compiler-emitted AMDGPU instructions while retaining the original implementation as the performance baseline. Preparation captures compiler output, binds it through the existing launcher, and verifies the replacement path before the optimizer edits the selected `.s`. There is no LLM PORT phase or additional rewrite command.

Automatic capture supports one direct FlyDSL compile call, one direct Triton/Gluon JIT launch, or standalone HIP source built through the explicit `compile_hip` entry point. Each campaign selects one specialization. Triton/Gluon reuse the original signature, argument marshaller and current stream through an independent compiled kernel; warmup compiles without executing the source. HIP retains the caller's symbol, argument ABI and launch geometry. Existing `.s`/launcher pairs remain supported. Arbitrary linked libraries and Python extensions require an explicit extraction and launcher contract; resolve autotuning and library dispatch before capture.

A source backend's selected implementation can enter a separate ASM campaign with its own budget. Source campaigns retain their original language. The FlyDSL prompt recommends `--kernel-backend assembly` for a follow-on campaign rather than installing an assembly replacement inside the source loop. ASM is not automatically appended after every backend.

Preparation and acceptance:

- Validate the source and unchanged compiler roundtrip, canonical compilation/correctness commands, a deliberate assembler error, and a no-op execution control beyond warmup. Timeouts, crashes and missing metrics cannot establish execution. The probe handles LLVM entry-label comments and gfx950 kernarg-preload entry blocks that bypass the ordinary prologue.
- Run the same no-op through the canonical numerical suite. Every declared candidate case must expose the disabled implementation, while both source legs remain finite and inside the original source's oracle/repeat-error envelope. This rejects source/candidate callback wiring errors and cached source results masquerading as candidates. Equal errors in a correct roundtrip, including all-zero errors, remain valid: numerical equality alone does not imply shared execution.
- Require a protected `config.yaml` numerical contract for ASM KEEP. Fresh, complete source-before/candidate/source-after measurements must meet mathematical limits and source-relative oracle/repeat-error bounds. Repeated timing must beat the original implementation. No tolerances are relaxed by this change.
- Freeze frontend definitions, launcher, driver, reference helpers and numerical contract during instruction optimization. Resume requires schema-4 preparation evidence; older campaigns must prepare again to establish source independence.
- Publish winning patches with their binding, manifest and editable assembly. Without an accepted winner, select the original source and publish no assembly solution patch. Temporary Git index copies preserve their modification time so Git's racy-index detection does not omit same-size source edits from patches containing new assembly files.

The minimal FlyDSL and Triton vector-add examples cover an independent exact oracle, fresh outputs, input preservation, changed inputs, nondefault streams, graph replay and repeated numerical measurements. Assembly knowledge follows `INDEX.md`, `API_docs/` and `skills/{profile,bottleneck,optimize}`. It contains eight concise documents and one consolidated historical Kimi-K3 A16W4 case. Personal NFS paths and unavailable artifact-directory instructions have been removed from shipped knowledge. The six placeholder `Test <test@example.com>` commit authors were corrected without changing the source tree.

Validation, rebased on main `0af1c0969` (Forge sources, tests, examples and project configuration are unchanged by the final rebase):

- 619 Linux/Python 3.10 tests passed across assembly preparation/adapters, numerical contracts, canonical acceptance, KEEP/REVERT and publication, measurement guards and backend prompts; six tests/modules were skipped in the CPU environment, and two asyncio subprocess-cleanup warnings were emitted.
- Eight MI355X GPU tests passed: FlyDSL, Triton, Gluon and standalone HIP preparation, numerical execution controls, wrong-instruction rejection and clean patch replay; three deliberately miswired numerical drivers rejected during preparation; atomic-output execution checked for eager and graph replay. Versions: ROCm 7.2, Triton 3.7.0, FlyDSL 0.3.2. The image emits an unknown `asyncio_mode` configuration warning because pytest-asyncio is absent; these tests use `asyncio.run` explicitly.
- Four driver-wiring regressions were demonstrated failing on the pre-fix code. Additional tests reject missing numerical evidence and driver crashes. A deterministic same-timestamp regression reproduced the incomplete Git patch before preserving index metadata and passed afterward.
- Repository-wide Ruff lint/format checks passed. All 20 local knowledge links resolve. A clean-source wheel passed the packaging checker and contains both new adapters and exactly eight ASM documents, with no NFS paths in those documents.
- Test logs, tested-source hashes, generated assembly, preparation records, clean replay workspaces and the final wheel are preserved at `/shared_nfs/chenyi/forge-asm-adapters-20260916/review-fixes/`. GPU-tested preparation, numerical-acceptance and adapter modules are verified against the final source. The publication-index correction has separate CPU regression coverage.

These tests establish integration and acceptance behavior, not new model E2E gains. Historical Kimi-K3 serving validation measured 477.45 tok/s for pooled FlyDSL source versus 488.90 tok/s with repaired Stage1 plus Stage2 ASM (+2.40%). Stage1 required manual numerical repair; Stage2 passed a frozen, source-calibrated precision envelope while retaining approximately 9.7% higher mean repeat RMS. Fixed 512-question replicas did not reproduce persistent quality loss. The retained case documents these limits; this is not a new autonomous search, unchanged-repeatability claim or general model-quality guarantee. The old numerically failing +3.32% result remains rejected.


chennyiiis:
@jiaqiang-dot-liu This draft adds a working FlyDSL → compiler-generated AMDHSA assembly → reassembly → original-launcher path. The MI325X roundtrip, changed-instruction negative control, and nondefault-stream graph tests pass; representative-kernel performance is still open.

Could you point us to the intended FlyDSL kernel(s), preferably a repository path/branch or PR, along with the representative shapes/dtypes, target GPU, and existing correctness/benchmark entry point? That will let us compare the original FlyDSL kernel, unedited reassembly, and actual assembly optimizations under the same driver.

jiaqiang-dot-liu:
Reviewed the current head (`2a5c471`, based on `d086e425`). The core design — deterministic preparation, the three negative controls, and falling back to the original implementation when no candidate wins — is a real improvement over the previous revision, and the honesty of the retained case is appreciated.

Three blocking items below. Non-blocking observations omitted.

---

### 1. The `repeat_errors` criterion can be silently voided by a driver wiring mistake

This is the one place in the chain where a critical gate is hand-written and unverified.

The two families of checks in `judge_evidence` have very different strength ([`loop/numerical.py#L87-L97`](https://github.com/AMD-AGI/Hyperloom/blob/2a5c4712b4dcbb18b9d657d3a068f26f64f95b54/src/kernelforge/loop/numerical.py#L87-L97)):

```python
if max(item["oracle_errors"]) > limits["max_oracle_error"]:   # absolute bound, oracle only
...
for metric in ("oracle_errors", "repeat_errors"):
    source_error = max(max(measurements[role][metric]) for role in ("source_before", "source_after"))
    bound = max(source_error * limits["max_error_ratio"], limits["error_floor"])
    if max(measurements["candidate"][metric]) > bound:
```

`repeat_errors` has **no absolute limit at all**. Its only bound is the source-relative one. And the source leg is hand-written per task — the example has to compile a second, independent copy:

```python
source = flyc.compile(_vector_add, *args, N, torch.cuda.current_stream())
```

If that leg is ever wired to the candidate instead (copying `candidate_run` and renaming it is an easy mistake, and much less obvious in a real task than in vector-add), then `source_repeat ≈ candidate_repeat` and `max_error_ratio >= 1`, so `bound` is always at least the candidate's own value. The repeated-output stability check passes unconditionally, and everything still reports green.

The `s_endpgm` probe does not cover this. A no-op kernel makes both legs no-ops, both oracle errors blow past tolerance, `accept_candidate` fails — and the probe only requires failure, so it reports success. The SNR/allclose path has three negative controls; the numerical contract, which is what actually authorizes an ASM KEEP, has none.

This is not hypothetical for this workload. Repeated-output variability is exactly what the retained Kimi-K3 case identifies as the decisive signal (~27 dB between repeated outputs, ~9.7% higher mean repeat RMS).

**Suggested minimal fix** — no new probe needed, the data is already collected. Preparation stores both `source_numerical_evidence` and `roundtrip_numerical_evidence` in the record but never compares them. In the roundtrip run, if the candidate leg's `oracle_errors` are bit-identical to `source_before`, the driver is almost certainly measuring the same implementation twice; reject with a clear message. That closes the gap at the same place every other invariant is already enforced.

---

### 2. The prompt relaxation leaks into every backend, in the opposite direction from this PR

Two changes:

[`orchestrator/agent.py#L388`](https://github.com/AMD-AGI/Hyperloom/blob/2a5c4712b4dcbb18b9d657d3a068f26f64f95b54/src/kernelforge/orchestrator/agent.py#L388) removes a rule that applied to **all** backends:

```diff
-- Keep the kernel in its original backend/DSL (do not rewrite in another language).
+- Implementation language may change through the supported routes in the selected
+  backend expertise, unless the task explicitly restricts languages.
```

and `kernel_backends/flydsl/prompts.py` adds a "Moving between FlyDSL and assembly" section that explicitly instructs a FlyDSL agent to export compiler assembly, reassemble it, and establish parity itself when profiling points at spills, barriers, or wait placement.

The combined effect is that an agent in an **ordinary FlyDSL campaign** is now steered toward hand-rolling assembly with none of the guards this PR builds: no build-failure probe, no execution probe, no specialization fingerprint, and no mandatory `numerical_validation` (that is gated on `kernel_backend == "assembly"`).

That is precisely the failure mode the preparation stage exists to eliminate, re-opened for existing users in the same change that removes it for the new path. Compiler-generated spills are a common profiling outcome, so this is likely to trigger rather than theoretical.

**Request:** scope both changes to the assembly backend. The FlyDSL section should point at `--kernel-backend assembly` as a follow-on campaign rather than describe a do-it-yourself route.

---

### 3. Personal NFS paths ship inside the wheel

Both files live under `src/kernelforge/data/local_knowledge/`, which is package data — the PR description confirms the clean-source wheel contains these 8 ASM documents.

- `.../assembly_levers/instruction_optimization.md:69` — `/shared_nfs/<user>/forge-neha-repro-.../moe-forge-r3b`
- `.../assembly_levers/kimi_k3_moe_a16w4_gfx950.md:102` — `Evidence root: /shared_nfs/<user>/forge-kimi-k3-asm-12h-.../`

Every existing `/shared_nfs` reference on `main` is a shared path (models, caches); none is a personal directory and none is in `local_knowledge`. These are unreachable for anyone who installs the wheel, so labelling them "Evidence root" will send agents looking for paths that do not exist. Please drop the paths and keep the conclusions, or move the artifact pointers to an internal ticket.

Also worth cleaning up before merge: six commits are authored by `Test <test@example.com>`, which will land in `main`'s history with the wrong attribution.

jiaqiang-dot-liu (approved):
LGTM

PR #1430, opened 2026-09-07
Preserve environment values in the GEAK handoff

### LINKED ISSUE #1429: Preserve environment values across the GEAK handoff and return path
The GEAK handoff writer can corrupt a current-best environment value before GEAK parses it. It joins assignments with spaces without quoting each complete assignment.

For example, `extra_envs={"CUSTOM_CONFIG_PATH": "/models/shared config.json"}` is forwarded as `CUSTOM_CONFIG_PATH=/models/shared config.json`. The actual GEAK parser rejects the second word as an environment entry without `=`. Apostrophes can produce an unmatched quotation, and backslashes can be removed during shell tokenization.

The structured `baseline_env_spec.config.extra_envs` still contains the original value. The defect is in the redundant shell-string serialization of `accepted_env`, so GEAK can also report a conflict between two fields that were meant to describe the same environment.

A regression invoking the real Hyperloom handoff writer reproduces four failures on main `f02fd43659ebf13b563bbe4c079041d1417847fb`: spaces, JSON text, apostrophes, and backslashes. Empty and plain values are controls. Serialize the complete assignment tokens with `shlex.join` so shell parsing recovers their original values.

GEAK also independently canonicalizes some JSON environment strings during parsing. That consumer defect needs its own correction; this issue concerns Hyperloom's serialization. Neither environment defect has been attributed to a saved production run, so historical funnel impact is unquantified.

The return path has a separate, now-reproduced consumer defect. In Kimi CI session `20260824T090746Z-dcb5bfc6`, task `eb34e60877cd49f382ba207bf9ea779e`, prose in GEAK's `accepted_config.env` became `RUN_EVAL=true;` and `EXTRA_ENV=).` in the actual warmup YAML, with no retained evaluation output. GEAK has published validated `env_map` since commit `5f71eccc`, while retaining raw `env` for provenance. Hyperloom's consumer still ignored that map.

PR #1428 now uses the structured map when present, including empty maps, and retains legacy parsing only when the map is absent. The materiality gate, KernelPhase and dispatched GridVariant share that normalization. Invalid map types fail explicitly; existing environment-key filtering remains.

Native CPU replay through the actual producer, consumer, GridVariant and YAML writer yields `true;`, `true;`, and `true` for historical/current, current/current and current/corrected combinations. A retained local Magpie evaluation predicate enables evaluation only in the last case. Exact historical CI-generated script bytes were not recovered, so this is not claimed as a full historical executable reconstruction. Replay SHA256: `630332b19407431afcfcd8f8649886c6d17ad03e4c6742cf081c6b4326b8d6bc`.

Current PR head `ecf07f0b7bd8ffb2ae65bcafea1927e0275d1d8f` passes 20,681 local full-suite tests, 43 skips, 8 expected failures and 15 subtests, plus Ruff, format and REUSE. The return-map correction has 80 focused passing tests and an independent review. This establishes configuration recovery; it does not establish a new GPU gain or dashboard contribution.

iraj465:
The GEAK handoff's `accepted_env` string could split values containing spaces, misread apostrophes, or remove backslashes even though the parallel `extra_envs` mapping held the original value. Serialize complete `KEY=value` tokens with `shlex.join`; plain and empty values retain their existing meaning.

- Description: one serialization change plus a regression invoking the real handoff writer for spaces, JSON/equals, apostrophes, backslashes, empty values, and plain values.
- Linked issue: fixes #1429.
- Tests: the regression gives **4 failures and 2 passes on main**, and all six pass with the correction. Focused handoff/identity/fidelity/GPU-pin/gain tests: **126 passed**. Full suite with two workers: **20,654 passed, 43 skipped, 8 xfailed**, plus 15 passing subtests. Whole-repository Ruff check/format, REUSE, and whitespace checks passed. Advisory Mypy produced identical baseline and candidate diagnostics, with no new findings.
- Breaking changes: no; the handoff fields and schema are unchanged.
- Single concern: yes, lossless producer serialization.
- Upstream root cause: this defect is in Hyperloom. The independent GEAK JSON-environment normalization defect is handled separately.

The two-repository audit runs the actual handoff producer, resolver, argument mapper, and measurement-protocol mapper without launching the external runner. With the separate GEAK environment correction and #452, all six argv/state cases and all six environment cases pass, including composition with #437. No GPU performance or historical funnel recovery is claimed.

### DIFF
--- a/src/hyperloom/inference_optimizer/tests/test_geak_handoff_env.py
+++ b/src/hyperloom/inference_optimizer/tests/test_geak_handoff_env.py
@@ -1,0 +1,43 @@
+# SPDX-FileCopyrightText: 2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+"""The GEAK handoff's shell environment preserves current-best values."""
+
+from __future__ import annotations
+
+import json
+import shlex
+from pathlib import Path
+
+import pytest
+
+from hyperloom.orchestrator.loop.coordinator import Coordinator
+from hyperloom.orchestrator.state.shared_state import SharedState
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("value", ["two words", '{"key": "a b", "eq": "a=b"}', "user's config", r"a\b", "", "plain"])
+async def test_handoff_preserves_environment_values(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch, value: str
+) -> None:
+    expected = {"SGLANG_USE_AITER": "1", "CUSTOM_CONFIG": value}
+    coord = Coordinator.__new__(Coordinator)
+    coord.session_dir = tmp_path
+    coord.shared_state = SharedState(
+        model_path="/models/example",
+        current_best={"extra_envs": expected, "optimization_stack": []},
+    )
+    coord.phase_kernel._record_geak_kernel_journey = lambda _result: None
+    monkeypatch.setenv("FRAMEWORK", "sglang")
+
+    def stop_after_handoff(_name: str) -> Path:
+        raise RuntimeError("stop after handoff write")
+
+    monkeypatch.setattr(
+        "hyperloom.orchestrator.kernel.request_handlers._kernel_agent_tool_path",
+        stop_after_handoff,
+    )
+    await coord._run_geak_kernel_phase(from_phase="KERNEL")
+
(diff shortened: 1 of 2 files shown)


iraj465:
Superseded by #1428, which now includes these exact fixes and regression tests. Consolidated into one PR per repository for this delivery; the combined source passed fresh local checks. Please review #1428.

PR #1434, opened 2026-09-08
Allow KernelForge Claude fallback model via environment

weilei0120:
## Summary
- KernelForge Claude model fallback is configurable through `FORGE_CLAUDE_FALLBACK_MODEL` (default remains `claude-opus-4-8`; empty/`none`/`off` disables it).
- Agent model probe failures now emit warning logs, including fallback retry and fallback-probe failure.

## Test plan
- [ ] `pytest -n auto src/kernelforge/tests/test_provider_registry.py`
- [ ] Confirm unset env still falls back to `claude-opus-4-8`
- [ ] Confirm `FORGE_CLAUDE_FALLBACK_MODEL=none` does not retry another model on probe failure

Made with [Cursor](https://cursor.com)

PR #1437, opened 2026-09-08
Remove KernelForge same-provider model fallback

weilei0120:
## Summary
- KernelForge no longer retries a second Claude/Codex model after probe failure (no `claude-opus-4-8` / `gpt-5.5` ladder).
- Claude Agent SDK sessions no longer pass `fallback_model`; the requested id is the only model used.
- Provider fallback (`FORGE_AGENT_FALLBACK_PROVIDER`) is unchanged.

Supersedes #1434.

## Test plan
- [ ] `pytest -n auto src/kernelforge/tests/test_provider_registry.py src/kernelforge/tests/test_claude_resume.py`
- [ ] Confirm a failing primary-model probe raises instead of switching ids


xiaofei-zheng:
Reviewed at 45cd198 (merged up to 0f3c1b5). No blocking issues — good to merge from my side.

What I checked:

- `git grep` over `src/` finds **zero** remaining references to `fallback_model`, `model_fallback_reason`, `_prepare_with_model_fallback`, `FALLBACK_CLAUDE_MODEL` or `FALLBACK_CODEX_MODEL`. The follow-up commit that dropped the two now-unused dataclass fields (`AgentRuntimeConfig.fallback_model`, `AgentProvider.fallback_model`) and the stale test assertions closed out the last leftovers, so nothing dangles.
- Control flow in `create_registered_backend` is intact: the `try` block now returns `_prepare_backend(...)` directly, and the cross-provider fallback path (`FORGE_AGENT_FALLBACK_PROVIDER`) is untouched.
- The new `log.warning` reads `agent provider unavailable provider=... model=...`, which is accurate — the earlier `probe failed` wording would have misreported `preflight()` failures at the many call sites that never pass `probe_cwd`.
- Tests: 433 passed across the agent_backends / registry / claude / codex / orchestrator-agent selection.
- Docs need no update: the `claude-opus-4-8` mentions in `docs/install/slurm.md` and `docs/reference/authentication.md` describe the AMD gateway allowlist for user-selected models, not this automatic in-code degradation.

One behavioural note for the record, since the PR body only says "the requested id is the only model used": dropping `ClaudeAgentOptions.fallback_model` also removes the `claude --fallback-model` flag, which is the CLI's own per-request bypass when the primary model is overloaded or unavailable mid-session — a separate layer from the start-up probe ladder, though both fed off the same hardcoded constants so removing the constants necessarily took both. Confirmed offline that dropping both is intended.

xiaofei-zheng (approved):
Re-reviewed at aeda324. Still no blocking issues — approving.

The new commit is test-only (+51 lines in `test_provider_registry.py`, no source change) and the three cases it adds are meaningful rather than placeholders:

- `test_unavailable_provider_logs_and_does_not_probe` pins the corrected log wording (`agent provider unavailable`), which matters because the warning also covers `preflight()` failures at the many call sites that never pass `probe_cwd`.
- `test_fallback_provider_failure_raises_combined_error` covers the both-providers-down path and asserts both names surface in the combined error.
- `test_broken_entry_point_is_isolated` covers plugin-load isolation via the entry-point group.

Verified on a clean checkout of aeda324: `grep` over `src/` finds zero remaining references to `fallback_model`, `model_fallback_reason`, `_prepare_with_model_fallback`, `FALLBACK_CLAUDE_MODEL` or `FALLBACK_CODEX_MODEL`, and `test_provider_registry.py` + `test_claude_resume.py` pass (54). Leaving the rest to CI.

PR #1438, opened 2026-09-08
Honor --no-eval on Magpie explore rematerialization

weilei0120:
## Summary
- Propagate session `--no-eval` / `disable_run_eval` into explore Magpie YAML as `RUN_EVAL=false`.
- Keep warmup and decision rounds throughput-only when eval is opted out.

## Test plan
- [x] `pytest -n auto src/hyperloom/inference_optimizer/tests/test_explore_executor.py::test_explore_no_eval_disables_magpie_warmup_and_decision`
- [x] Existing warmup-keeps-eval and cold-decision-keeps-eval cases still pass


xiaofei-zheng (approved):

xiaofei-zheng:
Reviewed — no blocking issues. LGTM.

**What this does:** propagates the session-level `--no-eval` (and the `disable_run_eval` task param for internally queued explores) into the explore path, so Magpie gets `RUN_EVAL=false` both in the re-materialized base YAML and in every grid variant's `extra_envs` (warmup and decision rounds alike).

**Correctness check I ran:** the one path that could regress is the explore accuracy gate at `explore.py:1603` — it reads eval results whenever `scriptable or baseline_accuracy > 0` and fails closed (`accuracy_unavailable` -> REVERT) when there is no score. Forcing `RUN_EVAL=false` would break every variant if eval-off could coexist with a non-zero baseline accuracy. It can't:

- fresh session: with `--no-eval` the baseline itself runs `RUN_EVAL=false`, so `baseline_accuracy` stays 0;
- resume: `cli/__init__.py:2315-2325` explicitly refuses to enable `--no-eval` on a session that already anchored an accuracy;
- scriptable (xdit): the image-quality gate is driven by `HYPERLOOM_QUALITY_REF` / the `quality_gate` block in `benchmark_report.json`, not by `RUN_EVAL`, so it is unaffected by this change.

Also confirmed the forced `RUN_EVAL` does not leak into the promoted stack — `next_envs` is rebuilt from `stack_extra_envs + gv.extra_envs`, and the recorded variant metadata uses `gv.extra_envs`.

**Tests:** `test_explore_executor.py` 59 passed; the related `accuracy_gate` / `workload_env` / `no_eval` selection 239 passed.

PR #1439, opened 2026-09-08
refactor(kernelforge): remove three superseded subsystems (-1.4k production lines)

xiaofei-zheng:
## What this does

Removes three KernelForge subsystems that had been superseded in place rather than deleted, so the older generation kept shipping with nothing reading it. Net **-1.4k lines of production code**, plus the deprecated CLI alias, a de-duplicated MCP transport, and a test-helper cleanup.

**The one observable difference** is at the end of a `forge-loop` run: it no longer writes lesson markdown under the writable knowledge base's `learned/` directory, and no longer prints `Lessons learned: N`. Nothing in this repository ever read that directory, and the `Transfer rules discovered: N` line beside it could not fire at all, because it is derived from the tuning DB whose writes have been disabled. The docs claimed those lessons were injected into the next campaign's prompt; they were not, and the quickstart is corrected here to describe the mechanism that does carry state between runs (the workspace experience ledger and the knowledge-base experience store). Everything else in this PR is reference-dead code with no reachable call site.

`python -m kernelforge.cli` remains the single external CLI surface.

## Not in this PR: `gemm_tune/tier3/`

The audit that found the subsystems below also found the generated-tuner tier unreachable, and an earlier revision of this PR deleted it. **That was wrong and has been reverted** — `gemm_tune/` is now byte-identical to the merge base.

What the audit found is still true, and is a bug worth filing: `warrants_generated_tuner` gates the tier on a demanded table having no owner, while `dispatch.py`'s `SUPPORTED_TABLES` admits exactly one table, so the two predicates accept structurally disjoint sets and the gate cannot open through normal dispatch. On the one path where it does fire, `generate.py` has the model author a tuner and `runner.py` discards it unconditionally at the referee stage. But tier3 is meant to run, so that is a defect to fix rather than dead code to remove. Being handled separately.

## The three dead subsystems

**`learning/` (4 files, ~0.9k lines).** Everything it produced went through `tuning_db.py`, whose `_TUNING_DB_WRITE_ENABLED` has been `False` since `knowledge/experience_sink.py` took the job over; the output files it would have written have no reader anywhere in the tree. `IterationLoop`'s `evolver` constructor parameter and its two call sites (`on_benchmark`, `on_experiment_complete`) go with it, as does the now-unreferenced `resources.writable_knowledge_root()`.

**The tracker's scoring layer.** `Iteration.summary_row`, the `KernelScoringView` dataclass, and `best_iteration` through `summary_table` in `tracker/schema.py`, plus `get_best` / `summary` in `tracker/experiment.py`. This is the pre-`loop/scoring.py` scorer: production reads only `.iterations` and `.checkpoint` off the tracker (`cli.py`), `loop/scoring.py` owns `DEFAULT_SNR_THRESHOLD_DB`, and `loop/runner.py` carries its own `_is_gate_met` and tracks `best_mean_case_speedup` itself.

`list_experiments` is deliberately **kept**. It is an independent state-inspection helper — ten tests use it to read tracker state, not to exercise scoring — and deleting it would only force those tests to hand-roll glob-and-parse.

**The reachability island in `fusion/validate.py` (~248 lines).** A closed seven-function cluster (`unreached_fusion_symbols` and its six private helpers) whose only references were each other's definitions. `fused_symbol_invocation_evidence`, which `fusion/command.py` does call, is untouched.

## Also in this change

- **Drop the deprecated `kernel-agents` console script.** The rename to `kernelforge` shipped in v1.0.0b2 and the alias was kept for one release. Nothing in this repository, the docs, or the example scripts invokes it, and the orchestrator dispatches `python -m kernelforge.cli` directly. The `kernel_agents.agent_providers` entry-point group **stays**: that is how third-party provider plugins published before the rename are still discovered, and it is not a CLI surface.

- **Factor the shared MCP stdio transport** into `mcp_server/stdio_transport.py`. The newline-delimited JSON-RPC envelope and read loop existed as byte-identical copies in `pr_stdio_server.py` and `probe_stdio_server.py`; only one copy was tested, and `InvalidParamsError` was defined twice, so an `isinstance` check could not span the two servers. The probe server now inherits the ten transport tests the PR server already had.

- **De-duplicate test helpers that had drifted into three or five copies**: `_git` / `_GIT_IDENTITY` across five `kernel_rewrite_controller` test modules, the KB Store run-config builder across three modules, and the agent-provider registry isolation fixture across three. Remove one test that was a byte-identical copy of another (it even imported `_make_loop` from the module that owns the original).

## Verification

| Check | Result |
|---|---|
| Full suite | 4470 passed, 1 skipped, 1 deselected |
| `scripts/tests/test_forge_e2e_gate.py` | 11 passed |
| `ruff check .` / `ruff format --check .` | clean |
| Import every production module in `kernelforge` | 0 broken |
| The 18 `kernelforge.*` modules Hyperloom imports | all resolve |
| `python -m kernelforge.cli --help` + all 5 subcommands | respond |

The one deselected test is `test_controller_full_path.py::test_runtime_failure_is_not_mislabeled_as_handoff_validation`, which hangs on `main` as well: it does not monkeypatch the analysis agent, so `run_controller` spawns a real `claude` CLI subprocess. Unrelated to this change.

A full branch-coverage run was also used to look for dead branches inside surviving functions. It found 50 fully-uncovered functions (834 lines), and **every one has a live production caller** — they are GPU-, network-, or subprocess-gated paths, not residue. Nothing was removed on that basis.

## Diff

```
44 files changed, +300 / -3165
  production  +223 / -1578   net -1355
  tests        +42 / -1587   net -1545
  CHANGELOG      ~250
```


jiaqiang-dot-liu:
LGTM

jiaqiang-dot-liu (approved):

chennyiiis:
On the `gemm_tune/tier3/` part of this PR — the factual claim that it never ran was correct, and I've now fixed the reasons rather than removed the code: #1452 (gate reaches dense bf16, referee-verified) and #1458 stacked on it (fused MoE, 3 of 3 shapes improved on MI355X, best 1.1528x).

Two corrections to the description while we're here:

1. "`runner.py` discards it unconditionally at the referee stage — the result was never returned to a caller" isn't accurate. `runner.py:186-213` returns a full `Tier3Outcome` and writes `outcome.json`; the caller `cli.py:711` was the one dropping it on the floor.
2. That isn't an independent second reason — it's downstream of the first. The gate only opened for ownerless tables, `adapters_for()` therefore always returned `None`, and `runner.py:162-167` returned before the referee.

There's also a third closed set the description doesn't mention: `utils.py:58`'s `TUNER_ENV_VARS` and `phases/kernel.py:142`'s `_AITER_ENV_TO_TABLE`, which is why a verified win had nowhere to be deployed even if it got that far.

Could you drop the tier3 chunk from this PR? It has to come out as a unit — deleting `tier3/` while leaving `warrants_generated_tuner` and the `router.py` comment rename behind breaks `tier3/gate.py:96`. The other three subsystems (`learning/`, the tracker scoring layer, the `fusion/validate.py` island) are unaffected by my branches; no objection to those going.

PR #1441, opened 2026-09-08
fix(agentx): isolate aiperf in a compatible Python environment

zengleixin-amd:
## Description: what and why

AgentX dispatches #55042 and #55047 stopped at the first baseline preflight because the sandbox installer selected Python 3.10.12, while the pinned aiperf revision requires Python >=3.11,<3.14. The existing automatic repair correctly ran `install.sh --only-aiperf`, but retried the same incompatible interpreter.

This change supplies the client with its own compatible environment instead of upgrading the sandbox GPU stack or downgrading the pinned client.

### Scope

Four implementation files and three existing test files; no new service, feature flag, model-server configuration, or aiperf revision change.

- **Installer:** install aiperf into `${HYPERLOOM_STATE_DIR:-$HOME/.hyperloom}/aiperf-venv`. Reuse an available Python 3.11–3.13 interpreter, or provision private managed Python 3.11 through uv. If uv is absent, bootstrap pinned uv 0.12.3 into a private tools directory with `pip --target`.
- **Preflight and client:** consistently resolve explicit `AIPERF_BIN`, then the managed CLI, then PATH. Keep Python-path isolation local to managed client/probe subprocesses; do not activate the venv or change the primary interpreter, parent PATH, server environment, or GPU packages.
- **Runtime repair:** keep installation and post-repair lookup on the same state directory, including missing/empty HOME. A broken managed interpreter cannot be declared healthy by importing aiperf from the host Python.
- **Install safety and reuse:** protect unknown directories and symlinks, serialize shared-state installation, validate the recorded pin/spec against the actual environment, and reuse healthy installs. Preserve explicit overrides, dry/check modes, pre-warm versus required-install error handling, and network/index/mirror/offline environment policy.

## Linked issues

Follow-up to #1387. No separate issue is closed by this PR.

## Tests

Regressions were demonstrated failing before the corresponding fixes.

### Focused Linux regression suite

```bash
python -m pytest -o addopts= \
  src/hyperloom/inference_optimizer/tests/test_agentx_repair.py \
  src/hyperloom/inference_optimizer/tests/test_agentx_preflight.py \
  src/hyperloom/inference_optimizer/tests/test_aiperf_client_sh.py \
  src/hyperloom/inference_optimizer/tests/test_setup_cli.py \
  src/hyperloom/inference_optimizer/tests/test_agentx_runtime.py -q
```

**311 passed** in 375.16 seconds. The isolated test environment reported one `asyncio_mode` configuration warning because plugin autoload was disabled; these selected tests do not require async execution. Local and Linux-tested versions of all seven changed files were matched by SHA256 after LF normalization.

### Real isolated installation checks

- Original full installer with an actual Python 3.10.12 caller: reproduced the exact Requires-Python failure, exit 1.
- Real installer block with only Python 3.10 discoverable and uv absent from PATH: bootstrapped private uv, downloaded Python 3.11.15, and installed aiperf 0.12.0 from the unchanged commit `754356e9a39acc6cc6afb242d123bb57c3fb6f75`; exit 0. A second invocation reused the healthy installation. The primary Python version and package inventory remained unchanged.
- Full `--only-aiperf` entry point with a Python 3.10 caller and an existing compatible Python 3.12: installed into the dedicated venv and reused it on the second invocation.
- Actual runtime repair from Python 3.10, starting without aiperf: installer, managed-path re-resolution, capability/progress preflight, and second-call reuse passed. Parent and caller-supplied environments remained unchanged.
- Managed Python 3.11 preflight also passed with deliberately conflicting host `PYTHONPATH` entries.
- Real disposable-directory checks confirmed rejection of a tools-root symlink and refusal to bootstrap over the network under `UV_OFFLINE`.

### Static checks and limits

- Targeted Ruff lint/format, `bash -n` for both changed shell scripts, and `git diff --check` passed.
- ShellCheck passed for the complete client script and the changed installer block. Whole-installer ShellCheck remains blocked by a pre-existing malformed directive at baseline line 1081 (`# shellcheck disable=SC2086 -- ...`); it was not changed here.
- No GPU baseline, complete optimization run, or end-to-end run in the exact incident image was performed. The real checks above exercise installation, CLI capability checks, and runtime repair without starting a model server.

## Breaking changes

No public API or benchmark-scenario changes. Automatic installs now use a dedicated environment rather than the primary Python environment, and the managed CLI takes precedence over a legacy PATH entry. Explicit `AIPERF_BIN` retains precedence and remains operator-managed. A custom `HYPERLOOM_STATE_DIR` must be absolute.

When the necessary interpreter/packages are not already available and downloads are disallowed or unreachable, the installer reports the failure; it does not bypass offline policy or replace the system interpreter.

## PR addresses a single concern

Yes: provide and consistently use a compatible, isolated aiperf environment for AgentX. Directory protection, environment handling, and tests support that same installation boundary.

## Upstream root cause / ticket

Not an upstream component defect. The pinned client legitimately requires Python >=3.11; Hyperloom must provide a compatible client environment while retaining its supported Python 3.10 sandbox stack.

### DIFF
--- a/src/hyperloom/inference_optimizer/agentx/preflight.py
+++ b/src/hyperloom/inference_optimizer/agentx/preflight.py
@@ -34,11 +34,32 @@ class AgentXPreflightError(RuntimeError):
         self.repairable = repairable
 
 
+def _aiperf_state_dir(env: Mapping[str, str]) -> Path:
+    state = env.get("HYPERLOOM_STATE_DIR")
+    if not state:
+        home = env.get("HOME")
+        if not home:
+            if os.name == "posix":
+                import pwd
+
+                home = pwd.getpwuid(os.getuid()).pw_dir
+            else:
+                home = str(Path.home())
+        state = str(Path(home) / ".hyperloom")
+    path = Path(state)
+    if not path.is_absolute():
+        raise AgentXPreflightError(f"HYPERLOOM_STATE_DIR must be an absolute path: {state}")
+    return path
+
+
 def resolve_aiperf_bin(env: Mapping[str, str]) -> Optional[str]:
-    """Return ``AIPERF_BIN`` (operator override) else a PATH lookup else None."""
+    """Return the explicit override, managed CLI, or a PATH lookup, in that order."""
     override = (env.get("AIPERF_BIN") or "").strip()
     if override:
         return override
+    managed = _aiperf_state_dir(env) / "aiperf-venv" / "bin" / "aiperf"
+    if managed.is_file() and os.access(managed, os.X_OK):
+        return str(managed)
     # Resolve against the SAME PATH the benchmark subprocess will use (the child
     # env), not this process's os.environ, so preflight probes the binary that
     # actually runs. Falls back to os.environ PATH when env has none.
@@ -61,13 +82,29 @@ _ALLOWLIST_SNIPPET = (
 )
 
 
-def _default_probe(aiperf_bin: str) -> str:
+def _is_managed_aiperf(aiperf_bin: str, env: Mapping[str, str]) -> bool:
+    if (env.get("AIPERF_BIN") or "").strip():
+        return False
+    managed = _aiperf_state_dir(env) / "aiperf-venv" / "bin" / "aiperf"
+    return Path(aiperf_bin).resolve() == managed.resolve()
+
+
+def _probe_env(aiperf_bin: str, env: Mapping[str, str]) -> dict[str, str]:
+    child_env = dict(env)
+    if _is_managed_aiperf(aiperf_bin, env):
+        for key in ("PYTHONHOME", "PYTHONPATH", "PYTHONUSERBASE", "PYTHONPLATLIBDIR", "__PYVENV_LAUNCHER__"):
+            child_env.pop(key, None)
+    return child_env
+
+
+def _default_probe(aiperf_bin: str, *, env: Optional[Mapping[str, str]] = None) -> str:
     """Run ``aiperf profile --help`` and return combined stdout+stderr."""
     out = subprocess.run(
         [aiperf_bin, "profile", "--help"],
         capture_output=True,
         text=True,
         timeout=60,
+        env=_probe_env(aiperf_bin, os.environ if env is None else env),
     )
     return (out.stdout or "") + (out.stderr or "")
 
@@ -75,9 +112,8 @@ def _default_probe(aiperf_bin: str) -> str:
 def _interpreters_for(aiperf_bin: str) -> list[str]:
     """Interpreters that might have the probed aiperf importable.
 
-    install.sh pips aiperf into Hyperloom's own environment, so ``sys.executable``
-    is the usual hit; an ``AIPERF_BIN`` pointing at another venv is served by the
-    python sitting next to it.
+    The managed CLI uses the python next to it. ``sys.executable`` remains a
+    fallback for legacy installs and operator-provided wrappers.
     """
     import sys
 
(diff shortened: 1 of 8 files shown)


zengleixin-amd:
Pushed CI follow-up in `71fb54421` (test-only).

The Python 3.11 shard failed in the existing specialist-concurrency test: all four tasks completed, but a fixed 0.3-second probe lifetime recorded a peak of three when startup was staggered. Python 3.11 coverage then failed because it enforces shard success, not because coverage was below its threshold.

The fix synchronizes expected probe entrants with an Event before allowing each wave to finish. Existing concurrency assertions remain unchanged; missing-participant, serialized-execution and wave-reset checks prevent false passes. No production dispatcher or aiperf behavior changed.

Reproduction: the old staggered case failed with the same peak=3 assertion on both Python 3.10 and 3.11. After the fix, the complete test file passed all 11 cases on each version; Ruff lint/format and diff checks also passed. New CI is running; its final result is not yet confirmed.

xiaofei-zheng (approved):
**No blocking issues.** LGTM.

### What this PR does

The pinned `aiperf` build requires Python `>=3.11,<3.14`, but the GPU stack runs on 3.10.12, so installing it into the main environment is impossible. This PR installs it into a dedicated managed venv at `${HYPERLOOM_STATE_DIR:-$HOME/.hyperloom}/aiperf-venv` (created by `uv`, with a private `uv` bootstrapped via `pip install --target`), and makes every consumer agree on the same resolution order: `AIPERF_BIN` override -> managed CLI -> `PATH`. Python-path isolation (`PYTHONHOME/PYTHONPATH/PYTHONUSERBASE/PYTHONPLATLIBDIR/__PYVENV_LAUNCHER__`) is applied only when the managed binary is the one being executed. The silent flag-probe degrade is replaced by a hard, repairable `AgentXPreflightError`. The second commit is test-only (wave synchronization in `_ConcurrencyProbe`).

### What I verified

- Resolution order is consistent across `preflight.py` (`_aiperf_state_dir` / `resolve_aiperf_bin` / `_is_managed_aiperf` / `_probe_env` / `_default_loader_probe`), `repair.py` (freezes `HYPERLOOM_STATE_DIR` before supplying `HOME`, so preflight and installer cannot disagree), `aiperf_client.sh`, and `install.sh`. Relative `HYPERLOOM_STATE_DIR` is rejected in both the Python and the shell path.
- `AIPERF_PACKAGE_SPEC` (`install.sh:234`) is byte-identical to the string `_aiperf_healthy` (`install.sh:1220`) compares against, so the commit-pin check is not silently skipped for the default spec.
- New `flock` uses fd 8; the pre-existing checkout lock uses fd 9 — no collision. `--only-aiperf` gates (`install.sh:620`, `install.sh:755`) correctly skip the apt bootstrap and the `--break-system-packages` probe.
- No stale `$AIPERF` references remain in `aiperf_client.sh`; `bash -n` passes on both shell files; `ast.parse` clean on the Python files.
- Ran the real install block in a sandbox: `uv` bootstrap via `pip install --target` produces `bin/uv` and bypasses PEP 668; `uv` writes `direct_url.json` with `vcs_info.commit_id` matching the pin; the ownership marker survives `uv venv --allow-existing`; the unowned-directory guard correctly refuses to `rm -rf` a directory without the marker.
- Isolation is scoped to the `aiperf` exec only — `map_aiperf.py` and `aiperf_phase_gate.py` still run under the host Python (covered by `test_managed_client_sanitizes_only_aiperf_python_environment`).
- Audited every `_ConcurrencyProbe` usage against its capacity/task counts: waves are always full, no partial-wave deadlock.
- The branch is behind `main` but merges cleanly; the only overlapping file is `install.sh`, where main's `#1417` hunk (~line 797) is ~40 lines away and semantically independent.

### Tests

All unit shards on both py3.10 and py3.11, plus coverage/ruff/mypy/pylint/bandit/CodeQL/Gitleaks/reuse/byte-compile/docs, are green in CI. Locally: 222 + 77 passed on the branch, 94 + 137 passed on the merge tree, and `test_specialist_concurrent_dispatch.py` 11 passed across 6 consecutive runs (no flake from the new wave gating).

The red `ci-e2e/run` / `e2e` check is a runner queue timeout — all 220 polls report `phase=Queued node=- job=-` and it fails with "not terminal after 13200s". Unrelated to this change.

Issue #51, opened 2026-04-14
HF Token Leak


blakechi:
### How were you running Hyperloom?

Local

### Optimization domain

N/A

### Issue type

Other

### Which phase failed?

Other

### Short Summary

HF token leak

### Expected behavior

HF token should be set by the environment, not hard-coded.

### Actual behavior

HF token leaks at [here](https://github.com/AMD-AGI/Hyperloom/blob/main/inference_optimization/InferenceX/run_local_glm5_mi355x.sh#L13)

### Logs

```shell

```

### Data handling

- [x] I confirmed the attached logs/artifacts contain no sensitive information

chaojhou:
https://github.com/AMD-AGI/Hyperloom/pull/54  close this issue

PR #1442, opened 2026-09-08
Forge takes its model and effort from the runtime, and Hyperloom's env names

xiaofei-zheng:
## What

Three related things, all about where a Forge session's model and reasoning effort come from.

**1. Port [KernelForge#51](https://github.com/AMD-BRAIN-Internal/KernelForge/pull/51).** The runtime now settles model and reasoning effort; a call site that passes `reasoning_effort=` is ignored rather than obeyed. `AgentRunSpec.resolved()` takes the runtime's value when it names one and falls back to the spec's when it does not. The startup probe no longer pins `low`, and the Codex default moves to `gpt-5.6-sol` — the id the gateway actually serves — with `gpt-5.5` as the fallback rung.

**2. Stop Forge spelling the Hyperloom env vars its own way.** `FORGE_CLAUDE_MODEL` / `FORGE_CODEX_MODEL` named exactly what Hyperloom already names with `CLAUDE_MODEL` / `CODEX_MODEL`; `FORGE_AGENT_MODEL` was a third spelling of the same decision. Now that Forge ships inside Hyperloom, two spellings for one setting means an operator sets both and gets whichever they did not expect. Deleted outright, no deprecation window. The tests that covered them are kept and inverted: they now assert the old names are ignored, so re-adding one fails the suite.

**3. One effort vocabulary, one ceiling.** New `hyperloom/common/reasoning_effort.py` — a name list with no transport — is now the single definition both components import: `low` / `medium` / `high` / `xhigh` / `max`, default `high`. `max` is Claude's deepest; the OpenAI-compatible gateway rejects the name, so Codex sends it as `xhigh` rather than erroring. #51 deleted 13 call-site effort literals. Twelve were opinions and deserved to go. One — the plan-critic width repair — was a statement of fact: no tools, two turns, restating a decision already made. It has no business burning the campaign's `high`. New `max_reasoning_effort` field, which can only ever *lower* (an operator running at `low` still gets `low`), used at exactly one call site. An AST test asserts that call site is the only one, so a second has to be argued for by editing the test.

## Not ported: the context window

#51 added a `CLAUDE_CONTEXT_WINDOW` env that appends a window suffix to the model id (`claude-opus-5[1m]`). Measured against the gateway this PR targets, `claude-opus-5[1m]` answers `400 Invalid model name`, as does `claude-opus-5[200k]`; `/v1/models` publishes 23 ids and none of them is windowed. Forge also has no compaction and no token budget keyed off a window, so nothing downstream would read the number.

Hyperloom's `MODEL_CONTEXT_WINDOWS` is not the same setting and is not affected: it maps a model id to an int used purely locally (`coordinator.py` → `orchestration_memory.py`) to decide when to compact the orchestrator's own conversation, and never reaches the wire. A test pins the removal — no window suffix can reach a model id.

## Breaking change — please tell your deployments

These are gone. A box that sets only the old name now falls back to the provider default, visible at preflight (the catalog check prints the model it is about to run) rather than silently.

| Removed | Use instead |
| --- | --- |
| `FORGE_CLAUDE_MODEL` | `CLAUDE_MODEL` |
| `FORGE_CODEX_MODEL` | `CODEX_MODEL` |
| `FORGE_AGENT_MODEL` | `CLAUDE_MODEL` / `CODEX_MODEL` |
| `CLAUDE_CONTEXT_WINDOW` | nothing — never worked against this gateway |

`HYPERLOOM_REASONING_EFFORT=minimal` (and `=none`) also stops working: it is not a level any more, so Hyperloom's own chat.completions runs at the gateway default instead — deeper and more expensive, with no error. A box sitting on `minimal` has to move to `low` by hand.

`FORGE_AGENT_REASONING_EFFORT` and `HYPERLOOM_REASONING_EFFORT` are unchanged, and now accept the same five level names on both sides.

## Tests

Local, on `152abb39c`:

- `pytest src/kernelforge/tests src/hyperloom/common/tests --ignore=src/kernelforge/tests/fusion` — **3405 passed**
- `ruff format` / `ruff check src/` clean.

Gateway behaviour above was measured directly against the deployment, not inferred: `claude-opus-5` → 200, `claude-opus-5[1m]` → 400, `claude-opus-5[200k]` → 400, `gpt-5.6` → 400 (deployment not found), `gpt-5.6-sol` → 200, `gpt-5.5` → 200.


jiaqiang-dot-liu:
## Review: blocking items before merge

Verified against the PR head (`a04c2b9f3`), full checkout. The direction is right and the
measured-against-the-gateway evidence is convincing — the items below are the ones I think
have to be resolved first.

### 1. The convergence is incomplete: two call sites bypass the switch

There are 4 non-test constructions of a runtime in the tree (`resolve_agent_runtime(` /
`AgentRuntimeConfig(`). Two of them do not go through `config.agent_runtime()`:

- `src/kernelforge/fusion/command.py:179` — `_create_agent_backend()` passes model, timeout
  and sandbox but **not** `reasoning_effort`, so it is pinned to `DEFAULT_REASONING_EFFORT`
  forever. This PR removed the literal `reasoning_effort="high"` from that call, which makes
  it read as if the runtime now settles it; it does not. `FORGE_AGENT_REASONING_EFFORT` and
  `HYPERLOOM_REASONING_EFFORT` have no effect on any forge-fusion session.
- `src/kernelforge/gemm_tune/tier3/generate.py:103` — builds its own runtime via
  `select_default_agent_provider` + `resolve_agent_runtime`, passing no effort; and
  `gemm_tune/tier3/runner.py:100` calls `generate_tuner(mandate, work_dir, retry_note=...)`
  with no model. So this session reads **neither** switch: the model is always the provider
  default and the effort is always `high`, regardless of `CLAUDE_MODEL` / `CODEX_MODEL` /
  either effort variable.

Net: effort converges at 2 of 4 sites, model at 3 of 4.

### 2. The AST guard asserts the wrong property, which is why (1) went unnoticed

`tests/test_agent_model_and_effort.py` asserts that no call site writes a
`reasoning_effort=` literal. The property that actually protects convergence is that no call
site builds a runtime outside `config.agent_runtime()` — deleting the literal satisfies the
current guard while leaving the bypass in place, which is exactly what happened in
`fusion/command.py`.

Suggested replacement: assert that `resolve_agent_runtime(` / `AgentRuntimeConfig(` appear
only in an allowlist of `{config.py, agent_backends/registry.py, agent_backends/*}`. That
one test would have caught both bypasses in (1) and the defect in (3), and would keep
catching them.

### 3. `auto` backend can hand a Claude model id to Codex (regression introduced here)

`Config.from_env` resolves the model *before* the provider is settled:
`resolve_agent_model("auto")` reads `CLAUDE_MODEL` (`src/kernelforge/config.py:66`), and only
afterwards does `agent_runtime()` call `select_default_agent_provider(self.agent_model)`
(`src/kernelforge/config.py:245`). Ownership is a preference, not a constraint: when the
Claude SDK is not importable (`_claude_available()`), the selection falls to codex and we end
up at `resolve_agent_runtime("codex", model="claude-opus-5")`.

Before this PR `agent_model` came from `FORGE_AGENT_MODEL` / `KERNEL_AGENTS_MODEL`, which
nothing set in practice, so the value was `""` and the provider default applied. Both env
templates set `CLAUDE_MODEL` unconditionally, so the triggering configuration is not exotic.

The cross-provider fallback in `create_registered_backend` does reset
`model=fallback_registration.default_model`, so the "Claude installed but probe fails" path
is safe; the unsafe path is the one where auto selects codex directly.

Resolving the model after the provider is settled fixes it:

```python
provider = self.agent_backend
if provider == "auto":
    provider = select_default_agent_provider(self.agent_model).name
model = self.agent_model or resolve_agent_model(provider)
```

(with `from_env` storing `""` for `auto`).

### 4. `minimal` becomes a silent no-op and is not in the breaking table

`apply_reasoning_effort` used to accept `minimal | low | medium | high`; `minimal` and `none`
are now dropped without a diagnostic, because that function has to stay a no-op for
non-reasoning models. A deployment currently running `HYPERLOOM_REASONING_EFFORT=minimal`
gets silently promoted to the gateway default — the expensive direction, with no signal. That
contradicts the PR's stated position that the removals are loud at preflight. Please list it
in the CHANGELOG breaking table alongside the removed variables.

### 5. One documentation surface still names a removed variable

`examples/hyperloom-qwen3-14b-fp8-12h-forge/SKILL.md:32` still says "`FORGE_CLAUDE_MODEL`
falls back to `CLAUDE_MODEL`". It is the only remaining non-test occurrence in the tree.

xiaofei-zheng:
Thanks — I verified all five against the tree rather than taking them as read. Four hold and are fixed in `3a08bc17a`; one is half right. Details below.

### 1. Two call sites bypass the switch — confirmed, fixed

Both real, and the split you describe is exact.

`fusion/command.py:179` — this branch deleted the literal `reasoning_effort="high"` (visible in `git diff d5b62a3fc..HEAD -- src/kernelforge/fusion/command.py`), which left the call reading as if it deferred while `resolve_agent_runtime` just re-applied `DEFAULT_REASONING_EFFORT`. Net behaviour was unchanged (`high` before, `high` after), so not a regression, but it did mean this PR claimed a convergence it had not made. Now passes `reasoning_effort=resolve_agent_reasoning_effort()`.

`gemm_tune/tier3/generate.py:103` — confirmed, and `git diff d5b62a3fc..HEAD -- src/kernelforge/gemm_tune/` is empty, so this one predates the branch. Fixed anyway, since leaving it is the same defect: it now takes `model or resolve_agent_model(chosen.name)` and the effort ladder.

### 2. The AST guard asserts the wrong property — confirmed, fixed

Agreed, and this is the more useful of the two findings: deleting the literal satisfies the existing guard while leaving the bypass, which is exactly the failure mode in (1). Added `test_every_runtime_is_built_where_the_switches_are_read`, which closes the set of modules constructing a runtime (`config.py`, `fusion/command.py`, `gemm_tune/tier3/generate.py`, `orchestrator/agent.py`, `orchestrator/supervisor.py`). A sixth has to be added to the allowlist, which is where the "does it read both switches?" check happens.

### 3. `auto` can hand a Claude model id to Codex — confirmed regression, fixed

Traced it exactly as you describe: `from_env` stores `resolve_agent_model("auto")` → `CLAUDE_MODEL` (`config.py:279`), and only `agent_runtime()` settles the provider (`config.py:246`). At merge-base `agent_model` came from `FORGE_AGENT_MODEL` / `KERNEL_AGENTS_MODEL`, neither of which anything sets, so it was `""` and the provider default applied — the regression is this branch's.

I had previously written this off because every Hyperloom-launched path pins the backend (`forge_submit.py:2568/2761/3497`, `forge_collective.py:188/190`, `forge_fusion.py:120`), but `Config.agent_backend` defaults to `auto` and the forge CLI is reachable directly, so the path is real. Your ordering fix is what I took: `resolve_agent_model("auto")` now returns `""` and `agent_runtime()` reads the pair after `select_default_agent_provider` has run. Two existing tests asserted the old `auto` → `CLAUDE_MODEL` behaviour and were inverted; added `test_auto_never_hands_a_claude_model_id_to_codex`.

One correction on the supporting detail: `.env.template:71` has `CLAUDE_MODEL` commented out, not set unconditionally. Doesn't change the conclusion — a box that sets it and lacks the Claude CLI is an ordinary configuration — just noting the premise.

### 4. `minimal` is a silent no-op — confirmed, documented

Confirmed against merge-base: `apply_reasoning_effort` accepted `{minimal, low, medium, high}` and forwarded `minimal` verbatim; it now falls out of `gateway_reasoning_effort` and the call runs at the gateway default. You are right that it is the expensive direction with no signal, and right that it contradicts the "loud at preflight" claim — that claim only holds for Forge's side, which refuses to start. Hyperloom's side cannot be loud without breaking non-reasoning models, so I documented it rather than changed it: it is now a **BREAKING** entry in the changelog saying a deployment on `minimal` has to move to `low` by hand. Re-adding `minimal` as a level is not an option — the Claude CLI does not know it and there is no level below `low` to project it onto.

### 5. `SKILL.md:32` — confirmed, fixed

Correct, and it was the only remaining non-test occurrence (`config.py:58-59` names the removed variables in prose explaining the removal). Rewritten to name `CLAUDE_MODEL` / `CODEX_MODEL`.

### Verification

`pytest src/kernelforge/tests src/hyperloom/common/tests --ignore=fusion` — 3403 passed. `ruff check` / `ruff format --check` clean. The 5 failures in `tests/fusion/test_harness_noise.py` when the fusion directory runs as one process reproduce identically at merge-base `d5b62a3fc` (same 5, `json.decoder.JSONDecodeError`), so they are pre-existing test-ordering pollution, unrelated to this PR.

jiaqiang-dot-liu:
Re-reviewed at `3a08bc1`. All five items from the previous round are resolved — `auto` now
resolves the model after the provider, both bypass sites read the switches, the closed-set
test is in, and the `minimal` breaking note and the SKILL.md line are fixed.

One blocking item left, found while checking the new invariant.

### The backend-switch paths do not read the switches, in opposite directions

Both files are on the allowlist of the new
`test_every_runtime_is_built_where_the_switches_are_read`, so the suite is green, but the
property that test asserts — every module in the set reads the env pair — does not hold for
these two branches.

**a) `src/kernelforge/orchestrator/supervisor.py:211` — no model is passed, so `CODEX_MODEL`
is ignored**

```python
runtime = resolve_agent_runtime(
    backend,
    executable="",
    timeout_sec=config.agent_timeout_sec,
    ...
)   # no model= -> registration.default_model
```

This path is live and ordinary: `--supervisor-backend` defaults to `codex` on the rewrite
command (`cli.py:2198`) and is forwarded into forge-loop (`fusion/campaign.py:232`). With a
Claude implementer, `sup_backend` at `cli.py:1697` differs from `runtime.provider`, the switch
branch fires, and the supervisor always runs `gpt-5.6-sol` regardless of what `CODEX_MODEL`
says. Per the Codex-default commit message, the development box sets `CODEX_MODEL=gpt-5.5` —
on that box the supervisor is not running 5.5.

**b) `src/kernelforge/orchestrator/agent.py:118` — the *previous* provider's model id is
carried across the switch**

```python
runtime = resolve_agent_runtime(
    agent_backend,              # the new provider
    model=config.agent_model,   # the old provider's id
    ...
)
```

With `FORGE_AGENT_BACKEND=claude`, `config.agent_model` is `CLAUDE_MODEL`, so switching to
codex sends `claude-opus-5` to the OpenAI-protocol gateway — the same defect just fixed in
`Config.agent_runtime`, relocated. The only in-tree caller (`cli.py:1619`) passes
`selected_runtime.provider`, which equals `runtime.provider`, so the branch is currently
unreachable; the code is live, though, and a caller that genuinely switches would hit it.

Same fix in both places — read the pair for the provider that is actually going to run:

```python
model=resolve_agent_model(agent_backend)   # agent.py
model=resolve_agent_model(backend)         # supervisor.py
```

With that, all five runtime construction sites read the switches on every branch, not just on
the path where the provider does not change.

xiaofei-zheng:
Both confirmed and fixed in `152abb39c`.

**(a) `supervisor.py:211`** — verified reachable exactly as described: `--supervisor-backend` defaults to `codex` on `forge-rewrite` (`cli.py:2198`), is forwarded by `fusion/campaign.py:232`, and `cli.py:1697` sets `sup_backend = supervisor_backend or effective_implementer`, so a Claude implementer takes the switch branch on an ordinary run and the supervisor gets `registration.default_model`. Your reading of the dev box is right too — with `CODEX_MODEL=gpt-5.5` set, the supervisor was running `gpt-5.6-sol`. Now passes `model=resolve_agent_model(backend)`.

For the record on severity: at merge-base this branch also produced the registry default (`config.agent_model` came from `FORGE_AGENT_MODEL`, which nothing sets), so it is not a regression this PR introduced — but it is exactly the invariant this PR claims, so it should not ship unfixed. Fixed rather than argued.

**(b) `agent.py:118`** — confirmed, and confirmed unreachable today for the reason you give: the only caller passes `selected_runtime.provider` (`cli.py:1619`), which the branch then compares equal. Fixed anyway, since it is live code and the defect is the one just fixed in `Config.agent_runtime`, one level up. Now `model=resolve_agent_model(agent_backend)`.

**On the guard.** You are right that the closed-set test could not see either of these — it says *where* runtimes are built, not *what model they name*. Added `test_every_runtime_names_a_model_it_resolved_for_that_provider`, which walks every `resolve_agent_runtime(` call and flags two shapes: no `model=` at all, and a `model=` naming `config.agent_model` (the already-resolved value, carried across a switch). I checked it actually bites by reverting each fix in turn — both revert to a failing suite. `config.py` is exempt with a comment, since it is the one place that legitimately holds the resolved value on the dataclass and re-reads the pair once the provider is settled.

**Verification:** `pytest src/kernelforge/tests src/hyperloom/common/tests --ignore=fusion` — 3405 passed. `ruff check` / `ruff format --check` clean. Changelog updated with both branches.

jiaqiang-dot-liu (approved):
LGTM

PR #1443, opened 2026-09-08
Forge token efficiency: attribute the spend, then cut the context

xiaofei-zheng:
## What

Token-efficiency work for Forge, stacked on #1442 (which this needs for the `AgentRunSpec` field it appends).

Motivation: a price-weighted split over 316 real Forge e2e runs (`/shared_nfs/chenyi/kernel-arena-daily-ci/reports/`, $81,922 total, $259.25/run) shows where the money goes — cache_read 39.5%, cache_write 32.9%, output 25.9%, input 1.6%. About 72% of the bill scales with how large each session's context is, not with how many sessions run. So the lever is prompt size, and the prerequisite for pulling it is knowing which role is paying.

**1. Attribute spend to the role that incurred it.** A campaign's cost was one number. `AgentRunSpec` gains a `role` string (accounting only — no provider reads it), every call site names itself, and the tracker breaks the total down by role. Empty folds into `unattributed`, so a caller that names nothing still has its tokens counted. This is what made the rest of the PR measurable rather than guesswork: it showed the implementer role alone carries 75% of a campaign's cache_read (16.96M of 22.58M over 6 calls), which is why every change below targets the prompt that role carries.

**2. Point at the knowledge maps instead of inlining them.** This is the largest change in the PR. `to_prompt_dict` inlined the whole knowledge index into every specialist, lane-synthesis and analysis payload, with another copy in the system prompt — so the maps were paid for repeatedly, in the cached prefix that is re-read on every turn of every session.

Under the new default an agent gets a pointer to each pillar's `INDEX.md` and reads what it needs.

Measured as the first-turn prefix of an analysis session. Deferral and the tool set of change 3 both land in the same prefix, so the campaigns were run as a 2×2 rather than reported as one pair — the headline 51,275 → 8,487 (−83.4%) is what the two changes do **together**, and neither one owns it:

| analysis prefix | default tool set | named tool set |
|---|---|---|
| maps inlined | 51,275 | (not measured) |
| maps deferred | 35,555 | 8,487 |

Deferring the maps is the −15,720 (−30.7%) down the first column; the tool set is the −27,068 (−76.2%) across the second row. The ordering is not commutative, which is why the table has a hole: while the maps were still inlined they dominated the prefix, so the tool set measured against that baseline would report a smaller number than the −27,068 it is actually worth. Implementer lanes split the same way, falling from a mean 62,022 (n=18) to 17,704 (n=4) with both changes on, ranges disjoint.

(These per-session prefix figures were read from the campaigns' Claude session transcripts, which have since been deleted; the campaign logs and their `__FORGE_RESULT__` records remain, so the speedups below re-derive but the prefix sizes do not.)

Deferring only helps if agents still consult the knowledge base once it is a pointer, and that is a behaviour question, so it was decided by a four-a-side A/B on the same kernel:

| | speedup | improved |
|---|---|---|
| inlined | 1.0800 / 1.0481 / 1.1447 / 1.0000 | 3 of 4 |
| deferred | 1.2036 / 1.1562 / 1.1068 / 1.1193 | 4 of 4 |

The deferred arm's worst run (1.1068) beats the inlined arm's mean (1.0682). Agents do follow the pointer: both `INDEX.md` reads in the deferred arm were each followed by a card read, where the inlined arm read `INDEX.md` zero times across 98 sessions.

The tool-set A/B (change 3, four campaigns) shows the same absence of a regression from its side: 1.0708 / 1.0527 naming the tools against 1.0552 / 1.0546 with the default set.

Two caveats stated plainly. This is one kernel at n=4 a side — hence `KERNELFORGE_DEFER_KNOWLEDGE_MAPS=0` still inlines the maps. And the saving does **not** show up as a campaign-level cost difference: per-call cost and cached reads land inside the inlined arm's spread either way, because campaign totals vary by more than the effect. The hard evidence is the per-session prefix above. Change 5 is why the summary could not have shown it in any case.

A related, much smaller change drops each map's own YAML front matter from the prompt — bookkeeping the model never uses, repeated per map.

**3. Name the built-in tool set a session actually uses.** `allowed_tools` was passed as `--allowedTools`, which is only a permission list: the CLI still loads and describes every built-in tool, and those schemas sit in the cached prefix. Naming the set a role actually uses (`--tools`) removes the rest. Measured by a four-campaign A/B whose only difference is this flag, both arms deferring the knowledge maps: an analysis session's first-turn prefix is 35,555 / 35,507 tokens with the default set and 8,487 / 8,439 naming the six an implementer uses — **−27,068, or −76.2%**, with the two replicates in each arm within 50 tokens of each other. The implementer lanes in the same runs split the same way, 44.9–47.1k against 17.4–19.6k. Note the ordering matters: while the maps were still inlined they dominated the prefix and hid most of this, so the two changes have to be read together rather than added up from separate baselines. MCP tools are not part of the built-in set and are still carried by `allowed_tools`.

**4. Remove rtk.** `rtk` is an external CLI that trims tool output before it enters the context. The tree carried a wrapper (`rtk.py`), a pinned-version downloader with per-platform sha256 (`rtk_install.py`), a `kernelforge install-rtk` command, an `ensure_forge_rtk` step in `install.sh`, two build-command call sites, docs, and — the part that was actually being paid for — a 350-token (cl100k) paragraph in the system prompt of every session of every campaign telling the model to prefix its shell commands with `rtk`.

It did not work, for a structural reason. `smart_wrap` only ever wrapped the *build* command, which fires on runs that compile: 41 of 316 real runs, 5.5% of spend. Everything else was the prompt paragraph asking the agent to opt in by hand. Agents essentially never did: across the six e2e campaigns below — 93 sessions, ~6 GPU-hours, with rtk 0.48 on `PATH` in every arm — `rtk` was invoked **once**. Five of the six campaigns used it zero times. So the ledger was 350 tokens of prompt per session, resident in the cached prefix that every turn re-reads, against one invocation in six campaigns.

Making it work would mean a different mechanism: a `PreToolUse` hook that rewrites `Bash` tool input to insert the wrapper, so the filtering happens whether or not the model cooperates (the hook plumbing exists — `AgentHooks` in `agent_backends/base.py`, and `opportunity_agent.py` already rewrites tool input this way). That is a real design with real risk — it changes what every shell command in the loop actually executes, and adds a mandatory external binary to the critical path. It is not this PR's subject, and shipping the advertisement without the mechanism was worse than shipping neither. 12 files, −929 lines.

One fix from that work is kept because it is independently correct: `_build_failure_tail` now reads stdout as well as stderr. ninja prints compiler output on stdout, so a build failure used to reach the agent as `BUILD FAILED:` and nothing else.

**5. Report the two token columns the bill is actually in.** A campaign's run summary printed `in` / `out` tokens and a total. Output is 400–780k tokens a campaign against 5–90k of uncached input, so that line is dominated by `out` and moves with how much the agents wrote — while `cache_read` (39.5% of the bill) and `cache_creation` (32.9%) went untracked in the summary even though the accumulator already carried them. Every change in this PR moves exactly those two columns, which is why the campaign whose prefix fell 83% printed the same `in`/`out` line as the one whose prefix did not. The summary now prints four columns for the total and for each role.

**6. Stop paying for prompt text that corrects the prompt beneath it.** Same shape as change 2, applied to the kernel-backend prompts. Four of them (triton, hip, ck, aiter) told the session to build with a `build` tool, time with a `bench` tool and profile with a `pmc` tool; none exist — this loop hands a session Bash and the measurement driver — so `make_agent_fn` carried a sentence naming all five and explaining that they meant shell commands, in the system prompt of every session of every campaign. The prompts name the mechanism now, the correcting sentence is gone, and a contract test refuses a prompt that invents a tool again. Separately, Triton's Gluon-escalation section restated its two knowledge cards inline for an escalation most sessions never reach; the trigger, the ownership claim and the route stay resident and the mechanics defer to the cards (grepped: every deferred claim is in them). Per session, cl100k on the rendered prompt: triton −269 tokens, ck −71, hip −67, aiter −58, every other backend −49.

**7. Merge notes.** Rebased onto the current #1442 (which since dropped the context-window rung entirely) and merged `main` three times, most recently at #1475. Three of main's changes met this branch head-on and are worth naming:

- #1437 removed KernelForge's same-provider model fallback on the same lines this branch was rewriting; the fallback goes, and nothing in the tree still reads `fallback_model`.
- #1410 gave the round partition its own `reasoning_effort`, through a `_run` parameter #1442 had already removed — under the new ranking the runtime's effort wins, so a call-site value is ignored rather than honoured. Dropping the parameter would have silently discarded #1410's decision that dividing ground the analyses already name is a reading task, so it is re-expressed as `max_reasoning_effort`, a ceiling that only ever lowers: an operator running the campaign below it now keeps their own value, which `reasoning_effort=` did not allow. That makes a second capped call site, and `test_agent_model_and_effort.py`'s allowlist exists to refuse one until the exception is argued — the argument is written there.
- The last merge brought main's own fix for the model switch: `resolve_agent_model("auto")` now returns `""` and `Config.agent_runtime` reads `CLAUDE_MODEL` / `CODEX_MODEL` once the provider is settled, so a Claude id no longer reaches the Codex gateway. That is main's change, not this branch's; it is named here because it lands on the same function this branch's role/effort plumbing edits. Main's `UsageAccumulator` `on_update` callback and this branch's `by_role` bucketing were merged as a union — the callback fires after the per-role counters are incremented, so an external ledger reading it sees the call already attributed.

## End-to-end: this branch against its base

The per-change A/Bs above each isolate one change. This one runs the whole branch: six campaigns of the shipped `triton-softmax-forge-loop` example, one hour and one GPU each, `claude-opus-5` at `high` effort (rtk 0.48 on `PATH` in every arm — the control that produced the usage count in change 4), the five task files verified byte-identical between the two trees before launch.

| arm | runs | tree |
|---|---|---|
| base | B1, B2 | the commit this PR is based on, checked out in its own worktree |
| PR | D1, D2 | this branch, defaults (`KERNELFORGE_DEFER_KNOWLEDGE_MAPS=1`) |
| PR, maps inlined | I1, I2 | this branch, `KERNELFORGE_DEFER_KNOWLEDGE_MAPS=0` |

**First-turn prefix**, the whole cached prompt a session opens with — `input + cache_creation + cache_read` on the first assistant message, read from the Claude CLI transcripts:

| session | base | PR, maps inlined | PR |
|---|---|---|---|
| analysis | 59,488 (n=2) | 24,632 (n=3), −58.6% | **8,780** (n=2), **−85.2%** |
| workspace, small | 30,932 (n=12) | 26,386 (n=19) | **8,920** (n=6), **−71.2%** |
| workspace, large | 57,737 (n=7) | 49,713 (n=8) | **28,154** (n=6), **−51.2%** |

Workspace sessions fall into two clusters an order of magnitude apart in count, so they are reported per cluster; the aggregate (40,808 → 18,537, −54.6%) moves with the cluster mix as well as with the prompt. Every arm's range is disjoint from every other arm's on both session kinds, at n≥6 per cell. These numbers were measured with change 4's 350-token rtk paragraph still in the prompt. One further campaign was run after the removal, same task and same budget, as a check rather than a seventh arm: its four small-workspace sessions opened at 8,251 / 8,826 / 8,836 / 8,859, every one below the 8,920 this table records — which is the cluster that paragraph was attached to, so that is where its removal shows. The other two clusters from that run are not comparable and are not read as movement either way: its single analysis session (9,067) ran `static-only` under a one-hour budget, a different prompt that never carried the rtk paragraph at all, and its three large-workspace sessions (34,101 / 34,639 / 34,645) each open on that run's own `analysis_evidence` payload — 78,116 / 79,617 / 79,655 characters, about 19.9k tokens — leaving roughly 14.7k of scaffolding underneath. That cell tracks how much the analysis agent wrote on the day, not what this PR ships, which is why it is reported at n=6 rather than from any single run.

The split matches the per-change results: base → maps-inlined is changes 3 and 6 (−58.6% on analysis), maps-inlined → PR is change 2 (a further −64.4%).

**Where the prefix saving goes.** A smaller prefix is paid on every turn, not only the first, so the same transcripts give the per-turn figure directly — every assistant message's `input + cache_creation + cache_read`, averaged over all sessions in an arm:

| arm | sessions | first-turn prefix | turns/session | context/turn | context/session | output/session |
|---|---|---|---|---|---|---|
| base | 23 | 42,014 | 19.0 | 55,314 | 1,182,513 | 39,314 |
| PR, maps inlined | 30 | 32,431 | 18.3 | 45,471 | 888,494 | 35,882 |
| PR | 14 | 17,143 | 19.3 | **33,303** (−39.8%) | **833,568** (−29.5%) | 44,802 |

The saving carries through in absolute terms and is diluted in relative terms, for a reason worth stating: of base's 55,314 tokens per turn, 42,014 is the prefix and 13,300 is what the session grows during its work — tool output, its own writes. Under this PR the grown part is 16,160, essentially unchanged; the 22,011-token drop per turn is 88% of the 24,871-token prefix drop. Nothing else moved, and nothing else was expected to.

Turns per session are flat (19.0 → 19.3), so this does not come from shorter sessions. Output per session is up 14%, which is what a time-boxed budget does with reclaimed context: `--max-hours` bounds the campaign, not tokens, so the room a smaller prefix frees is spent writing more rather than finishing cheaper. That is the intended trade, and it is also the second reason campaign totals are the wrong place to look for this — the first being that a campaign's session count ranged 8 to 17 across these six runs, which swamps a 29.5% per-session difference outright. Note too that most of what is saved is `cache_read`, billed at a tenth of the input rate, so a 30% token saving is not a 30% cost saving; the bill's centre of mass is `out`, the column this PR does not touch.

**Campaign totals are not conclusive at this n and are not claimed to be.** Cost ran $8.85–$15.25 in the base arm and $7.98–$16.88 within the maps-inlined arm alone, driven by how many turns a campaign happens to take; that spread is larger than any difference between arms. Arm means came out `cache_read` 5.31M → 4.28M and `cache_write` 742k → 494k for PR against base, in the expected direction, but with n=2 that is a coincidence-compatible number, not a result. The prefix table is the measurement that isolates the change.

Change 5 in action, one PR campaign's summary line: `206 in / 164,181 out / 489,203 cache-write / 4,173,281 cache-read tokens, $9.51 (9 calls)` — 4.83M tokens, of which the two columns printed before this PR were 3.4%. By role: analysis $3.02, implementer $2.55, specialist compute $0.93, specialist algorithm $0.91, lane 2 synthesis $0.87, orchestration $0.62, lane 1 synthesis $0.61. The same line from the post-removal campaign: `188 in / 162,830 out / 541,621 cache-write / 3,198,866 cache-read tokens, $10.21 (10 calls)` — eight roles, all attributed, no `unattributed` bucket.

## Tests

`PYTHONPATH=src pytest src/kernelforge/tests --ignore=src/kernelforge/tests/fusion` — 3206 passed, 8 skipped, on the merge of `main` at #1409. Change 4 landed after that run and deletes 31 tests; it was checked by running the same suite on this branch and on the untouched base tree with one interpreter and one dependency set, which differed by exactly one failure — a file that does not exist on base, failing on a package missing from the shared venv, and passing once it was installed. `ruff check .` and `ruff format --check .` clean. After the #1475 merge, `test_agent_env_contract.py`, `test_agent_model_and_effort.py`, `test_usage*.py` and `test_forge_orchestration.py` were rerun on the merge result — 152 passed — and ruff is clean over the merged files; the rest is CI's. (`fusion/test_harness_noise.py` has 5 failures that reproduce on `origin/main`.) New: `test_usage_attribution.py`, `test_context_cost_guards.py`, `test_deferred_knowledge_maps.py`. `test_build_output_filter.py`/`test_rtk.py`/`test_rtk_install.py` are deleted with change 4.


jiaqiang-dot-liu:
LGTM

jiaqiang-dot-liu (approved):

PR #1444, opened 2026-09-08
[Feat] Grade AgentX on the axis InferenceX ranks

ZhengGong-amd:
## We graded the fastest users, on a number made of cache hits

Three defects, each confirmed against the archived AgentX session
`Hyperloom-Sessions/vllm/Kimi-K3/_archive/20260831T124523Z-983e17b2` and the
pinned upstream checkout at `.cache/InferenceX@3d5581562`.

**The percentile pointed at the wrong tail.** `mapping.py` read `p90` of aiperf's
`e2e_output_token_throughput`, which is a per-request rate flagged
`LARGER_IS_BETTER`. aiperf takes percentiles on the raw values, so its p90 is the
*snappiest* decile. Upstream computes the percentile in seconds-per-token and
then inverts (`MODELS.md:78`), which keeps the slow tail; that is `p10` of the
rate. In the real export: `p10=22.56`, `p90=371.23`. We were grading a number
16x more optimistic than the one a submission is ranked on.

**The objective was mostly accounting.** `total_token_throughput` is 99.2% input
tokens on this corpus, and 97.5% of those are prefix-cache hits, so the graded
quantity was dominated by how fast the server can hit cache.

**The workload description was fiction.** `state.isl/osl` stayed at 1024/1024
while the corpus runs p50 95k / p99 506k in and p50 333 / p99 6386 out. Every
specialist prompt, the Critic's context, the manifest and the roofline read those
placeholders.

The module docstring also claimed upstream "compares throughput per chip *at a
fixed interactivity target*". It does not: `MODELS.md` publishes a Pareto
frontier with interactivity as the x-axis and no target anywhere in the master
configs. That sentence had propagated into `docs/reference/environment-variables.md`.

## The verdict is two-dimensional

`resolve_graded_comparison` is the one chokepoint every lane already passed
through, so the rule lives there: KEEP when interactivity clears the threshold
with per-chip throughput inside the noise band, REVERT when both axes regress,
and RECORDED when neither dominates.

RECORDED exists because the asymmetry is real. Upstream stores a dominated point
and shows it in the unfiltered scatter, so keeping one costs a row; discarding a
config is final, and `configs/nvidia-master.yaml:8788` documents an arm that was
"strictly dominated" below conc 16 whose frontier peak sat at conc 48. A false
REVERT is the expensive mistake, so REVERT requires both axes to fail.

The threshold is floored at 2% for AgentX, because the slow-tail percentile's own
run-to-run variance is unmeasured and the default 1% sits inside the band. That
floor is a placeholder, recorded as debt rather than as a measurement.

Fixed concurrency approximates what upstream maintainers do to decide whether a
change is an improvement — `configs/amd-master.yaml:1704` selects "the
Pareto-optimal arm per concurrency". It does not approximate what gets published;
that needs the ladder, and a candidate losing at conc 8 can still win at another
rung. Hence RECORDED rather than REVERT.

## The corpus describes itself

`SharedState.agentx_corpus_shape` is seeded from the canonical corpus and
replaced by the measured distribution after every aiperf run.
`prompts/agentx_context.py` renders the workload and grading blocks from it, so
Orchestration and the specialists read one source instead of two copies of the
same literals, and the numbers track the corpus the session actually replayed.

`map_aiperf` carries the distributions, the loader from
`metadata.dataset.loader`, and `request_error_rate`. `benchmark_mode` joins the
profile-workload identity and the variant fingerprint, so an AgentX trace and a
synthetic 1024/1024 trace can no longer collide.

Four consumers stop pretending they have a shape. The roofline abstains — its
`kv_seq = isl + osl//2` is 65x short, and the conc-sweep chart already refused to
draw the ceiling for that exact reason. `conc_sweep` defaults off, since each rung
is a 3600s window and the session grades one concurrency. `target_analysis` skips:
upstream agentic rows carry null isl/osl by design and the strict filter can never
match one. Recipe-KB reads are blocked alongside the writes, because a recipe
validated at 1024/1024 passes the donor shape gate and would warm-start an
agentic session onto the wrong regime.

## Eval becomes the validity gate

AgentX used to set `eval_disabled` to dodge the missing-accuracy guard. It now
runs the check upstream actually gates on: `request_error_rate <= 0.10` over
completed requests, mapped pass/fail against a fixed reference through the
existing scriptable branch. It fails closed — a run that reported no rate is not
comparable, and calling that a pass is how an incomparable measurement reaches
the leaderboard set.

## The verdict is legible

`intvty_regression` used to reach the agent as a bare token, and
`integrate_patch`, the `current_best` lift and `kernel_stack` wrote only
`log.info`, so a correctly-rejected patch was indistinguishable from a failed
measurement. The measured pair now travels in the ledger `reason` that SEED
renders, variant lines carry `graded_objective` so a gain column is never read
against an unnamed axis, and `kernel_stack` records `graded_verdict` (the stack
still rolls back either way — it cannot be left half-applied).

## Second commit: the compatibility layer

The first commit kept deprecated aliases, a `.vetoed` shim, two accessors with no
caller, a legacy key fallback, three `except` blocks and a dead variable renamed
to silence the linter — one of which was hiding a real kwarg mismatch. All
deleted. The corpus numbers were also still literals in two prompt builders while
`agentx_corpus_shape` was written and never read; that is what
`prompts/agentx_context.py` fixes.

## Template

- **Linked issues:** none. Three untracked notes at the repo root record the
  follow-ups: `kb-agentX-schema.issue.md` (the recipe id needs a mode dimension,
  which is what forces the read block), `agentX-compareGPU.issue.md` (match
  agentic rows on null isl/osl and compare on interactivity),
  `agentX-kernel-evidence.issue.md` (route prefix-cache hit, the TTFT/E2EL split
  and the OSL distribution into the profiler digest; also `gemm_tuning.py:54`,
  which still hardcodes ISL/OSL into a GEAK prompt). `.gitignore` now covers
  `*.issue.md`.
- **Tests:** `perf_metric`, `graded_comparison`, `map_aiperf`, `conc_sweep`,
  `workload_envs`, `shared_state_units`, plus the
  agentx/prompt/manifest/bootstrap/accuracy/critic/specialist selection — 1961
  passed. New coverage for the p10 read, the three-way verdict, the corpus-shape
  record and the mode-branched prompts. `ruff check` and `ruff format --check`
  clean. `test_agentx_preflight::test_capability_absent_raises` fails on this
  branch before the change (environment-dependent).
- **Breaking changes:** yes, internal. `intvty_p90` is `e2e_norm_intvty_p90`,
  `composite_v1` is `intvty_v1`, `total_tput_grading_enabled` and
  `total_tput_serving_grading_enabled` are gone, `GradedComparison.vetoed` is
  gone and `graded_on_total` is `graded_on_intvty`, `agentx_kb_write_blocked` is
  `agentx_kb_blocked`, and `graded_metric_key` returns the interactivity field
  under AgentX. `conc_sweep` defaults off under AgentX. No serving or CLI surface
  changes; no resume path was kept, by instruction.
- **Single concern:** yes — AgentX grades on the axis it is judged by, and the
  prompts describe the workload it runs. The prompt and legibility work is in
  scope because an axis the agent cannot see is an axis it cannot optimise.
- **Upstream root cause:** no. The one upstream-shaped item is the Recipe KB
  identity lacking a mode dimension; it is filed in
  `kb-agentX-schema.issue.md` for the KB owners rather than worked around here
  beyond the read/write block.


src/hyperloom/orchestrator/actions/executors/_accuracy_gate.py
```diff
@@ -621,6 +655,21 @@ def parse_eval_results(
 
     scriptable = framework_registry.is_scriptable(framework)
 
+    # AgentX has no lm-eval; its correctness signal is the error rate upstream
+    # gates a submission on. Fails closed like the scriptable gate below: a run
+    # that reported no rate is not comparable, and treating that as a pass is
+    # how an incomparable measurement reaches the leaderboard set.
+    if is_agentx_mode(benchmark_mode):
+        rate = parse_agentx_error_rate(workspace)
+        passed = rate is not None and rate <= AGENTX_ERROR_RATE_THRESHOLD
```
xiaofei-zheng: **Blocking: unit mismatch — the gate is 100x too strict, and it never actually fails closed.**

aiperf's `request_error_rate` is a **percentage**, not a fraction: `RequestErrorRateMetric._derive_value` returns `100.0 * errors / total` with `unit = GenericMetricUnit.PERCENT` (confirmed in two aiperf checkouts). Comparing it against `AGENTX_ERROR_RATE_THRESHOLD = 0.10` rejects anything above **0.1%**, where upstream allows 10%.

Reproduced end to end (`map_aiperf` -> `inferencex_result.json` -> `parse_eval_results`):

```
mapped rate: 5.0        # 5% error rate; valid upstream
{'accuracy': 0.0, 'task': 'agentx_error_rate', 'error_rate': 5.0}   # judged fail
```

This is not confined to a variant REVERT. Since this PR also drops `eval_disabled=... or _agentx_enabled()` in `bootstrap.py`, an AgentX **baseline** that returns accuracy 0.0 gets stamped as an eval failure by `_maybe_stop_on_missing_baseline_accuracy`, which blocks `baseline_tput` from anchoring and stalls the session — exactly the scenario the deleted comment described.

Second half of the same problem: `mapping.py` uses `stat(m, "request_error_rate")`, which coalesces a missing metric to `0.0` rather than `None`. So `inferencex_result.json` always carries a number, `parse_agentx_error_rate` never returns `None`, and the documented fail-closed path is dead:

```
missing -> mapped: 0.0
{'accuracy': 1.0, 'error_rate': 0.0}    # a run that reported no rate passes
```

That path is reachable in practice: none of the aiperf exports I could find locally carry `request_error_rate` at all — upstream `validate_agentic_result.py` derives the rate itself from `error_request_count / completed_request_count`. Deriving it the same way (and leaving it absent when the counts are missing) would fix both halves.

src/hyperloom/orchestrator/actions/executors/explore.py
```diff
@@ -1750,15 +1761,18 @@ def _stopped_by_the_run(result: Any, *, variant: GridVariant, idx: int, round_la
...
-                            running_base_perf = None
+                        if grade_on_intvty:
+                            # The KEEP's own axes become the next variant's
+                            # anchor. A KEEP that could not supply them clears
+                            # the anchor so the rest of the round grades on
+                            # output too, rather than against a stale pair.
+                            running_base_perf = perf_snapshot_from_mapping(variant_meas)
+                            if running_base_perf is None:
+                                log.info(
+                                    "explore: KEEP %r had no graded axes; grading the rest of "
```
xiaofei-zheng, line 1401: **Blocking: clearing the anchor no longer degrades to output grading — it silently re-resolves the session anchor.**

The log says the rest of the round grades on output throughput, but `resolve_graded_comparison` treats `anchor_perf=None` as "not supplied" and falls back to `resolve_grading_anchor_perf(state)`:

```
resolve_graded_comparison(state, meas, anchor_perf=None, anchor_tput=40.0)
-> objective='e2e_norm_intvty_p90', reference=20.0 (from current_best), verdict='KEEP'
```

The pre-PR code short-circuited explicitly (`if grade_on_total and running_base_perf and cand_snap:`), so a `None` round anchor always meant the output axis. This is a behaviour change introduced here.

Trigger: an AgentX round where one variant lacks the graded axes and KEEPs on the output axis. Every later variant in that round is then graded on interactivity against the anchor the round opened on, while stacking on top of that KEEP — so the earlier KEEP's gain is credited to the later variant and can carry it over the threshold on its own. The log line also no longer matches the axis actually used.

A sentinel distinct from `None` for "anchor deliberately cleared", or passing the degrade decision down explicitly, would restore the old semantics.

xiaofei-zheng (commented):
Reviewed against `git merge-base origin/main` (`0a8edb157`), with both findings reproduced locally. `test_perf_metric` / `test_graded_comparison` / `test_map_aiperf` / `test_agentx_context_and_eval` pass (89), so neither is covered by the new tests.

Two blocking items, both inline. No other blocking issues found: the p10 read, the renamed `e2e_norm_intvty_p90` key (no stale readers left), the three-way verdict, the corpus-shape plumbing and the four abstaining consumers all check out.

ZhengGong-amd:
## Scanner findings: none are actionable

Three inline findings landed on `87f8866f7`. Two report the same thing. I checked each against the current head (`6767a77c4`) rather than against the commit they were raised on.

### CodeQL — Cyclic import, `_accuracy_gate.py:657`

Stale, and the literal remedy would have broken the module.

The cycle was real: `_workload_envs` imports `_accuracy_gate` at module scope, so the back-reference to `agentx_active` had to be deferred into the function body. Dropping that import, which is what the finding asks for, raises `ImportError` at load. The deferred import was the remedy, not the disease.

It is already gone as of `6767a77c4`. `parse_eval_results` now takes `benchmark_mode` as a parameter and imports only `is_agentx_mode` from `hyperloom.common.perf_metric`, which is a leaf. No cycle remains. That change also closed a fragility the scanner did not see: the old call passed no `shared_state`, so it read `HYPERLOOM_AGENTX` alone and would have missed the AgentX branch in any subprocess that did not inherit the variable.

### CodeQL + code-quality — Implicit string concatenation, `specialist_prompt_builder.py:2350`

No defect. The construct is the house norm in this file, and the line was already written this way before this PR.

- The pre-existing version of this exact entry was itself a two-literal implicit concatenation. This PR changed the wording and made it three. That is the only reason it appears in the diff.
- Line 2346 immediately above it is the same pattern and is not flagged, because it is untouched.
- There are 35 such sites in the files this PR touches alone, nearly all pre-existing.

The rendered string is correct, so there is nothing to fix behaviourally. Converting all 35 is an unrelated style sweep and does not belong in a PR about the AgentX grading axis.

### Separately

While verifying the above I did find a real problem on that same line, which neither scanner reports: it hardcodes `~142:1 (ISL avg 114k / OSL avg 806)`, duplicating corpus numbers that now live in `SharedState.agentx_corpus_shape` and are rendered from it by `prompts/agentx_context.py`. It also sits in `_section_pd_disaggregation`, which is gated on PD topology rather than on AgentX, so a synthetic disaggregated session reads AgentX corpus figures as well. I will drop that clause; the corpus shape already reaches an AgentX specialist through Section 2.

xiaofei-zheng (approved):
## LGTM — both blocking findings are fixed

Re-reviewed at `61e663e32` (was `6767a77c4` when I filed the two inline comments). Both are addressed, and I re-verified each by running the code rather than reading the diff.

### 1. `request_error_rate` unit mismatch — fixed

`AGENTX_ERROR_RATE_THRESHOLD = 0.10` → `AGENTX_ERROR_RATE_THRESHOLD_PCT = 10.0`, and the missing-metric half is fixed too: both `agentx/mapping.py` and the standalone `assets/agentx/map_aiperf.py` now pass `default=None`, so an export without the metric fails closed instead of coalescing to a perfect `0.0`.

Driven end to end through `map_aiperf` → `inferencex_result.json` → `parse_eval_results`:

```
threshold = 10.0
5%  valid  : mapped=5.0  -> accuracy=1.0 error_rate=5.0
12% invalid: mapped=12.0 -> accuracy=0.0 error_rate=12.0
missing    : mapped=None -> accuracy=0.0 error_rate=None
```

Previously 5% (a valid submission under upstream's 10% gate) was stamped 0.0, and a missing metric scored a perfect 1.0. Both directions are correct now.

### 2. Grading anchor silently falling back — fixed

The `ANCHOR_DEGRADED` sentinel separates "this round already degraded to the output axis" from "no explicit anchor supplied", and `resolve_graded_comparison` branches on it ahead of the `anchor_perf is not None` check. Both `explore.py` sites set it — the round-opening path and the post-KEEP path where the KEEP could not supply axes — and the value does reach the resolver via `anchor_perf=running_base_perf`.

```
anchor=None            -> objective='e2e_norm_intvty_p90'  graded_on_intvty=True   reason=''
anchor=ANCHOR_DEGRADED -> objective='output_throughput'    graded_on_intvty=False  reason='round_degraded'
```

That is exactly the silent fallback path I flagged; it is now cut off.

### Also checked

- `9e926c803` (hardcoded `~142:1` corpus ratio dropped from `_section_pd_disaggregation`): prompt text only, no correctness impact. Agreed it did not belong on a PD-topology-gated section.
- Ran the directly affected suites — `test_accuracy_gate_units.py`, `test_graded_comparison.py`, `test_map_aiperf.py`, `test_agentx_context.py`, `test_specialist_prompt_builder_coverage_unit.py`: all green.

No remaining blocking issues from my side.

PR #1445, opened 2026-09-08
fix(telemetry): record each round's GPU telemetry instead of losing it to a name mismatch

luochen-amd:
## Why

GPU power, temperature and clock have been sampled all along -- Magpie's own monitor on one node, the pod-side `rocm-smi` sampler on several -- and land in each round's `benchmark_report.json` under `gpu_monitor`. Nothing read them correctly.

The read side looked for `power_w` while the producer writes `power_watts`, and the alias chain it used could not express "absent" at all:

```python
"avg_power_w": _avg("power_w") or _avg("power"),
```

`_avg` returned `0.0` for a missing metric, so a genuine `0.0` also fell through to the next alias, and an entirely unsampled metric shipped as a plausible-looking zero. Across **101 archived sessions the GPU section read all zeros** -- not missing, which someone would have questioned, but zero.

## What changed since the first version of this PR

The original fix repaired `_aggregate_gpu_monitor` in `breakdown/collectors/telemetry.py`. #1455 then retired the whole read side, including that file and the `telemetry` section it fed, so the fix had no remaining home.

Rewritten as a **per-round artifact**. That is the better shape anyway: the session-level aggregate averaged baseline, explore and roofline rounds together and described none of them. Power and thermals only mean something against the round that produced them.

`gpu_metrics.json` lands beside `benchmark_report.json` in the round workspace and joins the session bundle. Recording its path in the session breakdown is a follow-up PR, alongside the same for `kv_metrics.json` from #1446.

## Design

**Both producer shapes are read.** Magpie writes a pre-aggregated `{min,max,avg}` block; the multi-node harvester writes flat per-sample scalars. For a flat sample the scalar is that sample's mean and its peak alike.

**One alias wins per block, chosen once rather than per statistic.** Resolving per statistic let a mean come from `power_watts` and a peak from a stale `power_w` in the same block -- which can report a maximum below the average, a self-contradictory number of exactly the kind this exists to stop emitting.

**Means are weighted by `sample_count`.** A Magpie block summarises many samples while a flat sample is one; unweighted, a 35-sample block would pull the round's mean as hard as a 1355-sample one.

**Every metric is `float | None` and is never coerced.** A card that drew no power and a card nobody sampled must not produce the same number. `samples` counts only blocks that yielded a reading, so a large sample count can never sit beside a row of nulls.

## Utilization and VRAM

These answer what power and temperature cannot: whether the GPU was compute-idle, and whether it was memory-constrained. Both are reported, and both come back `null` on a single-node round.

That is the honest answer, not a gap in the reader. Across 13 production `benchmark_report.json` files, Magpie's `GPUMonitor` emits exactly `sample_count`, `duration_sec`, `temperature_c`, `gpu_clock_mhz`, `mem_clock_mhz` and `power_watts` -- no occupancy reading at all. Magpie is external (`github.com/AMD-AGI/Magpie`), so that is not fixable from here. The multi-node harvester does sample both, and those rounds report them.

`null` rather than `0.0` because 0% utilization asserts the GPU sat idle through the round, which is the opposite of what an unsampled metric means. The percent aliases are percent-named only: folding an absolute `vram_used_mb` into a `_pct` field would put 81920 where a percentage belongs, which is the same class of defect this PR started from.

`mem_clock_mhz` is now read too -- present in every real Magpie block and previously ignored.

## Verification

Run against **every `benchmark_report.json` on the cluster**: 8 of 8 now yield real figures where all 8 previously read `0.0`.

| round | samples | mean / peak power | mean / peak temp | mean clock |
| --- | --- | --- | --- | --- |
| AgentX baseline | 1355 | 632.0 W / **1001.0 W** | 67.3 C / 71.0 C | 2241.9 MHz |
| roofline | 969 | 306.1 W / 838.0 W | 57.3 C / 62.0 C | 1457.7 MHz |
| short round | 31 | 251.1 W / 262.0 W | 57.7 C / 58.0 C | 110.5 MHz |

The first row is the point of the per-round shape: a session-level mean would have flattened a 1355-sample round under load against the 31-sample rounds beside it.

17 tests, with the block fixtures taken verbatim from production reports rather than invented -- including the regression that this exact Magpie shape used to aggregate to all zeros, and that a real `0.0` is a reading rather than a miss.


haishuok0525:
Reviewed for correctness only. I read the diff plus `_common._to_float`, the `collect_telemetry` → `exporter` call chain, and every consumer of `gpu_monitor_aggregate` / `avg_power_w` in the tree, then extracted the new `_gpu_metric` / `_aggregate_gpu_monitor` into a standalone harness and ran it. All eight test assertions hold, including the weighted `298.02`.

The core fix looks right and I did not find a regression. The multi-node injector's flat `power_w` / `temperature_c` / `clock_mhz` are still covered by the alias tuples; nothing downstream actually reads the five metrics (the `avg_power_w` in `benchmark_result.py` belongs to its own `gpu_monitor_by_role` aggregate), so `float` → `float | None` breaks no consumer; `schema.py` has `from __future__ import annotations`, so `float | None` in the TypedDict is fine on 3.10; ruff only selects E/F/W here, so the bare `zip` is not a B905 hit.

Four things I'd call bugs.

**1. The doc claims the key is absent, but it is always present as `{}`**

The new doc paragraph says `gpu_monitor_aggregate` "is absent entirely when no benchmark report carried a `gpu_monitor` block". `collect_telemetry` writes the key unconditionally:

```python
"gpu_monitor_aggregate": _aggregate_gpu_monitor(all_reports, warnings),
```

`_aggregate_gpu_monitor` returning `{}` still leaves the key in the section, and nothing on the `exporter._pick` / `_safe_collect` path prunes empty values. A consumer following the doc with `if "gpu_monitor_aggregate" not in telemetry` gets the wrong answer. Either say "empty object" in the doc, or actually omit the key in `collect_telemetry`.

**2. `samples` can be large while every metric is `None`**

`samples` is `int(sum(weights))`, summed over all blocks regardless of whether a block contributed any metric at all:

```
in:  {"sample_count": 27000, "duration_sec": 10.0}
out: {'samples': 27000, 'blocks': 1, 'avg_power_w': None, ...}
```

That is the same "well-formed section with heavy-looking sample counts behind nothing measured" shape this PR sets out to remove. `_avg` already computes a per-metric `weight_sum`; either surface that, or have `samples` count only blocks that contributed at least one reading.

**3. `weights` uses the falsy fallback the PR argues against**

```python
weights = [max(1.0, _to_float(b.get("sample_count")) or 1.0) for b in blocks]
```

`sample_count: 0` — a monitor that started and sampled nothing — takes the `or 1.0` branch and is reported as one sample. This is the same pattern as the `_avg("power_w") or _avg("power")` form the description calls out. An explicit `is None` check before deciding whether to clamp would be consistent with the rest of the change.

**4. `avg` and `max` can resolve to different aliases within one block**

`_gpu_metric` does `continue` when the key is present but `field` is not, so it falls through to the next alias, and the two statistics can come from different keys:

```
in:  {"power_watts": {"avg": 300.0}, "power_w": 12.0}
out: avg_power_w = 300.0 (from power_watts), max_power_w = 12.0 (from power_w)
```

The result is a max below the avg. A related variant, `{"power_watts": {"avg": None, "max": 310.0}}`, yields `avg=None, max=310.0`. Magpie writes all three statistics today so this cannot fire yet, but the failure shape — a producer omitting one field and the aggregator silently emitting a self-contradictory number — is the same class of defect this PR fixes. Safer to select the alias once, on "key present and at least one statistic parses", and not fall through afterwards.

Minor and ignorable: `int(sum(weights))` truncates rather than rounds, which only matters if `sample_count` is ever non-integral.

I'd fix #4 before merge since it is cheap and squarely on-topic. #1–#3 are fine to fold into a follow-up.

haishuok0525:
One remaining telemetry gap: this fixes parsing for power, temperature, and GPU clock, but it still does not provide GPU utilization or VRAM usage.

The multi-node injector already emits flat `gpu_util_pct` and `vram_pct` samples, but `_aggregate_gpu_monitor` ignores both fields. The single-node Magpie block used by the regression test contains only `power_watts`, `temperature_c`, and `gpu_clock_mhz`, so this change also does not add utilization or VRAM collection on that path.

As a result, `gpu_monitor_aggregate` remains unable to report whether the GPU was compute-idle or memory-constrained even when the multi-node source carried those readings, and single-node sessions still have no such readings at all. Please extend the producer/normalization path as needed and add aggregate fields for average/peak GPU utilization and VRAM usage, with coverage for both the Magpie nested shape and the multi-node flat shape.

luochen-amd:
Added, with one finding that changes what "coverage for the Magpie nested shape" can mean.

**Multi-node: done.** `_aggregate_gpu_monitor` now reads `gpu_util_pct` and `vram_pct` and reports `avg_gpu_util_pct` / `max_gpu_util_pct` / `avg_vram_pct` / `max_vram_pct`, through the same alias resolution, sample-count weighting and tri-state path as power and temperature. A block that yields only occupancy now counts as contributing, so a util-only source is no longer scored as an empty block.

**Single-node: the readings do not exist to collect.** I checked 13 production `benchmark_report.json` files rather than reasoning from the fixture. Every Magpie `gpu_monitor` block carries exactly:

```json
{"sample_count": 59, "duration_sec": 124.02,
 "temperature_c": {"min": 46.0, "max": 68.0, "avg": 61.3},
 "gpu_clock_mhz":  {"min": 131, "max": 2096, "avg": 1411.6},
 "mem_clock_mhz":  {"min": 900, "max": 1300, "avg": 1171.2},
 "power_watts":    {"min": 149.0, "max": 751.0, "avg": 524.3}}
```

No utilization, no VRAM, in any of them. The regression fixture was thin because the producer is thin. And Magpie is external (`github.com/AMD-AGI/Magpie`, installed by `install.sh`) -- its `GPUMonitor` is not in this tree, so the producer side of single-node cannot be extended from this PR. The only in-repo lever is `_magpie_patcher.py`, which patches Magpie's `benchmarker.py` at runtime; extending that to rewrite its telemetry collection is a materially riskier change than this one and I would rather not smuggle it in here. Happy to open it as a follow-up against Magpie (or as a patcher change) if you want that path -- say which.

So single-node reports `None` for both, and the docs now state that plainly with a per-topology table. `None` is load-bearing here: 0% utilization asserts the GPU sat idle through the round, which is the opposite of "nobody sampled it", and that conflation is what this branch started from.

**Coverage.** Both shapes are tested: the flat multi-node samples, and the nested `{min,max,avg}` form for util/VRAM (which pins the parsing so it works the day a producer emits them). Plus: a real Magpie block reports `None` rather than 0.0; a genuinely idle GPU at 0.0 is kept as a reading; a mixed-topology session keeps the multi-node occupancy; and `vram_used_mb` / `memory_used_bytes` are *not* accepted into a `_pct` field -- an absolute MiB reading is a different quantity, and letting it fall through would put 81920 where a percentage belongs, which is the same class of bug as the original one.

End to end over those 13 real reports:

```json
{"samples": 1984, "blocks": 13,
 "avg_power_w": 347.8, "max_power_w": 753.0,
 "avg_temp_c": 57.52, "max_temp_c": 78.0, "avg_clock_mhz": 1732.54,
 "avg_gpu_util_pct": null, "max_gpu_util_pct": null,
 "avg_vram_pct": null, "max_vram_pct": null}
```

luochen-amd:
Reworked, because #1455 removed what the previous version fixed. Title, description and implementation all changed; the diff is not an increment on what was reviewed before.

**What happened.** The earlier version repaired `_aggregate_gpu_monitor` in `breakdown/collectors/telemetry.py` and extended `GpuMonitorAggregate`. #1455 retired the read side -- that file, the `telemetry` section and its schema are all gone from main, and the V6 recorder carries no GPU telemetry of its own. So the fix had no delivery vehicle, and rebasing it would have meant resurrecting a deleted subsystem.

**What it is now.** The same normalisation, writing a per-round `gpu_metrics.json` beside `benchmark_report.json` rather than a session-level aggregate. That is the better shape independently of #1455: the aggregate averaged baseline, explore and roofline rounds together, and on the cluster that means a 1355-sample round under load was being flattened against 31-sample rounds beside it.

**Where the earlier review comments landed.** The four GPU-aggregate correctness fixes are carried over unchanged -- one alias resolved per block rather than per statistic, `sample_count` weighting, tri-state metrics, and `samples` counting only blocks that contributed. The follow-up asking for utilization and VRAM is implemented, with the caveat that only the multi-node harvester samples them: across 13 production reports Magpie's monitor emits power, temperature and both clocks and no occupancy reading at all, so single-node rounds report `null`. Magpie is external, so that is not fixable from this repo.

**Recording the artifact path in the session breakdown is deliberately not here.** It belongs with the same change for `kv_metrics.json` from #1446 (now merged), as one follow-up that adds both paths to the recorder's round event, next to the `benchmark_report_path` it already carries.

Verified against every `benchmark_report.json` on the cluster: 8 of 8 now yield real figures where all 8 read `0.0` before.

luochen-amd:
### The red shard 4 is not from this PR

`test_target_analysis_executor.py::test_agentx_state_to_external_reference_and_final_report[True]` fails with `assert 'no_match' == 'ok'`. It is a pre-existing order dependency in the suite, which any PR that changes the number of tests will hit.

**Control experiment.** I branched off `main` and added a single test file with 17 trivial `assert i == i` tests -- the same count as this PR's, and no product code whatsoever. Shard 4 failed the same test with the same message ([#1498](https://github.com/AMD-AGI/Hyperloom/pull/1498), now closed):

```
FAILED src/hyperloom/inference_optimizer/tests/test_target_analysis_executor.py::
  test_agentx_state_to_external_reference_and_final_report[True] - AssertionError: assert 'no_match' == 'ok'
```

Tests are distributed by pytest-split across 6 shards and then by xdist across 2 workers per shard, so the test count alone decides who shares a worker with whom. Adding tests -- any tests -- reshuffles that, and this one fails when it lands next to whatever leaks the state it depends on.

**Supporting evidence.** The `[False]` parametrization passes while `[True]` fails, which is what worker placement looks like rather than a semantic difference: under `--dist load` the two parametrizations go to whichever worker is free, so they are not on the same one. The test passes on `main`, passes alone, and passes with this PR's changes applied locally -- including all 32 tests in its own file.

I could not reproduce it on Windows: I swept all 493 test files in `inference_optimizer/tests`, running each one ahead of the target test, and none of them reproduced it. That points at a polluter among the tests skipped on Windows, so it needs a Linux session to pin down.

The owner of the target-analysis test is better placed to fix the leak than I am -- happy to file it as its own issue if that is preferred.

### Unrelated to the above, one real fix in the latest push

`write_gpu_metrics` was the only step in `harvest_leaked_artifacts` not wrapped, so a read-only or full round workspace would have raised out of the harvest and discarded the artifacts collected above it. Telemetry describes a round rather than forming part of it, and failing to describe one must not discard it. Now guarded, with a regression test (18 tests in the file, all green).

luochen-amd:
All four addressed, and CI is green -- 30/30, including the shard 4 and coverage jobs that were red.

**CI root cause: confirmed, and it was the `PRECISION` leak.** Thank you for pinning it; my own control experiment had only got as far as proving the failure was not caused by this PR's product code. Reproduced exactly as you described: with `PRECISION=fp8` in the environment both parametrizations fail, and with the fixture in place they pass. Fixed as test isolation, not by touching the production precedence -- an autouse fixture in `test_target_analysis_executor.py` clears `PRECISION`, `FRAMEWORK`, `MODEL_PATH`, `ISL` and `OSL` per test, so the executor answers from the arguments each test sets rather than from whatever the worker inherited. `params > env > SharedState` is unchanged, since other tests require the explicit override to beat stale state.

**P1, artifact written before the report settles.** Fixed in `_settled_measurement`, which is the one path every grid round settles through, so it is a single place rather than eight call sites. It writes from the report already in hand instead of re-reading the file. The harvest-time call stays, because rounds that never settle still need it, and it is now explicitly best-effort: a missing report there is the expected case, not a fault. Regression test covers exactly the delayed-report sequence -- harvest with no report on disk produces no artifact, the report lands afterwards, the settled write produces it.

**P2, silent failure and the redundant guard.** You were right that the outer handler was unreachable. Removed it; the guarantee now lives in one place, and that place distinguishes the two cases the caller cannot: a round that carried no telemetry stays quiet, a round whose telemetry was read but could not be written logs a warning. That distinction is the point -- the GPU section read zeros for 101 sessions precisely because nothing ever complained, and a silent writer would have set that up to happen again. The import moved to module scope so an import failure cannot take the harvest with it either. Two tests: one asserts the warning fires on a failed write, one asserts the no-telemetry case stays silent so the warning keeps meaning something.

**P2, min-only blocks.** Fixed. `_source` now only accepts a block offering a statistic that is actually emitted, so `{"sample_count": 27000, "power_watts": {"min": 100}}` contributes nothing rather than crediting 27000 samples to a row of nulls. Two tests: the block alone reports `samples == 0`, and beside a real block it neither contributes nor distorts the weighting.

22 tests in `test_gpu_metrics.py`, all green. I also diffed the full `inference_optimizer` failure set against clean `main` locally: identical, so nothing here introduces a new failure.

haishuok0525 (approved):
LGTM. Per-round gpu_metrics.json from benchmark_report.gpu_monitor; no merge conflicts with main.

PR #1446, opened 2026-09-08
feat(kv): collect KV cache occupancy and pressure from the engine

luochen-amd:
## Why

Nothing in Hyperloom has ever read how full the engine's KV pool is. A round can
spend its entire budget retracting requests and the breakdown reports only
"throughput low" — with the actual bottleneck invisible to the optimizer
choosing what to try next.

This matters more than it used to. On production AgentX traces the input length
median is ~62k tokens and p90 is ~339k, against a fixed 8192 on the synthetic
path, which moves KV from a background resource to the main constraint.

## What this produces

One `kv_metrics.json` per benchmark round, written into that round's workspace
next to `server.log`: phase-tagged samples, pool capacity, and the
retract/preemption counters. On an AgentX round it is joined by an
`agentx_timeline.jsonl` carrying the workload's own event stream. Both are added
to `PACKAGE_GLOBS`, since the round workspace is not otherwise reachable by the
session bundle.

Rows come from one of two collectors and `sample_source` says which. Where aiperf
is driving the workload it is already scraping the same endpoint, so its records
are used and this one stands down; everywhere else the watchdog collects.

**Breakdown integration is deliberately not in this PR.** The `telemetry.kv_cache`
aggregation is stacked on `feat/kv-metrics-sbd-section` and will follow once this
merges. This PR ships the raw artifact only.

## Design

**`/metrics`, not `server.log`.** The log cannot express the thing this is for.
Occupancy is two readings: tokens held by running requests (pressure) versus
tokens physically resident including what the prefix cache holds but would hand
back. A pool at 100% physical can be under no pressure at all — that is prefix
cache working — and only `kv_evictable_tokens`, which exists solely on
`/metrics`, separates them. On a measured saturated server the two read 0.85 and
0.91.

**Riding the existing watchdog loop, not a sampler process.** A sampler competing
for CPU would show up in the very latency numbers the round exists to measure.
The loop itself polls at 0.5s; scrapes are rate-limited on top of it to one
every 2s (`INFERENCE_OPTIMIZER_KV_SCRAPE_INTERVAL_SEC`), because a scrape is a
synchronous HTTP round trip plus a parse of the engine's whole exposition, and
the engine's gauges do not refresh anywhere near 0.5s -- most of those samples
would be the same numbers read again. That loop is also the only place that sees both the server-ready marker
and the eval-start marker, so it is the only place that can date the measured
window — and an engine process outlives that window by a wide margin. Boot
carries no traffic, warmup runs a deliberately cold cache, and the accuracy eval
drives request shapes unrelated to the benchmark. Averaged together they describe
no phase at all, so every sample carries its phase.

**AgentX needed no marker of ours.** aiperf already prints structured phase
transitions at NOTICE with millisecond stamps. Its log did have to be added to
the scan set explicitly — an AgentX workspace has no `benchmark_stderr.log`, so
those lines were simply unreachable.

**On AgentX rounds, aiperf's own scrapes are used instead.** aiperf scrapes the
engine's `/metrics` for its own purposes -- enabled by default, deriving the URL
from `--url` -- so every AgentX round was already doing it alongside this
collector. Its export is now read at close and the live poller backs off to one
scrape a minute, so the endpoint is not scraped twice to produce rows that get
discarded.

Its records are the better source besides. `SlimRecord` carries `benchmark_phase`,
stamped at collection time by the process that owns the phase transition, which
removes the boundary lag rather than correcting for it. It also carries
`request_sent_ns` and `endpoint_latency_ns`, so the scrape window is measured
rather than reconstructed.

The normalisation is not duplicated: `SlimRecord.metrics` is
`{name: [{labels, value}]}`, the same information `parse_prometheus_text`
produces, so records are rebuilt into that shape and go through the existing
`sample_from_families`. Per-rank assembly, the active/physical split and the
counter grouping are untouched, and the two sources cannot disagree about what a
number means.

Adoption replaces only the window aiperf actually covers, judged on time rather
than on the phase label. It starts after the server is up and exits before the
accuracy eval, so readings outside that span -- boot, and the whole eval phase --
are kept. The client asks for the `jsonl` format, which is the only one carrying
per-scrape records, and widens aiperf's own 333ms cadence to 2s to match.

**The watchdog path stays** because that export does not cover everything: there
is none for a synthetic benchmark, for boot, for the accuracy eval, or for a
round killed before aiperf flushed. Across 105 archived sessions and both
validation runs on crusoe, every round was synthetic.

**Phase boundaries come from aiperf's own stamps.** `phases.<name>.start_ns` is
recorded when the phase actually begins. The log line the watchdog would
otherwise grep for is written after the transition and read on the next poll, and
both lags land on the boundary -- so the samples either side of it were
systematically credited to the phase that had just ended, which is exactly where
the engine's behaviour changes most. The progress API was already being polled to
gate trace capture; it is now enabled on every round, its address published for
the watchdog to find, and polled alongside the `/metrics` scrape.

At close the rows are re-labelled from those stamps *and* the counter windows are
re-bracketed from each row's own raw per-series counters, so the phase totals move
with the boundary rather than only the labels. Adjacent phases share the reading
at their boundary, keeping the gap-free property the boundary scrapes give. A
`start_ns` that is not on the Unix epoch is rejected rather than trusted -- a
monotonic stamp would place every boundary in 1970 and sweep the round into one
phase -- and the artifact then reports `phase_source: log_markers`.

**The workload timeline, below the phase.** aiperf writes `profile_export.jsonl`
-- one record per request, with epoch-nanosecond `request_start_ns` /
`request_end_ns`, `x_correlation_id`, `conversation_id`, `turn_index` and
`x_request_id` -- at its default `--export-level records`. It has been landing in
`aiperf_artifacts/` on every AgentX round all along, next to the aggregate export
we were already reading. So the event stream needs no upstream change and no
added flag, only a reader.

At close we now emit `agentx_timeline.jsonl` beside the KV artifact:
`phase_start`/`phase_end` (from the progress API's authoritative bounds),
`trajectory_start`/`trajectory_end`, `turn_start`/`turn_end`, and
`request_start` / `first_token` / `request_end`, each carrying trajectory, turn
and request identity, time-ordered so it merges against the KV rows in one pass.
Each KV row also gets the exact set in flight during its own scrape window --
request and turn counts plus the live trajectory ids -- so "which trajectory was
running when the pool spiked" is answerable without doing the join.

`first_token` comes from the TTFT metric rather than `request_ack_ns`: the ack is
when the server accepted the request, and dating a decode-phase KV rise from it
would be wrong by the whole prefill.

Two scale guards. The correlation sweeps rather than scanning per row (a few
thousand rows against a few hundred thousand records is otherwise hundreds of
millions of comparisons in the path that finishes a round). And the request layer
is stride-sampled past a cap, since a long high-concurrency round would otherwise
outweigh every other artifact in the bundle; the phase, trajectory and turn layers
stay complete, the in-flight counts are computed from every record before any
sampling, and the summary says whether the stream was sampled.

**What this still does not do.** An exact counter total at a boundary needs the
client to quiesce, take one snapshot, and resume; aiperf exposes no such control,
so the boundary is exact to its stamp and the counter attribution to the scrape
interval either side of it.

**Nothing can fail a round.** Sampling exceptions are swallowed, an unwritable
artifact is logged at debug, and a poller that cannot reach the endpoint parks
itself rather than retrying twice a second for the whole round. Kill switch:
`INFERENCE_OPTIMIZER_KV_METRICS=0`.

**Nothing missing becomes zero.** These gauges legitimately read 0.0 — an idle
engine reports 0.0 for minutes — so absent stays `None` throughout, and poller
reachability is tri-state.

## Traps handled

Each of these produces plausible-looking wrong numbers rather than an error:

- `num_retracted_requests_total` (cumulative, 48) vs `num_retracted_reqs` (last
  batch, pid-labelled, 1). One word apart.
- Counters kept per label series, not summed flat, so an engine restart shows as
  a series resetting instead of a total going backwards.
- Hybrid pools take the max across subpool gauges, matching how SGLang's own
  scheduler judges pressure. Reading `full_token_usage` alone understates it.
- Port comes from the materialized config's `benchmark.envs.PORT`, which is a
  per-session ephemeral port, not a constant.

`sglang:cache_hit_rate` is deliberately **not** read: observed at 0.0 on a server
whose `cached_tokens_total` had already reached 4.6M, so it is not a cumulative
rate, and whether it is instantaneous or windowed is unresolved. A field nobody
can interpret is worse than an absent one.

## Watchdog changes

Kept deliberately small — it is a production loop:

- one optional duck-typed `kv_recorder`, `None` by default and then a strict
  no-op;
- one `tick` per pass, rate-limited on the monotonic clock (the slice shrinks
  below its nominal 0.5s near a deadline, so counting passes would sample
  fastest exactly when the run is most loaded), and held off until the server is
  up;
- phase notes at markers already being detected;
- the loop body wrapped in `try/finally`.

The `finally` is the part that matters. There was none, and the loop has seven
exits: five gate exceptions, the hard timeout, and the normal return. A window
that never closes is the one failure a consumer cannot recover from, because it
cannot tell a round still running from one that died mid-window.

`_scan_server_log_increment` now returns a NamedTuple instead of a 4-tuple — the
marker set grew here once and will again. Its existing test moves to attribute
access, and the truncation branch, reachable since it was written and never
covered, finally has a test.

## Two defects found by wiring it up

Both are about the scrape being a blocking call inside a watchdog, and neither
was reachable by unit tests:

- **The scrape honoured the ambient proxy.** `urlopen` reads `http_proxy`, so a
  loopback scrape on a corporate host routes through an external proxy — wrong
  by construction and slow enough to visibly stretch the loop.
- **It ran during boot.** Before the engine binds its port every attempt is a
  refused connection paid for inside the loop, during the phase when the ready
  marker and death watchdog most need it responsive. It broke a real test: a
  child talking only through its redirected stderr stopped being reported alive.
  Sampling now waits for the server, which costs nothing — boot has no traffic,
  and capacity is readable the moment the server answers. Warm-reuse rounds write
  no ready marker and are covered by `server_already_ready`.

## Prerequisite

SGLang exposes `/metrics` only with `--enable-metrics`; vLLM has it on by
default. Until that flag is set, this collects nothing and records
`available: false` — correctly, rather than reporting an idle pool. Measured
overhead of the flag is below noise (5×2 runs, the flagged arm 0.35% *higher*
than the control, against ~1.5% within-arm spread).

## Verification

Failure set on the touched suites is unchanged against `origin/main` (five
pre-existing WSL-dependent failures); 92 pass, including an end-to-end case that
runs a real round and asserts the artifact exists on disk. That last one covers
the gap this PR's final commit closes: the collector and the loop hook were both
built and tested while nothing ever constructed a recorder, so the feature
collected nothing with every test green.

Not yet validated against a live engine — the `/metrics` field names and values
come from expositions captured off MI355X nodes and used as fixtures, but
end-to-end confirmation waits on the flag above.

Refs: `kvcache-metrics-plan.md` §1.2, §3.2, §4.1, §4.3, §5.2, §6.1, §6.5, §7.4, §8.3, §8.5

### DIFF
--- a/src/hyperloom/inference_optimizer/assets/agentx/aiperf_client.sh
+++ b/src/hyperloom/inference_optimizer/assets/agentx/aiperf_client.sh
@@ -375,6 +375,11 @@ export AIPERF_DATASET_WEKA_LIVE_ASSISTANT_RESPONSES="${AGENTX_LIVE_ASSISTANT:-0}
 # measurement window then emits nothing until it ends, so a run that is merely
 # slow is indistinguishable from one that has wedged. Upstream exports it too.
 export AIPERF_UI_REALTIME_METRICS_ENABLED="${AGENTX_REALTIME_METRICS:-true}"
+
+# aiperf scrapes the engine's /metrics itself, on by default, deriving the URL from --url. Its own default cadence is
+# 333ms, which is far tighter than the KV signal needs and six times what the Hyperloom-side collector settled on;
+# widen it so the round is not paying for readings nobody reads. There is no CLI flag for this, only the env var.
+export AIPERF_SERVER_METRICS_COLLECTION_INTERVAL="${AGENTX_SERVER_METRICS_INTERVAL_S:-2.0}"
 # Content-addressed mmap cache: on a hit this skips loader + tokenizer +
 # composer entirely, turning that 4-14 min into ~0 for every run after the
 # first. Soft default -- never required, so a bare environment still works.
@@ -547,6 +552,7 @@ run_aiperf() {
     --stats-interval 30 \
     --slice-duration 1.0 \
     --no-gpu-telemetry \
+    --server-metrics-formats json csv parquet jsonl \
     ${CTX_ARGS[@]+"${CTX_ARGS[@]}"} \
     ${SMOKE_ARGS[@]+"${SMOKE_ARGS[@]}"} \
     ${AIPERF_PROGRESS_ARGS[@]+"${AIPERF_PROGRESS_ARGS[@]}"} \
--- a/src/hyperloom/inference_optimizer/assets/agentx/aiperf_client.sh
+++ b/src/hyperloom/inference_optimizer/assets/agentx/aiperf_client.sh
@@ -499,6 +499,32 @@ log "aiperf model=${SERVE_MODEL} corpus=${DS} entries=${NENT} conc=${CONC} durat
 #                        /v1/models id is more robust when a server is reused.
 #   --max-context-length omitted (see the replay-context note above).
 AIPERF_PROGRESS_ARGS=()
+AIPERF_PROGRESS_PORT=""
+PHASE_GATE="${BENCH_DIR}/aiperf_phase_gate.py"
+
+# Enable AIPerf's progress API on every round, not just trace-capture ones. It
+# is the only authoritative source of phase timing: `phases.<name>.start_ns` is
+# stamped by aiperf when the phase actually begins, whereas the human-readable
+# log line the KV collector otherwise greps for is written afterwards and read
+# on the next watchdog poll, so warmup traffic lands in the measured window.
+#
+# The endpoint costs nothing when nobody polls it, and the address is published
+# as a file because the reader -- Hyperloom's watchdog -- is a different process
+# that only sees this round's result directory.
+if [ -f "$PHASE_GATE" ]; then
+  if AIPERF_PROGRESS_PORT="$(python3 "$PHASE_GATE" pick-port)"; then
+    AIPERF_PROGRESS_ARGS=(--api-host 127.0.0.1 --api-port "$AIPERF_PROGRESS_PORT")
+    printf '{"url": "http://127.0.0.1:%s", "schema_version": 1}\n' \
+      "$AIPERF_PROGRESS_PORT" > "${ART}/progress_api.json" 2>/dev/null \
+      || log "WARN could not publish the AIPerf progress API address"
+  else
+    AIPERF_PROGRESS_PORT=""
+    log "WARN failed to allocate an AIPerf progress API port; phase timing falls back to log markers"
+  fi
+else
+  log "WARN missing AIPerf phase gate ${PHASE_GATE}; phase timing falls back to log markers"
+fi
+
 run_aiperf() {
   "${AIPERF_CMD[@]}" profile \
     --scenario inferencex-agentx-mvp \
@@ -540,12 +566,10 @@ if [ "${PROFILE:-0}" = "1" ]; then
   _require_uint AGENTX_PROFILE_WINDOW_S "$PWIN"
   : "${AGENTX_CAPTURE_ID:?AGENTX_CAPTURE_ID required for AgentX profiling}"
   : "${AGENTX_CAPTURE_STATUS_PATH:?AGENTX_CAPTURE_STATUS_PATH required for AgentX profiling}"
-  PHASE_GATE="${BENCH_DIR}/aiperf_phase_gate.py"
   PHASE_WAIT_TIMEOUT="${AGENTX_PHASE_WAIT_TIMEOUT_S:-$(( DATASET_CONFIG_TIMEOUT + WARMGRACE + DURATION ))}"
   _require_uint AGENTX_PHASE_WAIT_TIMEOUT_S "$PHASE_WAIT_TIMEOUT"
   CAPTURE_STATUS_FILE="$AGENTX_CAPTURE_STATUS_PATH"
   rm -f "$CAPTURE_STATUS_FILE"
-  AIPERF_PROGRESS_PORT=""
   PHASE_GATE_FAILURE_REASON="profiling_phase_unavailable"
   _write_profile_capture_status() {
     _status="$1"
@@ -570,15 +594,13 @@ if [ "${PROFILE:-0}" = "1" ]; then
   if [ -n "${AGENTX_PROFILE_WARMUP_S:-}" ]; then
     log "WARN AGENTX_PROFILE_WARMUP_S is ignored: profiling now starts from AIPerf's measured-phase signal"
   fi
-  if [ -f "$PHASE_GATE" ]; then
-    if AIPERF_PROGRESS_PORT="$(python3 "$PHASE_GATE" pick-port)"; then
-      AIPERF_PROGRESS_ARGS=(--api-host 127.0.0.1 --api-port "$AIPERF_PROGRESS_PORT")
-    else
-      PHASE_GATE_FAILURE_REASON="api_port_allocation_failed"
-      log "WARN failed to allocate an AIPerf progress API port; the measurement will run without trace capture"
-    fi
-  else
+  # The progress API is brought up for every round further above; capture only
+  # needs to know whether that succeeded, since it cannot gate a window without it.
+  if [ ! -f "$PHASE_GATE" ]; then
     log "WARN missing AIPerf phase gate ${PHASE_GATE}; the measurement will run without trace capture"
+  elif [ -z "$AIPERF_PROGRESS_PORT" ]; then
+    PHASE_GATE_FAILURE_REASON="api_port_allocation_failed"
+    log "WARN no AIPerf progress API port; the measurement will run without trace capture"
   fi
   # A 200 OK from /stop_profile means the tracer was TOLD to stop, not that the
   # trace is on disk. MEASURED on GLM-5.3 (sglang, TP=8, one 20s window): the
--- a/src/hyperloom/inference_optimizer/breakdown/session_package.py
+++ b/src/hyperloom/inference_optimizer/breakdown/session_package.py
@@ -70,6 +70,7 @@ PACKAGE_GLOBS: tuple[str, ...] = (
     "reports/kernel_roofline.json",
     "reports/conc_sweep_summary.json",
     "runs/**/kv_metrics.json",
+    "runs/**/agentx_timeline.jsonl",
     "reports/sbd_v6/timeline/*.json",
     "reports/sbd_v6/write_warnings.jsonl",
     "reports/trace/*.jsonl",
--- a/src/hyperloom/inference_optimizer/breakdown/session_package.py
+++ b/src/hyperloom/inference_optimizer/breakdown/session_package.py
@@ -69,6 +69,7 @@ PACKAGE_GLOBS: tuple[str, ...] = (
     "reports/kernel_optimization_summary.json",
     "reports/kernel_roofline.json",
     "reports/conc_sweep_summary.json",
+    "runs/**/kv_metrics.json",
     "reports/sbd_v6/timeline/*.json",
     "reports/sbd_v6/write_warnings.jsonl",
     "reports/trace/*.jsonl",
--- a/src/hyperloom/inference_optimizer/tests/test_agentx_timeline.py
+++ b/src/hyperloom/inference_optimizer/tests/test_agentx_timeline.py
@@ -228,8 +228,14 @@ def test_request_events_are_sampled_but_the_stream_says_so(tmp_path, monkeypatch
     assert len([e for e in events if e["event"] == "turn_start"]) == 10
 
 
-def test_rows_learn_which_trajectories_were_in_flight(tmp_path):
-    """The question this exists for: which trajectory was live when the pool spiked."""
+def test_rows_learn_how_much_was_in_flight(tmp_path):
+    """Counts on the row; the identities are the timeline's job.
+
+    Repeating the ids on every row copied the same set dozens of times over -- a
+    trajectory lasts minutes, rows are seconds apart -- and made 65% of a 4.6 MB
+    artifact that ships in the session bundle. The spans in the timeline answer
+    the same question exactly, and without a truncation cap.
+    """
     records = parse_profile_export(
         _export(
             tmp_path,
@@ -248,11 +254,33 @@ def test_rows_learn_which_trajectories_were_in_flight(tmp_path):
     annotated = correlate_rows(rows, records)
 
     assert annotated == 2
-    assert rows[0]["workload"]["in_flight"]["requests"] == 2
-    assert rows[0]["workload"]["in_flight"]["trajectory_ids"] == ["traj-1", "traj-2"]
-    assert rows[1]["workload"]["in_flight"]["trajectory_ids"] == ["traj-3"]
+    assert rows[0]["workload"]["in_flight"] == {"requests": 2, "trajectories": 2, "turns": 2}
+    assert rows[1]["workload"]["in_flight"] == {"requests": 1, "trajectories": 1, "turns": 1}
     # A window with nothing running is left alone rather than annotated with zeros.
     assert "workload" not in rows[2]
+
+
+def test_the_timeline_still_names_the_trajectories_a_row_counted(tmp_path):
+    """The join the row no longer duplicates: spans overlapping its scrape window."""
+    records = parse_profile_export(
+        _export(
+            tmp_path,
+            [
+                _record(request_id="r1", correlation="traj-1", turn=0, start_offset_ms=0, duration_ms=1000),
+                _record(request_id="r2", correlation="traj-2", turn=0, start_offset_ms=500, duration_ms=1000),
+            ],
+        )
+    )
+    row = {"scrape_start_unix": BASE + 0.6, "scrape_end_unix": BASE + 0.7}
+    correlate_rows([row], records)
+    events, _ = build_events(records)
+
+    live = {
+        e["trajectory_id"] for e in events if e["event"] == "trajectory_start" and e["ts"] <= row["scrape_end_unix"]
+    } & {e["trajectory_id"] for e in events if e["event"] == "trajectory_end" and e["ts"] >= row["scrape_start_unix"]}
+
+    assert live == {"traj-1", "traj-2"}
+    assert row["workload"]["in_flight"]["trajectories"] == len(live)
 
 
 def test_correlation_counts_a_request_that_spans_the_whole_window(tmp_path):
--- a/src/hyperloom/inference_optimizer/tests/test_agentx_timeline.py
+++ b/src/hyperloom/inference_optimizer/tests/test_agentx_timeline.py
@@ -1,0 +1,301 @@
+# SPDX-FileCopyrightText: 2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Tests for the AgentX workload timeline built from aiperf's per-request export.
+
+The record shape below is aiperf's own documented one (``docs/tutorials/working-with-profile-exports.md`` at the pinned
+commit), not one invented here: getting the field names wrong is the whole risk in reading somebody else's artifact.
+"""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+from hyperloom.orchestrator.actions.executors._agentx_timeline import (
+    TIMELINE_ARTIFACT_NAME,
+    build_events,
+    correlate_rows,
+    find_profile_export,
+    parse_profile_export,
+    write_timeline,
+)
+
+BASE_NS = 1_759_813_207_000_000_000
+BASE = BASE_NS / 1e9
+
+
+def _record(
+    *,
+    request_id: str,
+    correlation: str,
+    turn: int,
+    start_offset_ms: float,
+    duration_ms: float,
+    ttft_ms: float | None = 250.0,
+    conversation: str = "conv-1",
+    error: dict | None = None,
+) -> dict:
+    """One line of aiperf's ``profile_export.jsonl``, in its documented shape."""
+    start_ns = BASE_NS + int(start_offset_ms * 1e6)
+    metrics: dict = {"input_sequence_length": {"value": 550, "unit": "tokens"}}
+    if ttft_ms is not None:
+        metrics["time_to_first_token"] = {"value": ttft_ms, "unit": "ms"}
+    return {
+        "metadata": {
+            "session_num": 45,
+            "x_request_id": request_id,
+            "x_correlation_id": correlation,
+            "conversation_id": conversation,
+            "turn_index": turn,
+            "request_start_ns": start_ns,
+            "request_ack_ns": start_ns + 100_000_000,
+            "request_end_ns": start_ns + int(duration_ms * 1e6),
+            "worker_id": "worker_359d423a",
+            "record_processor_id": "record_processor_1fa47cd7",
+            "benchmark_phase": "profiling",
+            "was_cancelled": False,
+            "cancellation_time_ns": None,
+        },
+        "metrics": metrics,
+        "error": error,
+    }
+
+
+def _export(tmp_path: Path, records: list[dict], *, nested: bool = False) -> Path:
+    """Write an export where aiperf would put it."""
+    art = tmp_path / ("benchmark_agentx_1" if nested else ".") / "aiperf_artifacts"
+    art.mkdir(parents=True, exist_ok=True)
+    path = art / "profile_export.jsonl"
+    path.write_text("".join(json.dumps(r) + "\n" for r in records), encoding="utf-8")
+    return path
+
+
+def test_export_is_found_beside_the_round(tmp_path):
+    """The same resolution the progress address and the aiperf log use."""
+    _export(tmp_path, [_record(request_id="r1", correlation="t1", turn=0, start_offset_ms=0, duration_ms=500)])
+    assert find_profile_export(tmp_path) is not None
+
+
+def test_export_is_found_in_a_nested_benchmark_directory(tmp_path):
+    """An AgentX round nests its benchmark one level below the server log."""
+    _export(
+        tmp_path,
+        [_record(request_id="r1", correlation="t1", turn=0, start_offset_ms=0, duration_ms=500)],
+        nested=True,
+    )
+    assert find_profile_export(tmp_path) is not None
+
+
+def test_absent_export_is_not_an_error(tmp_path):
+    """Every synthetic (non-AgentX) round is this case."""
+    assert find_profile_export(tmp_path) is None
+
+
+def test_records_parse_with_the_identifiers_the_hierarchy_needs(tmp_path):
+    """Trajectory, turn and request identity all come off one record."""
+    path = _export(
+        tmp_path,
+        [_record(request_id="req-a", correlation="traj-1", turn=2, start_offset_ms=1000, duration_ms=400)],
+    )
+    (record,) = parse_profile_export(path)
+
+    assert record.request_id == "req-a"
+    assert record.trajectory_id == "traj-1"
+    assert record.conversation_id == "conv-1"
+    assert record.turn_index == 2
+    assert record.start == BASE + 1.0
+    assert record.end == BASE + 1.4
+    # Derived from TTFT, not from the ack: the ack is acceptance, not the first token.
+    assert record.first_token == BASE + 1.25
+
+
+def test_a_truncated_last_line_does_not_cost_the_rest(tmp_path):
+    """The export is written as the round ends; a partial final line is normal."""
+    path = _export(tmp_path, [_record(request_id="r1", correlation="t1", turn=0, start_offset_ms=0, duration_ms=100)])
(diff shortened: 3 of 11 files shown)


haishuok0525:
I think the sampling and phase-accounting design should be adjusted before this is used for AgentX measurements.

The fixed 0.5-second scrape interval is unnecessarily aggressive. The engine may not refresh these metrics at the same frequency, and serving plus parsing `/metrics` can itself take over 100 ms, so this can add measurable overhead while producing duplicate observations. I suggest making the interval configurable and using a 2–3 second default.

Periodic scrapes should still capture both gauges and cumulative counters. Each stored sample should retain:

- wall and monotonic timestamps, scrape start/end timestamps, and scrape duration;
- KV usage/capacity gauges;
- raw per-series retract, preemption, and prefix-cache counters.

Keeping the raw counter series in every periodic sample allows consumers to derive interval deltas, identify spikes, and detect series resets or engine restarts. Only storing phase-level first/last counters would lose that trend information.

Phase accounting should be driven by explicit workload boundaries rather than by the time the watchdog happens to observe human-readable log markers. AgentX/AIPerf should emit a timestamped event timeline for phase, trajectory, turn, and request start/end events. This does not need to attribute global KV usage to individual concurrent requests; it only provides the workload timeline against which the engine-wide KV series can be aligned.

For phase-level counter totals, phase transitions should use a synchronous boundary handshake:

1. Complete the previous phase's final request.
2. Emit a transition event and pause before starting the next phase.
3. Force one `/metrics` snapshot, bypassing the 2–3 second periodic interval.
4. Use the same counter snapshot as both the previous phase's endpoint and the next phase's baseline.
5. Acknowledge the transition and allow the next phase to start.

A final forced boundary snapshot should be taken before closing the recorder. This produces gap-free `phase_delta = end_boundary - start_boundary` values while the periodic samples preserve the within-phase gauge and counter trends.

The existing AIPerf progress API already exposes phase timing (`start_ns` and `requests_end_ns`) and is a better authority than scanning `aiperf.log`. However, observing that API after a phase has started is still not an exact counter boundary; exact phase totals require the synchronous boundary hook above. The resulting supported granularity would be exact warmup/measured/eval boundaries plus a 2–3 second engine-wide KV/counter timeline correlated with the active trajectory/turn/request set, without attempting per-request KV attribution.

luochen-amd:
Phase attribution is now driven by aiperf's own stamps rather than by log markers, and the counter totals move with them. Two of your points need aiperf changes we do not control; details at the end.

**Authoritative boundaries.** `phases.<name>.start_ns` / `requests_end_ns` are now the source. We were already polling that endpoint to gate trace capture, but only under `PROFILE=1`, so ordinary measured rounds never had it. It is now enabled on every round, its address published to `aiperf_artifacts/progress_api.json` (the watchdog is a different process and only sees the round's result dir), and polled alongside the existing `/metrics` scrape.

**The counter fix, not just the labels.** You were right that boundary scrapes alone only closed the storage-layer gap. At close, rows are re-labelled from the stamps *and* the counter windows are re-bracketed from each row's own raw per-series counters -- which is what those raw maps in every sample are for. Adjacent phases share the reading at their boundary, so this keeps the gap-free property rather than trading it for accuracy. There is a test asserting the retract that happened after the real transition is booked to `measured`, not to the warmup window the watchdog had not yet closed.

A `start_ns` that is not on the Unix epoch is rejected rather than trusted: the API reports nanoseconds without naming an epoch, and a monotonic stamp would put every boundary in 1970 and sweep the whole round into one phase. The artifact then falls back and says so, via `phase_source: log_markers`.

**Per-sample timestamps.** Rows now carry `scrape_start_unix`, `scrape_end_unix`, `scrape_start_mono`, `scrape_end_mono` alongside `scrape_sec`. A gauge read across a 300 ms round trip describes some instant inside that window; a single stamp would invite aligning KV against a workload timeline to a precision nobody measured.

**Workload correlation.** Each row records the aiperf phase in flight and that phase's own counters at scrape time.

**PR description** updated -- it said 0.5s sampling, which has been 2s since the interval was made configurable.

**Two things I did not implement, because they are not ours to implement.**

1. *Synchronous boundary handshake.* Steps 1, 2 and 5 (finish the last request, pause, acknowledge, resume) are aiperf-side controls. Nothing in our integration can quiesce its request generation -- `aiperf_client.sh` can only launch, observe and post-process it, and aiperf is pip-installed from `SemiAnalysisAI/aiperf` at a pinned commit. Steps 3 and 4 (force a snapshot, reuse it as both endpoints) are done and were already in the previous round. So the boundary is now exact to aiperf's own stamp, with counter attribution accurate to the scrape interval either side of it -- better than log markers, short of exact.

2. *Trajectory / turn / request event timeline.* The progress API reports per phase; nothing we can reach exposes per-trajectory/turn/request events, and `profile_export_aiperf.json` is aggregate metrics with no event stream. Whether aiperf could expose them is a question for that repo -- I could not confirm it from here, since it is not vendored.

If you want either, the shape of the upstream ask is now concrete and I am happy to file it against aiperf. Both are also cleanly separable from this PR: the artifact already records `phase_source`, so a consumer can tell which authority produced any given round's numbers.

luochen-amd:
You are right, and my previous answer was wrong on the key point. This is now implemented -- no upstream change was needed.

**The data was already on disk.** I had checked the progress API and `profile_export_aiperf.json` and concluded per-request data did not exist. I had not checked what else aiperf writes into the artifact dir. It writes `profile_export.jsonl` -- one record per request -- at its default `--export-level records` (`OutputDefaults.EXPORT_LEVEL = ExportLevel.RECORDS`), which is exactly the "CLI default". Every AgentX round has been producing one all along, right next to the aggregate export we were already reading; we simply never opened it. Each record carries:

```json
{"metadata": {"x_request_id": "...", "x_correlation_id": "...", "conversation_id": "...",
              "turn_index": 0,
              "request_start_ns": 1759813207532900363,
              "request_ack_ns":   1759813207650730976,
              "request_end_ns":   1759813207838764604,
              "benchmark_phase": "profiling", "was_cancelled": false},
 "metrics": {"time_to_first_token": {"value": 255.88, "unit": "ms"}, ...},
 "error": null}
```

Epoch nanoseconds -- the same clock the progress API stamps phases on -- plus every identifier the hierarchy needs. So no aiperf change, no added flag.

**What now ships.** `agentx_timeline.jsonl`, beside `kv_metrics.json` and added to `PACKAGE_GLOBS`, with exactly the event set you specified:

- `phase_start` / `phase_end` -- from the progress API's authoritative bounds, not from the records (their `benchmark_phase` reads `profiling` for everything aiperf exports)
- `trajectory_start` / `trajectory_end`
- `turn_start` / `turn_end`
- `request_start` / `first_token` / `request_end`

Each event carries `trajectory_id` / `turn_index` / `request_id` and a timestamp, and the stream is time-ordered so it merges against the KV rows in one pass. Trajectory identity is `x_correlation_id` -- the live session, constant across a multi-turn conversation -- with `conversation_id` carried alongside, since that names the dataset entry being replayed and can recur across trajectories.

`first_token` is derived from the TTFT metric, not from `request_ack_ns`. The ack is when the server accepted the request; dating a decode-phase KV rise from it would be wrong by the whole prefill.

**The correlation is also folded into the rows,** so the common questions do not require the join: each KV row now carries the exact set in flight during its own scrape window -- request and turn counts, plus the live trajectory ids. Your three examples are answerable:

- *which trajectory was live when KV spiked* -- `samples[i].workload.in_flight.trajectory_ids`
- *which turn a retract happened during* -- rows carry raw per-series retract counters, so the interval that incremented is identifiable and its `in_flight` names the turns
- *which requests were running in a sampling window* -- the row gives the counts and trajectories, the timeline gives the individual requests over `[scrape_start_unix, scrape_end_unix]`

**Two scale guards,** both stated in the artifact rather than left implicit. The correlation is a sweep, not a scan per row: a few thousand rows against a few hundred thousand records would otherwise be hundreds of millions of comparisons inside the path that finishes a benchmark round. And the request layer is stride-sampled past a cap, because a three-hour high-concurrency round is well over 100k requests and three events each would outweigh every other artifact in the session bundle combined. The phase, trajectory and turn layers are always complete, the per-row in-flight counts are computed from every record before any sampling, and the summary reports `request_events_complete` / `request_event_stride` so a reader counting `request_start` events can tell a sampled stream from a whole one.

16 new tests, with the record fixture taken verbatim from aiperf's documented shape rather than invented.

**Still outstanding, and still genuinely upstream:** the synchronous boundary handshake. Steps 3 and 4 (force a snapshot, use it as both endpoints) are done; steps 1, 2 and 5 need aiperf to quiesce and resume its request generation, and it exposes no such control. So phase boundaries are exact to aiperf's own stamp, with counter attribution accurate to the scrape interval either side. I will not claim that one is unavailable without checking the source this time -- I looked, and there is no such control at the pinned commit.

luochen-amd:
Ran this on a real crusoe session (Qwen3-0.6B, vLLM, single node) and it found three defects that no unit test could have. Fixed in the latest push; the evidence is below because it changes what the artifact means.

**The session collected nothing on two of its three rounds, including the measured baseline** -- the only phase allowed into a comparison.

| round | result | why |
| --- | --- | --- |
| warmup | `available: false`, 0 samples | scraped :8888, server bound :34407 |
| **measure** | `available: null`, 0 samples | warm reuse: no log, so the tick gate never opened |
| roofline | `available: true`, **812 samples** | bound :8888 by coincidence, matching the default |

**1. The port was wrong.** It is pinned in the round's materialized `benchmark.envs.PORT`, and that is never exported into the subprocess environment, so `resolve_metrics_port` fell through to the 8888 default. The warmup and measure rounds pinned `PORT: 34407`; the roofline round pinned nothing and bound 8888, which is the only reason one round worked. The docstring already claimed the config was the authoritative source -- it just never received it. Now the round's own YAML is read; verified against the exact production file that caused this (resolves 34407).

**2. Giving up needed a clock, not just a count.** Three consecutive misses parked the poller permanently, which at a 2s interval is about six seconds -- an engine that has logged `Application startup complete` can still be inside that. Now parking also requires a minute of continuous failure, so it means "nothing answered here" rather than "the first few tries missed".

**3. A warm-reuse round never ticked.** The measure round re-attaches to the server the warmup round booted (`launch_evidence.actual_server_log_path` is empty, and the round has no `server_ready_at`). It owns no log, so no ready marker ever lands in its scan set -- and the tick was gated on exactly that marker. It produced a well-formed artifact containing zero samples, which is the worst possible failure: it looks like a round with no KV pressure. Now a round that still has no resolvable server log well after it started is treated as re-attached, and collects.

**What the one working round confirms about the rest of the design:**

- 812 samples, gap between rows `min 2.01s / p50 2.02s / max 305.7s` -- the configurable 2s interval behaves exactly as specified (the max is a long stall in the round, not a sampler problem).
- `scrape_sec` `min 4.6ms / p50 5.5ms / max 36ms`, comfortably under the 400ms timeout -- the scrape is not costing the benchmark anything measurable.
- **0 inverted scrape windows** across 812 rows, confirming the enclosing-millisecond fix from the previous round of review.
- Phase attribution behaved: the boundary reading is tagged with the phase that is ending (1 `boot` row at the exact instant of the `measured` mark), then 811 `measured`.
- `phase_source: log_markers` and `workload_timeline: null` throughout, which is correct -- this is a synthetic vLLM run, so there is no aiperf progress API and no `profile_export.jsonl`. The AgentX paths degrade the way they are supposed to; they still need an AgentX round to be exercised end to end.

**One inherent limitation this run made concrete:** on vLLM, `capacity_tokens` / `used_tokens` / `available_tokens` all come back null and only `active_pool_usage` (0.0–0.0505 here) and the preemption counter are populated. vLLM exposes an occupancy ratio, not token counts, so there is nothing to derive a capacity from. That is the engine's exposition, not a collection bug, but it means the token-level fields are SGLang-only in practice.

Also dropped `agentx_timeline.jsonl` to `0o600` to match the KV artifact beside it -- `atomic_write_text` masks everything it writes to `0o700` deliberately, and a sibling from the same collector should not be the one world-readable file in the directory.

luochen-amd:
Adopted. I checked every load-bearing claim against aiperf at the pinned commit (`754356e9`) first, and the core insight holds -- including one thing I had previously got wrong.

**The normalisation reuse works exactly as §6 describes.** `SlimRecord.metrics` is `{name: [MetricSample(labels, value)]}`, and `MetricSample.labels` is a real `dict[str, str]`. That is isomorphic to what `parse_prometheus_text` produces, so the records rebuild into the same internal shape and go straight through the existing `sample_from_families`. Per-rank assembly, the active/physical split and the counter grouping are untouched, and there is no second normalisation to keep in sync.

**It is better than the proposal argues, on the point the review has been circling for three rounds.** `SlimRecord` carries `benchmark_phase`, and `ServerMetricsManager` sets it on `CREDIT_PHASE_START` -- so the phase is stamped at collection time by the process that owns the transition. That does not correct the boundary lag, it removes it. It also carries `request_sent_ns` and `endpoint_latency_ns`, so the scrape window is measured rather than reconstructed. The `progress_api` re-attribution is now only needed on the non-aiperf path.

**I was wrong in my earlier assessment that aiperf scrapes nothing here.** I had read `ServerMetricsConfig.urls` defaulting to an empty list and `ServerMetricsDiscoveryMode.AUTO` being Kubernetes-only, and concluded it was inert on Slurm. The CLI docs say otherwise: `--server-metrics` is *enabled by default* and "automatically collects from inference endpoint `base_url` + `/metrics`" -- `urls` is for *additional* endpoints. So it has been scraping our engine on every AgentX round all along, and the duplicate-load argument is real. Your §2.1 was right and my objection was not.

**Three corrections to the proposal's mechanics, one of which would have broken every AgentX round:**

1. There is no `--server-metrics-collection-interval` CLI flag -- only `--server-metrics`, `--no-server-metrics` and `--server-metrics-formats`. I had written the flag in before checking; aiperf would have rejected the command line and failed the round. The interval is env-only, `AIPERF_SERVER_METRICS_COLLECTION_INTERVAL`.
2. Passing `--server-metrics <our own url>` is unnecessary and risks scraping the same endpoint twice, since the base URL is already collected automatically. Only `--server-metrics-formats ... jsonl` is needed.
3. aiperf's default cadence is **333ms**, which is 6x the 2s this collector settled on after the earlier review round said 0.5s was too aggressive. Adopting it as-is would have quietly raised the sampling rate well past what was rejected. I set the interval env to 2s.

**What I did not do: remove the watchdog path.** The export does not exist for a synthetic benchmark, for the accuracy eval (aiperf has already exited), for boot, or for a round killed before aiperf flushed. Concretely: across 105 archived session breakdowns there are **zero** rounds with any aiperf artifact -- 9,964 benchmark directories, all Magpie synthetic -- and both of this week's validation runs on crusoe were synthetic vLLM too. (Caveat: those archives are 2026-08-18..21, so they predate the AgentX push; they show what has dominated historically, not necessarily what will.) §5.4 leaves this as a product decision and I took the thin-fallback option, because removing it does not shrink the PR -- it keeps both ingestion paths and drops coverage on every round observed so far.

**Shipped** (`9a3aa6f1e`):

- `read_aiperf_server_metrics()` + `families_from_aiperf_record()`; on a round with an export, aiperf's rows *replace* the live ones (not merged -- two collectors on one endpoint produce interleaved readings of the same counters, and a window bracketed across both would credit an increment to whichever was sampled either side of it)
- phase taken verbatim from `benchmark_phase`; no re-attribution on that path
- adjacent phases still share the boundary reading: aiperf takes its own scrape at each transition but tags it with the phase that is *ending*, so the windows are rebuilt from the rows the same way the re-attribution path does
- `sample_source` in the artifact, so a consumer comparing rounds can tell the two collectors apart
- `aiperf_client.sh`: `--server-metrics-formats json csv parquet jsonl` and the interval env
- 6 new tests using the documented record shape

Open items on your §11, with my answers: AgentX post-process only -- **yes**. Synthetic without KV -- **no**, thin fallback kept. Eval without aiperf KV -- **no**, the watchdog covers it. Force jsonl -- **yes**. Breakdown stays a follow-up -- **yes**, `kv_metrics.json` schema is unchanged apart from the two additive fields.

Still unverified end to end: none of this has run on an actual AgentX round. Both crusoe validations were synthetic, so the aiperf path has never executed against a real export. Worth one AgentX run before merge -- check `sample_source` in the artifact to confirm which collector actually produced the rows.

luochen-amd:
Following up on the refactor proposal with an explicit account of where I did **not** follow it, so it is a decision rather than an omission.

| §10 step | proposal | state |
| --- | --- | --- |
| 1. keep parser / timeline / progress API | keep | done |
| 2. add jsonl ingestion | add | done |
| **3. remove the watchdog recorder wiring, post-process only** | **remove** | **not done** |
| 4. update the PR description | update | done |
| **5. delete the watchdog e2e tests** | **delete** | **not done** |

Nothing on the §5.1 deletion list was removed either: the recorder is still wired into the watchdog at 8 call sites, `tick()` / `note_phase()` still take boundary scrapes, `AiperfProgressPoller` is still polled inside the loop, and `test_kv_metrics_wiring.py` has 73 tests rather than fewer.

**Why.** The export only exists where aiperf runs, and aiperf runs on AgentX rounds only. Across 105 archived session breakdowns there are zero rounds carrying any aiperf artifact -- 9,964 benchmark directories, all Magpie synthetic -- and both of this week's crusoe validation runs were synthetic vLLM as well. (Those archives are 2026-08-18..21, so they predate the AgentX push; they show what has dominated historically, not necessarily what will.) Beyond that, aiperf starts after the server is up and exits before the accuracy eval, so even on an AgentX round it cannot cover boot, cannot cover eval, and produces nothing for a round killed before it flushes. Removing the watchdog path would drop collection on all of those, and §5.4 leaves that as a product decision with a fallback permitted.

**What I did instead, for the goal behind step 3.** The duplicate `/metrics` load was real and I had made it worse: the first cut adopted aiperf's records at close while still polling at the full rate, so an AgentX round scraped twice and discarded one set. Fixed in `3bed99c55` -- the poller backs off to one scrape a minute once the round's `aiperf_artifacts` directory appears, and resumes the full rate at eval, when aiperf has exited. Adoption now replaces only the window aiperf actually covers, judged on time rather than on the phase label, so boot and eval readings survive. So the duplicate-load goal is met; the wiring removal is not.

**Where that leaves it.** I am treating "keep a fallback" as the §5.4 option rather than as a rejection of the proposal, but it is a deviation on the two steps that shrink the PR, and it is yours to overrule. If you would rather have the full refactor, the honest cost is: synthetic rounds, boot and eval lose KV entirely, a round that dies before aiperf flushes produces nothing, and the 73 wiring tests are replaced by jsonl-fixture tests. I would rather do that deliberately than have it land by attrition -- say which and I will make it match.

Two caveats that apply either way.

**None of the aiperf path has run end to end.** Both validation runs were synthetic, so `read_aiperf_server_metrics` has never been fed a real export and the fixtures behind it are written from aiperf's documentation rather than from a real file. Worth weighing against this PR's record: the port bug, the warm-reuse bug and the give-up bug all had passing unit tests and still left two of three rounds with zero samples on the first real run, and the scrape-window rounding bug only appeared on CI. Green tests have not been predictive here. It needs one AgentX round before merge -- `sample_source` in the artifact will say which collector actually produced the rows.

**Enabling `jsonl` has an unmeasured cost.** aiperf excludes that format by default, explicitly "to avoid large files", and §8 of the proposal requires it because the aggregated JSON cannot support per-phase counter boundaries. A long round writes one record per scrape carrying the whole exposition, so the file could be substantial in the round workspace. Widening aiperf's cadence from its 333ms default to 2s cuts it roughly sixfold and it is deliberately not in `PACKAGE_GLOBS`, but nobody has measured it on a real round yet.

luochen-amd:
## Verification status

Six AgentX rounds on crusoe, all Llama-3.1-8B-Instruct / vLLM / TP=1. Every defect below was found by a real round, not by the test suite -- which is why the unverified list at the end is worth taking seriously rather than treating as a formality.

### Found and fixed by those runs

| # | Defect | How it presented |
| --- | --- | --- |
| 1 | Counters spelled the exposition way, but aiperf names them by Prometheus *family* | `preempt_total` null on all 1656 rows; `retract_delta`/`preempt_delta` null. Of 47 families in a real record exactly one ends in `_total`, and that one is a histogram |
| 2 | Port resolved from config/default only | vLLM bound `:8000`, collection sat on `:8888`, `available: false` for the whole round |
| 3 | Bind-line pattern matched `Uvicorn running on` | These builds print `Starting vLLM server on http://0.0.0.0:PORT`; the fallback never fired |
| 4 | Unstamped aiperf records filed as `measured` | aiperf's pre-phase baseline captures landed in the only phase allowed into a comparison, producing a measured prefix-cache delta of 34,394 describing nothing |
| 5 | A `profiling` stamp arriving *before* warmup | Setup leak, seen on 3 of 6 rounds; same poisoning of `measured` |
| 6 | `available` described the watchdog, not the round | `available: false` on an artifact carrying 1656 aiperf samples; the breakdown folds that into "no round reached the endpoint" |
| 7 | Duplicate scraping | Adopting aiperf's records while still polling at full rate -- worse on the axis the refactor existed to improve |
| 8 | Per-row trajectory ids | 1280 bytes a row, 65% of a 4.65 MB artifact that ships in the session bundle |

### Verified on real hardware

| Behaviour | Evidence |
| --- | --- |
| aiperf export adopted | `sample_source: aiperf_server_metrics`, `phase_source: aiperf_server_metrics` |
| Port resolution | `url: http://127.0.0.1:8000/metrics` (was `:8888`) |
| Availability semantics | `available: true` on a round whose own poller never reached the engine |
| Scrape cadence | gap p50 **2.007s**; `AIPERF_SERVER_METRICS_COLLECTION_INTERVAL=2.0` overrides aiperf's 333ms default |
| Phase attribution | `{boot: 3, warmup: 1637}` on an aborted round; `{boot: 2, warmup: 3, measured: 3}` on one that reached profiling |
| Merge by time | 1638 aiperf records -> 1640 rows: the two live readings outside aiperf's window are kept, the ones inside are replaced |
| Counters resolve | 1541/1541 rows carry `preempt_total` after the family alias |
| Artifact size | 4.65 MB / 3017 B per row -> **1.82 MB / 1107 B per row** |
| Workload timeline | 2228 events, 111 trajectories, 401 turns/requests, 1632 of 1640 rows correlated |
| Aborted rounds | five rounds killed externally, each still wrote a complete artifact with `aborted: true` |
| **Measured phase under load** | occupancy to **0.9458**; measured prefix delta 2,150,935 queries / 335,280 hits (**15.6%**) against warmup's 14,928,610 / 1,167,584 (**7.8%**) -- cache warming visible across the boundary |

Cost, measured: `server_metrics_export.jsonl` runs ~20 MB/hour at 2s (11.9 KB a record). It is not in `PACKAGE_GLOBS`. `kv_metrics.json` and `agentx_timeline.jsonl` are.

### Not verified

| Gap | Why it matters |
| --- | --- |
| **SGLang, at all** | Every round was vLLM, which is the thin engine here: it has no token counts, no capacity, and no `kv_evictable_tokens` -- so **the active-vs-physical occupancy distinction this PR is built on has never been exercised against real data**. Nine SGLang constants are unexercised, including `num_retracted_requests_total`. Given defect #1, aiperf will present that as `sglang:num_retracted_requests`; the alias should cover it, but "should" has been wrong four times this week, and a miss presents as `retract_delta: null` -- indistinguishable from a round that never retracted |
| **TP > 1** | `series_count: 1` on every round. The per-rank assembly -- the blocking finding from an earlier review, where mixing fields across ranks produced `physical_pool_usage` above 1.0 -- has never seen a second series |
| A round closing normally | All six were killed; `aborted: false` is one flag away but unobserved |
| The eval phase | No round reached the accuracy eval, so the resume-to-full-rate path and the keeping of live eval rows are untested in production |

### Where that leaves it

The vLLM/AgentX path is well covered and the measured-phase numbers are sound. One SGLang round would close the largest gap in a single shot -- same procedure, different image, plus `--server-args "--enable-metrics"` -- and would exercise retract, evictable tokens, capacity, the hybrid pool, and the aiperf renaming on SGLang's own names.

Merging before that is defensible: collection is strictly observational, every failure path is swallowed, `INFERENCE_OPTIMIZER_KV_METRICS=0` disables it, and the artifact self-describes through `sample_source` / `available` / `phase_source`. But defect #1 is the counter-example to "unverified paths just degrade to null": a naming miss reads as a clean round, which is the one failure mode this artifact must not have. I would rather that gap be a decision than an assumption.

Two items outstanding outside the code: the deviation from the refactor proposal on its steps 3 and 5 (awaiting the author), and `feat/kv-metrics-sbd-section`, which #1455 invalidated by retiring the `telemetry` section it extended.

haishuok0525 (approved):

PR #1447, opened 2026-09-08
fix(agentx): align Hyperloom measurement and integration contracts

zengleixin-amd:
## Description

Keep AgentX measurement, promotion, reporting, and artifact fixes within Hyperloom. Merge the latest `origin/main` into the feature branch and remove reliance on unpublished GEAK/SGLang patches.

### Changes
- Preserve complete E2E measurements and compare candidates on the configured grading axis; missing required axes cannot authorize a performance KEEP.
- Share native performance gating with deferred accuracy scheduling, preserve accepted anchors, and update cumulative validation only from comparable accepted measurements.
- Freeze GEMM paired recipes and carry configuration removal/replacement controls without contaminating the reference leg.
- Keep report snapshots, recorder metric provenance, and ledger attribution consistent; distinguish applied source patches from attempted work artifacts.
- Use GEAK's existing supported client/native-launcher interfaces for proposals. Its AgentX-session throughput is a search proxy, not a canonical performance reference. Canonical AgentX promotion stays in Hyperloom's own same-workload revalidation; the GEAK harness cannot serve as a mismatched AgentX fallback.
- Follow SGLang's existing step-bounded capture contract in the Hyperloom caller: skip redundant stop only when fresh complete current-capture trace evidence proves automatic completion; otherwise retain explicit-stop failure handling.

### Scope
- Only Hyperloom sources and directly related tests are included.
- No GEAK or SGLang source changes, dependency pins, package installation, or runtime deployment are part of this PR.
- Existing isolated upstream experiments are not required or included.

## Linked issues

None linked; this PR consolidates the diagnosed Hyperloom-side AgentX contract defects.

## Tests and draft status

**Draft: implementation and targeted CPU validation are complete; fresh CI and review are pending.**

- Published the TP=1 unranked-trace fix, bounded streaming validation, per-capture metadata cache, and deadline enforcement. Complete JSON/gzip validation and explicit stop failure handling remain enforced.
- Published the Controller full-measurement writeback and GEAK refusal/recovery fixes, including retry when a previously unloadable overlay becomes usable.
- Merged `origin/main` at `cbcd5409a`; resolved patch-safety conflicts while retaining advisory root handling and work-artifact/annotation filtering.
- Post-merge targeted Linux run: **734 passed, 3 failed**. The failures were inherited help-probe tests consulting the host's installed AIPerf loader instead of their injected help text. After a test-only isolation fix, the affected preflight/repair suites passed **137/137**; other suites were not rerun after that test-only change.
- Ruff lint/format checks passed for PR-changed Python files; shell syntax and diff checks passed.
- Prior targeted verification: trace/deployment/budget regression **273 passed**; Controller/GEAK/promotion regression **264 passed**; pre-dispatch/refusal regression **133 passed**. Counts overlap and are not additive.
- CPU/mock HTTP capture loop **4/4 passed** using actual client/phase-gate/mapper/curl and real wall clock. Synthetic 2.4M-event trace validation used about **20.3 MiB** peak RSS; a single 64 MiB-string event used about **22.2 MiB**. These are not GPU benchmarks.
- No new real GPU or live-model deployment verification is claimed. Fresh GitHub checks belong to the pushed revision `8ff036a65`.

## Compatibility and limitations

- AgentX canonical performance is validated through Hyperloom, not GEAK's potentially different search workload.
- Explicit output/synthetic/legacy modes and existing runtime/accuracy/artifact checks are retained.
- Existing incomplete historical winner measurements are not populated with unrelated later baseline axes.
- A failed or unproven profiling stop is not converted into success merely because trace files exist.

## Checklist

- Single concern: AgentX contracts at Hyperloom-owned boundaries (split implementation batches).
- Breaking behavior: previously unsupported or unverified performance results can remain unvalidated rather than being promoted.
- External component boundary: existing GEAK/SGLang APIs are used; no unpublished component fix is assumed.

ðŸ¤– Generated with [Claude Code](https://claude.com/claude-code)


xiaofei-zheng:
Reviewed the full diff (20 non-test files) with a focus on the `degrade_reason` semantic flip and the new AgentX trace gate. Two blocking issues, both in the new trace-completeness gate.

### 1. `traces_complete` cannot rank vLLM traces, so every vLLM profile capture times out and is marked failed

`src/hyperloom/inference_optimizer/assets/agentx/aiperf_phase_gate.py` — `traces_complete()` requires a rank for *every* file (from `_TRACE_RANK_PATTERNS` on the filename, or `distributedInfo.rank` in the payload) and returns `False` if any file yields none.

Reproduced with the branch's own script:

- SGLang-style `2026-01-01-TP-0.trace.json.gz` → `traces-complete --tp 1` exits 0
- vLLM-style `host_12345.1700000000.pt.trace.json.gz`, containing a valid `cat=="kernel"` event → exits 1

This is not a synthetic case. A real trace produced by this repo on disk — `.../Qwen3-14B-FP8/.../pip-test-k88bb-master-0_84217.async_llm.1787731415290283310.pt.trace.json.gz` — has no rank token in its filename, and `grep -o '"distributedInfo":[^}]*}'` on the decompressed payload returns nothing.

Because `aiperf_client.sh:644` now ANDs `_trace_complete` into the flush success condition:

```sh
if [ "$_stable" -ge 3 ] && [ "$_want" -gt 0 ] && [ "$_cnt" -ge "$_want" ] && _trace_complete; then
```

such a run can never satisfy it. It burns the whole `AGENTX_TRACE_FLUSH_TIMEOUT_S` (default 1800s), then gets stamped `trace_flush_timeout` / `failed` — while the trace itself is intact and usable.

The rest of the codebase deliberately tolerates unranked traces: `src/hyperloom/agents/kernel/tools/_trace_rank.py:select_primary_trace` has an explicit `if tensor_parallel_size == 1 and unranked: return max(unranked, ...)` fallback. The new gate does not, so this is a compatibility break introduced by this PR (the pre-image `_trace_stat` only counted files and summed `wc -c`).

### 2. `traces_complete` full-`json.load`s every trace just to find one kernel event

Same file. To answer "does this trace contain at least one `cat=="kernel"` / `ph=="X"` event", the gate deserializes the entire payload:

```python
opener = gzip.open if path.suffix == ".gz" else open
with opener(path, "rt", encoding="utf-8") as handle:
    payload = json.load(handle, parse_constant=_reject_json_constant)
```

Measured on the real trace above (39.8 MB gzipped, ~2.4M events): **4.2 s and 2607 MB peak RSS for a single file**. The script's own comments cite a GLM-5.3 TP=8 capture totalling 5.1 GB gzipped — extrapolating, that is tens of GB of host RAM per file, on the benchmark node, while the server still holds the GPUs. And `_wait_for_trace_flush` polls every 10 s, so this can re-run once per iteration for the whole budget.

A streaming scan (or just reading the first few MB of the decompressed stream looking for `"cat":"kernel"`) is sufficient for the predicate being computed.

---

Checked and cleared (not reporting these):

- `report.py` `performance_comparison`: `comparison['reference']:.1f` and `baseline['intvty_p90']` cannot raise — `GradedComparison.candidate/reference` are plain floats defaulting to `0.0`, and `perf_snapshot_from_mapping` returns `None` unless both total and `intvty_p90` are strictly positive.
- The three removed `resume_pending_revalidation = False` sites in `writeback.py` do not deadlock the resume path — the flag is set once per resume reconcile and enqueues a single rebench (`writeback.py:5145`).
- `_want -gt 0` in `aiperf_client.sh` is safe: `TP` is always exported (`cli/__init__.py:1600`, `os.environ["TP"] = str(max(1, int(tp_resolved or 1)))`).
- `_run_gemm_e2e_sweep`: `total_gain` is now `None`-initialized inside the `kept` branch, but the `elif faults:` / `else:` branches still bind it to `0.0`, so the `if total_gain is None or ...` check cannot hit `UnboundLocalError`.

xiaofei-zheng:
Re-reviewed at `6f235d31` / `357b582e` / `f024eee6`.

**Finding 2 (unbounded memory) is fixed.** The streaming `_TraceJSONReader` brings the real 39.8 MB gzipped trace down from **2607 MB peak RSS to 18 MB** (measured, same file as before), and the shared `TRACE_FLUSH_DEADLINE_NS` budget plus `--timeout-seconds` bounds the wall-clock cost. The per-file/`stat`-signature cache also stops the 10s poll loop re-parsing the same files. No objection here.

**Finding 1 is not fixed for real vLLM captures — and `len(files) != tp` makes it stricter than before.**

A vLLM `/start_profile` capture writes *two* trace files into the single shared `VLLM_TORCH_PROFILER_DIR` (`bypass_runner.py:846` points all three profiler env vars at one `profile_dir`): the EngineCore worker trace, and a CPU-only frontend trace from the API-server process. This is what a TP=1 run of this repo actually produced:

```
$ ls session/Qwen3-14B-FP8/.../geak/e2e_cycle0/profile/round_tuning/profile/
pip-test-k88bb-master-0_84217.async_llm.1787731415290283310.pt.trace.json.gz   39,805,630
rank0.1787731397790965374.pt.trace.json.gz                                     15,080,208
profiler_out_0.txt
```

Running the new gate against that directory unmodified:

```
current_traces count: 2
traces_complete(tp=1) -> False

rank0.1787731397790965374.pt.trace.json.gz          {'rank': 0,    'has_kernel': True}
pip-...async_llm.1787731415290283310...json.gz      {'rank': None, 'has_kernel': False}
```

Three independent rejections, any one of which is enough:

1. `len(files) != tp` → `2 != 1` → `False` before anything is parsed. The pre-image was `len(files) < tp`, so this commit turned "at least tp" into "exactly tp"; an extra frontend trace is now fatal rather than ignored.
2. The frontend trace is genuinely CPU-only — I counted `cat` values over the whole decompressed 656 MB payload: `python_function` ×2,399,819 and `Trace` ×1, zero `kernel`. `traces_complete` requires `has_kernel` on **every** file, not on at least one.
3. Even with the count relaxed, the new `rank is None and tp == 1 → rank = 0` fallback assigns rank 0 to the frontend trace, colliding with the real `rank0.*` file → `rank in ranks` → `False`.

`_trace_files` only filters `graph_capture_*` and `merged-*`, so the frontend trace is in the watched set.

Consequence is unchanged from the first review: `_wait_for_trace_flush` cannot ever return 0 for a vLLM capture, so it runs out the whole `AGENTX_TRACE_FLUSH_TIMEOUT_S` (default 1800s) and returns 4 → `_write_profile_capture_status "failed" "trace_flush_timeout"`, on a capture whose GPU trace is complete and valid.

The new tests model TP=1 vLLM as a single unranked file (`test_current_trace_proof_accepts_single_unranked_tp1`, including an `async_llm_*` name), and `test_current_trace_proof_rejects_ambiguous_unranked_fallback[multiple_unranked]` explicitly asserts rejection for two files at tp=1 — which is the shape a real vLLM run always has. The predicate that matches reality is closer to "at least `tp` distinct ranks covering `range(tp)`, each with kernel events", with rank-less / kernel-less extras ignored rather than counted or fatal. `_trace_rank.py` already partitions traces this way for the consumer side.

Repro (no real artifact needed):

```python
w("rank0.1787731397790965374.pt.trace.json.gz",
  [{"cat": "kernel", "ph": "X", "name": "gemm"}])
w("pip-test-k88bb-master-0_84217.async_llm.1787731415290283310.pt.trace.json.gz",
  [{"cat": "python_function", "ph": "X", "name": "step"}])
traces_complete([d], snapshot, 1)   # -> False
```

Also checked in this pass, no issues found: `_trace_complete`'s `local` usage is fine (the script is `#!/usr/bin/env bash`); `TRACE_FLUSH_START_NS` / `TRACE_FLUSH_DEADLINE_NS` are set at the single call site before `_wait_for_trace_flush` runs, and `_now` is always assigned before the loop can break; `f024eee6`'s `record_keep` hook reaches `WritebackCollaborator._record_integrate_keep` correctly through `CoordinatorCollaborator.__getattr__`, and `_keep_result` supplies the keys `_record_integrate_keep` reads; `357b582e`'s `geak_verdict_is_terminal` / `geak_candidate_is_adjudicated` helpers and every symbol they import (`_geak_overlay_is_loadable`, `_normalize_geak_overlay_dir`, `is_agentx_mode`, `agentx_enabled`) exist and line up.

xiaofei-zheng:
Re-reviewed at `070242961` (the eight commits pushed after `8ff036a6`, plus the merges from main).

**The trace gate is untouched, so the one blocking issue stands.** `git diff 8ff036a6..070242961 -- assets/agentx/aiperf_phase_gate.py assets/agentx/aiperf_client.sh` is empty, and the real vLLM shape still fails on this head:

```
HEAD=070242961  traces_complete(tp=1) -> False
  rank0.<ts>.pt.trace.json.gz                    ranked, has kernels
  <host>_<pid>.async_llm.<ts>.pt.trace.json.gz   unranked, CPU-only
```

Details and the repro are in the previous comment; nothing about it has changed.

No new blocking issues in the eight new commits. What I checked:

- `6c8f90b6` reverts the GEAK handoff's `e2e_metric` to the constant `"output"` — that matches `origin/main:phases/kernel.py:791`, so it aligns with trunk rather than regressing.
- `a2791b47` (`_grid_base.py:204`) and `767872cc` (the revalidation axis-preservation list in `writeback.py:4099`) rename the emitted key `intvty_p90` → `e2e_norm_intvty_p90`. This is a real fix against the `#1444` merge: `GRADED_INTVTY` is `"e2e_norm_intvty_p90"`, so before the rename `perf_snapshot_from_mapping` found no interactivity axis and returned `None` for every grid variant / revalidation measurement. There are no remaining readers of a bare `"intvty_p90"` dict key in `src/` outside tests, and `_grid_runner.py:2231` already consumed the new name.
- `767872cc` also widens the `prebaseline_enablement` carve-out in `_lift_to_current_best` from "skip the comparability check" to "skip all three lift checks". Gated on `baseline_tput <= 0` + `baseline_enablement` + `attribution_eligible is False`, i.e. bring-up before any baseline exists, where there is nothing to compare against — consistent.
- `070242961` unset/override precedence: `resolved["unset_envs"]` is unconditionally assigned a few lines above every read, inside the same `isinstance(current_best, dict)` branch, so no `KeyError`; the materializer's docstring and code now agree that an explicit unset applies *after* `extra_envs`, and the handler correspondingly strips explicitly-unset keys out of `requested_envs`. The two `materialize_config_with_envs` edits across `a2791b47` and `070242961` converge back to the unconditional `envs.pop(...)`.
- `6e44181f` only drops `scratch` from `_SPECIALIST_SCRATCH_DIRS`.

Tests run locally on this head: `test_integrate_payload_defaults.py`, `test_kernel_integrate_and_report.py`, `test_geak_revalidation_dispatch.py`, `test_grid_runner.py`, `test_explore_executor.py`, `test_promote_shared_state_lock.py` — **561 passed, 2 skipped**. The two AgentX trace suites from the previous round (`test_aiperf_phase_gate.py`, `test_aiperf_client_sh.py`) were **243 passed** and are unchanged since; as noted before, they model a TP=1 vLLM capture as a single unranked file, which is why they stay green through the issue above.

xiaofei-zheng:
Re-reviewed at `760abac7`.

**The vLLM blocking issue is fixed.** `len(files) < tp` is restored and `has_kernel is False` files are skipped for rank accounting while still being fully parsed, so corruption/freshness checks stay intact. Verified against the real capture directory and synthetic shapes:

| case | result |
|---|---|
| real TP=1 vLLM dir (`rank0.*` + `*.async_llm.*`, 55 MB) | `True` — 7.3s, 18 MB peak RSS |
| CPU companion only, no GPU trace | `False` |
| TP=2, rank1 present but no kernel events yet | `False` |
| TP=2, both ranks + companion | `True` |

One new problem, in the tests this commit adds.

### `test_cpu_frontend_does_not_mask_incomplete_or_ambiguous_gpu_traces[truncated_frontend]` fails

`src/hyperloom/inference_optimizer/tests/test_aiperf_phase_gate.py:233`, on this head:

```
$ python -m pytest src/hyperloom/inference_optimizer/tests/test_aiperf_phase_gate.py -q -p no:randomly
1 failed, 136 passed in 4.87s

E  AssertionError: assert not True
E   +  where True = traces_complete([...], {'started_ns': 1788957465573715285, 'files': {}}, 1)
```

Cause is the test, not the gate. `_write_trace` stamps `os.utime(path, ns=(time.time_ns(),) * 2)` on every file it writes, but the `truncated_frontend` branch afterwards does `frontend.write_bytes(frontend.read_bytes()[:-8])`, which re-stamps mtime from the kernel's coarse timestamp clock. On this host that clock trails `time.time_ns()` by ~0.2 ms:

```
started_ns            = 1788957508655408321
after utime  mtime_ns = 1788957508655655686   >= started: True
after write  mtime_ns = 1788957508655182206   >= started: False   (delta -226 µs)
```

so `current_traces`' `stat[0] >= snapshot["started_ns"]` freshness filter drops the truncated file entirely. Only the valid `rank0` trace is left, the gate legitimately returns `True`, and the assertion fires. Adding the same `os.utime` stamp after the truncation makes it behave as intended (`files seen: 2`, `traces_complete -> False`), and confirms the gate itself does reject a truncated companion.

Note the failure is order-dependent in wall-clock terms, not in test order — it reproduces deterministically under `-p no:randomly` and passed when I ran the same node id with the random plugin active, so CI can go either way. On the hosts where it passes, the assertion is vacuous rather than correct: it is asserting on a directory the gate never sees the truncated file in.

Everything else in this commit checks out. The rest of the two AgentX trace suites is green (**256 passed** alongside that one failure), and the six suites I ran on `070242961` (integrate payload defaults, kernel integrate & report, GEAK revalidation dispatch, grid runner, explore executor, promote shared-state lock) were **561 passed, 2 skipped**.

xiaofei-zheng:
Re-reviewed at `a2cfd0895`. The previously-flagged `truncated_frontend` flaky test is fixed (the fixture is now re-stamped after truncation and asserted inside `current_traces`), and I re-verified the vLLM trace gate against the real capture shape. One new blocking issue, introduced by an encoding regression in this branch.

### Blocking: `writeback.py` / `integrate_patch.py` were re-saved as double-encoded UTF-8, corrupting a persisted KB string

This PR re-encoded two source files so that every `—`/`→` became mojibake. Byte-level proof from the committed blobs (`git show <sha>:<path> | xxd`), merge base is `d0b16f73a`:

- base `writeback.py:1611`: `change} E2 86 92 {sever`  → correct `→` (U+2192)
- head `writeback.py:1632`: `change} C3 A2 E2 80 A0 E2 80 99 {sever`  → mojibake `â†’`

Occurrence counts (base had **0** in both): `writeback.py` = 70, `integrate_patch.py` = 37.

106 of those 107 are inside comments/docstrings — cosmetic only, no runtime effect. **One is not**: `writeback.py:1632`, the `pitfall` branch of `_build_statement`:

```python
return f"{fw_tag}{change} â†’ {severity or '?'} on {model}/{hw}"
```

This return value is passed straight to `_kb_amend_recipe(append_pitfall={"description": description, ...})` and persisted into the recipe KB. Per the function's own docstring, this string is the identity key for dedup ("... MUST exclude volatile fields ... so N sessions merge instead of producing N rows"). Consequences:

1. Corrupted, non-recoverable text (`â†’` instead of `→`) is written into the persistent KB pitfall descriptions.
2. The same logical pitfall now renders to a different string than pre-PR sessions produced, so it no longer merges with existing rows — exactly the "producing N rows" failure the docstring is written to prevent. Only the `pitfall` branch is affected (the `lesson` branch has no arrow).

Fix: re-save both files as UTF-8 (this restores all 107 sites, including the functional one). No logic change is needed.

### Checked and cleared (no blocking issues)

- **AgentX trace gate** (`aiperf_phase_gate.py` / `aiperf_client.sh`): ran `traces_complete` against real shapes — vLLM TP=1 (`rank0.*` + CPU-only `*.async_llm.*`) → True with the companion skipped; SGLang unranked TP=1 → True; TP=2 missing/duplicate rank → False; truncated gzip → False; >1 MB single-event streaming path classifies kernel vs CPU-only correctly. Failure paths return non-zero and surface `trace_flush_timeout`; nothing is silently swallowed. `test_aiperf_phase_gate.py` passes deterministically under `-p no:randomly` (the local `test_wait_for_phase_fails_when_aiperf_process_exits` failure is a Windows-only artifact of `process_alive` using `/proc` + `os.kill`; it passes on Linux/CI).
- **Graded comparison / cumulative validation** (`writeback.py`, `report.py`, `perf_metric.py`, `integrate_performance.py`): `not comparable` is fail-closed in all three sinks (`assess_integrate_performance` → NEEDS_REVIEW, `_lift_to_current_best` → hold, `_update_cumulative_gain_validated` → returns False and callers gate the flag clear on it); no missing-axis path can authorize a KEEP. `integrate_performance.py` is wired in via `baseline.py` and `request_handlers.py` (not dead code).
- **`intvty_p90` → `e2e_norm_intvty_p90`** rename: readers/writers are consistent; `_grid_base.to_summary_dict` emits the new key and `perf_snapshot_from_mapping`/`graded_axes_of` read only the new key.
- **GEAK rebench / handoff** (`kernel.py`, `geak_rebench.py`, `_geak_sweep.py`): `no_promote`/fallback/incomparable all record status and clear `geak_pending`; overlay recovery re-opens adjudication when a previously-unloadable overlay becomes loadable.
- **env unset precedence** (`_workload_envs.py`, `request_handlers.py`): explicit unset is popped after `extra_envs` merge; docstring now matches code.

No blocking issues other than the encoding regression above; not approving until `writeback.py:1632` (and the file-wide re-encode) is fixed.

xiaofei-zheng (approved):
Re-reviewed at `779407d52`. The blocking encoding regression is fixed by `01addd307` ("restore UTF-8 punctuation and stable KB pitfall identities").

Verified:
- `writeback.py:1632` pitfall return is back to the correct arrow — byte-level `change} E2 86 92 {severity` (U+2192), not the mojibake `C3 A2 E2 80 A0 E2 80 99`.
- mojibake count is now **0** in both `writeback.py` and `integrate_patch.py` (was 70 and 37).
- The fix is punctuation-only: stripping all non-ASCII bytes from the commit's 107 removed and 107 added lines yields byte-identical text, so no logic changed — the persisted KB pitfall description / identity string is restored.

All other paths were cleared in the prior pass (AgentX trace gate against real vLLM/SGLang shapes, fail-closed graded comparison and cumulative validation, `e2e_norm_intvty_p90` rename, GEAK rebench/overlay recovery, env unset precedence).

No blocking issues remain. LGTM.

PR #1448, opened 2026-09-08
fix(kb): bank a losing operator's progress, and pick a warm start by measurement

jiaqiang-dot-liu:
## The ratchet that would not catch

An operator that cannot yet beat its source baseline could never record anything, so every run restarted from the same losing seed.

Measured on `mxfp4_moe_e65_i1024`:

| Round | Candidate read | Seed | Result | Write |
|---|---|---:|---:|---|
| 1 (48h) | `rewrite-...-5a8168c860d1` | 0.0047 | 0.9887 | refused, `no_improvement` |
| 2 (24h) | **the same one** | 0.0047 | 0.1595 | refused again |

Round 1 spent 48 hours climbing from a seed 200x off the pace, missed 1.0 by 1.1%, and had the result discarded. Round 2 then read that same 0.0047x record and, with half the time, only reached 0.1595.

Two things caused it: the sink's `speedup <= 1.0` refusal, and the rewrite writer's `allow_non_improving=port.attempts > 0` — which meant a *successful* warm start (PORT skipped, `attempts == 0`) was exactly what made the run unable to bank anything.

## What changed

**Recording is on correctness, not speed.** Both writers now record any run with a diff. The champion pointer stays gated on speedup, and a run that merely reproduces the solution it warm-started from is still refused as a duplicate (`no_improvement_over_reuse`).

**Warm start chooses by measurement.** A claim is computed over whatever cases produced it, so it cannot order candidates for a task that scores different ones — yet the rewrite path adopted the first candidate that merely passed correctness. It now times every admissible candidate on this task's own driver and adopts the fastest, which is what the forge-loop path already did.

**The field is ten wide because arena reuses an operator across shape sets.** The same operator task is run against different test shape sets, and a record's claim was computed over the set its producing task scored. So the top 3 by claim is not a shortlist of the 3 best starts for *this* task — it is 3 records ranked on a number that does not transfer. Which one suits the shapes in front of us is only knowable by measuring, and a field of 3 is too narrow to expect the right one to be in it.

**That width is what needs bounding.** One trial is a compile plus a correctness suite plus a benchmark — median 43s but up to 953s on `gemm_a16w16_nt_n6144_k6144` cold. A trial's cost also scales with how slow the candidate is, since both the suite and the benchmark run the kernel, so a record claiming under 0.3x is skipped rather than measured: the claim does not have to be right about the ranking for it to be right about the magnitude. The search closes after 30 minutes, adopting the best already measured, and each stage inside it takes whatever is left of that budget rather than its own ceiling — otherwise the first trial could spend the whole search and leave a ten-wide field with one member measured. All three bounds live in one module both warm starts read, with env overrides.

**Every rewrite KEEP is banked when it happens.** Rewrite runs forge-loop under `--no-experience-kb` because the two address different identities, which left a rewrite's KEEPs unbanked until the run ended — and an OPTIMIZE session can be cut off at its cutoff or killed outright. The rewrite layer now watches forge-loop's `--result-json` (refreshed on every KEEP) and publishes each new best to its own store, reading the kernel from the kept commit rather than the workspace the agent is still editing. A record is named after the session that produced it — `session_key`, supplied by the caller — so every publication in one OPTIMIZE session replaces the last instead of burying the identity's history under one run, and a run whose OPTIMIZE published nothing lands on the PORT record rather than beside it.

**One knock-on fix.** With the speedup gate gone, `no_improvement_over_reuse` becomes the common refusal in the campaign publication state machine, where it was classified as transient and would have been retried forever. It is a decision, so it is now settled.

## Test plan

- [x] `test_rewrite_by_flydsl_kb.py`, `test_rewrite_by_flydsl_pipeline.py`, `test_experience_sink.py`, `test_kb_measured_evidence.py`, `test_warmstart_policy.py`, `test_cli_max_hours.py` pass locally
- [x] `ruff check` clean
- New coverage: fastest-measured wins over best-claimed; below-floor candidate never downloaded and never referenced; a field entirely under the floor leaves the workspace untouched for PORT; one trial cannot spend the whole search budget; each KEEP handed over exactly once; a KEEP after the last poll still lands; a stale result file is never published; a failing publish does not cost the run; a KEEP whose commit lacks the kernel is skipped; a run whose OPTIMIZE published nothing lands on the PORT record
- [x] CI green on the merge with `main`

## Risks

- **Warm start gets slower on slow kernels.** 10 candidates is ~7 min on a median task and would be ~2.6h on the worst one measured; the 30-minute ceiling is what makes this safe, not optional.
- **Cross-task claim incomparability is not eliminated, only contained.** The 0.3 floor screens for catastrophe; comparability comes from measuring each survivor. Both are needed — the floor alone would not be enough. The floor does rest on the claim being right about the magnitude even when it is wrong about the ranking, which is the weaker of the two assumptions this PR makes.
- **The forge-loop path bounds a trial more coarsely than the rewrite path.** Its probes cap at 120s correctness plus 3×300s benchmark, so one candidate can overrun the 30-minute budget by up to 17 minutes. That is within the budget's order rather than able to consume it outright, and tightening it means threading a deadline through two shared probe wrappers, so it is left for a follow-up.
- **Candidate count multiplies with write frequency.** Per-run-stable session ids keep 10 candidates meaning roughly "the last 10 runs". If that ever reverts to content-hash naming, a batch this size would produce ~716 records against a 200-record index cap.


src/kernelforge/knowledge/experience_integration.py
```diff
@@ -40,16 +42,20 @@
...
+# something to fall back to, and so a record whose claim does not survive
+# measurement can lose to one that does.
+#
+# The two are equal on purpose: reading more than can be evaluated only pays for
+# ranked metadata nothing will act on. Each evaluation costs a correctness run
+# plus KEEP_MEASUREMENT_COUNT benchmark runs on the real driver, so the search is
+# bounded by ``warmstart_policy.budget_sec()`` as well as by this count -- on the
+# heaviest kernels one candidate is minutes, and the count alone would let a
+# well-populated identity spend hours before the agent's first edit.
+_WARMSTART_TOP_K = warmstart_policy.top_k()
```
ZhengGong-amd: `top_k()` is resolved at import time here, while `min_claimed_speedup()` and `budget_sec()` are resolved at call time in `kb_warmstart`, and `rewrite_by_flydsl/kb.py` resolves `top_k()` at call time too. So `FORGE_KB_WARMSTART_TOP_K` only takes effect if it is set before this module is imported — the override is effectively broken on this path, and the three knobs the new policy module was created to unify behave differently from each other and across the two warm starts. Suggest calling `warmstart_policy.top_k()` inside `kb_warmstart()` instead of binding it to a module constant.

src/kernelforge/rewrite_by_flydsl/runner.py
```diff
@@ -493,6 +493,50 @@ def _setup_failed(reason: str, failure_class: str) -> dict:
...
+        optimize_session_digest = hashlib.sha256(str(payload.get("experiment_id") or commit).encode()).hexdigest()
+        write = write_flydsl_kb_solution(
+            spec,
+            driver_path,
+            config,
+            source_ms=source_ms,
+            flydsl_best_ms=payload.get("best_ms"),
+            best_commit=commit,
+            framework=framework,
+            snr_db=port.snr_db,
```
ZhengGong-amd: `port.snr_db` is the SNR measured during PORT, but this record is being written for a KEEP artifact read out of `commit` — a different artifact. Every KEEP published in the run therefore carries accuracy evidence about something else, which a later warm start reads as if it belonged to the kernel it is adopting. Either measure the SNR for this KEEP or pass `snr_db=None` so the record does not claim what was not measured.

ZhengGong-amd (commented):
Reviewed the diagnosis and the four changes — the direction is right and the docstrings are unusually well kept in sync. Two things I think should be fixed before merge; the rest I'll leave out of the way.

src/kernelforge/knowledge/experience_integration.py
jiaqiang-dot-liu: Fixed in 70e51d1 — you are right, and the override was dead on this path.

`top_k()` is now read inside `kb_warmstart()`, where the floor and the budget are already read, so all three knobs resolve at the same moment on both warm starts. New coverage in `test_warm_start_reads_the_candidate_count_the_environment_asks_for` sets `FORGE_KB_WARMSTART_TOP_K` after import and asserts it governs the whole boundary — how many are read, measured, and offered to the author — which is exactly what an import-time binding could not do.

Chasing this also turned up that `_WARMSTART_MAX_MEASURED_CANDIDATES` was unreachable, and had been since before this PR when both constants were 3. `read_top_solutions` ends at `read_top_n(limit=top_k)`, so `len(sols) <= top_k`, and `measurements` grows by at most one per index, so `len(measurements) >= top_k` cannot hold at any `idx < len(sols)`. The only thing keeping it alive was the test pinning the two constants apart. It is gone; the bounds that actually close a field are the width and the wall-clock budget, and the replacement test drives the budget through a stubbed clock so the assertions do not depend on how fast the test machine is.

src/kernelforge/rewrite_by_flydsl/runner.py
jiaqiang-dot-liu: Fixed in 70e51d1. Measuring the SNR per KEEP would cost a validation run for a field nothing reads, so I took the `None` branch — but not unconditionally, because one case turned out to need the reading kept.

Per-KEEP publications now pass `snr_db=None`. forge-loop's `--result-json` does not carry the accuracy it measured for a KEEP (`_build_result` reports `best_ms` and `best_commit`; the reading only reaches the recovery checkpoint), so there is nothing honest to put there.

The run's final write had the same defect and predates this PR, so I fixed it too — with one exception. When OPTIMIZE produces no best, `opt["best_commit"]` falls back to `port_commit`, and with an empty `session_digest` that write lands on the same session id as the PORT publication and replaces it. Passing `None` there would have erased a measurement that was genuinely taken. So it is now `port.snr_db if final_commit == port_commit else None`: the reading is recorded exactly when the artifact being recorded is the one PORT measured.

I put the rule on the parameter rather than at the call sites, since that is where the next caller will look:

> `snr_db` is the accuracy measured for *this* artifact. A caller that did not measure it passes `None`; a reading taken from a different artifact is not a substitute, because the record does not say which kernel it was taken from and a later reader has no way to tell that it does not belong.

Worth noting for anyone who reaches for this field later: on the rewrite path `metric.snr_db` currently has no reader at all. `_read_top_candidates` hands `value` through as `attrs`, warm start uses only `flydsl_kernel` and `implementation_signature`, and `_reference_context` shows the agent `speedup` and `reason`. So this is a record-integrity fix, not a live misread — and if a consumer ever does want SNR, it should be plumbed from forge-loop rather than inferred, since like the speedup claim it is measured by the producing task's driver on its own cases.

Two tests cover it: `test_run_rewrite_records_accuracy_only_for_the_artifact_it_measured` and `test_run_rewrite_keeps_the_port_reading_when_optimize_never_moved_the_best`.

ZhengGong-amd:
Both fixes look right, and dropping the dead count constant plus adding the budget tests was the right call.

Stepping back though: if none of this code existed, I don't think most of it is what I'd write. Separating the defect from the reaction to it.

**The defect is two booleans.** `this_speedup <= 1.0` in the sink and `allow_non_improving = port.attempts > 0` in the rewrite writer fully explain the 48h loss and the repeated 0.0047x seed. Deleting them plus the `cli.py` reason update is the fix — roughly five lines. Nothing else in the PR addresses the defect; the rest is a chain reacting to it:

```
drop the write gates -> the store accumulates weak records
                     -> top_k 3->10
                         -> the 0.3x floor (10 candidates will contain junk)
                         -> the 30-minute budget (10 x up to 953s is unaffordable)
                             -> warmstart_policy.py, 3 env vars, 3 status strings, 66 lines of tests
```

The root of the chain is `top_k 3->10`, and that step is unrelated to the defect. Replay the incident with the gates dropped and `top_k` left at 3: round 1 banks 0.9887x, round 2 reads a field ranked with 0.9887 on top and climbs from there; the 0.0047x record sits below it and is never measured. **The gate removal alone fixes the incident.** With `top_k` unchanged, neither the floor nor the budget has a reason to exist.

**The floor is the speed gate reinstalled at the read boundary.** `write_flydsl_kb_solution`'s docstring now says correctness is what makes an artifact reusable — on the flydsl path a record's whole value is skipping PORT, which a correct-but-slow port still does. Then `try_flydsl_kb_warmstart` filters it back out on speed. The write side says speed is not a condition and the read side says it is. Concretely: round 2's own 0.1595x result — exactly the progress this PR set out to preserve — would be permanently unreadable.

**And it substitutes an incomparable number for a comparable one that is already in hand.** `DEFAULT_MIN_CLAIMED_SPEEDUP`'s own comment concedes a claim is not comparable across tasks, so it can only be a coarse screen. But the thing actually worth preventing — don't start from a seed slower than your own starting point — has an exact answer: compare the candidate's measurement against *this* task's baseline. The integration path already does this (it benchmarks pristine first and gates on it). The flydsl path holds `source_ms` and does `del source_ms`. Replacing that `del` with "measured slower than baseline is not adopted" needs no new module, no env var, no tunable constant, no `below_claim_floor` status, no policy test file — and is strictly more accurate, because it compares two real numbers from the same machine and driver.

**The exhaustive search duplicates a bounded one next door.** "A claim cannot order candidates for a task that scores different cases" is correct and stands on its own. But the flydsl implementation measures the entire field with no early exit, while `experience_integration` already has `_measurement_confirms_rank` and `_outranks_remaining`: measure in rank order, close the field once no remaining claim can beat the best measured. Rank order + early exit + `top_k=3` costs at most 3 trials, which is what already ships. The 30-minute ceiling is only load-bearing because the field was widened to 10 *and* the search refuses to stop early. Two mechanisms for one problem — keep the bounded one.

**`session_digest` patches a problem that need not exist.** Its entire job is making repeated publications in one run replace each other. It is needed only because candidates are named after the artifact while per-KEEP publication inherently wants them named after the run. The fix is one line at the naming site — name candidates by run — rather than an override parameter threaded through two call sites plus a `nonlocal` in the runner plus a fallback branch for "OPTIMIZE published nothing." One rule instead of keeping two consistent.

Per-KEEP banking itself should stay: an OPTIMIZE session cut off at its cutoff is a second, genuinely independent defect. Polling `--result-json` is defensible — it keeps the identity boundary that `--no-experience-kb` draws — at the cost of a weaker second copy of a publication path forge-loop already has. Worth naming as a deliberate trade.

**Summary**

| Mechanism | Suggestion | Why |
|---|---|---|
| Drop the two write gates + `cli.py` reason | keep | this is the fix |
| Per-KEEP publication (the concept) | keep | second, independent defect |
| `session_digest` | delete | name candidates by run at the naming site |
| `top_k` 3->10 | delete | unrelated to the defect; root of the chain |
| 0.3x floor | delete | reinstates the speed gate at the read boundary; replace with measured-vs-this-task's-baseline |
| 30-minute budget | delete | only needed at top_k=10 without early exit |
| `warmstart_policy.py` + 3 env vars | delete | all three values are gone |
| flydsl exhaustive measurement | rework | reuse `_outranks_remaining`, don't reimplement it unbounded |

Net: roughly +956/-63 down to +250/-80, one fewer module, three fewer env knobs, one fewer tunable constant, and no disagreement between the write side and the read side about whether speed is a condition.

jiaqiang-dot-liu:
Thanks — this is the right level to argue at, and it moved the PR. Three of your points landed, one of them turned up a defect neither of us had named, and I'm pushing back on the chain argument at its root.

Pushed in `4a1231e`.

## Where the chain breaks: `top_k` is not a reaction to the defect

You are right that widening the field is the root of the chain, and right that nothing in the PR explains why it is there. That is my fault — the reason was never written down, so the review can only read it as a reaction.

Arena runs the same operator task against **different test shape sets**. A record's claim was computed over the set its producing task scored. So "top 3 by claim" is not a shortlist of the three best starts for *this* task; it is three records ordered by a number that does not transfer to it. Which one suits the shapes in front of us is knowable only by measuring, and three is too narrow to expect the right one to be in the field at all.

This is the same incomparability you accept elsewhere in the review — *"A claim cannot order candidates for a task that scores different cases is correct and stands on its own."* Your replay leans the other way: round 1 banks 0.9887x, round 2 reads a field with it on top. That holds across a 200x gap. It does not hold across the 0.5x-versus-0.8x distinctions arena spends most of its time in, and those are the ones that decide which candidate is worth adopting.

The rationale is now in the PR description, where it should have been from the start.

## Rejecting "reuse `_outranks_remaining`" — it is unsound under that premise

This is the one suggestion I think is actively wrong, and for the reason you yourself gave.

`_outranks_remaining` (`experience_integration.py:790`, used at :1249) compares a **measurement** against the remaining candidates' **claims**:

```python
for sol in remaining:
    ranked = _ranked_speedup(sol)
    if ranked is not None and ranked > measured_mean_case_speedup:
        return False
return True
```

If claims do not transfer across shape sets, a candidate claiming 0.5x on the set that produced it can be the fastest on the set in front of us, and this closes the field before reaching it. The guard in front of it (`_measurement_confirms_rank`) only establishes that *this* candidate reproduced *its own* claim; it licenses nothing about the others'. So the flydsl path measuring the admissible field exhaustively is not a worse copy of the bounded one — it is the one that does not silently drop the candidate we widened the field to find.

That cuts the other way too: the forge-loop path has the same exposure today. It predates this PR (both helpers are in the merge base), so under *one concern per change* it is not this PR's to fix, but it is worth a follow-up and I'd rather name it than leave it implied.

## Accepted — the floor's stated reason was wrong

You demolished the argument the code actually made, and you were right to: *"not instructive"* does not survive the observation that a slow-but-correct port still saves the next run its PORT.

That was never the real reason. A trial's cost scales with how slow the candidate is, because the correctness suite and the benchmark both run the kernel — a record orders of magnitude off the pace can spend the whole search budget on itself. The claim only has to be right about the *magnitude* for skipping to be the right trade, which is a much weaker demand than being right about the ranking.

The floor stays, with that reason now stated at both sites and on the constant. Two things I'll own rather than paper over:

- **The 0.1595x example is real.** A record claiming under 0.3x is banked and then not read back. The incident itself is still fixed — round 1's 0.9887x is well clear of the floor — but you have correctly identified the residue, and it rests on the magnitude assumption above. That is the weaker of the two assumptions this PR makes, and it is now listed under Risks.
- **The floor currently excludes more than the cost argument requires.** It drops the candidate at the read boundary, so it is not offered as reference material either — and showing a record to the author costs tokens, not minutes. If you think reference-only reading should survive the floor, say so; I think that is a defensible narrowing and I did not make it because it changes what the author sees, which deserves its own argument.

Your alternative — measured against *this* task's baseline — is not a substitute for the floor, because it can only reject after paying for the measurement the floor exists to avoid. It is a good adoption gate, and the forge-loop path already applies one.

## Accepted — `session_digest`

Right about the smell, and the fix is better than a rename. `session_digest or best_commit or content_hash` meant a record's name depended on which caller wrote it, with `best_commit` doubling as metadata and as a name. There is now one rule: `session_key` names the session a record belongs to, and each caller says which. `best_commit` is metadata only.

One correction to the diagnosis: it is not a one-line change at the naming site. `identity.session_id` has two callers with two different notions of the third argument (`kb.py` and `agent_kb.py:231`), and that layer has no concept of a run. So the plumbing does not vanish — but it collapses from three interacting rules to one, and the "OPTIMIZE published nothing" branch is now a stated rule rather than the coincidence it was relying on (`final_commit == port_commit`).

Production naming is byte-identical to before. A test that had been leaning on `best_commit` to keep two same-content records apart now says so explicitly.

## Found while checking the above — a real defect

A candidate's own stage ceiling is 1800s. The whole candidate search is also 1800s, and the budget is only checked *between* candidates. So the first trial could spend the entire budget and leave a field widened to ten with one member measured — the failure mode is exactly the one widening the field was meant to avoid, which is how it escaped notice.

Each stage now takes whatever is left of the search rather than whatever its own ceiling allows (`kb.py:41`). Covered by `test_one_trial_cannot_spend_the_whole_search_budget`.

The forge-loop path bounds a trial more coarsely — 120s correctness plus 3x300s benchmark, so one candidate can overrun by up to 17 minutes. That is within the budget's order rather than able to consume it outright, and tightening it means threading a deadline through two shared probe wrappers and their doubles across five test files. Left as a follow-up and listed under Risks.

## Where that leaves the diff

Down by the `session_digest` plumbing and its fallback branch; up by the budget clamp and its test. `top_k`, the floor, the budget and `warmstart_policy` stay, now with the reason for the first written down and the real reason for the second stated in the code.

Happy to keep arguing any of these — particularly the reference-material narrowing, where I think you may be right and I have deliberately not acted.

xiaofei-zheng (approved):
No blocking issues — LGTM.

Reviewed at 4a1231ec5 (re-reviewed after 70e51d1ee / 4a1231ec5 landed on top of the earlier head).

What I checked, and how:

**Write-policy change (`experience_sink`, `write_flydsl_kb_solution`)**
- Confirmed the champion pointer is still gated on speedup on both paths after the write gates were removed: `agent_kb._maybe_promote` (`speedup is None or <= 1.0` -> no promote) and `kb.py:460` (`speedup is not None and speedup > 1.0`, plus incumbent comparison). A losing record is stored but never becomes the identity's champion.
- Grepped the whole tree for producers of the reason string `"no_improvement"` now that `cli._record_remote_publication_result` no longer lists it as settled: neither writer emits it any more, and the only input to that state machine is `write_experience_to_kb` (cli.py:2056). So no refusal can fall through into `pending_retry` forever. `gemm_tune`'s unrelated `"no_improvement"` status never reaches this state machine.

**KEEP-stream publication (`optimize.run_optimize` + `runner._publish_keep`)**
- `_result_for_this_run` rejects a stale result file on the announced experiment id, and refuses to trust the file before an id has been announced; a partially written file raises `JSONDecodeError` and is simply retried on the next poll.
- The post-loop `_publish_new_best()` call covers a KEEP recorded between the last poll and process exit, including the deadline-termination path.
- Checked the case I was most worried about: the final `write_flydsl_kb_solution` reuses the OPTIMIZE session key and reads content off the workspace, so it overwrites the last KEEP record. That is safe because `run_optimize` calls `_restore_best_kernel(best_commit=...)` (optimize.py:370) before returning, on both the clean-exit and the deadline-terminated path, so the workspace holds the best commit's kernel at that point.
- Callback exceptions are caught and logged inside `_publish_new_best`, so a store failure cannot take down OPTIMIZE.

**Record naming (`session_key`)**
- All three production call sites are in `runner.py` (415 / 484 / 536); nothing else in the tree calls this writer, so replacing the old `session_digest or best_commit or content_hash` chain with `session_key or content_hash` cannot silently rename records written by some other caller. PORT -> port_commit, each KEEP -> the OPTIMIZE session hash, final -> `optimize_session_key or port_commit`, which keeps the "no KEEP means land on the PORT record" behaviour explicit.

**Bounds**
- `_stage_timeout` behaves as intended: with 300s of budget left and no run deadline it returns 299; with a 60s run deadline it returns 60; when the budget is already spent it clamps to 1 rather than going negative.
- `below_floor(None)` is False (an unrecorded claim is not a low claim), `below_floor(0.29)` True, `below_floor(0.3)` False; `_positive_int/_positive_float` fall back to the default for empty, non-numeric and non-positive env values.
- `top_k()` is now read at call time on both warm-start paths, so `FORGE_KB_WARMSTART_TOP_K` actually takes effect; no stale `_WARMSTART_MAX_MEASURED_CANDIDATES` references remain.
- The widened field is still bounded end to end by `stop_at_unix=search_stop_unix` passed at the call site (runner.py:329), so a 10-candidate search cannot outlive the run budget.

**Tests run locally** (minimal scope; CI covers the rest): `test_rewrite_by_flydsl_kb.py`, `test_kb_measured_evidence.py`, `test_rewrite_by_flydsl_pipeline.py`, `test_experience_sink.py`, `test_warmstart_policy.py`, `test_cli_max_hours.py` — 168 passed. `ruff check src/kernelforge` clean.

xiaofei-zheng (approved):

PR #1421, opened 2026-09-07
refactor: compact Python comments and docstrings

chennyiiis:
## Summary

- Compact verbose Python docstrings and comment blocks across Hyperloom, the inference optimizer, orchestration code, and KernelForge.
- Replace 1,038 mechanically truncated first-line docstrings with complete semantic summaries while keeping them compact.
- Preserve every existing docstring slot and standalone comment block, including security invariants and external model/runtime constraints.
- Leave standalone documentation files unchanged and preserve the runtime text of every CLI that uses `description=__doc__`.
- Redact KernelForge recovery exception text consistently from logs and durable controller artifacts while retaining source, operator, commit, and exception-type diagnostics.

## Scope

- 1,376 Python files changed
- 121,294 lines removed and 21,013 concise or test lines added
- **100,281 net lines removed**
- 18,219 original docstring locations preserved
- 16,354 original standalone comment blocks preserved
- Executable Python structure is unchanged by the documentation cleanup. The only intentional runtime change is the recovery-error redaction in both logs and durable state.

## Validation

- `python -m ruff check .`
- `python -m ruff format --check .`
- `python -m compileall -q .github/scripts scripts src`
- `git diff --check`
- AST audit: zero remaining first-physical-line docstring truncations.
- CLI audit: every changed `description=__doc__` module has its original runtime docstring/help text.
- 138 targeted KernelForge recovery/controller and Hyperloom server-argument/workload tests passed under Linux/WSL.

The broader component suites were also exercised before the final rebase: orchestrator (385 passed), agents/common (3,463 passed, 26 skipped, 15 subtests passed), KernelForge (3,711 passed, 6 skipped), and shell client (72 passed).

PR #1452, opened 2026-09-09
gemm-tune: actually open the tier3 path, and let it reach fused MoE

chennyiiis:
## What this is

The tier3 path — "let a model write a tuner on the spot" — was asked to be opened
in August. The change that landed flipped `FORGE_TIER3_DISABLE` from opt-in to
opt-out and stopped there, so the switch said open and the path stayed shut. It
has now run end to end on an MI355X, from the fleet's own demand document to a
promoted candidate, and this is what it took.

#1439 proposed deleting `gemm_tune/tier3/` on the grounds that it never ran.
That grounds was correct. This is the other answer to it; #1439 has since merged
with `tier3/` left in place.

This was three PRs -- #1452 (M0 + M1), #1458 (M2, fused MoE) and #1461 (the JIT
build cap). They are now one branch, because they were always one story: the MoE
half was stacked on this one and never independently reviewable, and the build
cap was the prerequisite for the last unfixed path inside tier3 itself. Both are
closed and point here.

## The acceptance run

Real box, real credential, real 42-shape demand taken from an aiter serving log.
No stubs at any stage.

```
== the gate      allowed=True  gap=bf16_tuned_gemm.csv (42 misses / 42 keys)
== the adapter   _Bf16DenseAdapter
== generate      a real authoring session wrote tuner.py (817 lines)
== the attempt   stage=referee  ok=True  36 of 42 shape(s) improved
== the result    tier3_generated_bf16_tuned_gemm  status=ok  candidate=True
                 env=AITER_CONFIG_GEMM_BF16  shapes=36/42  best=1.5216x
== promotion     per_tuner_candidates -> ['tier3_generated_bf16_tuned_gemm']
```

Two independent runs: 35/42 and 36/42 improved, best 1.52x in both. The noise
floor did its job — `16x3456x1152` re-timed at 1.008x and was recorded as not
improved.

The tuner the session wrote searched all five dispatchable backends, screened
every one of the 2084 hipBLASLt solutions the library offered for each shape,
timed with captured graphs, and audited correctness 8 times on fresh inputs.
Winners: hipblaslt 32 shapes, torch 7, aiter_asm 3.

## Six things were in the way

Each one is a wall immediately behind the gate, and every one of them only
became visible with real hardware and a real provider behind it.

1. **The referee approved the baseline over itself.** No noise floor, and the
   unmodified path beat itself by up to 1.00925x across a 100-trial null
   distribution. `MIN_SPEEDUP = 1.01`, measured rather than picked.
2. **The hipBLASLt guard did not prevent the abort it existed for.** An invented
   `solidx` fails inside `hipbsolgemm.cu:945` and never raises in Python.
   Candidates are now checked against the per-shape `hipb_findallsols` set.
3. **The mandate never said what a dispatchable candidate looks like.** It asked
   for candidates "carrying enough detail to be dispatched by code that did not
   write your script" without naming the five backends, the `k=v;k=v` grammar or
   the `MxNxK` key. An author following it exactly proposes candidates
   `dispatch._build` cannot parse, all of which are recorded as not dispatchable.
   The brief is now generated from the same table `_build` refuses against, so
   the two cannot drift.
4. **The authoring session had nowhere it was allowed to write.** A writable
   session runs under the workspace guard, which requires a git worktree to
   snapshot and roll back; `work_dir` is an output directory and never was one.
   On a GPU box the session died at `WorkspaceSafetyError('not a git
   repository')` before the model was asked anything. This is upstream of the
   gate, so no amount of gate work could have exposed it.
5. **The session had no pre-approved tools, asked for permission, and stopped.**
   With `tool_policy=None` the backend sets no `allowed_tools` at all. Measured:
   it asked for Write and for `python3`, explained why it could not enumerate
   hipBLASLt solutions without them, and ended its turn waiting for an answer
   nobody was there to give. The failure was reported as "the session ended
   (agent_stopped) without writing tuner.py" — the agent's own explanation was
   thrown away, which cost a full GPU run to rediscover by hand. It is now
   quoted.
6. **A bad launch kills the process, and the brief never said so.** The first
   tuner written from this mandate died on shape 2 of 42 with a GPU memory
   fault while sweeping one backend's kernel ids — no traceback, nothing for
   `except` to see, four minutes of correct work reduced to one row. The mandate
   warned that an *invented* hipBLASLt index does this; it did not say a
   legitimate sweep of the other backends does too, that the fault surfaces
   asynchronously so the candidate it lands on is not reliably the guilty one,
   or that the only defence is subprocesses and on-disk state. The session
   worked all of that out by itself and rewrote the tuner around a
   driver/worker split — the tier working as intended, but two GPU runs spent
   learning something already in a log.

### M0 — a table we deployed ourselves stopped reading as a table nobody owns

An aiter miss line names the file the runtime looked in, and after a deploy that
file is no longer named after the table. Two renames compose: a tuner's artifact
is named after the tuner (`sglang_dense_bf16` writes `tuned_dense_bf16.csv` for
what aiter calls `bf16_tuned_gemm.csv`), and deploy lands the merged result as
`merged_<artifact>.csv`. `TABLE_TO_TUNER` knows neither name, so the demand came
back with `tuner=None` — the same signal a table nobody implements gives.

Across the fleet's 0903–0906 serving logs that is **28,818 misses filed as
ownerless**, every one of them against an artifact we tuned and deployed
ourselves.

Verified on an MI355X against a real aiter log, produced by pointing
`AITER_CONFIG_GEMM_BF16` at a deployed-style path and letting aiter print its
own miss lines:

```
before: table='merged_tuned_dense_bf16.csv' tuner=None env_var=None schema=M,N,K
        GAP kind=no_tuner warrants=True        <- 60 misses, above the floor of 25
after:  table='bf16_tuned_gemm.csv' tuner='sglang_dense_bf16'
        env_var='AITER_CONFIG_GEMM_BF16'
        schema=M,N,K,dtype,otype,bias,scaleAB,bpreshuffle
        (no coverage gaps)
```

This is a blocker for M1 rather than a nicety: opening the gate without it means
the one thing tier3 reliably fires on is the table `sglang_dense_bf16` already
owns and just deployed — precisely what gate condition 2 exists to prevent. The
key schema also stops degrading to a bare `M,N,K` fallback.

### M1 — the gate can now reach a table whose owner left it untuned

`warrants_generated_tuner` accepted `no_tuner` only. That sounds stricter and was
in practice absolute: across 0903–0906 **every demanded table had a registered
owner**, so no run ever produced a gap it would accept, while the runtime went on
missing millions of keys those owners never covered. "A tuner exists" was
standing in for "the table got tuned", and only the second is what the runtime
experiences.

It now also accepts:

- `empty` — an owner ran and handed back nothing the integrate lane can apply.
- `skipped` — an owner declined, for a reason that is not a capability refusal.
  `not supported` / `unavailable on` / `already at peak performance` are filtered
  out by the existing `_NOT_A_COVERAGE_GAP` before a gap of that kind is built.

`not_selected` is still refused. That is a routing bug, and generating a second
tuner would paper over the first.

Second half of M1: `cli.py:711` called `_attempt_tier3(...)` and dropped the
return value, so a verified result reached no caller, entered no report, and was
merged by nobody. It is now mapped to an ordinary `TuneResult` with
`candidate=True`, which puts it on the same e2e-validated path as split-K — the
referee's micro speedup is measured on graph replays of individual kernels and
is not by itself a claim about end-to-end throughput.

## M2 — fused MoE, the table the fleet actually loses time on

Dense bf16 proved the path. This is the table that costs real time: `fmoe_ck`
burns 621-707s per GLM session and hands back `empty_output`.

Fused-MoE misses were parsed, counted, and then dropped on the floor --
`evidence.py` filed them under `dispatch["moe"]` while `coverage_gaps()` reads
only `demands`, so `tuned_fmoe.csv` could not become a gap at all. Behind that
sat a second wall: no adapter for the table, and `runner.py` paid for a
generated tuner *before* checking whether one existed, so the miss would have
surfaced as a wasted authoring call rather than a refusal.

Building the demand has a real and deliberate blast radius -- `router.py` now
selects `fmoe_ck` from log evidence, and `demand_for_tuner(report, "fmoe_ck")`
stops returning `None`. Both are fixes in their own right, which is why they are
their own commit behind `FORGE_MOE_DEMAND_DISABLE`.

### Two ways a fused-MoE candidate lies about winning, both measured

**A row aiter reads and then ignores.** `fused_moe.py:2048-2051` logs the tuned
pair *before* asking `if is_flydsl1 or is_flydsl2`. A pair with neither name
FlyDSL's falls past the entire tuned branch into the heuristic chain -- but the
log says it was used. Measured: a candidate naming `nope1`/`nope2` was logged
verbatim, ran the default kernels, and timed **1.1% faster than baseline**, over
the referee's 1.01 floor. `aiter_honours_kernel_pair()` refuses the pair before
any GPU work.

**A `block_m` that does not match the tile.** `block_m` is the granularity
tokens are sorted into before either kernel indexes them, so a mismatch does not
fault -- it reads the wrong rows, fast. Nine measurements at token=512, no
exceptions:

| stage1 tile_m | stage2 tile_m | block_m | us | mean err |
|---|---|---|---|---|
| 128 | 128 | 32 | 71.70 | 1.404 (wrong) |
| 128 | 128 | **128** | 156.74 | **0.0016 ok** |
| 32 | 32 | **32** | 121.12 | **0.0002 ok** |
| 32 | 32 | 128 | 91.10 | 1.096 (wrong) |

Baseline 111.67us. The wrong pairs were the *fastest things in the search* --
71.70us, a 1.55x. The adapter's mean-error gate caught every one, but a tuner
that pins `block_m` and trusts its clock spends its whole budget ranking
nonsense, so the mandate now states the constraint.

### Acceptance, on hardware

MI355X / gfx950, aiter `d9e5ef7ce`. The demand comes from a log aiter itself
wrote, not a fixture; the authoring stage uses `--stub-generate` because that box
had no provider credential.

```
demand:  tuned_fmoe.csv  tuner=fmoe_ck env=AITER_CONFIG_FMOE  36 misses / 3 keys
gap:     ('tuned_fmoe.csv', 'empty', True)      gate: allowed=True
adapter: _FusedMoeAdapter                       shapes: 3
attempt: stage=referee ok=True  reason="3 of 3 shape(s) improved"
result:  tier3_generated_tuned_fmoe status=ok candidate=True shapes=3/3 best=1.1528x
```

Per shape, referee-verified under `time_paired` (interleaved, min-of-repeats,
best-vs-typical consistency gate):

| token | baseline | candidate | speedup |
|---|---|---|---|
| 512 | 2212.0us | 2170.3us | 1.0192 |
| 1024 | 3051.7us | 2651.4us | **1.1510** |
| 2048 | 4208.6us | 3650.6us | **1.1528** |

Zero candidates discarded on correctness -- the search groups by tile_m and sets
`block_m` to match, so every pair it can propose is legal.
`requires_e2e_validation` stays true: micro is a screen, not the verdict.

## What review changed

Two things that were real, and neither of them visible in the diff.

**The router starts booking `fmoe_ck` by itself.** `router.py` is not in this
diff, but the new `tuned_fmoe.csv` demand reaches
`_tuners_the_demand_says_are_needed`, which adds `fmoe_ck` at `priority=10` --
about fifteen minutes at the head of every MoE run -- and it did so on any number
of misses at all. Both switches involved (`FORGE_MOE_DEMAND_DISABLE`,
`FORGE_TIER3_DISABLE`) are opt-*out*, so the default path was: one miss, fifteen
minutes of `fmoe_ck`, quite possibly `empty_output`, then a tier3 authoring
session and a sandbox GPU run. Demand-driven selection now needs
`DEMAND_MIN_MISSES = 25` lookups before it books anyone -- the same floor tier3's
own gate applies to the same demand report. The demand is still recorded whatever
the count; only the time booking is gated.

**The MoE adapter did not give the process back.** It steers aiter through
`AITER_CONFIG_FMOE`, a process-wide variable; it captured the prior value and
never restored it. So the last candidate tried stayed selected for the rest of
the process, including report generation and e2e validation -- which would then
be served by a config the referee may well have rejected, attributed to whatever
the report said was deployed. The `mkdtemp` scratch tables leaked with it.
`_FusedMoeAdapter.close()` restores the variable (deleting it when it was
previously unset), drops aiter's three dispatch caches so the restored name is
the one actually read back, and removes the temp dir. `_attempt_tier3` calls it
in a `finally`, so a failed attempt cleans up too; it is a context manager as
well.

Plus the mechanical bot findings: module-import style in two tests, four
`staticmethod(lambda: _FakeTorch())` reduced to `staticmethod(_FakeTorch)`, and
a redundant local `import re`.

## The JIT build cap (was #1461)

aiter's JIT sizes its `ninja -j` from the CPU count, which on a fleet box is the
*host's* -- 236 here -- while the memory those jobs consume is capped by the
*container's* cgroup, 128 GiB here. Nothing reconciles the two. Measured, that is
`-j 188` at 1.64 GiB a job (three CK-tile fused-MoE compiles, 30s each, within
0.01 GiB of each other), or about 308 GiB against a 128 GiB ceiling. The build is
not slow or flaky; it is OOM-killed and does not finish.

The cap computes 44 for that box. Verified on hardware: running all 44 at once
took the cgroup's `memory.current` from 1.0 to 58.4 GiB -- 1.30 GiB a job in
aggregate, since the peaks do not coincide -- leaving 69.6 GiB of headroom, 44/44
compiling clean in 177s. It does nothing when `MAX_JOBS` is already set, and
nothing on a box with no cgroup ceiling or a generous one. Both "unlimited"
sentinels are tested: cgroup v2's literal `max` and v1's `9223372036854771712`,
which taken at face value would compute a limit of several billion.

Applied in `run_subprocess`, which every tuner goes through, and -- the last
commit here -- in `tier3/sandbox.py`, which builds its own environment and calls
`subprocess.run` directly. That was the one path still asking for a job per host
CPU, and the one least able to absorb the result: an aiter tuner that gets
OOM-killed is a tuner that failed and can be re-run, whereas a generated tuner
gets one sandbox run and a dead build is recorded as "the script the model wrote
does not work". Wrong lesson, wrong evidence, about the leg of this pipeline that
is hardest to judge in the first place.

## Tests

Full `gemm_tune` suite: **988 passed, 8 skipped**, plus 3 pre-existing Windows
`os.killpg` failures in `test_utils_extra.py` that are on `main` too. `ruff
format` and `ruff check` clean. `test_moe_dispatch.py` and
`test_build_parallelism.py` are new files. The new tests assert the things that
were silently false: that every backend named in the brief reaches its own branch in `_build`
and none falls through, that an unadvertised backend is refused before any
operand is allocated, that the work_dir becomes a worktree and the session is
not started when it cannot, that the tool policy grants write and shell, that
the agent's own words survive into the failure reason, and that the brief names
the fault, its asynchrony, and what to do about it.

## Not in this PR

**M3 — deploying a genuinely unmapped table.** `TUNER_ENV_VARS` and
`_AITER_ENV_TO_TABLE` are closed sets keyed by name, so a table with no owner has
`env_var=None` and a win has nowhere to land. No such table exists on the fleet
today. When one does, the env should be derived *and then verified* against the
runtime's own hit lines before anything is deployed — `utils.py:64-68` records
what happens otherwise: a guessed `AITER_CONFIG_GEMM_A4W4_BLOCKSCALE` silently
discarded every fp4 tuned GEMM.

**Four aiter findings**, all upstream rather than patchable here: a tuned config
naming `tile=256` for `inter_dim=384` silently runs 128 (aiter's own tuner picks
these, and they fail correctness at 0.14); `AITER_CONFIG.get_config_file`'s
`lru_cache` key omits the env var its body reads; the tuned row logged before the
branch that discards it (above); and the JIT's `-j` sizing itself, which the cap
here works around rather than fixes.

## The one thing not yet exercised

`generate` -- the leg where the model actually writes the tuner -- ran under
`--stub-generate` for the MoE acceptance, because that box had no agent provider
installed. The dense acceptance at the top of this PR did run it for real, with a
real credential, and the 817-line tuner it produced is where those numbers came
from. So the leg works; it has not yet been run against the MoE mandate
specifically. Everything else in both paths ran with real components.

## On #1439

#1439 has merged, with `tier3/` correctly left in place, so there is nothing left
to sequence. Two corrections to its description, for the record:

- "`runner.py` discards it unconditionally at the referee stage — the result was
  never returned to a caller" is not accurate. `runner.py:186-213` returns a
  complete `Tier3Outcome` and writes `outcome.json`. The call site at
  `cli.py:711` is what dropped it.
- That is not an independent second reason. The gate only opened for ownerless
  tables, so `adapters_for()` always returned `None` and `runner.py:162-167`
  returned before the referee. It is downstream of the first reason, not beside
  it.


chennyiiis:
Follow-up stacked on this branch: #1458 takes the same gate to `tuned_fmoe.csv`. Referee-verified on MI355X, 3 of 3 shapes improved, best 1.1528x. Review this one first.

chennyiiis:
Rebased onto current `main` (was CONFLICTING, now MERGEABLE).

Every conflict came from #1421, the repo-wide comment compaction — not from #1439, which merged with `tier3/` correctly left in place. On the eight files that collided I rewrote our rationale in the new compact style rather than taking "ours", since restoring the long docstrings would have silently reverted #1421 on exactly those files. The measured evidence is all still there, just shorter: the referee's 94MHz idle ramp and 1269us-vs-517us neighbour swing, the mandate's four clauses, the gate's four conditions and the 0903–0906 fleet-log reason for widening condition 2.

One test needed a real fix, not just a re-wording. `main` brought in `tests/test_tier3_dispatch.py`, whose fake `aiter` returns `None` from `hipb_findallsols` and never defines `hipb_create_extension`. Both were fine against the old `_build`, which passed any `solidx` straight through; this branch validates it first, because `hipb_mm` on an index hipBLASLt never offered does not raise — it aborts the process from C++ with `INVALID_VALUE`, and the referee runs in-process at the tail of a session, so that death costs the whole run its report. Against a fake answering `None` every hipBLASLt candidate is now correctly undispatchable, which the test was reading as a regression. The fake now has the handle and a real solution list, and asserts the ordering it was named for; an invented index gets its own test instead of riding along.

Local: `ruff check` clean, `3 failed, 866 passed, 6 skipped` — the three are pre-existing Windows `os.killpg` failures in `test_utils_extra.py`, present on `main` too.

chennyiiis:
**Folded #1458 and #1461 into this PR** — one branch, 15 commits, rebased onto current `main` (`670b67d10`). Both are now closed and point here.

The three were always one story and splitting them made each piece read as unmotivated:

- **#1461** caps the aiter JIT build to the container's cgroup memory instead of the host's CPU count. It was filed separately because it is not tier3-specific — every tuner benefits. But the follow-up it needed (below) could not land until it did, and the reviewer would have had to hold both in their head anyway.
- **#1458** is the fused-MoE half. It was stacked on this branch, so it was never independently reviewable in the first place.

Folding them in also let me finish the one piece that was blocked on ordering: `tier3/sandbox.py` builds its own environment and calls `subprocess.run` directly, so it was the last path still asking for a job per host CPU. It is also the path least able to absorb an OOM-killed build — an aiter tuner that gets killed is a tuner that failed and can be re-run, whereas a generated tuner gets one sandbox run and a dead build is recorded as "the script the model wrote does not work". Wrong lesson, wrong evidence, about the leg of this pipeline that is hardest to judge. That is `5beadc63e`, with two tests: the cap reaches the child, and an explicit `env_overrides={"MAX_JOBS": ...}` still wins.

What the combined PR contains, in order:

1. `b0de15f07` — size the aiter JIT build to the container (was #1461)
2. – 9. — M0 + M1: resolve a deployed table back to its owner, widen the gate to an owner that produced nothing, wire the outcome back into `results`, and the four findings from running it on an MI355X
3. 10. – 14. — M2: fused-MoE demand, the MoE adapter, and the two dispatch findings (was #1458)
4. `5beadc63e` — the sandbox cap above

Local: `ruff format`/`check` clean, `3 failed, 940 passed, 8 skipped` — the three are pre-existing Windows `os.killpg` failures in `test_utils_extra.py`, present on `main` too.

jiaqiang-dot-liu:
Two things I'd want resolved before this merges. Everything else I found is non-blocking.

## 1. `fmoe_ck` becomes auto-selected on every MoE run, at top priority, for ~10min of likely-empty work

`evidence._moe_demand()` turns fused-MoE misses into a real entry in `demands`. `router._tuners_the_demand_says_are_needed()` on `main` already walks that list and adds a `TunerSpec` for every entry with a registered owner — with **no `miss_count` floor**, and `fmoe_ck` hardcoded to `priority=10`, ahead of everything else:

```python
priority=10 if name == "fmoe_ck" else 20,
estimated_minutes=15 if name == "fmoe_ck" else 20,
```

A single aiter_ck MoE miss in the log is therefore enough to schedule `fmoe_ck` first. This PR's own description says that tuner "burns 621-707s per GLM session and hands back `empty_output`".

That `empty_output` then feeds the second half: it becomes a `KIND_EMPTY` gap, `warrants_generated_tuner` now accepts `empty`, and `FORGE_TIER3_DISABLE` is unset by default — so the run continues into a model authoring session plus a sandbox GPU run.

Net effect after merge: an ordinary tuning run against any aiter_ck MoE serving log picks up ~10 minutes of probably-unproductive work at the head of the queue (eating `global_deadline` from tuners that would have produced something), then an LLM call and a sandbox run behind it. Both `FORGE_MOE_DEMAND_DISABLE` and `FORGE_TIER3_DISABLE` are opt-out, so this is the default path.

Suggest either landing both flags as opt-in for now, or adding a miss floor to the demand-driven selection in `router.py`.

Related note on reviewability: the description says "`router.py` now selects `fmoe_ck` from log evidence", but `router.py` is not in the diff. The behaviour change is real and entirely indirect, so there is nothing for a reviewer to look at — worth calling out explicitly in the description.

## 2. `_FusedMoeAdapter` mutates `AITER_CONFIG_FMOE` and never restores it

In `tier3/dispatch.py`:

```python
self._restore = os.environ.get("AITER_CONFIG_FMOE")
```

`_restore` is assigned in `__init__` and never read anywhere in the PR — the restore was clearly intended and never wired up. `_point_at()` writes `os.environ` directly on every call, so once an attempt finishes the variable is left pointing at `/tmp/forge_tier3_fmoe_*/candidate.csv` (whichever candidate or the empty baseline was last in play) for the remainder of the process, including report generation and any e2e validation that follows. The `mkdtemp` directory is not cleaned up either.

A `close()`/context manager on the adapter, called by the runner when the attempt ends, would cover both.

chennyiiis:
Both fixed in 00ecd5f0b — thanks, the second one would have been very hard to see from a report.

**1. `fmoe_ck` at the head of every MoE run.** You're right that the flags are both opt-out and that the cascade is one miss → 15 min → likely `empty_output` → a tier3 session. I took the floor rather than reverting the flags to opt-in, since opt-in is what kept this door shut for the last six months. `router._tuners_the_demand_says_are_needed` now requires `DEMAND_MIN_MISSES = 25` lookups before it books anyone — the same number `tier3/gate.py`'s `DEFAULT_MIN_MISSES` applies to the same demand report, so the two gates on the same evidence now agree. The demand is still recorded at any count; only the time booking is gated. Tests: 24 misses records the demand and selects nothing, 25 selects `fmoe_ck`. And the PR description now says outright that the router behaviour changes even though `router.py` isn't in the diff — that was my omission.

**2. `AITER_CONFIG_FMOE` left pointing at a temp CSV.** Worse than dead code, as you spotted: the process kept running with whatever candidate was tried last, through report generation and e2e, so a config the referee may have *rejected* could be what actually served, labelled as whatever the report said was deployed. `_FusedMoeAdapter.close()` restores the variable (deleting it when it was previously unset), clears aiter's three dispatch caches — without that, the restored name is not the table actually in use, `get_config_file`'s `lru_cache` still holds the candidate — and removes the `mkdtemp` dir. `_attempt_tier3` calls it in a `finally`, so a failed attempt cleans up too; `__enter__`/`__exit__` are there as well. Four tests: set → restored, unset → stays unset, workdir gone, context manager.

Bot findings all applied too (import style ×2, `staticmethod(_FakeTorch)` ×4, redundant `import re`).

Full `gemm_tune` suite 988 passed / 8 skipped, ruff clean. Not re-run on hardware: the GPU box this was accepted on has been released, and neither change touches a measured path — the floor is a selection decision and `close()` runs after the last measurement.

jiaqiang-dot-liu (approved):
LGTM

PR #1454, opened 2026-09-09
fix(gemm-tune): stop the demand ranking from deciding whether decode is tuned

jiaqiang-dot-liu:
## What happened

A 12h Qwen3-14B-FP8 / sglang / MI355X session with `KERNEL_OPT_BACKEND_ORDER=forge` produced **zero** KERNEL-phase KEEPs. The gemm-tune stage worked mechanically — it routed to the right op (`a8w8_blockscale_bpreshuffle`, the same op GEAK tunes on sglang), the artifact engaged (40 lookups hit it), and it won its microbenchmark at **1.86×** — but it handed the tuner **14 shapes, all at M ∈ {1024, 2048, 4096, 8192}, and no decode-band row at all**. On an ISL/OSL 1024/1024 workload that moved e2e by **+1.30%** and was reverted (`micro_decision: candidate_no_e2e_gain`).

For scale: GEAK's own gemm tuning on the same framework accepted at **+2.77%**, so this is a real shortfall against the reachable result, not against a hypothetical one.

## Root cause

`demand_shapes` ranks lookup buckets by `requests`, which counts **log lines**. `get_CKGEMM_config` is `lru_cache`-d, so a shape is logged once per process however often its kernel runs. Summing that per `padded_m` bucket scores a bucket by *how many distinct M happen to land in it* — and `nextPow2` buckets double in width. Replayed from that run's own `demand.json`:

| bucket (padded M) | requests | distinct raw M | rank |
|--:|--:|--:|--:|
| 2048 | 551 | 391 | 1–4 |
| 1024 | 176 | 128 | 5–8 |
| 4096 | 57 | 54 | 9–12 |
| 8192 | 19 | 12 | 13–16 |
| **1–64 (decode)** | **1–2** | **1–2** | **33–56 of 56** |

The metric is inverted relative to GPU time: M=64 runs once per output token (~327k executions in this bench) and logs 3 lines; a one-off chunked-prefill M=1203 runs once and also logs 3 lines. A budget of 14 then cut the decode band entirely.

`_ensure_decode_m_coverage` exists for exactly this failure — its docstring even describes it as "a micro win that regresses E2E because the throughput-dominant small-M decode GEMMs get the prefill-tuned tile". It never ran, because `_demand_input_csv(...) or _resolve_input_csv(...)` short-circuits the function the guard lives in, while `_resolve_input_csv`'s own docstring promises "every recorded source gets a decode-band guarantee".

## Changes

1. **Demand path gets the decode-band guarantee** (`_aiter_dense_common._demand_input_csv`). Applied *after* the budget cut on purpose: the decode band is mandatory, so it must not compete for a discretionary budget on a metric that cannot see it. It appends at most one row per (dispatch group, lookup bucket) it is missing.

2. **`demand_shapes` docstring corrected.** It claimed that summing per bucket "repairs" the tie problem; it does not, it substitutes bucket width for frequency. The quoted validation ("share of logged misses served": 95.6% @ budget 24) is the same log-line metric and structurally cannot see this failure, so it is now labelled as such rather than read as coverage of GPU time.

3. **split-K forwarding contract** (`_SPLITK_FORWARDING_LIBTYPES` + `_cap_splitk_to_serve_safe`). Adding `--splitK` to the bpreshuffle tuner is the obvious-looking fix here — split-K is worth **1.47× at decode M=64** on the non-preshuffle op (same-build A/B on this hardware). **It would have been wrong.** `gemm_a8w8_blockscale_bpreshuffle`'s ck/cktile wrappers accept no `splitK` argument at all; only `asm` does. That run's rows were `ck: 2, cktile: 12`, so 12 of 14 would have deployed a config benchmarked *with* split-K and served *without* it — slower than measured, with every engagement gate still green. The table records which `(op, libtype)` pairs forward it, fails closed for anything unverified, and the cap now runs unconditionally so such a row cannot reach serving. `splitK=0` rows take a fast path, so a table without split-K pays nothing.

4. **split-K trial is op-aware** (`aiter_splitk_validate`). It hardcoded `gemm_a8w8_blockscale_ck` behind an equality check on `script_key`, with a comment noting other ops "would be validated against the WRONG kernel". A registry makes an unregistered op answer "unknown" so the caller keeps its static cap, instead of being handed another kernel's limit — which would either license a splitK the real kernel rejects (engine-init crash) or tighten one it accepts (silent loss).

5. **`zero_hit` vs `inconclusive_no_hit_logging`** (`evidence.parse_log`). That run's demand report labelled 3428 conclusive misses `inconclusive_no_hit_logging`. A caller that set `AITER_LOG_TUNED_CONFIG` itself can now say so, mirroring `apply_verification.verify_applied`, which already made this distinction. It matters: an op with genuinely zero shipped coverage is the strongest reason to tune it (verified — all 8 shipped bpreshuffle tables have **0 rows** for this model's four (N,K)), while "not recorded" is no reason at all.

## Verification

Replayed against the reverted run's real `demand.json`:

```
before: 14 rows, M [1024, 2048, 4096, 8192], decode rows 0
after:  26 rows, M [16, 32, 64, 1024, 2048, 4096, 8192], decode 12 / prefill 14

  N=5120   K=5120   decode=[16, 32, 64]  prefill=[1024, 2048, 4096, 8192]
  N=5120   K=17408  decode=[16, 32, 64]  prefill=[1024, 2048, 4096]
  N=7168   K=5120   decode=[16, 32, 64]  prefill=[1024, 2048, 4096, 8192]
  N=34816  K=5120   decode=[16, 32, 64]  prefill=[1024, 2048, 4096]

decode grid [1, 4, 16, 32, 64]: unserved (group, M) pairs = none
```

Prefill selection is unchanged; the decode band is additive. 26 shapes at the ~40 s/shape this op measured is ~1040 s, inside the run's 1216 s lane budget.

- `src/kernelforge`: **4548 passed**, 1 skipped, 8 xfailed
- affected Hyperloom suites (`orchestrator/measurement`, `test_coordinator_gemm_promote_units`, `test_fmoe_ck_coverage`, `test_gemm_shape_coverage`): **248 passed**
- `ruff check` / `ruff format --check`: clean
- New tests: decode band survives a budget that only funds prefill; guarantee holds per dispatch group; prefill not crowded out; forwarding contract (forwarded / unforwarded / drop / fail-closed / `None` back-compat); op-awareness (unregistered op never dispatches another op's kernel); `zero_hit` verdict incl. that the flag cannot manufacture a hit

## Not in this PR

- **Per-shape cost constant.** `_DEMAND_PER_SHAPE_COST_S = 74` against ~40 s measured on this op at `mp=1` (~1.85× pessimistic), and it does not scale with `mp`. Retuning a shared constant from one op on one box could truncate other ops' runs, and with the decode band no longer competing for the budget the remaining harm is a shorter prefill tail. Wants more data points.
- **Duplicate op across lanes.** The Kernel Rewrite Controller rejected one of its two tasks as `operator task is already published` and then spent 2h09m / $41.91 on `gemm_a8w8_blockscale_bpreshuffle` for `patch_count: 0`, after gemm-tune had already failed that op's e2e gate 561 s in. Feeding the gemm-tune verdict in as a suppression signal is orchestration policy in `opportunity_agent`, not a correctness bug in this path.
- **`sglang_dense_bf16` being skipped as a fallback was correct** and is deliberately unchanged: the baseline log carries **0** bf16 misses (all 3428 are bpreshuffle), so bf16 dense is already covered by the shipped tables and tuning it would write a table nothing reads.

## Risk

`_cap_splitk_to_serve_safe` now runs on every dense tuner instead of only those passing `--splitK`. For the existing `--splitK` caller (`a8w8_blockscale`) behaviour is unchanged: its forwarding set is `{ck, cktile}`, which is what its production dispatch actually forwards. For every other dense tuner the tables today carry `splitK=0` everywhere (no `--splitK`, so aiter sets `maxsplitK=0`), and those rows take the no-op fast path.

### DIFF
--- a/src/kernelforge/gemm_tune/aiter_splitk_validate.py
+++ b/src/kernelforge/gemm_tune/aiter_splitk_validate.py
@@ -1,6 +1,16 @@
 # SPDX-FileCopyrightText: 2026 Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
-"""Per-shape production split-K support, by trial-dispatch."""
+"""Per-shape production split-K support, by trial-dispatch.
+
+The trial is per-op: a limit measured on one op's kernel says nothing about
+another's, so :data:`_TRIAL_OPS` registers the production callable per op and an
+unregistered op answers ``None`` (caller keeps its static cap). Answering with
+the wrong kernel's limit would license a splitK the real kernel rejects (crash)
+or tighten one it accepts (silent loss). ``a8w8_blockscale_bpreshuffle`` is
+absent on purpose: its ck/cktile wrappers take no ``splitK`` at all, so
+``_SPLITK_FORWARDING_LIBTYPES`` in ``tuners._aiter_dense_common`` is the gate
+that matters there.
+"""
 
 from __future__ import annotations
 
@@ -29,7 +39,7 @@ def _resolve_device(gpu_ids: str = "") -> str:
     return f"cuda:{first}"
 
 
-def _supports(m: int, n: int, k: int, split_k: int, device: str = "cuda") -> bool:
+def _supports_a8w8_blockscale(m: int, n: int, k: int, split_k: int, device: str = "cuda") -> bool:
     """True if the production a8w8_blockscale CK kernel dispatches (m,n,k,split_k)."""
     import torch  # noqa: PLC0415
     import aiter  # noqa: PLC0415
@@ -47,17 +57,44 @@ def _supports(m: int, n: int, k: int, split_k: int, device: str = "cuda") -> boo
     return True
 
 
-def max_supported_splitk(m: int, n: int, k: int, ceiling: int = 6, device: str = "cuda") -> int | None:
-    """Max splitK in ``0..ceiling`` the production kernel accepts for (m,n,k)."""
+#: Ops whose production split-K limit this module can measure, keyed by the
+#: tuner ``script_key``. Absent => no trial, caller keeps its static cap.
+_TRIAL_OPS = {
+    "a8w8_blockscale": _supports_a8w8_blockscale,
+}
+
+
+def supported_ops() -> frozenset[str]:
+    """Op keys :func:`make_support_fn` can build a real trial for."""
+    return frozenset(_TRIAL_OPS)
+
+
+def max_supported_splitk(
+    m: int,
+    n: int,
+    k: int,
+    ceiling: int = 6,
+    device: str = "cuda",
+    op: str = "a8w8_blockscale",
+) -> int | None:
+    """Max splitK in ``0..ceiling`` the production kernel accepts for (m,n,k).
+
+    ``None`` when ``op`` has no registered trial, or when the splitK=0 control
+    fails, so the caller keeps its static fallback rather than trusting an
+    unvalidated trial.
+    """
+    probe = _TRIAL_OPS.get(op)
+    if probe is None:
+        return None
     try:
-        if not _supports(m, n, k, 0, device=device):
+        if not probe(m, n, k, 0, device=device):
             return None
     except Exception:  # noqa: BLE001 — no GPU / import error / tensor issue
         return None
     best = 0
     for sk in range(1, max(0, ceiling) + 1):
         try:
-            ok = _supports(m, n, k, sk, device=device)
+            ok = probe(m, n, k, sk, device=device)
         except Exception:  # noqa: BLE001 — treat a hard error as "unsupported"
             ok = False
         if not ok:
@@ -66,15 +103,23 @@ def max_supported_splitk(m: int, n: int, k: int, ceiling: int = 6, device: str =
     return best
 
 
-def make_support_fn(ceiling: int = 6, gpu_ids: str = ""):
-    """Return an (m,n,k)->int|None callable memoized per shape for reuse as the ``support_fn`` of ``_cap_splitk_to_serve_safe``."""
+def make_support_fn(ceiling: int = 6, gpu_ids: str = "", op: str = "a8w8_blockscale"):
+    """Return an (m,n,k)->int|None callable memoized per shape for reuse as the ``support_fn`` of ``_cap_splitk_to_serve_safe``.
+
+    ``op`` selects the production kernel to trial; an unregistered op answers
+    ``None`` for every shape so the caller falls back to its static cap instead
+    of being handed another op's limit.
+    """
+    if op not in _TRIAL_OPS:
+        return lambda m, n, k: None
+
     cache: dict[tuple[int, int, int], int | None] = {}
     device = _resolve_device(gpu_ids)
 
     def _fn(m: int, n: int, k: int):
         key = (m, n, k)
         if key not in cache:
-            cache[key] = max_supported_splitk(m, n, k, ceiling=ceiling, device=device)
+            cache[key] = max_supported_splitk(m, n, k, ceiling=ceiling, device=device, op=op)
         return cache[key]
 
     return _fn
--- a/src/kernelforge/gemm_tune/cli.py
+++ b/src/kernelforge/gemm_tune/cli.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 import sys
 import time
 from pathlib import Path
@@ -58,7 +59,13 @@ def _demand_from_serving_log(server_log: str, output_dir: Path) -> str:
     try:
         from .evidence import moe_dispatch_keys, parse_log_file, write_demand
 
-        report = parse_log_file(server_log)
+        # Hyperloom's workload env sets AITER_LOG_TUNED_CONFIG=1 for every
+        # serving run, and it is inherited here, so when it is on we can say a
+        # zero-hit table really had zero coverage instead of leaving the verdict
+        # inconclusive. Absent/0 stays unknown -- an operator-supplied log may
+        # have been produced without it.
+        hit_logging = os.environ.get("AITER_LOG_TUNED_CONFIG", "").strip() not in ("", "0")
+        report = parse_log_file(server_log, hit_logging=hit_logging or None)
     except Exception:  # noqa: BLE001 - deriving demand must never fail tuning
         log.debug("could not parse %s for demand", server_log, exc_info=True)
         return ""
--- a/src/kernelforge/gemm_tune/evidence.py
+++ b/src/kernelforge/gemm_tune/evidence.py
@@ -214,8 +214,16 @@ def _record_moe_key(moe: dict[str, Any], parts: list[str], *, miss: bool) -> int
     return token
 
 
-def parse_log(text: str) -> dict[str, Any]:
-    """Parse a serving log into demands, an apply verdict and dispatch facts."""
+def parse_log(text: str, *, hit_logging: bool | None = None) -> dict[str, Any]:
+    """Parse a serving log into demands, an apply verdict and dispatch facts.
+
+    Args:
+        text: The serving log.
+        hit_logging: Whether ``AITER_LOG_TUNED_CONFIG`` was on for the run that
+            wrote this log. ``None`` means unknown, which keeps a zero-hit
+            result ``inconclusive_no_hit_logging``; ``True`` resolves it to
+            ``zero_hit``. A caller that set the variable itself should say so.
+    """
     demands: dict[str, Demand] = {}
     key_counts: dict[str, dict[tuple, int]] = {}
     hits = 0
@@ -380,7 +388,7 @@ def parse_log(text: str) -> dict[str, Any]:
             "hit": hits,
             "miss": misses,
             "hit_ratio": (hits / total) if total else None,
-            "verdict": _apply_verdict(hits, misses),
+            "verdict": _apply_verdict(hits, misses, hit_logging),
         },
         "merged_tables": sorted(set(merged)),
         "consulted_tables": sorted(consulted),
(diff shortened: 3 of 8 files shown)


xiaofei-zheng:
## Review

Read the whole diff, replayed the demand path against synthetic `demand.json`s, and checked the split-K facts against the installed aiter (`/usr/local/lib/python3.12/dist-packages/aiter/ops/gemm_op_a8w8.py`).

Verified as correct:

- `_SPLITK_FORWARDING_LIBTYPES` matches the runtime. `gemm_a8w8_blockscale` forwards `splitK` on `ck` and `cktile` (and asserts on anything else); `gemm_a8w8_blockscale_bpreshuffle` passes only `kernelName` to its `ck`/`cktile` wrappers and forwards `splitK` only on `asm`. The table is right, and failing closed for unlisted ops is the right default.
- Running `_cap_splitk_to_serve_safe` unconditionally is safe for the other dense tuners: in the aiter tune scripts `maxsplitK` is `... if useSplitK else 0`, so without `--splitK` every row is `splitK=0` and takes the fast path.
- The `zero_hit` verdict cannot leak into a keep/revert decision. `apply_verification.verify_applied` recomputes its own verdict from `hit`/`miss`; the only consumers of `evidence.parse_log`'s `apply_verdict["verdict"]` are log lines in `cli.py`. `hit_logging` defaults to `None` for every existing caller.
- The decode guarantee itself is per dispatch group and preserves the original rows verbatim, and `work_dir` is per-tuner (`base.py:171`), so the fixed `decode_covered_dense.csv` name cannot collide across tuners.

### Blocking

**1. The decode rows are appended with no time accounting, so the tuner can now overrun its hard timeout and return nothing.**

`_demand_input_csv` (`src/kernelforge/gemm_tune/tuners/_aiter_dense_common.py:282-347`) spends the *full* budget on the prefill selection (`demand_shapes(entry, limit=budget)`) and then appends up to one row per (dispatch group, missing lookup bucket). Nothing is reserved — the docstring's "The decode band is reserved before the budget is spent" does not match the code, and the comment at the return says the opposite ("Applied AFTER the budget cut on purpose").

`ctx.timeout_s` is not a soft target: it is the `communicate(timeout=...)` deadline in `utils.run_subprocess`, which returns `124`, and `run_aiter_dense_tuner` maps `rc == 124` to `TuneResult(status="failed", error_class="timeout")` with no candidate. Per-shape isolation (which could salvage partial work) is opt-in and off by default (`tune_robustness.is_isolation_enabled`). So an overrun does not degrade the table — it loses the lane, including the prefill rows that used to complete.

Replayed on this branch:

```
# 6 dispatch groups, 60 prefill M each, conc=64
timeout_s=1216  budget=14  -> 30 rows written   (14 prefill + 16 decode)
                              30 * 74s (_DEMAND_PER_SHAPE_COST_S) = 2220s  vs 1216s
                              30 * 40s (the rate measured in the PR body) = 1200s vs 1216s

# 4 dispatch groups, thorough=True
timeout_s=3600  budget=8   -> 20 rows written   (8 prefill + 12 decode)
                              20 * 420s (_DEMAND_PER_SHAPE_COST_THOROUGH_S) = 8400s vs 3600s
```

The 4-group Qwen3-14B case in the PR body fits only because `_DEMAND_PER_SHAPE_COST_S` happens to be ~1.85x pessimistic for that op; the margin is consumed by group count, and it is gone at 6 groups even at the optimistic 40 s/shape. In `thorough` mode the budget is 8 shapes and the guard can more than double it, which no amount of pessimism in the constant covers. Note also that `_resolve_input_csv` deliberately does *not* run the decode guard in `thorough` mode (it runs `_augment_with_config_m_values` instead), while the new demand path runs it unconditionally.

Suggested fix: make the reservation real — estimate the decode rows first (distinct dispatch groups x missing buckets) and pass `limit=max(1, budget - reserved)` to `demand_shapes`, so the total still respects the time the lane actually has.

jiaqiang-dot-liu:
Fixed in 80a52f0b6. You were right on every point: the reservation was not real, and `ctx.timeout_s` being the `communicate(timeout=...)` deadline with no salvage path is what makes it a lane-loss rather than a shorter table.

## What changed

`_demand_shapes_within_budget` walks the ranking and takes a shape while the total — shapes taken plus the band rows their dispatch groups still lack — stays inside the budget. `_decode_band_buckets` is factored out of `_ensure_decode_m_coverage`, so sizing the reservation and appending the rows read the same plan and cannot drift.

Replayed on your two counterexamples:

```
# 6 dispatch groups, conc=64, timeout_s=1216, budget=14
before: 30 rows (14 prefill + 16 decode)
after:  14 rows (5 prefill + 9 decode, 3 of 6 groups)

# 4 dispatch groups, thorough=True, timeout_s=3600, budget=8
before: 20 rows (8 prefill + 12 decode)
after:   8 rows (2 prefill + 6 decode, 2 of 4 groups)
```

No group that gets tuned has a hole in its band in either case.

## Why not `limit = max(1, budget - reserved)`

I tried the suggested shape first and dropped it, because `reserved` has to be estimated before the selection exists, and the only pre-selection estimate available is the group count of the *whole* demand report — which systematically over-reserves, since the selection usually touches only some of those groups. On the 6-group case: `reserved = 6 x 3 = 18`, so `limit = max(1, 14 - 18) = 1`, one prefill shape, one group in the CSV, 3 band rows — 4 rows against a budget of 14. Safe, but it leaves 10 shapes of the lane's time unspent.

A prefix cut sized by the actual selection does better but still strands budget: it stops at the first prefix that does not fit, giving 12 rows (3 prefill + 9 band) on the same case. The greedy fill is what gets to 14/14 — a shape that would open a new group carries that group's whole band with it, so it is passed over in favour of a lower-ranked shape in a group already paid for, instead of ending the walk.

The concession is trimming **groups**, never the band inside a group: a group holding part of its band serves the uncovered decode M with a prefill tile, which is the regression the band exists to prevent, while a group left out entirely keeps whatever the shipped tables already give it. When not even the top-ranked group's band fits, it is still tuned whole and the overrun is logged rather than silently shipping a band with holes.

One honest consequence worth flagging: with the reservation real, `_DEMAND_PER_SHAPE_COST_S = 74` no longer costs only the prefill tail — it now costs group coverage. At the measured ~40 s/shape the same 1216 s lane would fund 27 shapes rather than 14, which covers all 6 groups instead of 3. That strengthens your original point about the constant; it is still out of scope here for the reasons in the PR body, but it is now the binding constraint rather than a margin.

## On the thorough-mode asymmetry

I looked at this and deliberately left the behaviour as it is. `_augment_with_config_m_values` calls `compute_dense_m_values(..., thorough=True)`, which seeds from `_SGLANG_CUDAGRAPH_BS_THOROUGH = [1, 2, 4, 8, 12, 16, ...]` and writes every M against every `(N, K)` — so the thorough `_resolve_input_csv` path already covers the decode band per dispatch group, and running the guard there would be redundant, not missing. The demand path has no augmentation step at all, so it is the only source of decode coverage in both modes, which is why it runs the guard unconditionally. What was wrong was that it ran unaccounted; in thorough mode it is now sized the same way (budget 8 -> 2 prefill + 6 band).

Happy to change this if you read the intent of the two paths differently.

## Tests

New `TestTheBandIsPaidForOutOfTheBudget`: the budget invariant parametrized over 5 group counts x 3 modes, every tuned group gets a whole band, the prefill tail yields as the group count grows, an unaffordable group is skipped rather than ending the walk, and an unaffordable band is still kept whole.

Two existing budget parameters moved because those values are no longer affordable under honest accounting — `test_decode_rows_are_per_dispatch_group` 2 -> 8 (two prefill shapes plus two bands of three) and `test_decode_band_survives_a_budget_that_only_funds_prefill` 1 -> 4 (one band plus one prefill row). The assertions are unchanged.

All checks pass, including both coverage jobs and all eight test shards.

The CodeQL cyclic-import note on the new local import is a pre-existing cycle — `dense_shapes` reaches into `tuners._aiter_dense_common` for `_aiter_fp8_dtype_str` while three existing local imports in `_aiter_dense_common` already go the other way. Breaking it properly means moving the aiter dtype helpers out to their own module, which is a separate concern from this fix.

xiaofei-zheng (approved):
## Re-review of 80a52f0b6

The budget finding is fixed, and fixed better than the shape I suggested — reserving from the whole report's group count would have over-reserved, and the greedy fill does spend the lane.

What I checked:

- **Re-ran my two counterexamples on this branch.** 6 groups / `timeout_s=1216`: 14 rows (5 prefill + 9 band, 3 of 6 groups) against budget 14, was 30. 4 groups / `thorough=True` / `timeout_s=3600`: 8 rows (2 prefill + 6 band, 2 of 4 groups) against budget 8, was 20. Also 4 groups at `conc=256` (28 rows, budget 47) and a non-thorough 4-group 3600 s lane (20 rows, budget 47) — under budget in every case, and no group that gets tuned has a hole in its band (checked by replaying `_dispatch_lookup_ms` over `compute_decode_m_values(conc)` for every group in the emitted CSV).
- **The deliberate overrun is bounded.** With `_DEMAND_MAX_SHAPES_ENV`-style tiny budgets (budget 2) the fallback emits `ranked[:1]` plus that one group's band — 4 rows, and the warning fires. That is 1 + at most one group's band, not the unbounded `budget + 3 x groups` this replaces, so the failure mode is gone rather than moved.
- **Reservation and emission cannot drift**: `_decode_band_buckets` is the single planner, and both callers feed it the same `(N, tuned M)` view — `_demand_shapes_within_budget` keys groups on `(N, K)` while `_ensure_decode_m_coverage` keys on `(N, K, q_dtype_w)`, which coincide here because the demand CSV writes one `q_dtype_w` for the whole file. `int(shape["N"])` in the new sizing loop is safe: `demand_shapes` drops any key whose M/N/K is not an int.
- **Nothing else moved.** AST-compared every file in the PR against the head I reviewed (`fb9230923`), ignoring docstrings: `aiter_splitk_validate.py`, `evidence.py` and `cli.py` are logic-identical, and in `_aiter_dense_common.py` the only changed definitions are `_demand_input_csv`, `_demand_shapes_within_budget`, `_decode_band_buckets` and `_ensure_decode_m_coverage`. So my earlier verification of the split-K forwarding contract against the installed aiter, of the unconditional cap being a no-op for the tuners that pass no `--splitK`, and of `zero_hit` not reaching any keep/revert decision, still holds.
- `src/kernelforge/gemm_tune/tests/test_demand_budget.py`: 33 passed. Full CI on this head is green.

No blocking issues. LGTM.

Agreed on the thorough-mode asymmetry and on `_DEMAND_PER_SHAPE_COST_S` — the constant is now the binding constraint on group coverage rather than on the prefill tail, which is a better place for it to be wrong, but it is worth a follow-up with real per-op timings.

PR #1455, opened 2026-09-09
refactor(breakdown): record the Session Breakdown at author time, and retire the read side it replaced

haishuok0525:
## What this is

The Session Breakdown used to be assembled at export time: a set of collectors
walked the session directory and re-derived what had happened from whatever
artefacts survived. This branch finishes moving every fact to the moment it is
decided -- the recorder writes a durable fragment at author time -- and then
deletes the collectors that were answering the same questions a second time.

Two answers to one question is the failure mode being removed here. The export
side could disagree with the run, and when the two disagreed the exported
breakdown was the one people read.

Net: 197 files, +33320 / -32957. The exporter went from 34 `collect_` calls to
11.

## What is gone

Nine legacy top-level keys and their collectors (`geak`, `kernels`, `timeline`,
`telemetry`, `roofline`, `explore`, `optimizations`, `decision`, `v6_stages`),
the pre-recorder fallback projections that fired when fragments were absent, and
the schema TypedDicts describing V5 shapes with no V6 wire presence.

`timeline` is now pure playback of the durable events. It used to be a
projection whose single failure discarded every recorded event.

Terminal configuration moved to close-out: `outcome.final` reads
`close.final_recipe` instead of walking the reports directory. No single event
can hold the whole terminal answer -- throughput, latencies, action path, server
args, env -- and close-out is when it is known.

`write_decision_trace` moved to `orchestrator/trace/decision_trace.py`, next to
its real readers (the Langfuse emitter and the backfill tool). It was never part
of the session-breakdown contract. The Langfuse receipt stays behind in a
collector of its own, since another process writes it.

Write-side orphans went too: `record_phase_event` and the explore_search and
roofline item snapshots wrote fragments no V6 collector reads.

## One bug fixed along the way

Tool provenance was being silently discarded. It was written into the `metadata`
singleton under a second producer -- but a singleton is one file per producer and
assembly keeps only the newest, and the Coordinator reissues its own metadata
write on every state save, so it was always the newer of the two and dropped the
tool versions whole. The rows now go to a `versions` item stream keyed by tool
name, folded into `metadata.versions.tools` at assembly, which is what the
section registry already claimed happened.

## Merge note: #1321 was preserved by hand

main's #1408 moved kernel selection into KernelForge and met this branch on the
same code in nine places. Eight resolutions were a straight question of which
side still has a reader. One was not:

#1321 fixed the run's GPU pin not reaching GEAK, which made its servers land on
physical card 0 and collide with a foreign co-tenant -- surfacing as a `no_gain`
indistinguishable from a real one (#1312). That fix's core is elsewhere and
merged clean, but it also surfaced `gpu_ids` / `gpu_pin` in the geak collector so
the degradation was diagnosable from the session artifacts, and that collector is
deleted here. Those two fields are now recorded onto the geak handoff event at
author time, where the rest of the handoff already lands, with tests. Without
that port this merge would have silently regressed #1321's diagnosability.

## Validation

A 6h formal run on 1x MI355X (Qwen3-8B, TP=1, bf16) exercised the full pipeline
end to end. Baseline 7357.9 tok/s; an explore variant (`--kv-cache-dtype
fp8_e4m3`) reached 8622.3 tok/s (+17.18%), so the run stopped early on
`target_reached`. The exported breakdown carried V6 keys only, `outcome.final`
agreed with `validation.validated_total_gain_pct`, the timeline was 13 durable
events across 4 phases with no reconstruction, and 108 fragments across 31
section types assembled cleanly.

Suite: 12172 passed. 14 failures are environment-bound (LLM credentials,
multi-node cluster) and fail identically on an unmodified checkout of either
parent -- verified against a clean worktree at both.

`ruff check`, `ruff format --check` (with the 0.16.2 the workflow pins) and
byte-compile all pass.


ZhengGong-amd:
## Review notes (written against `4a82360fc`, before `c27e3e61c`)

I wrote these up while reviewing the branch at `4a82360fc` and never posted them. Since then `c27e3e61c` has landed and resolves a good number of them — I've marked those inline rather than dropping them, because the reasoning behind each is still the reason the fix was the right one, and a few are only partly closed.

Legend: **[Resolved in c27e3e61c]** · **[Partly resolved]** · **[Still open]**

---

# 1. Is the change complete for the goal it states?

**Not completely — and three of the gaps happen to contradict the PR's own central argument.**

### [Still open] "Two answers to one question" was not eliminated, only relocated

`exporter.build()` still has four collectors rebuilding facts at export time:

```python
# breakdown/exporter.py:88-112
session_section = _merge_session(
    assembled.get("session"),
    _safe_collect("session", lambda: collectors.collect_session(sd, state, manifest, warnings), warnings),
)
workload = _safe_collect("workload", lambda: collectors.collect_workload(state, manifest, warnings), warnings)
model_info = _safe_collect(
    "model_info", lambda: collectors.collect_model_info(state, warnings), warnings, default={}
)
langfuse = _safe_collect("langfuse", lambda: collectors.collect_langfuse(sd, manifest, warnings), ...)
```

The `session` block is a textbook case of the very pattern being removed: the recorder writes a fragment, the collector rebuilds a second one, and `_merge_session` (`exporter.py:37`) overlays them field by field — "fragment wins when non-empty." This is exactly the `_pick` / `_merge_phase_timeline` shape the PR description names as the thing to delete; it survived under a different name. On top of that, `manifest.py:334` adds a *new* `record_metadata_identity(sd, manifest)` call, so image/host/pid now have both an author-time record and a collector rebuild.

The stated justification — "task config is authored before the recorder exists, so read it from the launch inputs" — holds for `workload` and `model_info`. It does not hold for `session`: `ended_at_utc`, `tick_count` and `elapsed_minutes` are all runtime facts. Either the recorder owns `session` outright and `collect_session` goes, or this block is openly declared derived-at-export and the fragment goes. Keeping the merge keeps the thing this PR set out to eliminate.

### [Resolved in c27e3e61c] Two mutually inconsistent `warnings` fields in the output

```python
# breakdown/exporter.py:113-189
v6_warnings = list(warnings)          # copied here
...                                    # every V6 collector afterwards appends only to v6_warnings
metadata["warnings"] = list(v6_warnings)
breakdown = { ..., "warnings": warnings }   # top-level: only what existed before the copy
```

The top-level `warnings` was a strict subset of `metadata.warnings`. Two identically-named fields in one document giving different answers means which one a consumer reads is luck. Since `metadata.warnings` is already described in the code as the only place a V6 failure is allowed to surface, the top-level one should go rather than persist as a partial copy.

> Confirmed fixed: the top-level key is gone from `build()`, `metadata.warnings` is now the single channel.

### [Resolved in c27e3e61c] `build()` claimed to be pure but writes to disk

```python
# breakdown/exporter.py:61-116
def build(session_dir: Path | str) -> dict[str, Any]:
    """Build a complete :class:`SessionBreakdown` for ``session_dir`` (pure; reads disk)."""
    ...
    _safe_collect("timeline_finalize", lambda: finalize_events(sd), v6_warnings, default=[])
```

`finalize_events` binds a `session_scope(session_dir)` and calls `finish_event(...)` for every unclosed event, **writing new fragments** (`recorder/event_finalize.py:139-171`). So `build()` mutates the session it is reading. `tools/dump_session_breakdown.py`, a read-only diagnostic, therefore changes the state of the session being diagnosed — a second dump does not match the first. The `(pure; reads disk)` docstring was simply wrong.

> Confirmed fixed, and fixed better than what I would have suggested: rather than moving `finalize_events` out, the docstring now states *"Not pure: … this closes any event whose phase was killed before it could close itself"* and argues the close is idempotent so repeated builds still agree. Stating the contract is the stronger answer.

### [Resolved in c27e3e61c] `critic` / `robustness` were recorded but nothing rendered them

`build()` emitted two new top-level keys, `critic` and `robustness` (`exporter.py:186-187`), and the assembler folded the data correctly via `_compose_critic` / `_compose_robustness`. But `reporters/compose.py:32-41` had neither:

```python
SECTION_GROUPS: list[tuple[str, list[str]]] = [
    ("Session & Workload", ["session", "workload"]),
    ("Performance Results", ["baseline", "final", "roofline", "optimizations", "attribution"]),
    ("Capability Search", ["capability_summary", "param_search"]),
    ("Kernel Optimization", ["kernel_lifecycle"]),
    ("Run Trace", ["phase_timeline"]),
]
```

`git grep '"critic"\|"robustness"' -- reporters/` returned nothing. The old `_renderers/critic_robustness.py` was deleted and no replacement renderer was written. The result: the Critic's per-iteration rulings and the Robustness signals stopped appearing in the markdown report at all, while the underlying data was not only still there but better structured than before. This was not "the data is gone so the renderer should go" — it was the renderer going while the data stayed. The one outright functional regression in the PR.

> Confirmed fixed: `_renderers/critic.py` and `_renderers/robustness.py` added, `("Review & Robustness", ["critic", "robustness"])` group added, with `test_critic_robustness_renderers_unit.py` covering it.

---

# 2. Is there a simpler way? Was existing infrastructure reused?

### [Still open] The instrumentation has no boundary — it grows inside business functions

The orchestrator gained a **net +3312 lines** and **41 new `_record_*` shim functions**. `loop/intent_router.py` went from 1170 to 1566 lines, and *every* function it gained is a recorder adapter: `_record_config_proposal`, `_record_phase_proposal`, `_record_phase_proposal_review`, `_record_critic_review`, `_record_phase_proposal_outcome`, `_record_review_outcome`, `_record_trace_analyze_on_timeline`, plus the helpers `_variant_review_rows`, `_review_subject`, `_phase_scope`. A routing module is now a third observability code.

The clearest sign of a missing abstraction is `phases/prelude.py`: `_record_warm_replay_skip(...)` is called **14 times** in that one file, once at each early-return guard in warm replay, 4-8 lines each, wedged between the state write and `state.save()`. The skip code is a derivable property of the guard condition; it is now hand-written fourteen times.

The accounting: 8669 lines of collectors deleted, replaced by 4144 lines of recorder **plus 3312 lines of instrumentation inside business logic** — the mechanism itself shrank by roughly 1200 lines. What "net −2037" actually means is that the complexity did not disappear; it moved out of an isolated read-side package and into the orchestrator's hot paths.

A note on the 3846 lines removed from `instrument.py`, since "it all moved" is not quite the right reading: roughly 1600 lines were the V4 schema layer (`record_operation` / `record_measurement` / `record_adoption` / `record_artifact` / `record_trace_event`) **deleted outright**, roughly 1800 moved into the per-domain modules, and roughly 550 are newly written. The file went from 3957 to 285 lines, but it is **not** a pure façade — it still holds `snapshot_state_sections` and `record_backend_versions_and_timeline`, both coordinator-specific. The split is a genuine improvement; only the framing needs correcting.

If the author-time model is to stay, the right shape is **one event bus and one subscriber layer**: business code emits a domain event (it already has `phase` / `action` / `proposal` as first-class concepts), and the recorder, as a subscriber, translates events into fragments at a single boundary. Then the orchestrator's diff is one `emit` line per site rather than 41 shims that each know the fragment shape. As it stands, every business module is coupled to the breakdown's wire shape — changing a field on one section means editing business files.

### [Partly resolved] The read-back is O(n²) and runs on the write path

`Recorder.record_item` writes **one file per row** into a single flat directory (`recorder.py:262-283`). And `event_parts((section,))` was:

```python
# breakdown/recorder/assembler.py:375-380
def event_parts(sections: tuple[str, ...]) -> dict[str, list[dict[str, Any]]]:
    assembled = assemble_parts(bound_session(), warnings=[], keep_event_rows=True)
    return {section: ... for section in sections}
```

`assemble_parts` globs `*.json` and **parses the entire spool directory**, then discards everything but one section. `phase_event.py`'s `_finish` / `_open_segment` / `_action_event` / `_proposal_event` (`phase_event.py:565-655`) each call it before writing, just to locate which event a row belongs to. A 6-hour run produces thousands of fragment files, and every record re-reads all of them — expensive in general, and especially so on a networked filesystem.

`finalize_events` is the extreme case: ten event types each call `event_parts`, so an export costs at least ten full spool parses (`event_finalize.py:145`).

`event_parts` also passes `warnings=[]`, so **assembly warnings raised during read-back are discarded outright.**

The right fix is an index in the spool (or event-id subdirectories) so that "which event does this row belong to" is an O(1) lookup instead of a full replay.

> `c27e3e61c` narrows the glob by section prefix and takes an optional `event` argument, which the commit message measures at ~6x less work on 18k-fragment sessions. That is a real improvement, but each write still re-reads a filtered subset with no index, so the cost still grows with session length.

### [Resolved in c27e3e61c] The same constants were copied three times

The `*_EVENT_SECTIONS` tuples were defined ten times in `assembler.py:246-330` and copied again in `phase_event.py:659`, `enablement_event.py:512` and `stack_event.py:341`. The code admitted it:

```python
# breakdown/recorder/phase_event.py:657-659
#: Every section a phase event assembles from. Declared here as well as in the
#: assembler so :func:`_finish` reads its parts without closing an import cycle.
```

"Copied to dodge an import cycle" records a design problem as a comment. Event types were enumerated in two further places (`event_finalize.py:_EVENT_TYPES` and `session/sbd_v6.py:_EVENT_TYPES`). AGENTS.md's *Derive over hardcode — a single computed source beats duplicated constants* is about exactly this. The fix is to sink the section registry into a dependency-free module (or derive it from `SECTION_SHAPES`) and have all four read from it.

> Confirmed fixed via the new `recorder/sections.py` leaf module. `_EVENT_TYPES` is still enumerated in two places (`event_finalize.py:55`, `session/sbd_v6.py:47`).

---

# 3. Redundant logic, excessive fallback, silent failure

### [Partly resolved] ~54 net-new broad exception swallows, all at DEBUG level

In production code (tests excluded), `except Exception` / `except BaseException` / `suppress(Exception)`: **119 added, 65 removed, net +54.** Densest in `phase_event.py` (+14), `enablement_event.py` (+11), `phases/framework.py` (+10), `phases/kernel.py` (+8/−13), `loop/intent_router.py` (+7), `tool_versions.py` (+6), `close_out.py` (+6).

AGENTS.md states: *No new broad `except Exception` / bare `except` — catch the specific error or let it raise.* This PR adds 54 in one go.

Most take this shape:

```python
# breakdown/recorder/phase_event.py:164-165
except Exception:  # noqa: BLE001 — a phase change outranks its own record
    log.debug("phase event: entry record failed", exc_info=True)
```

The rationale ("recording must not affect the run") is defensible on its own. The problem is what it becomes **in combination with deleting the collectors**:

- Before: recorder write fails → the collector rebuilds it from artifacts at export → the fact is still in the report.
- After: recorder write fails → **nothing takes its place** → that fact silently vanishes from the report.

And `metadata.warnings`, the channel that already exists for exactly this, is never used — `git grep 'warnings' recorder/` only hits the assembler's **read** side. The module docstring in `recorder/__init__.py` states the semantics outright:

> *Recording is best-effort by design, so a fact that never arrived leaves nothing behind to explain itself.*

In a world with a fallback, "best-effort and silent" is reasonable. In a **single-source** world it means the report is silently missing a piece, and its reader cannot tell "this never happened" from "the record was lost." That is precisely the failure this PR set out to fix — only transformed from *disagreement* into *absence*.

**And the author already knows the right shape — it was just done once.** `writeback.py:857-867`:

```python
except Exception as exc:  # noqa: BLE001
    log.debug("stack validation record failed", exc_info=True)
    # Losing this one costs the export its only independent check on
    # the ledger: with no promoted figure to compare against, the
    # session total falls back to the sum of the very steps it is
    # meant to be checking.
    trace_recording_skipped(
        "session_validation", reason="caller raised before the recorder", ...
    )
```

That is correct: it says why this particular loss has a cost, and leaves a structured diagnostic. Whereas `stack_event.record_adoption()` at `writeback.py:3322` — which records *which KEEPs were actually taken*, no less important — has no `try` at the call site at all and relies on `stack_event.py`'s internal `log.debug`. `trace_recording_skipped` has 2 call sites in the whole repo against 54 new swallows.

**Must fix**: every swallowed write failure has to leave a trace — simplest is a `warning` fragment in the spool that assembly folds into `metadata.warnings`, so that `log.debug` (effectively silent in production) stops being the only outlet.

> `c27e3e61c` adds `recorder_warnings.py`, and the design is right — deduped per (section, error class), guarded against recursion, folded into `metadata.warnings`. Coverage is partial though: `phase_event` 14/14, `enablement_event` 11/11, `stack_event` 4/4, `warm_start_event` 2/2 are wired up, while `tool_versions` 0/6, `session_metadata` 0/2, `kernel_event` 0/3, `baseline_event` 0/2, `roofline_event` 0/1, `conc_sweep_event` 0/1, `critic_out` 0/1, `robustness_out` 0/1, `warm_replay_event` 0/1 and `event_sink` 0/3 are still `log.debug` only — and **none of the ~54 orchestrator call sites are wired at all** (`note_failure` has zero hits under `src/hyperloom/orchestrator/`). The commit message says "roughly seventy handlers now park through the sidecar"; the actual `note_failure` call-site count is 36. The mechanism is in place, so finishing the wiring is mechanical.

### [Resolved in c27e3e61c] `suppress(Exception)` in `write_breakdown_json` was fully silent

```python
# breakdown/exporter.py:268-274
with suppress(Exception):
    from hyperloom.orchestrator.trace.decision_trace import write_session_decision_trace

    for warning in write_session_decision_trace(sd):
        log.debug("decision_trace: %s", warning)
```

A failed `decision_trace.jsonl` write left **nothing** — no `warnings` entry, no WARNING line — while the downstream Langfuse flush and the backfill tool both depend on that file. There is also a design issue here: the comment itself says it lives here "because every path that writes a session's artifacts goes through this function." That is coupling by convenience, not because it belongs. A function called `write_breakdown_json` should not also produce a file that is not part of the breakdown.

> Confirmed fixed for the silence: now `except Exception` with `log.warning("decision_trace write failed …")` and a comment explaining that it is scored downstream so a silent loss must be visible. The coupling point stands.

---

# 4. Are docstrings / comments / prompts updated in step?

### [Resolved in c27e3e61c] `breakdown/SKILL.md` had one line changed and was almost entirely wrong

`build()` emitted only ten keys: `schema_version` / `exported_at_utc` / `exporter_version` / `metadata` / `outcome` / `timeline` / `close` / `critic` / `robustness` / `warnings`.

`SKILL.md` (+1/−1 — and the single changed line was collateral from the unrelated `discovered_flags` removal) still instructed the agent to read sections that no longer existed:

| SKILL.md line | Section still documented | Actual state |
|---|---|---|
| 32, 135 | `session` | folded into `metadata.session` |
| 33, 136 | `workload` | folded into `metadata.task_config` |
| 34, 137 | `baseline` | **deleted** |
| 35, 138 | `final` | **deleted** |
| 36, 139 | `phase_timeline` | **deleted** |
| 37, 140 | `capability_summary` | **deleted** |
| 38 | `geak` | **deleted** |
| 39, 141 | `optimizations` | **deleted** |
| 40, 142 | `kernel_lifecycle` | **deleted** |
| 41, 143 | `param_search` | **deleted** |
| 42 | `sweep` | **deleted** |
| 43, 144 | `critic_robustness` | **deleted** |
| 44, 145 | `telemetry` | **deleted** |
| 50 | `source_files` | **deleted** |

The troubleshooting table at lines 161-169 rested entirely on `optimizations.available`, `optimizations.validation.unclaimed_integration_count` and `kernel_lifecycle.detected` — all gone. Line 48 still read "`close`: currently empty until the close-stage collector is implemented," which is precisely what this PR implemented. And the two new top-level keys, `critic` and `robustness`, were not documented at all.

This is not documentation fastidiousness: SKILL.md is read by the agent at runtime, so it was directing the agent at a document structure that no longer exists.

> Mostly fixed: the section table and the "Read from" table are both rewritten around the nine real keys. **The troubleshooting table (lines 146-154) is still untouched** and still references `optimizations.available`, `optimizations.validation.unclaimed_integration_count`, `optimizations.entries[].gain_method`, `kernel_lifecycle.detected` and a top-level `warnings: [...]` — all removed by this PR.

### [Resolved in c27e3e61c] `docs/reference/session-breakdown.md` was equally stale

The change was +89/−57 and only swapped `session`/`workload` for `metadata`. The top-level shape at lines 96-131 still listed `baseline` / `final` / `phase_timeline` / `capability_summary` / `geak` / `kernel_lifecycle` / `param_search` / `critic_robustness` / `telemetry` / `optimizations` / `source_files`, plus the entire optional block (`phase_segments` / `explore_search` / `perfskills` / `specialist_runs` / `kernel_roofline` / `kernel_optimization_summary` / `conc_sweep_summary` / `roofline` / `roofline_progress` / `decision_trace` / `token_usage` / `kernel_journey` / `versions` / `enablement`) — **not one of those 25 keys is emitted any more.**

Each also kept a full dedicated section: `## baseline` (421), `## final` (447), `## phase_timeline` (473), `## capability_summary` (485), `## kernel_lifecycle` (518), `## param_search` (534), `## critic_robustness` (547), `## telemetry` (555), `## enablement` (589), `## source_files` (750), `## optimizations` (146). All left as-is. `critic` / `robustness` remained undocumented.

A PR that removes 25 wire keys moved the documentation by 32 net lines. That ratio is itself the finding.

> Confirmed fixed: 784 lines rewritten, the top-level block now lists exactly the nine emitted keys, and the sections are restructured into `outcome.baseline` / `outcome.final` / `critic` / `robustness`.

---

# 5. Working backwards: if none of this code existed, what would be the right implementation?

### Is the mechanism worth keeping? — Yes, but what's worth keeping is the argument, not this shape

**The argument is correct.** For a 6-hour autonomous optimization run, reconstructing "what happened" from artifacts afterwards is necessarily lossy and will lie. `state.json` only holds the terminal state, benchmark reports only hold numbers, and the parts that matter — why this path was chosen, why that candidate was rejected — have nowhere to land at all. So author-time recording is the right direction, and 34 collectors doing archaeology on the wreckage did deserve to go.

**But this PR's shape implements a correct argument as three mutually contradictory things:**

1. It says it eliminates "two answers to one question", yet keeps `_merge_session`, *adds* a second writer via `record_metadata_identity` alongside `collect_session`, and emits two inconsistent `warnings` fields.
2. It says facts get a single source, yet runs that single source behind 54 new silent `except Exception` handlers and reports no write failure through the `warnings` channel that already exists.
3. It says it deletes a net 2037 lines, while actually pushing 3312 lines of complexity into the orchestrator's hot paths.

### What a from-scratch design looks like

**A. Business code emits once and knows nothing about fragments.**

Today every business module knows section names, row shapes and event-id rules, which is why it needs 41 shims and 54 try/excepts. The right boundary: the orchestrator already has phase / action / proposal / attempt as first-class concepts — have them emit a **domain event** on state change, and let the recorder, as a subscriber, translate events into fragments **in one place**.

Then:
- the orchestrator's diff is one `emit` line per site, not +3312 lines;
- the "recording must not cost the run" try/except is needed **once**, at the subscriber's dispatch point, not 54 times;
- changing the breakdown's wire shape stops requiring edits to business files.

**B. The spool should not be "one file per row, re-read in full before every write."**

`record_item` writes one JSON file per row, and locating "which event does this row belong to" globs and parses the whole directory. That is O(n²), on the business thread. An append-only NDJSON per producer plus an in-memory event-id → offset index makes writes O(1) and reads a single sequential scan, while naturally preserving order, tolerating concurrency and surviving restarts — and it is **less** code than what is there now: the whole slug/digest/legacy-name compatibility path in `_stable_item_filename` (`recorder.py:285-312`) disappears.

**C. "Best-effort and silent" must become "best-effort but traceable" once the fallback is gone.**

Deleting the collectors is right, but deleting the fallback obliges you to add the ability to say "I know I lost one." A write failure should drop a warning fragment in the spool that assembly folds into `metadata.warnings`, so a reader can distinguish "this never happened" from "the record of this was lost." That is the correct form, in a single-source world, of this PR's own thesis (*two answers to one question is the failure mode being removed here*). It costs a few lines, and it turns the mechanism from "trust that the write succeeded" into something verifiable.

**D. Things that get deleted should be deleted all the way, or not at all.**

`session` keeping its merge, the top-level `warnings` keeping a partial copy, `DISABLE_RECORDER` surviving with changed meaning — all three are half-deletions. A half-deletion costs more than no deletion: it keeps the maintenance burden of two paths while leaving neither trustworthy. Either the recorder owns these outright and the collector and the switch go, or these blocks are openly declared derived-at-export and the corresponding fragments go.

---

## Where that leaves things

`c27e3e61c` closes most of §1, §3's silent-failure core, and all of §4, and in two places (`build()`'s honesty about not being pure; the dedup and recursion guard in `recorder_warnings`) the fix is better than what I would have proposed. What remains open is §2's instrumentation boundary, the read-back cost, `_merge_session`, and finishing the `note_failure` wiring — plus the two items I'd still consider blocking, which I've raised separately: splitting out the `discovered_flags` / `synergy_attempted` removal, and updating the PR description (the "one bug fixed along the way" section, and declaring the `telemetry` / `token_usage` removals).

ZhengGong-amd (approved):
Revised. approved

PR #1457, opened 2026-09-09
fix(aiter): rebuild serving .so when tuned CSV kernels miss the JIT registry

BaoYunkai:
## Problem
GEMM E2E starts sglang with a merged CSV while the serving `.so` was compiled from a narrower kernel set. aiter JIT does not treat CSV content as a cache key, so lookup fails with `not present in the compiled registry` (wrapped as Capture cuda graph failed).

## Fix
Before GEMM E2E, check CSV `kernelName` values against serving `.so` files and unlink those modules (and invalidate `jit/build`) when they do not cover the CSV. Classify the error as `aiter_jit_registry_mismatch` ahead of `cuda_graph_capture_failed`, retry integrate once after dropping the `.so`, and drop serving `.so` on integrate fault and REVERT.


xiaofei-zheng:
Reviewed the classifier + `.so` invalidation paths against a real aiter install (`/usr/local/lib/python3.12/dist-packages/aiter`, gfx950). Two blocking issues.

**1. `serving_modules_cover_csv()` can never return True for the blockscale / bpreshuffle CSVs, so the "skip" path is dead and every GEMM E2E integrate pays a full CK recompile**

`_aiter_jit.py:serving_modules_cover_csv` decides coverage by byte-searching each `kernelName` inside the serving `.so`. `libtype=asm` rows carry mangled ASM symbol names (`_ZN5aiter42fp8gemm_bf16_blockscale_BpreShuffle_64x128E`); those kernels ship as prebuilt hsaco/asm objects and their names are **never** linked into any JIT `.so`. Grepping the whole aiter package for that symbol finds it only inside the CSV files themselves.

Stock `configs/a8w8_blockscale_bpreshuffle_tuned_gemm.csv`: 18 distinct kernel names, 8 of them asm → 6 names absent from `module_gemm_a8w8_blockscale_bpreshuffle.so`. Running the PR's own function against the installed tree:

```
AITER_CONFIG_GEMM_A8W8_BLOCKSCALE_BPRESHUFFLE -> False
AITER_CONFIG_GEMM_A8W8_BLOCKSCALE             -> False
AITER_CONFIG_GEMM_A8W8_BPRESHUFFLE            -> True   # only because no .so exists
AITER_CONFIG_GEMM_A8W8                        -> True   # ditto
AITER_CONFIG_GEMM_A4W4                        -> True   # ditto
```

Every tuned CSV under `configs/model_configs/` for these two envs misses names too, including ones with zero asm rows (`a8w8_blockscale_bpreshuffle_tuned_gemm_qwen3_235b.csv`: 0 asm rows, 4 of 10 names absent).

Consequence at `phases/kernel.py:3033` — `prepare_serving_so_for_csvs` runs before every integrate attempt and always takes the `invalidate` branch: it unlinks the serving `.so` and moves `jit/build` aside on a perfectly healthy registry. On this node that is `module_gemm_a8w8_blockscale_cktile.so` (76 MB) + `module_gemm_a8w8_blockscale.so` + `..._bpreshuffle.so`, plus a 539 MB `jit/build` tree, per GEMM candidate. That is a full CK codegen+compile before each candidate, and the moved `jit/build` copies accumulate under `session_dir/runs/aiter_jit_backup` unrestored. The unit tests pass only because they write synthetic `.so` bytes containing the literal CSV names.

The coverage signal needs to exclude asm rows (or key on something the serving build actually records — e.g. the generated `blob/*_lookup.h` / manifest under `jit/build/<module>/`) before it can gate anything.

**2. `_run_integrate_rebaseline_with_lock_retry` reads `extra_envs` from the wrong place, so it always drops every GEMM module**

`request_handlers.py:6228-6234`:

```python
extra = getattr(ctx, "extra", None) or {}
envs = extra.get("extra_envs") if isinstance(extra, dict) else None
```

On this path `ctx = RunnerContext(task=fake_task, lease=None)` (`request_handlers.py:6801`) — `RunnerContext.extra` defaults to `{}` and is only ever populated with `shared_state`/`session_dir`/`workspace`/`mn_round_restarted`. The envs live in `fake_task.params["extra_envs"]` (`request_handlers.py:6786`). So `envs` is always `None`, `_modules_for_envs(None)` returns the union of all 12 mapped module names, and the retry unlinks every GEMM serving `.so` rather than the ones tied to the failing CSV. Should read `getattr(getattr(ctx, "task", None), "params", {}).get("extra_envs")`. `test_integrate_retries_once_after_aiter_jit_registry_mismatch` passes `ctx=object()` and only asserts `dropped` is non-empty, so it does not catch this.

Minor note, not blocking: after the retry, the `error_class` re-stamp at `request_handlers.py:6248` re-reads the same workspace logs with no time bound, so a retry that fails for an unrelated reason still gets labelled `aiter_jit_registry_mismatch` from the first attempt's log lines. The baton-stall path next to it uses `since_unix` for exactly this reason.

Classifier ordering in `baseline.py` (registry mismatch winning over `cuda_graph_capture_failed` in both the no-workspace and invalid-measurement branches) and `_is_integrate_fault` handling of the new class both look right — a `status: failed` result is already a fault, so the phase-level retry arms correctly.

xiaofei-zheng:
Re-checked `5cac5f750` against the same aiter install (gfx950). Items 2 and 3 are fixed. Item 1 is only half fixed — one blocking issue remains.

**Fixed**

- `_integrate_extra_envs` now reads `ctx.task.params["extra_envs"]`; verified that is where `integrate_handler` puts it (`request_handlers.py:6786`). Correct.
- The `after_sizes` offset stops the first attempt's log lines from re-stamping the retry's `error_class`.
- The asm filter works. Stock `configs/a8w8_blockscale_bpreshuffle_tuned_gemm.csv` now returns `cover=True` (18 names → 12 JIT names, 0 missing).

**Still blocking: `libtype=cktile` rows in a bpreshuffle CSV resolve to a module that is not in the bpreshuffle module list**

`AITER_ENV_TO_SERVING_MODULES["AITER_CONFIG_GEMM_A8W8_BLOCKSCALE_BPRESHUFFLE"]` lists `module_gemm_a8w8_blockscale_bpreshuffle{,_cktile,_asm}`. But the bpreshuffle tuned CSVs carry `libtype=cktile` rows whose `kernelName` is `a8w8_blockscale_cktile_*` (no `bpreshuffle` infix), and those instances are compiled into `module_gemm_a8w8_blockscale_cktile.so` — which belongs to the *other* env's list, so `serving_modules_cover_csv` never opens it.

`configs/model_configs/a8w8_blockscale_bpreshuffle_tuned_gemm_qwen3_235b.csv`, 87 rows: 67 `libtype=ck` (`a8w8_blockscale_bpreshuffle_*`) + 20 `libtype=cktile` (`a8w8_blockscale_cktile_*`). After the asm filter, 10 JIT names, 4 missing — and all 4 are present in an installed `.so`:

```
MISS a8w8_blockscale_cktile_192x256x128_4x2x1_16x16x128_intrawave_0x1x0_1
  -> actually in: module_gemm_a8w8_blockscale_cktile.so, module_gemm_a8w8_blockscale_cktile_tune.so
  (same for _2, _3, _4)
```

Across `configs/model_configs/*blockscale_bpreshuffle_tuned*.csv` (post-filter JIT names / missing): dsv3 18/5, glm5.2 11/4, qwen3.5_397b 30/13, qwen3_235b 10/4, dsv4 26/9, glm5 16/3, mm2.5 4/0. Six of the seven still report False, so on any of those model configs `prepare_serving_so_for_csvs` still takes the `invalidate` branch on a healthy registry and forces the full CK recompile plus the `jit/build` move before every candidate. The default (non-model_configs) CSV is the only one that now skips.

For contrast, `AITER_CONFIG_GEMM_A8W8_BLOCKSCALE` → False is a **true** positive and should stay: its 5 missing `a8w8_blockscale_1x128x128_*` names exist only in `module_gemm_a8w8_blockscale_tune.so`, never in the serving `.so`. So the check does work once the module set is right — the fix is to resolve the module from the row's `libtype`/kernel-name prefix rather than from the env var alone, or to have the bpreshuffle entry also cover `module_gemm_a8w8_blockscale_cktile`.

**Non-blocking, two notes**

- `_modules_for_envs` changed from `if not envs` to `if envs is None`, so an empty `extra_envs` dict now yields an empty module tuple and drops nothing. `integrate_handler` builds `params["extra_envs"] = dict(payload.get("extra_envs") or {})`, so a kernel-patch integrate with no envs hits a registry mismatch, drops nothing, and burns the one retry on an unchanged tree. Previously it fell back to dropping all GEMM modules.
- `_workspace_has_compiled_registry_error(..., after_sizes=...)` seeks to the pre-retry size. The magpie wrapper rewrites `server.log` per run, so if the retry's log is shorter than the first attempt's, the seek lands past a genuine second mismatch and it goes unclassified. Only affects the label, not the outcome.

xiaofei-zheng (approved):
Re-checked `c6056eb9d` (`7e4d1e018` on top of the merge). No blocking issues left — all three items from the earlier rounds are resolved. LGTM.

**The remaining item is fixed and verified against the real aiter install (gfx950, `/usr/local/lib/python3.12/dist-packages/aiter`)**

`serving_module_for_kernel` resolving from the kernelName prefix + libtype is the right key. `a8w8_blockscale_cktile_*` rows in a bpreshuffle CSV fall through the `a8w8_blockscale_bpreshuffle` prefix and land on `a8w8_blockscale` → `module_gemm_a8w8_blockscale_cktile`, which is where those instances actually live. Ran the shipped function over every tuned CSV on the box:

```
True   configs/a8w8_blockscale_bpreshuffle_tuned_gemm.csv
True   configs/a8w8_bpreshuffle_tuned_gemm.csv, a8w8_tuned_gemm.csv, a4w4_blockscale_tuned_gemm.csv
True   model_configs/*blockscale_bpreshuffle_tuned*  (glm5.2, mm2.5, qwen3.5_397b, qwen3_235b, dsv4, glm5)
True   model_configs/dsv3_a8w8_bpreshuffle, glm5_a8w8_bpreshuffle
False  configs/a8w8_blockscale_tuned_gemm.csv
False  model_configs/a8w8_blockscale_bpreshuffle_tuned_gemm_dsv3.csv
```

Both remaining `False` results are true positives, not the old false negative:

- `a8w8_blockscale_tuned_gemm.csv`: 5 `a8w8_blockscale_1x128x128_*` names exist only in `module_gemm_a8w8_blockscale_tune.so`, never in the serving `.so`.
- `dsv3`: one `libtype=ck` name, `a8w8_blockscale_bpreshuffle_1x128x128_256x16x256x128_8x16_...`, is in **no** `.so` on the box.

So the guard now skips on a healthy registry and only invalidates when the serving `.so` genuinely misses a kernel — which is exactly the case that produced the original `not present in the compiled registry`. Also confirmed `csv_jit_kernel_rows` leaves no unresolved prefixes across all of those CSVs (nothing silently falls back to the env module list).

**Other paths checked**

- `_integrate_extra_envs` reads `ctx.task.params["extra_envs"]`, matching where `integrate_handler` sets it (`request_handlers.py:6786`); `ctx.extra` on that path only ever holds `shared_state`/`session_dir`/`workspace`/`mn_round_restarted`.
- `_workspace_has_compiled_registry_error(after_sizes=...)` now falls back to the 64 KiB tail when `start >= size`, so a truncated/rewritten `server.log` on the retry no longer reads past the new content.
- `_modules_for_envs` back to `if not envs`, so an empty `extra_envs` keeps the drop-everything fallback.
- `baseline.py` classifier ordering: `aiter_jit_registry_mismatch` wins over `cuda_graph_capture_failed` in both the no-workspace early return and the invalid-measurement branch, so the disable-cuda-graph one-shot is not armed for this class. `_is_integrate_fault` treats `status: failed` as a fault, so the phase-level `_MAX_INTEGRATE_FAULT_ATTEMPTS` retry arms correctly and the result is not mistaken for a REVERT verdict.

**Tests run locally** (worktree at `c6056eb9d`): `test_aiter_jit_unit.py` 42 passed, `test_kernel_integrate_and_report.py` 44 passed, `test_profile_gpu_preflight.py` 12 passed, `test_coordinator_gemm_promote_units.py` 128 passed.

PR #1458, opened 2026-09-09
tier3 (2/2): fused MoE — the gate reaches the table the fleet actually loses time on

chennyiiis:
Stacked on #1452 (M0 + M1). **Review that one first** — this branch is 4 commits on top of it, and the diff below only makes sense against it.

#1452 opened the third gate and proved it on dense bf16. This one takes it to the table the fleet actually loses time on: `tuned_fmoe.csv`, whose owner `fmoe_ck` burns 621–707s per GLM session and hands back `empty_output`.

## What was in the way

Fused-MoE misses were parsed, counted, and then dropped on the floor: `evidence.py` filed them under `dispatch["moe"]` while `coverage_gaps()` reads only `demands`. So `tuned_fmoe.csv` could not become a gap, could not reach the gate, and nothing downstream of it had ever run. Behind that sat a second wall — no adapter for the table, and `runner.py` paid for a generated tuner *before* checking whether one existed, so the miss would have surfaced as a wasted authoring call rather than a refusal.

## The four commits

1. **`da8b98ec6`** — check `adapters_for()` before generating, not after. A table with no adapter now refuses at the gate instead of buying a tuner it cannot time.
2. **`abf9f2f05`** — build a `tuned_fmoe.csv` Demand from `dispatch["moe"]`. Blast radius is real and deliberate: `router.py` will now select `fmoe_ck` from log evidence, and `demand_for_tuner(report, "fmoe_ck")` stops returning `None`. Both are fixes in their own right, which is why they are their own commit behind `FORGE_MOE_DEMAND_DISABLE`.
3. **`c7edadbdd`** — `_FusedMoeAdapter`: baseline, dispatch, correctness, and the two guards below.
4. **`50461a9e2`** — mandate text for the `block_m` constraint. No behaviour change.

## Two ways a fused-MoE candidate lies about winning, both measured

**A row aiter reads and then ignores.** `fused_moe.py:2048-2051` logs the tuned pair *before* asking `if is_flydsl1 or is_flydsl2`. A pair with neither name FlyDSL's falls past the entire tuned branch into the heuristic chain — but the log says it was used. Measured: a candidate naming `nope1`/`nope2` was logged verbatim, ran the default kernels, and timed **1.1% faster than baseline**, over the referee's 1.01 floor. `aiter_honours_kernel_pair()` refuses the pair before any GPU work.

**A `block_m` that does not match the tile.** `block_m` is the granularity tokens are sorted into before either kernel indexes them, so a mismatch does not fault — it reads the wrong rows, fast. Nine measurements at token=512, no exceptions:

| stage1 tile_m | stage2 tile_m | block_m | us | mean err |
|---|---|---|---|---|
| 128 | 128 | 32 | 71.70 | 1.404 ✗ |
| 128 | 128 | **128** | 156.74 | **0.0016 ✓** |
| 32 | 32 | **32** | 121.12 | **0.0002 ✓** |
| 32 | 32 | 128 | 91.10 | 1.096 ✗ |

Baseline 111.67us. The wrong pairs were the *fastest things in the search* — 71.70us, a 1.55x. The adapter's mean-error gate caught every one, but a tuner that pins `block_m` and trusts its clock spends its whole budget ranking nonsense, so the mandate now states the constraint.

## Acceptance, on hardware

MI355X / gfx950, aiter `d9e5ef7ce`. The demand comes from a log aiter itself wrote, not a fixture; the authoring stage uses `--stub-generate` because the box has no provider credential — the same documented substitution #1452's dense acceptance ran under.

```
demand:  tuned_fmoe.csv  tuner=fmoe_ck env=AITER_CONFIG_FMOE  36 misses / 3 keys
gap:     ('tuned_fmoe.csv', 'empty', True)      gate: allowed=True
adapter: _FusedMoeAdapter                       shapes: 3
attempt: stage=referee ok=True  reason="3 of 3 shape(s) improved"
result:  tier3_generated_tuned_fmoe status=ok candidate=True shapes=3/3 best=1.1528x
promote: {'tuner': 'tier3_generated_tuned_fmoe',
          'env': {'AITER_CONFIG_FMOE': '.../out.csv'},
          'best_micro_speedup': 1.1528, 'requires_e2e_validation': True}
```

Per shape, referee-verified under `time_paired` (interleaved, min-of-repeats, best-vs-typical consistency gate):

| token | baseline | candidate | speedup |
|---|---|---|---|
| 512 | 2212.0us | 2170.3us | 1.0192 |
| 1024 | 3051.7us | 2651.4us | **1.1510** |
| 2048 | 4208.6us | 3650.6us | **1.1528** |

Zero candidates discarded on correctness — the search groups by tile_m and sets `block_m` to match, so every pair it can propose is legal. `requires_e2e_validation` stays true: micro is a screen, not the verdict.

## Not in this PR

Three aiter findings, all the same bug class and all worth an upstream issue rather than a patch here:

* a tuned config naming `tile=256` for `inter_dim=384` silently runs 128 — aiter's own tuner picks these, and they fail correctness at 0.14;
* `AITER_CONFIG.get_config_file`'s `lru_cache` key omits the env var its body reads;
* the tuned row logged before the branch that discards it (above).

Also: `MAX_JOBS` must be capped on containers with high `nproc` and a small cgroup memory cap — aiter's JIT picked `ninja -j 188` against a 128 GiB limit and OOMed. One env var, wide blast radius if the fleet shares that shape.

## Tests

`src/kernelforge/gemm_tune`: 862 passed, 7 skipped, 3 pre-existing Windows failures (`os.killpg`). `test_moe_dispatch.py` is new: 26 passed, 2 skipped.

### DIFF
--- a/src/kernelforge/gemm_tune/cli.py
+++ b/src/kernelforge/gemm_tune/cli.py
@@ -10,11 +10,14 @@ import logging
 import sys
 import time
 from pathlib import Path
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 import click
 
 from . import __version__
+
+if TYPE_CHECKING:
+    from .tuners.base import TuneResult
 
 log = logging.getLogger("kernelforge.gemm_tune")
 
@@ -104,20 +107,33 @@ def _load_demand_report(demand_json: str) -> dict | None:
         return None
 
 
-def _coverage_gaps(demand_report: dict | None, tuner_specs: list, output_dir: Path) -> list:
-    """Write the demanded tables no selected tuner will produce, and return them."""
+def _coverage_gaps(
+    demand_report: dict | None,
+    tuner_specs: list,
+    output_dir: Path,
+    results: list | None = None,
+) -> list:
+    """Write the demanded tables no selected tuner produced, and return them.
+
+    Called twice per run, and the difference is the point: before the tuners run
+    ``results`` is None and selection alone says what is uncovered; after, a
+    table whose owner came back empty-handed joins the list, which is the state
+    the runtime is in either way. Best-effort -- a report must not affect the
+    tuning it describes.
+    """
     if not demand_report:
         return []
     try:
         from .tier3 import coverage_gaps
 
-        gaps = coverage_gaps(demand_report, tuner_specs)
-        if not gaps:
-            return []
-        (output_dir / "coverage_gaps.json").write_text(
-            json.dumps([g.to_dict() for g in gaps], indent=2),
-            encoding="utf-8",
-        )
+        gaps = coverage_gaps(demand_report, tuner_specs, results)
+        # The second pass writes even when it finds nothing, or the pre-run
+        # file it supersedes would be read as this run's final answer.
+        if gaps or results is not None:
+            (output_dir / "coverage_gaps.json").write_text(
+                json.dumps([g.to_dict() for g in gaps], indent=2),
+                encoding="utf-8",
+            )
         return gaps
     except Exception:  # noqa: BLE001 - a report must never fail the run
         log.debug("could not record coverage gaps", exc_info=True)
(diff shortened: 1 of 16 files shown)


chennyiiis:
Rebased onto the rebased #1452. Two conflicts, both from #1421's comment compaction: `demand_shapes`'s docstring in `evidence.py` and the fused-MoE constants block in `tier3/dispatch.py`. Kept the measurements (`MAX_MOE_MEAN_ERROR`'s three-case error table, the `FLYDSL_KERNEL_PREFIX` finding that aiter logs the tuned row before deciding to ignore it), dropped the prose around them.

Local: `ruff format`/`check` clean, `3 failed, 920 passed, 8 skipped` — the three are the pre-existing Windows `os.killpg` failures.

Stacked on #1452; will retarget to `main` once that lands.

chennyiiis:
Folded into #1452, same five commits. It was stacked on that branch, so it was never independently reviewable. Closing in favour of one branch.

Issue #512, opened 2026-06-09
AMD MoE sglang baselines break: aiter CK 2-stage fused-MoE JIT fails on this image (inject --moe-runner-backend triton)


Umangatamd:
## Summary
MoE models served by **sglang on MI300X/MI355X (aiter)** fail their baseline because sglang's default `--moe-runner-backend auto` routes Mixture-of-Experts models through aiter's **CK 2-stage fused-MoE kernel**, whose first-request JIT build is broken in the current ROCm image.

## Root cause
The aiter CK fused-MoE JIT build (`module_moe_ck2stages_*`) fails to compile in this image: `thrust` pulls in a missing `<cub/detail/detect_cuda_runtime.cuh>`, so `hipcc` errors out. The killed build leaves a stale lock, and the next attempts hang on "waiting for baton release" until sglang's 600s warmup read-timeout fires → **`baseline_failed`** (throughput 0, the whole optimization run never starts).

## Impact (observed)
- Qwen3-30B-A3B (MoE) sglang baseline never came up; injecting `--moe-runner-backend triton` is what unblocked it.
- Any AMD MoE sglang baseline (e.g. Qwen3-MoE family) is affected on this image.

## Suggested fix
At the single sglang server-arg choke point (`materialize_config_with_envs`), detect MoE models on AMD/ROCm and append `--moe-runner-backend triton` (the ROCm-capable fused-MoE backend sglang itself falls back to) **unless** the operator already pinned `--moe-runner-backend`. Triton sidesteps the broken CK JIT. Make the backend overridable via `$HYPERLOOM_SGLANG_MOE_RUNNER_BACKEND`, and fail-safe (inject nothing) for non-sglang frameworks, non-AMD GPUs, or non-MoE models.

This is the same class of per-launch guard as the existing `--watchdog-timeout` / `--context-length` injections.

PR #1459, opened 2026-09-09
fix(sglang): stop forcing --moe-runner-backend, let sglang auto pick it

BaoYunkai:
## Problem

Hyperloom force-injects `--moe-runner-backend triton` for MoE sglang models on AMD/ROCm (added in #513) to sidestep an aiter CK fused-MoE JIT build that was broken in some ROCm images.

Magpie's `sglang_mi300x.sh` / `sglang_mi355x.sh` separately set `SGLANG_USE_AITER=1`, which pre-shuffles MoE expert weights into the aiter kernel layout at load time. When Hyperloom then forces `triton`, the triton runner reads that aiter-laid-out memory with the wrong kernel: weights are silently wrong and accuracy (e.g. gsm8k) drops to 0.

## Verification

Verified end-to-end on a real MoE checkpoint (`Qwen3-Next-80B-A3B-Instruct-FP8`, current sglang + ROCm image):

- `--moe-runner-backend auto` (sglang's default) correctly follows `SGLANG_USE_AITER` in both directions (aiter when set, triton otherwise) and produces correct output.
- Forcing `--moe-runner-backend triton` while `SGLANG_USE_AITER=1` reproduces the garbage-output bug (verified via direct chat completion requests).
- The aiter CK JIT crash that originally motivated the override (#512) does not reproduce on this image/sglang version.

## Fix

Remove `inject_sglang_moe_runner_backend()` and its call site in `materialize_config_with_envs`, so sglang's own `auto` resolution decides the backend. `moe_runner_requires_aiter()` and the grid-path / baseline-retry code that strip an inherited or operator-pinned `--moe-runner-backend` on an aiter-only quant scheme (e.g. Quark MXFP4) are unaffected and still apply.

## Test plan

- Updated `test_moe_runner_backend_injection.py` and `test_baseline_moe_runner_fallback.py` to match the new behavior (no default injection; retry tests now pin the flag explicitly to simulate an operator/reference-recipe pin).
- `python -m pytest test_moe_runner_backend_injection.py test_baseline_moe_runner_fallback.py test_workload_envs_branches_unit.py test_aiter_moe_pin_filter.py -q` → 121 passed.


xiaofei-zheng (approved):
LGTM — no blocking issues.

## What this PR does
Removes the forced `--moe-runner-backend triton` injection (`inject_sglang_moe_runner_backend()`, its `HYPERLOOM_SGLANG_MOE_RUNNER_BACKEND` env + `triton` default constant, and the call site in `_finalize_framework_server_args`), plus the re-exports in `_grid_runner.py`. The original override existed to dodge a broken aiter CK fused-MoE JIT build, but it silently corrupted results when Magpie set `SGLANG_USE_AITER=1` (aiter-laid-out weights read by the triton runner → gsm8k → 0). sglang's own `--moe-runner-backend auto` now follows `SGLANG_USE_AITER` correctly, so the choice is left to sglang.

## Review notes
Checked on branch `bugfix/yunkai/moe-runner-drop-default`:
- No dangling references to the deleted function/constants anywhere (`_grid_server_args.py`, `_grid_runner.py`, `_workload_envs.py`, and the two test files all grep clean).
- `os`, `log`, `merge_server_args` remain in use in `_grid_server_args.py` — no orphaned imports / NameError.
- The aiter-only strip paths are preserved and independent of the removed injection: `moe_runner_requires_aiter`, `_remove_moe_runner_backend_arg`, the grid-variant inherited-flag strip, and the baseline `drop_moe_runner_backend` retry.
- `_finalize_framework_server_args` still strips an inherited/pinned flag on `drop_moe_runner_backend`; with no injection and no pin the retry is a correct no-op.
- Updated tests are internally consistent — retry tests now pin the flag explicitly via `extra_server_args` to keep exercising the flagless-retry path.

Non-blocking only (not gating): a couple of docstrings/comments still mention the removed "injection" step.

xiaofei-zheng (approved):

PR #1461, opened 2026-09-09
gemm-tune: size the aiter JIT build to the container, not to the host

chennyiiis:
## What breaks

aiter's JIT picks its `ninja -j` from the CPU count. In a fleet container that is the **host's** count while the memory those jobs consume is capped by the **container's** cgroup. Nothing reconciles the two.

Measured on the MI355X dev box (ROCm 7.2):

| | |
|---|---|
| `nproc` / affinity | 236 |
| `/sys/fs/cgroup/memory.max` | 137438953472 (128 GiB) |
| host `MemTotal` | 2.75 TiB |
| aiter's chosen `-j` | 188 |
| one hipcc job, peak RSS | **1.64 GiB** (3 samples, ±0.01 GiB, ~30s each, `module_moe_cktile2stages`) |
| 188 × 1.64 GiB | ~308 GiB against a 128 GiB ceiling |

So the build is OOM-killed partway through. Not slow, not flaky — it does not finish, and every tuner downstream of it reports whatever a missing module looks like from where it stands. Every run on that box so far needed `MAX_JOBS=16` set by hand.

## What this does

One call in `run_subprocess`, plus the three helpers it needs:

- `cgroup_memory_limit()` — the container's ceiling, v2 then v1, `None` when there is none;
- `build_job_limit()` — `ceiling × 0.7 / 2 GiB`, capped at the CPU count, `None` when there is nothing to correct;
- `cap_build_parallelism(env)` — sets `MAX_JOBS` only if it is not already set.

Deliberately narrow. No-op when `MAX_JOBS` is set, no-op when `env_override` names one, no-op without a cgroup ceiling, no-op when the ceiling is roomy enough for a job per CPU. On a developer machine nothing changes.

## Verified on hardware, not just derived

The computed cap on that box is **44**. Launching 44 real concurrent hipcc jobs there:

```
memory.current   1.0 GiB before  ->  58.4 GiB peak
per job          1.30 GiB aggregate (the peaks do not coincide)
headroom         69.6 GiB under the 128 GiB ceiling
result           44/44 compiled clean, 177s
```

The 2 GiB constant is the 1.64 GiB measurement rounded up; the run above is what says the rounding is safe rather than merely plausible.

## Why the sentinels are tested

Both cgroup layouts spell "no limit" as a *value*: v2 writes `max`, v1 writes `9223372036854771712`. Taking either at face value computes a job limit in the billions, which is the same OOM with extra steps. `test_cgroup_v1_sentinel_is_not_a_ceiling` and `test_cgroup_v2_unlimited_is_not_a_ceiling` pin both.

`MAX_JOBS=` (empty, as a shell profile leaves it) is likewise treated as unset rather than as an operator's choice — ninja reads it that way too.

## Tests

18 new in `test_build_parallelism.py`; both numbers are injected, so the suite says the same thing on a laptop as on a fleet node. Full `gemm_tune` suite: 831 passed, 6 skipped, 3 failed — the 3 being pre-existing Windows-only `os.killpg` failures in `test_utils_extra.py`, present on `main` too.

## Not in this PR

tier3's sandbox builds its own child environment with `subprocess.run` and does not route through `run_subprocess`, so it does not pick the cap up. It gets the same call once this lands — it cannot be done here because the tier3 work is on a branch that predates this helper.

Upstream, the honest fix is for aiter's JIT to size `-j` from the cgroup rather than from `os.cpu_count()`; that will be reported separately to ROCm/aiter.

### DIFF
--- a/src/kernelforge/gemm_tune/tests/test_build_parallelism.py
+++ b/src/kernelforge/gemm_tune/tests/test_build_parallelism.py
@@ -1,0 +1,151 @@
+# SPDX-FileCopyrightText: 2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Sizing an aiter JIT build to the container rather than to the host.
+
+The measurement behind every number here, taken on an MI355X under ROCm 7.2:
+one ``hipcc`` job on a CK-tile fused-MoE source holds **1.64 GiB** and runs for
+30 seconds, three samples within 0.01 GiB of each other. The box shows 236 CPUs
+and a cgroup ceiling of 128 GiB, so aiter's ``-j 188`` asks for roughly 308 GiB
+and the build is killed rather than slowed.
+
+These tests inject both numbers instead of reading the machine, so they say the
+same thing on a laptop as on a fleet node.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from kernelforge.gemm_tune import utils
+
+GIB = 1024**3
+
+
+class TestReadingTheCeiling:
+    def test_cgroup_v2_reports_bytes(self, tmp_path, monkeypatch):
+        limit = tmp_path / "memory.max"
+        limit.write_text("137438953472\n", encoding="utf-8")
+        monkeypatch.setattr(utils, "_CGROUP_MEMORY_LIMITS", (str(limit),))
+        assert utils.cgroup_memory_limit() == 128 * GIB
+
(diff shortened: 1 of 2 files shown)


chennyiiis:
Folded into #1452 as commit `b0de15f07`, unchanged. It was filed separately because the cap is not tier3-specific, but the follow-up it enables — `tier3/sandbox.py` builds its own environment and was the last path still asking for a job per host CPU — could not land until it did, so reviewing them apart meant holding both in your head anyway. Closing in favour of one branch.

PR #1466, opened 2026-09-10
fix(state): stop the warm-start block claiming a cold start on a KB hit

rpoornac:

  `SharedState.to_warm_start_summary` — which feeds the `=== Warm start ===`
  block injected into the conversation, and the `warm_start` MCP context tool —
  read three fields that no writer produces.

  `recipe.get("raw")` is the damaging one. `recipe_kb_t0` builds
  `warm_start_recipe` as `{workload, hw, tier, confidence, recipe}` with the KB
  row nested under `recipe`, and no backend has ever written a flat `raw` text
  field. The lookup therefore always missed and fell to the `else` branch,
  printing **`(no recipe text — first session for this workload/hw)` on an exact
  KB hit carrying a full `best_config`**. That is not a silent omission — it
  asserts the opposite of what the KB found, on the one line the model reads to
  decide whether it has prior work to build on.

  The pitfall loop had the matching problem, reading
  `entry.get("raw") or entry.get("symptom")`. `symptom` is a robustness-agent
  concept unrelated to KB pitfalls, so the block emitted a `pitfalls (N):`
  header with nothing beneath it, for both the local and the remote shape.

  Against a real `LocalRecipeStore` row, before:

  ```
  recipe: workload=kimi-k3 hw=mi355x
    · (no recipe text — first session for this workload/hw)
  pitfalls (1):
  ```

  after:

  ```
  recipe: workload=kimi-k3 hw=mi355x
    · best_throughput=875.0
    · --attention-backend=AITER
    · VLLM_ROCM_USE_AITER=1
  pitfalls (1):
    · K=8 draft overprovision at conc=1 (severity=regress)
  ```

  The body is summarised from the nested row (`best_throughput` / `best_config`,
  still capped at five lines and 240 chars), pitfalls read
  `description`/`severity` through the same `(entry.get("attrs") or entry)` form
  used for wrapped-vs-flat points elsewhere, and the "first session" line is
  reserved for the case that actually is one. Flat `raw` text is read first, so
  a session resumed from the older snapshot shape renders unchanged.

  None filed. Adjacent to #1464 (same class of defect — a reader expecting
  fields the writer does not produce — but a different root cause: wrong field
  names rather than a missing `attrs` unwrap). Independent of it; no shared
  files except `CHANGELOG.md`.

  Four added in `test_role_realignment.py`, alongside the existing warm-start
  tests. Three fail on `main`. They build the pair from `LocalRecipeStore` the
  way `recipe_kb_t0` assembles it, rather than hand-writing a dict.

  Worth calling out why this went unnoticed for so long: the existing test
  invented the shape it asserted on, passing `{"raw": ...}` for both the recipe
  and the pitfalls — a shape matching the renderer and nothing else in the
  system. It stayed green while production rendered nothing. That test is kept
  as the backward-compatibility case and still passes.

  ```
  pytest src/hyperloom/inference_optimizer/tests/test_role_realignment.py
      38 passed

  pytest src/hyperloom -k "warm or render or shared_state or prompt or context or pitfall"
      1467 passed

  pytest src/hyperloom scripts
      15815 passed, 8 failed — the same 8 that fail on a clean tree with this
      change stashed (pre-existing, subprocess "No module named 'hyperloom'" in
      this environment)

  ruff check / ruff format --check   clean
  ```

  No. The only rendering that changes is rendering that was previously empty or
  wrong. Legacy flat `raw` snapshots are still read first and render as before.
  One message was reworded (`no recipe text` → `no recipe`) since it now fires
  only when there genuinely is no row.

  Yes — one function, one defect, in the two reads that share it.

  No — root cause is in this repo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,29 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Fixed
+
+- **The `=== Warm start ===` block told the model it was starting cold on top of
+  a matched recipe.** `to_warm_start_summary` read three fields no writer
+  produces — `recipe.get("raw")`, and `raw`/`symptom` on each pitfall — so an
+  exact hit carrying a full config printed
+  `(no recipe text — first session for this workload/hw)`, and a `pitfalls (N):`
+  header could appear with nothing under it. That is not a silent omission; it
+  asserts the opposite of what the KB found, on the line the model reads to
+  decide whether it has prior work to build on.
+
+  The block now renders `warm_start_context`, the model-facing view
+  `recipe_kb_t0` already builds and persists on every anchor, instead of
+  re-deriving a second one from the raw row. That view answers what this block
+  exists to answer and the row cannot: `status` distinguishes a hit from a
+  seed-only first session, `match.tier`/`confidence` qualify the match, and
+  `recommended_replay` carries the config already split into server args and
+  envs with its donor attached. A borrowed config is now labelled with the model
+  it came from, so another workload's throughput can no longer read as this
+  session's own history, and the pitfall header counts the rows it prints.<br/>
+  **Operator note**: affects the conversation warm-start block and the
+  `warm_start` MCP context tool, in both local and remote Recipe modes.
+
 ### Removed
 
 - **The `learning/` tuning database, the tracker's scoring layer, and the
--- a/src/hyperloom/inference_optimizer/tests/test_role_realignment.py
+++ b/src/hyperloom/inference_optimizer/tests/test_role_realignment.py
@@ -204,23 +204,121 @@ def test_shared_state_warm_start_summary_empty_when_no_recipe():
     assert SharedState().to_warm_start_summary() == ""
 
 
-def test_shared_state_warm_start_summary_renders_recipe_and_pitfalls():
-    s = SharedState()
-    s.warm_start_recipe = {
-        "workload": "deepseek-r1",
-        "hw": "mi300x",
-        "raw": "recipe_id=42 stack=sglang/0.4.10\nbest_config={'foo':'bar'}\nwhat_worked=[A, B]",
-    }
-    s.warm_start_pitfalls = [
-        {"raw": "OOM on fp8 expert_dtype — switch to fp4"},
-        {"raw": "TP=8 + ISL>=8k causes nccl hang"},
-    ]
-    out = s.to_warm_start_summary()
-    assert "workload=deepseek-r1" in out
-    assert "hw=mi300x" in out
-    assert "recipe_id=42" in out
-    assert "pitfalls (2):" in out
-    assert "OOM on fp8" in out
+def _t0_warm_started(tmp_path, **put_kwargs) -> SharedState:
+    """Seed a recipe, run the real T0 anchor, return the state the renderer sees.
+
+    Hand-built fixtures are how this block came to render fields no writer
+    produces — a ``raw`` blob that has never existed on ``warm_start_recipe``.
+    Driving ``run_t0_anchor`` means the context under assertion is the one a
+    session actually builds, for whichever shape the KB and T0 agree on.
+    """
+    from hyperloom.orchestrator.knowledge.recipe_kb import (
+        LocalRecipeStore,
+        RecipeKB,
+        recipe_canonical_id,
+    )
+    from hyperloom.orchestrator.knowledge.recipe_kb_t0 import run_t0_anchor
+
+    state = SharedState()
+    state.framework_name = "vllm"
+    state.framework_version = "0.11.0"
+    state.precision = "fp8"
+    kb = RecipeKB(local=LocalRecipeStore(root=tmp_path / "kb"))
+    if put_kwargs:
+        kb.put_recipe(
+            canonical_id=recipe_canonical_id(
+                model="kimi-k3",
+                hardware="mi355x",
+                framework_name=state.framework_name,
+                framework_version=state.framework_version,
+                precision=state.precision,
+            ),
+            model="kimi-k3",
+            hardware="mi355x",
+            framework_name=state.framework_name,
+            framework_version=state.framework_version,
+            precision=state.precision,
+            provenance={"source": "seed", "generator": "ut"},
+            **put_kwargs,
+        )
+    session_dir = tmp_path / "session"
+    session_dir.mkdir(exist_ok=True)
+    run_t0_anchor(
+        kb,
+        state,
+        workload="kimi-k3",
+        hw="mi355x",
+        extra_attrs={"framework_name": state.framework_name},
+        session_dir=session_dir,
+    )
+    return state
+
+
+def test_warm_start_summary_reports_the_recipe_a_hit_actually_matched(tmp_path):
+    """On a KB hit the block must not claim there is nothing to go on.
+
+    It read a ``raw`` text field that no writer produces, so a matched recipe
+    rendered as "first session for this workload/hw" — worse than saying nothing.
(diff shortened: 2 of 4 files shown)


ZhengGong-amd:
The defect is real and well diagnosed — printing `(no recipe text — first session…)` on an exact hit is worse than printing nothing. But the fix re-derives a model-facing view inside the renderer, and T0 already builds one. On this branch, three real input shapes still render false statements:

**1. Remote mode has no `best_config` at all.** `knowledge_to_warm_recipe` (`remote_recipe/values.py:1269`) never returns that key — remote replay data goes through the section SDKs, as `_build_warm_start_context`'s own docstring states. A remote V2 hit renders only `· best_throughput=875.0`, and with no published throughput:

```
recipe: workload=kimi-k3 hw=mi355x
  · (recipe matched but carried no config or throughput)
```

That is the original defect, reworded. The CHANGELOG claims remote mode is fixed.

**2. Non-exact tiers attribute a donor's number to this workload.** `_cascade_warm_start_search` returns another model's row at sibling tiers, while `workload`/`hw` at the top of the snapshot are *this* session's. With `tier=arch_sibling, confidence=0.4` and a llama-3-70b/mi300x donor:

```
recipe: workload=kimi-k3 hw=mi355x
  · best_throughput=875.0          ← llama-3-70b on mi300x
  · extra_server_args=--tp 8
```

`tier` and `confidence` sit two keys away in the same dict and are never rendered. "Wrongly says cold start" becomes "wrongly says this is your warm data", which is harder to catch.

**3. Production `best_config` renders a Python dict repr.** `writeback._build_recipe_attrs_from_state` produces `{extra_server_args, extra_envs, name, tput, accuracy}`:

```
  · extra_server_args=--attention-backend=AITER --enable-chunked-prefill
  · extra_envs={'VLLM_ROCM_USE_AITER': '1', 'VLLM_USE_TRITON_FLASH_ATTN': '0'}
  · name=aiter-v2
  · tput=875.0
```

The `after` block in the description cannot be produced by this code — it comes from the fixture `best_config={"VLLM_ROCM_USE_AITER": "1"}`, a shape no CLOSE-phase writer emits. `_config_replay_args_envs` (`recipe_kb_t0.py:167`) already does this split correctly.

Two more: `pitfalls (N):` can still print a header with nothing under it (the count is `len(pitfalls)`, the rows are the ones that render) — still reproducible here. And `warm_start_recipe` has never carried `raw` in 686 commits of history, so that branch is not backward compatibility; giving it top priority is a trap, since `.kb_warm.json`'s `raw` is `json.dumps({"points": [...]})`.

**The shorter fix is to render what T0 already computed.** `_build_warm_start_context` (`recipe_kb_t0.py:418`) runs on every anchor, is persisted, and already carries `status` (hit/seed_only/miss), `match.tier`/`confidence`, `recommended_replay.{extra_server_args, extra_envs}` split cleanly, and `donor_canonical_id`/`donor_model`. Its `status` is already surfaced as `kb_hit` in the flat status summary. Rendering it is a net deletion, and it answers every question this block exists to answer.

Suggested direction:

1. Delete `_warm_recipe_body` along with the `raw` / `symptom` / `attrs` fallbacks.
2. Render `self.warm_start_context`: `status` picks the three states, `match.tier`/`confidence` on the header line, `recommended_replay` for the config lines, `donor_model`/`config_tier` for a "borrowed from" line at non-self donors, `ctx["pitfalls"]` for the pitfall rows.
3. Make `pitfalls (N):` unable to lie — count the rows actually rendered, and elide the rest with the existing `to_gaps_summary` idiom (`(+N more elided)`).
4. Update the `SharedState` field comments: drop "KB point dicts" from `warm_start_pitfalls`; say `warm_start_recipe` is the `{workload, hw, tier, confidence, recipe}` envelope and not a prompt input. That comment is why the `attrs` unwrap looked necessary.
5. Update the `to_warm_start_summary` docstring to say it renders `warm_start_context` and that the three states come from `status`.
6. Drop the `{"raw": …}` legacy test and the new wrapped-pitfall test. Keep the "build fixtures from the real writer" idea — it is the best part of this PR — but run them through `run_t0_anchor` so the context is what reaches the renderer. That would have caught all three cases above.
7. Trim the CHANGELOG to what an operator needs: a real hit used to render as a first session; the block now renders the T0 warm-start context with tier/confidence and donor attribution; affects the conversation block and the `warm_start` MCP tool.

Fixing the field names turns the local demo green, but it leaves two answers to "what should this block say" — the one T0 computed and the one the renderer guessed. Worth converging on the first.

rpoornac:
You're right that the fix was in the wrong place. I was correcting field names inside a view the renderer should never have been deriving, and I checked the three shapes you list — all three reproduce.

`knowledge_to_warm_recipe` returns `best_throughput` and passes `lessons`/`pitfalls` through, and never a `best_config`, so my `_warm_recipe_body` would have rendered exactly the original defect for remote hits while the CHANGELOG claimed remote was fixed. The `raw` branch I gave top priority is worse than dead weight: as you say `warm_start_recipe` has never carried it, and `.kb_warm.json`'s `raw` is a `json.dumps({"points": [...]})` blob, so the one place that name does exist would have dumped JSON into the prompt. I'd taken "no writer produces this" as a reason to keep reading it defensively, which is backwards — an unreachable branch reading a field that means something else elsewhere is a trap, not compatibility.

Reworked to render `warm_start_context`:

**1–2.** `_warm_recipe_body` is deleted along with the `raw` / `symptom` / `attrs` fallbacks. The block renders `status` for the states, `match.tier`/`confidence` on the header, `recommended_replay` for config, and `ctx["pitfalls"]` for the rows. Net −40 lines of guesswork against the raw row.

Your point 2 was the one I'd have shipped without noticing. A sibling-tier donor now reads:

```
recipe: hit tier=arch_sibling confidence=0.40
  · best_throughput=875.0
  · extra_server_args=--tp 8
  · expected_gain=12.30%
  · borrowed from llama-3-70b (config_tier=arch_sibling, confidence=0.40)
```

And because `recommended_replay` comes from `_config_replay_args_envs`, the production `best_config` shape splits properly instead of rendering a dict repr:

```
  · extra_server_args=--attention-backend=AITER --enable-chunked-prefill
  · extra_envs=VLLM_ROCM_USE_AITER=1 VLLM_USE_TRITON_FLASH_ATTN=0
```

A remote hit has no `recommended_replay` by construction, so it renders the tier, confidence, prior counts and pitfalls, and says the match carries no replayable config — true for remote, rather than the old line's implication that nothing was found.

**3.** The pitfall header counts rendered rows and elides the remainder with the `to_gaps_summary` idiom.

**4–5.** `warm_start_recipe`'s comment now describes the `{workload, hw, tier, confidence, recipe}` envelope and says it's bookkeeping, not a prompt input. The `warm_start_pitfalls` "KB point dicts" comment is fixed in #1464 — same root cause, and leaving it there avoids a conflict between the two branches. The docstring says the block renders `warm_start_context` and that the states come from `status`.

**6.** Both the `{"raw": ...}` legacy test and my wrapped-pitfall test are gone, and the fixtures now run through `run_t0_anchor`. That paid for itself immediately: **a genuine first session is `seed_only`, not `miss`**, because T0 seeds a row for every session. My first version only special-cased `miss`, so every real first session would have rendered as a hit with no config — a new false statement, in the same block, introduced by the fix for the old one. No hand-built fixture would have caught that, which is your argument.

**7.** CHANGELOG trimmed to the operator-facing version.

Same 8 pre-existing failures as a clean `main` checkout, none in this path.

One judgement call worth your eyes: I kept the prior counts line (`· priors: proven=2 avoid=1 lessons=3`) from `ctx`, because otherwise a remote hit renders a header and little else. Happy to drop it if you'd rather the block stay strictly about replay.

ZhengGong-amd (approved):
Approved

PR #1472, opened 2026-09-10
fix(agents): unify agent backend selection on credentials, then SDK

fengshaoyi-amd:
## What

Four places decided "claude or codex" and three of them disagreed. This makes it one rule, shared by both packages: **a configured credential first, then an installed SDK, with Claude ahead of Codex.**

## Why

| Path | What it read | What went wrong |
|---|---|---|
| `select_default_agent_provider` (forge-loop) | SDK availability only | An OpenAI-only box resolved to Claude whenever both extras happened to be installed, then failed to authenticate |
| `_credential_shape` (forge-fuse) | Its own key set | Missed `OPENAI_BASE_URL`; handed dual-configured boxes back to the SDK order |
| Specialists / TraceLens runner / `_resolve_forge_agent` | `llm_config.is_openai_only` | Agreed with each other, but each re-derived the predicate locally |
| Robustness RCA (`_discover_llm_credentials`) | OpenAI side, unconditionally | The only place preferring Codex, so a dual-configured deployment ran RCA on GPT while every other role ran Claude |

## How

`llm_config.preferred_agent_backend` is the credential-shape answer for the whole repository. A provider declares its own side through the new `AgentProvider.credentialed`, so `select_default_agent_provider` ranks by three derived keys instead of a chain of provider names, and Claude-ahead-of-Codex falls out of registration order rather than being restated:

```python
ranks = {
    provider.name: (
        0 if _holds(provider, lambda: provider.credentialed(os.environ)) else 1,
        0 if _holds(provider, provider.availability) else 1,
        0 if model and _holds(provider, lambda: provider.owns_model(model)) else 1,
    )
    for provider in providers
}
```

Model ownership (added in #1442) stays the **last** key, not the first. An owner that cannot run is worse than a fallback that can — `test_select_skips_unavailable_model_owner` is the existing statement of that — and on the dual-configured box where a named model is actually worth routing, the first two keys tie and ownership is what decides.

`_credential_shape` and `resolve_specialist_agent_backend` are deleted; `fusion/command.py` loses its whole if/elif and just asks the registry.

## Behaviour changes

- `CLAUDE_CODE_USE_BEDROCK` / `CLAUDE_CODE_USE_VERTEX` now count as a configured Anthropic side. A Bedrock or Vertex box used to read as unconfigured and get redirected to Codex. (`forge-fuse` already counted them; `llm_config` did not — that gap was the divergence.)
- **A deployment with no credential either package can see is no longer refused up front.** `forge-fuse` drops its `--agent-backend auto` usage error and forge-fusion drops the `llm_provider_unconfigured` result. A runtime logged in by other means carries no credential these can read, and its own preflight is what reports a genuine authentication failure. A provider missing *both* a credential and its SDK is still refused, now naming both.
- An Anthropic-only box with the `claude` extra missing now resolves to Claude and reports the absent extra, instead of silently running Codex with no OpenAI key.

## Tests

Rewritten around the new tiers in `test_resolve_and_cli.py`, with an `installed_sdks` fixture that pins registry availability so the credential tier can be tested on its own: credentials outrank an installed SDK, SDK availability is the fallback, an unconfigured environment defaults to Claude, and only "neither" is refused. Three tests asserting the old refusals are gone (two in `test_kernel_request_handlers_units.py`, one in `test_resolve_and_cli.py`); two RCA discovery assertions flip from `openai` to `anthropic`.

Verified against `origin/main` on the same machine — failure sets are identical name-by-name (13/13 in the KernelForge batch, 18/18 in the Hyperloom batch; all pre-existing Windows path and permission issues), with `+3` and `-1` net passing tests matching the tests added and deleted. `ruff check` / `ruff format --check` clean.

> Reviewer note: this replaces an earlier attempt that was branched before #1424 and #1442. It was re-applied onto current `main` rather than rebased, because #1424 deleted `forge_submit.py` outright and a conflict resolution would have resurrected it.


xiaofei-zheng:
Reviewed the unified selection rule end to end: `select_default_agent_provider` ranking, `preferred_agent_backend`, the `forge-fuse` auto path, the RCA credential discovery, and the specialist/subprocess call sites. Verified against `git merge-base origin/main` with small repro scripts and by running `test_resolve_and_cli.py`, `test_specialist_codex_backend.py` and `robustness/tests/test_factory.py` on both trees (no new product-test failures; the deltas I saw were my own exported `ANTHROPIC_*` env leaking into the tests).

Two blocking issues.

### 1. `forge-fuse --agent-backend auto` now selects Codex on a box with `OPENAI_BASE_URL` and no OpenAI key

`src/kernelforge/fusion/command.py:97` replaces `_credential_shape()` with `select_default_agent_provider()`, which ranks on `has_openai_side()`. That helper is an `any()` over `("OPENAI_BASE_URL", "OPENAI_API_KEY")` (`src/hyperloom/common/llm_config.py:128,146`), so a bare gateway pointer with no key counts as a configured OpenAI side. The deleted `_credential_shape()` required `OPENAI_API_KEY`.

Repro on this branch:

```
env: OPENAI_BASE_URL=https://gw/v1   (no OPENAI_API_KEY, no Anthropic side)
preferred_agent_backend(...) -> codex
select_default_agent_provider().name -> codex
```

codex ranks `(0, ...)` and claude `(1, ...)`, so the credential key decides before anything else. On merge-base the same environment raised `click.UsageError("--agent-backend auto found no OpenAI or Anthropic credentials; ...")`. Now the run starts against an unauthenticated Codex backend and fails at the first API call instead of at CLI parse.

Two things suggest this widening was not intended:

- `docs/reference/environment-variables.md` (touched by this PR) still says `OPENAI_BASE_URL` is "Required together with `OPENAI_API_KEY` to enable Codex".
- `test_a_retired_key_does_not_configure_a_provider` (`src/kernelforge/tests/fusion/test_resolve_and_cli.py:148`) had its `monkeypatch.setenv("OPENAI_BASE_URL", ...)` line removed in this PR. Put that line back and the new assertion `provider == "claude"` fails — it returns `codex`. That is exactly the coverage that pinned the old semantics.

Suggested fix: make the OpenAI-side credential predicate used for *selection* require `OPENAI_API_KEY` (i.e. keep `_codex_credentialed` distinct from `has_openai_side`, which is also used for endpoint-shape questions where the URL alone is meaningful), and restore the `OPENAI_BASE_URL` setenv in that test.

### 2. RCA silently degrades to `NoopRcaEngine` on a subscription-token host that also has a working OpenAI key

`_discover_llm_credentials` (`src/hyperloom/agents/robustness/config.py:302`) now asks the preferred side first, but `_anthropic_rca_credentials` answers on credential *presence*, not usability: with `CLAUDE_CODE_OAUTH_TOKEN` set it returns `("", "", "anthropic")` at line 297 — a non-`None` tuple with an empty key — so the OpenAI resolver is never reached.

Repro, `CLAUDE_CODE_OAUTH_TOKEN` + a working `OPENAI_API_KEY`/`OPENAI_BASE_URL`:

```
merge-base: ('https://api.openai.com/v1', 'sk-openai-working', 'openai')
this PR:    ('', '', 'anthropic')
```

That is fine while the Claude CLI is present (`anthropic_transport_ready()` is True via the SDK transport). When it is not — a robustness container without the CLI — `factory._llm_credentials_ready` returns False and `_build_rca_engine` returns `NoopRcaEngine`. Since `llm_rca_enabled` defaults to `None`, the `log.warning` at `factory.py:289` is skipped, so RCA turns itself off with no log line at all, on a deployment where it worked before this PR.

The docstring claims the resolver "falls through to the other one when that side carries no usable key", but usability is never tested. Either gate the subscription-token branch on `anthropic_transport_ready(env)` before returning it, or drop through to the OpenAI side when the Anthropic tuple carries no key.

Related and worth folding into the same fix: the unconfigured fallback flipped from `openai` to `anthropic`, so a deployment with zero credentials now hits the `config.llm_provider == "anthropic"` arm of the startup log at `config.py:233` and prints `llm=(configured)`. The one operator-facing signal that RCA has no credential now says the opposite.

### Checked and not blocking

- `ClaudeBackend` missing the `claude` extra is **not** deferred to `run()`: `_load_claude_sdk()` runs in `__init__` (`claude.py:283`), `ClaudeUnavailableError` subclasses `AgentProviderUnavailableError`, and `_prepare_backend` constructs the backend inside the `try` in `create_registered_backend`, so the `fallback_provider` path still fires.
- Demoting `owns_model` below the credential key can route an explicitly named model to the other provider, but only when the named model's owner has no credential — where both the old and the new selection end in a failed run, so this is a different error message, not a new failure.

xiaofei-zheng:
Follow-up on the same PR, separate in kind from the two behaviour regressions above: this one is about the stated goal itself. The unification is only half done — the "credentials, then SDK" pair exists on the KernelForge side only. Hyperloom's side is a single-key rule that never looks at what is installed.

### Two entry points, two different rules

`preferred_agent_backend()` (`llm_config.py:162`) ranks on credentials alone. Every Hyperloom call site uses it:

- `orchestrator/specialists/subprocess_.py:1181`
- `inference_optimizer/cli/executors.py:105`
- `orchestrator/kernel/request_handlers.py:4753`
- `agents/kernel/tools/tracelens_skill_runner.py:442`
- `agents/robustness/config.py:312`

`select_default_agent_provider()` (`registry.py:167`) ranks on `(credentialed, availability, owns_model)`. Every KernelForge call site uses it:

- `kernelforge/config.py:251`
- `kernelforge/fusion/command.py:97`
- `kernelforge/gemm_tune/tier3/generate.py:103`

So the SDK key — the second half of "credentials, then SDK" — is absent from five of the eight call sites.

### It diverges on exactly the shape this PR added handling for

Pinning `availability` on both registered providers and clearing every provider env var:

```
no creds, only codex SDK    hyperloom=claude   kernelforge=codex    <-- diverge
no creds, only claude SDK   hyperloom=claude   kernelforge=claude
anthropic key, codex SDK    hyperloom=claude   kernelforge=claude
openai key, claude SDK      hyperloom=codex    kernelforge=codex
```

The "neither side configured" case is the one this PR deliberately routes to Claude ("a runtime logged in by other means carries no credential this can see") and covers with `test_auto_agent_backend_defaults_an_unconfigured_environment_to_claude` and `test_auto_agent_backend_falls_back_to_the_installed_sdk`. On such a box with only the `codex` extra installed, `forge-fuse` resolves Codex while the specialist dispatcher and the forge-fusion request handler resolve Claude and try to spawn a CLI that is not there. That is the same class of cross-component disagreement the PR description sets out to remove, just relocated.

The docstring on `preferred_agent_backend` acknowledges the gap ("Callers that also need to know whether a backend is *installed* rank this answer above SDK availability rather than below it") but the Hyperloom callers are not "callers that don't need to know" — `executors.py:105` immediately resolves the executable for the backend it just picked, and `subprocess_.py:_agent_backend` returns the CLI name to spawn.

Suggested fix: give `preferred_agent_backend` an optional installed-check (or have the Hyperloom call sites go through a shared helper that applies the same two keys), so that "no credential on either side" falls to whatever is actually installed everywhere rather than only under KernelForge.

### Checked, not a problem

`_resolve_forge_agent(payload, env=...)` threads an env mapping into `preferred_agent_backend` while `select_default_agent_provider` reads `os.environ` directly. The only production caller is `request_handlers.py:4872`, which passes no `env`, so both read `os.environ` today; the parameter is exercised only by `test_kernel_request_handlers_units.py`. Worth keeping in mind if that argument ever gets a real caller, but it is not a live divergence.

fengshaoyi-amd:
Both blocking issues are fixed, plus a third in the same chain that the review did not reach. CI is green on `e7a76143c` and main is merged in (CHANGELOG was the only conflict; both sides' sections are kept).

**1. Bare `OPENAI_BASE_URL`** — `openai_agent_credentialed` now requires `OPENAI_API_KEY`, kept distinct from `has_openai_side` exactly as you suggested. The `OPENAI_BASE_URL` setenv is back in `test_a_retired_key_does_not_configure_a_provider`, and `test_auto_agent_backend_ignores_a_bare_openai_base_url` pins it directly.

**2. Subscription-token RCA** — the oauth branch is gated on `anthropic_transport_ready(env)`. The `llm=(configured)` worry you folded in does not apply: the unconfigured fallback is still `("", "", "openai")`, so a zero-credential host logs `(not available)`.

**3. Not in your review, same failure mode** — `_llm_credentials_ready` called `anthropic_transport_ready()` for *every* Anthropic provider, reading `os.environ` while the provider had been resolved against the normalized copy from `_provider_env()`. A box configured through the retired `DEEPSEEK_API_KEY` carries no `ANTHROPIC_*` outside that copy, so RCA degraded to `NoopRcaEngine` — silently, since `llm_rca_enabled` defaults to `None` and skips the warning, while startup still logged `llm=(configured)`. Only the keyless shape is spent by the CLI; a side that did resolve a key is an HTTP Messages client and needs no transport probe. Measured on a `DEEPSEEK_API_KEY`-only host: merge-base `LlmRcaEngine`, before this fix `NoopRcaEngine`, now `AnthropicRcaEngine`. `test_factory_uses_anthropic_engine_for_provider` sets exactly that shape but was pinning `anthropic_transport_ready` to `True`, which is what hid it — that monkeypatch is gone, so the test now covers the regression.

### Please re-check one thing you signed off on

You read the `owns_model` demotion as "a different error message, not a new failure". That holds only when the named model's owner genuinely cannot run, and it does not hold for `codex login`: ChatGPT subscription auth lives in `~/.codex/auth.json` and sets no env var — the "logged in by other means" case this PR's own docstring says the credential key cannot see. There, merge-base ran the named model on its owner; the ranking sent it to Claude.

Measured on an Anthropic-key box with both extras installed:

```
select_default_agent_provider("gpt-5.6-sol")        merge-base -> codex   ranked -> claude
Config(agent_backend="auto", agent_model="gpt-5.6-sol")        -> (claude, "gpt-5.6-sol")
```

That second pair is what the comment above `Config.agent_runtime` exists to prevent, in the other direction. Two production callers pass a model: `Config.agent_runtime` and `gemm_tune/tier3/generate.py:103`.

So ownership is no longer a ranking key. Ownership answers what the caller wants; the credential and SDK keys answer what the box can run — ranking all three in one tuple is what made them fight. A named model now narrows the candidates to the providers that own it and can run, and the two ranked keys stay credential-then-SDK for every other call, which is the shape `forge-fuse` needs. `preferred_agent_backend` is untouched, so the two sides still read the same rule.

Consequences worth checking:

- `test_select_skips_unavailable_model_owner` keeps its meaning — an owner that cannot run is still worse than a fallback that can, now stated by the filter rather than implied by a tuple position.
- The refusal test loses its `[:2]` slice, which only existed because the third key had to be excluded from it.
- `test_select_prefers_model_owning_provider` was failing with `assert 'claude' == 'betacli'` for anyone with `ANTHROPIC_API_KEY` exported, and passed in CI only because CI has neither a credential nor either SDK installed. It is environment-independent again. `test_a_credential_does_not_take_another_provider_s_named_model` is the direct guard, since on a clean CI env neither fake provider is credentialed and nothing else would catch the regression.

The PR's own core case is unchanged: OpenAI-only with both extras installed and no model named still resolves Codex.

### One thing I checked and am not changing

I said above in an earlier revision of this comment that `preferred_agent_backend` reads whether a *Python package* is importable while several callers spawn a *CLI*, and that I was leaving the mismatch for a separate PR. Having looked properly, the current shape is right and I withdraw that.

The two channels really do disagree — this box has `claude_agent_sdk` with no `claude` on `$PATH`, and `codex` on `$PATH` with no `openai_codex` — but the disagreement does not produce a wrong decision. `executors.py` already separates *which side to run* from *which dispatch mode to use*: `preferred_agent_backend` picks the side, then `shutil.which` picks subprocess or in-process. On the box above with no credential, selection returns Claude because the package is present, `shutil.which("claude")` is empty, and the dispatcher falls back to the in-process Claude backend — exactly the path `claude_agent_sdk` serves. The package key is the right signal for the in-process path, and that is the path with no other escape hatch; each CLI path is gated by its own resolver with a specific error (`executors.py:116`, `subprocess_.py:1229`).

Making the key exact would mean threading a per-caller predicate through five call sites for three different notions of "available" — `subprocess_` and `executors` want a CLI, `tracelens_skill_runner` runs the SDK in-process, `robustness/config` is an HTTP call that wants neither — to refine a tie-breaker that only fires when the credential key ties, behind callers that already report what is missing. The docstring states the contract that makes that correct: a provider missing a key is still returned so its own preflight reports the absent extra.

One related note: the specific box in your second comment — no credential, only the `codex` extra installed — now agrees on both sides, measured `hyperloom=codex kernelforge=codex`. That paragraph described the code before the SDK key existed.

xiaofei-zheng (approved):
Re-reviewed at `1b3583d8d` (merge-base `888831aa7`). All three points from my earlier comments are fixed, each with a test that pins the behaviour. **No blocking issues — LGTM.**

### How the three findings were addressed

**1. Bare `OPENAI_BASE_URL` selecting Codex.** The selection keys no longer reuse the endpoint-shape helpers. `openai_agent_credentialed` requires `OPENAI_API_KEY`, `anthropic_agent_credentialed` is `has_anthropic_credential` widened only by the managed gateways, and `has_anthropic_side` is restored to its original meaning so the other consumers of the shape helpers are untouched. Verified: `OPENAI_BASE_URL=https://gw/v1` with no key now resolves `claude` on both sides. `test_auto_agent_backend_ignores_a_bare_openai_base_url` covers it, and the `OPENAI_BASE_URL` setenv is back in `test_a_retired_key_does_not_configure_a_provider`.

**2. RCA shadowing a working OpenAI key.** The subscription-token branch is now gated on `anthropic_transport_ready(env)` and falls through when the CLI is absent; the unconfigured tuple is back to `("", "", "openai")`, which also fixes the `llm=(configured)` startup log. `_llm_credentials_ready` additionally stops probing the CLI for a keyed Anthropic side. Verified:

```
oauth + working openai key, transport unusable -> ('https://api.openai.com/v1', 'sk-openai-working', 'openai')
no creds at all                                -> ('', '', 'openai')
```

**3. Half-done unification.** `preferred_agent_backend` now applies the same two ranked keys as `select_default_agent_provider`, and `_claude_model_should_follow_codex` / `_build_backends` were the last two places deriving the rule for themselves. Cross-checked both entry points over eleven credential/SDK shapes — they agree on every one:

```
case                             hyperloom  kernelforge
only OPENAI_BASE_URL (no key)    claude     claude
no creds, only codex SDK         codex      codex
no creds, only claude SDK        claude     claude
no creds, both SDKs              claude     claude
no creds, no SDK                 claude     RAISE (intentional)
openai key only                  codex      codex
anthropic key only               claude     claude
both keys                        claude     claude
bedrock only                     claude     claude
bedrock + openai key             claude     claude
openai key, only claude SDK      codex      codex
```

The one asymmetry is deliberate and documented: with neither credential nor extra, KernelForge raises `AgentProviderUnavailableError` while Hyperloom returns Claude so its own preflight reports what is missing.

### Also checked

- `_credential_unproven` returning `False` closes the hole where any installed plugin outranked both built-ins on an unconfigured box; `owns_model` moved from a third ranking key to a candidate filter, so a named model is no longer overruled by the other side's credential. Both have tests in `test_provider_registry.py`.
- `preferred_agent_backend` detects the SDKs with its own `find_spec` calls rather than the registry's `availability`. The two copies use identical module names (`claude_agent_sdk`, `openai_codex`), and the `llm` extra ships both while `runtime` pulls `[llm,web,forge]`, so the second key cannot disagree between the two entry points in any packaged install. Worth remembering that it is a proxy — the Hyperloom subprocess path spawns CLI binaries, not the Python packages — but it does not misfire as packaged today.
- No dangling references after the renames (`_official_openai_only`, `ANTHROPIC_MANAGED_GATEWAY_ENVS`), and `is_truthy` is still used.
- `docs/reference/environment-variables.md` now scopes Bedrock/Vertex to "agent backend selection only" and states they satisfy neither `--critic-protocol=anthropic` nor the credential preflight, which matches `anthropic_agent_credentialed`.

### Tests

Ran the changed files at `1b3583d8d` and at the merge-base, in parallel, and diffed the failures: **identical on both** (28 failures, all pre-existing — Windows path separators and my own exported `ANTHROPIC_*`/`OPENAI_*` leaking past the fixtures). No new failures from this PR.

`test_provider_registry.py`, `test_resolve_and_cli.py`, `test_llm_config.py`, `robustness/tests/test_factory.py`, `test_startup_robustness.py`, `test_specialist_codex_backend.py`, `test_kernel_request_handlers_units.py`.

PR #1473, opened 2026-09-10
fix(forge): let a hookless provider run concurrent lanes

jiaqiang-dot-liu:
## Summary

Concurrent Implementer lanes refused any agent provider that does not declare `stop_hooks`. Codex does not, and `rewrite_by_flydsl/optimize.py` never passes `--lanes`, so `forge-rewrite-by-flydsl` inherited the default 3 and its entire OPTIMIZE phase was unreachable on Codex — PORT succeeded, then `forge-loop` exited before its first iteration:

```
Error: --lanes 3 needs guarantees that agent provider 'codex' does not declare:
stop_hooks (it must run the callbacks in AgentRunSpec.hooks, ...).
Re-run with --lanes 1, or select a provider that declares them.
```

Two changes, one per commit:

1. **The lane gate refuses on `session_env` alone.** `stop_hooks` becomes an advisory that names what the run gives up.
2. **The gate's hooks are built only for a provider that runs them.** Claude gets every hook exactly as before; Codex, which reads `AgentRunSpec.hooks` nowhere, is no longer handed a group it drops in silence.

## Why the line falls where it does

The two guarantees fail differently, and only one of them fails silently.

Without `session_env` a lane has no private AITER build cache. Lanes sharing one cache load each other's JIT binaries — aiter resolves a module by name and never checks the `.so` against the source it was built from — so a lane measures a kernel it did not write and **nothing downstream can tell that from a real result**. That stays a refusal.

Without `stop_hooks` a lane loses in-session denial: the `PreToolUse` callbacks that refuse an edit to the driver, harness or oracle, and refuse a driver run that skips the shared device lock. Worth having, but no published number rests on it:

- `IterationRunner._lane_rejection` refuses a candidate on its patch paths before it reaches the canonical tree
- `_validate_driver_integrity` re-checks the canonical driver byte-for-byte once the patch applies
- every candidate is re-measured one at a time in the canonical tree, so a lane's own timings never decide a KEEP
- the lane teardown's contention report voids a round that benchmarked unlocked, out loud

So the cost of running without the hooks is wasted budget, reported when it happens. That is a cost to warn about, not a result to refuse.

This deliberately reverses the decision `test_lanes_are_refused_on_a_provider_that_does_not_run_our_hooks` recorded. The rationale is written into that test's docstring and into the constants, so the next reader sees why rather than merely that it changed.

## Why the second commit

Hook usage was already per-provider at runtime and this does not change that: Claude translates the groups through one path keyed on `spec.hooks is not None`, and Codex has no equivalent. What changed is that the call site now says so. Building a protection object for a backend that discards it is the same misreading the refusal was guarding against, one layer down — a reader of `make_agent_fn` would conclude every session is protected.

Keyed on the backend that was *resolved*, not the one requested, so a provider fallback carries the decision with it.

## Notes

- The advisory prints *before* any refusal, so an operator whose provider lacks both hears everything at once — switching to a provider that clears the refusal is not the moment they first learn what it still will not do. `test_a_lane_refusal_names_every_missing_guarantee` asserts both halves.
- `--lanes 1` is untouched, and `test_lanes_are_refused_rather_than_quietly_reduced` still passes unchanged: that rule holds for `session_env`.
- The check still reads what a provider *declares*, never its name, so a third-party provider registering `stop_hooks=True` is treated like Claude and one registering nothing like Codex.
- This does not make Codex lanes as reliable as Claude's. The device lock is still opt-in by filename — the wrapper installs as `forge_lane_driver.py` while the real driver keeps its own name — so a hookless lane can still bypass it and void the round. Making the lock non-bypassable is a separate change and is not attempted here.

## Test plan

- [x] `test_lane_session_guarantees.py`, `test_provider_registry.py`, `test_agent_run_spec_contract.py`, `test_insession_gate_logic.py`, `test_session_timeout_wiring.py`, `test_claude_resume.py`, `test_measurement_fidelity.py`: 156 passed
- [x] New `test_the_builtin_hookless_provider_runs_lanes_and_is_warned` ties the change to the real `codex` provider rather than a stand-in
- [x] New `test_a_hookless_provider_is_given_no_hooks_to_drop` locks the second commit, alongside the unchanged hook-capable assertion
- [x] New `test_a_provider_that_declares_both_is_warned_about_nothing` keeps the advisory from degrading into a banner on every run
- [x] Reproduced the original failure end to end: `forge-rewrite-by-flydsl` on Codex/`gpt-5.6-sol` reached `PORT OK (attempt 1, SNR=143.43)` and then died at the lanes gate
- [x] Re-ran the same command on this branch: the advisory prints and `forge-loop` proceeds into its iteration loop on Codex
- [x] Ruff clean on all changed source files
- [ ] CI for the full suite


xiaofei-zheng (approved):
Reviewed both commits against `git merge-base origin/main fix/lane-provider-hookless-advisory`.

**No blocking issues. LGTM.**

What I checked, and why each claim in the PR holds:

1. **Downgrading `stop_hooks` to an advisory does not let a bad number reach a KEEP.** Traced the lane candidate path end to end: `run_lanes._one` returns a diff only (`loop/fanout.py`), `_fill_lane_queue` queues `LaneResult.diff` (`loop/runner.py:1719`), and `_next_lane_candidate` re-checks the patch paths through `_lane_rejection` -> `is_protected_path` with the campaign driver as an exact path (`loop/runner.py:1762-1780`) before the patch touches the canonical tree. A lane's in-session timings are never carried into the queue at all, so they cannot decide a KEEP. That is the property that makes `session_env` (shared JIT build cache -> a lane measures a sibling's `.so`, indistinguishable from a real result) a refusal and `stop_hooks` an advisory. The split is drawn in the right place.

2. **Commit 2 is behaviour-preserving for both built-in providers.** `AgentCapabilities.stop_hooks` is documented as "whether the provider runs the callbacks in `AgentRunSpec.hooks`" (`agent_backends/base.py:79`) — the whole group, not just Stop — so gating `hooks=` on it in `orchestrator/agent.py:610` withholds nothing from a provider that would have run them. `claude` declares `stop_hooks=True` and is the only backend reading `spec.hooks` (`agent_backends/claude.py:406-411`), so its options (including the `bare` extra-arg interaction) are unchanged; `codex` declares `stop_hooks=False` and dropped the group already. `uses_outer_gate` at `orchestrator/agent.py:650` keys on the same capability, so the hookless path still gets Stop decisions driven from outside between resume turns — the two now agree instead of one building what the other assumes is absent.

3. **The advisory does not disturb the consumer that made this bug visible.** `rewrite_by_flydsl/optimize.py` streams forge-loop with `stderr=STDOUT` and parses only the `__FORGE_RESULT__` sentinel line, so the extra `[lanes] WARNING:` line passes through harmlessly.

4. **The advisory fires on capability, not name.** `_require_lane_provider_capabilities` reads `_LANE_PROVIDER_ADVISORIES` before building the refusal list, so a provider missing both is told both in one run, and the `fullcli` test pins that a fully-capable provider prints nothing.

One thing to be aware of rather than to change here: the teardown contention report cited as the after-the-fact safety net covers processes *still alive* when the lane session ends — `ReapReport.contended` is `unkillable | holding_device` (`llm/process_reaping.py:60-70`). A hookless lane that benchmarks around the `forge_lane_driver.py` wrapper and exits cleanly leaves `contended` False, so that particular wasted round is not reported. It does not change the correctness argument above (no lane timing reaches a KEEP), and the PR already states that making the device lock non-bypassable is a separate change.

Tests were not run locally (this suite is POSIX-only; `kernelforge.tracker.experiment` imports `fcntl`). CI covers them.

PR #1474, opened 2026-09-10
fix(forge): report complete FlyDSL rewrite LLM usage

jiaqiang-dot-liu:
## Summary
- thread a shared usage accumulator through FlyDSL driver preparation, PORT, and apply-back agent calls
- combine in-process usage with the nested forge-loop usage while preserving token, call, cost, and cost-availability semantics
- expose the cumulative `llm_usage` block in all final FlyDSL rewrite result payloads, including failure results

## Test plan
- [x] `python -m pytest -q src/kernelforge/tests/test_usage.py src/kernelforge/tests/test_rewrite_by_flydsl.py src/kernelforge/tests/test_rewrite_by_flydsl_pipeline.py`
- [x] `python -m pytest -q src/kernelforge/tests/test_flydsl_rewrite_driver_preparation.py src/kernelforge/tests/test_rewrite_by_flydsl_applyback.py`
- [x] `python -m ruff check` on all modified Python files
- [ ] full CI suite

### DIFF
--- a/src/kernelforge/rewrite_by_flydsl/applyback.py
+++ b/src/kernelforge/rewrite_by_flydsl/applyback.py
@@ -33,6 +33,7 @@ from kernelforge.rewrite_by_flydsl.budget import DEFAULT_REWRITE_BUDGET
 from kernelforge.rewrite_by_flydsl.protocol import validate_applyback_manifest
 from kernelforge.rewrite_by_flydsl.spec import RewriteSpec
 from kernelforge.durable_io import atomic_write_text
+from kernelforge.tracker import UsageAccumulator
 
 # Framework apply-back artifacts live beside, never inside, the artifact paths the nested standalone FlyDSL forge-loop
 # owns (``forge_experiments/best*``).
@@ -326,6 +327,7 @@ async def _run_agent(
     timeout_sec: int,
     progress_log: list[str],
     prior_failure: str = "",
+    usage: UsageAccumulator | None = None,
 ) -> tuple[str, str]:
     runtime = config.agent_runtime()
     backend = create_registered_backend(
@@ -366,7 +368,7 @@ async def _run_agent(
         ),
     )
     result = await asyncio.wait_for(
-        backend.run(run_spec),
+        backend.run(run_spec, usage=usage),
         timeout=watchdog_timeout_sec(timeout_sec),
     )
     # A turn cap or SDK error leaves a half-rewired integration that passes host validation and every gate after it,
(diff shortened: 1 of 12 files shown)


xiaofei-zheng (approved):
No blocking issues — LGTM.

**What this PR does:** threads a single `UsageAccumulator` through the FlyDSL rewrite pipeline (driver preparation, PORT, apply-back), adds `combine_usage_totals()` to fold the in-process totals together with the nested forge-loop's `llm_usage`, and exposes the cumulative `llm_usage` block on every `RewriteResult` payload (setup failure, PORT failure, interim, final).

**Paths I checked:**
- Agent call-site coverage: every LLM entry point reachable from `run_rewrite` now receives the accumulator — `prepare_rewrite_driver` → `_run_agent` → `backend.run(..., usage=usage)`, `run_port_loop` → `make_agent_fn(usage=usage)` (threaded to `backend.run`/`backend.resume` in `orchestrator/agent.py`), and `generate_applyback_patch` → `_run_agent`. `try_flydsl_kb_warmstart` makes no LLM calls, so nothing is missed there.
- Backend signatures: `AgentBackend.run`, `ClaudeBackend.run` and `CodexBackend.run` all already accept `usage`, and both concrete backends fold it in (`usage.add_from_message` / `usage.add_usage`). No signature mismatch at any registered backend.
- No double counting: OPTIMIZE runs forge-loop as a subprocess with its own accumulator and reports `llm_usage` in its result dict; the in-process accumulator is never passed to `run_optimize`, so the two records are disjoint before being combined.
- Result completeness: all four `report.build_result` call sites in `runner.py` pass `llm_usage=_total_usage(...)`, and the final one passes `opt` so the nested totals are included. `RewriteResult` gains a required field but `report.build_result` is its only constructor, and `validate_applyback_outer_result` does not reject extra keys.
- `combine_usage_totals` semantics: I executed `tracker/usage.py` standalone and exercised it against `UsageAccumulator.totals()` outputs — empty+empty → `unavailable`/calls 0; priced+priced → `provider` with summed tokens/calls/cost; priced + a record with calls but no cost keys → `partial` with `cost_available=False`. This matches `UsageAccumulator.totals()`'s own provenance rules, so downstream consumers (`forge_runner.llm_usage`, `controller._forge_llm_usage`) keep consistent semantics. Per-key `contextlib.suppress` is inside the loop body, so one malformed key does not drop the rest of the record.
- Compared against `git merge-base origin/main` — before this change `result["llm_usage"]` was absent for FlyDSL rewrites, so `ForgeRunner.llm_usage` always resolved to empty; that is the behavior this PR fixes.

**Tests:** I could not run the suite locally (Windows host — `kernelforge.tracker.experiment` imports `fcntl`), so I verified `combine_usage_totals` by loading the module directly and asserting the combinations above. The new tests in `test_usage.py`, `test_rewrite_by_flydsl.py` and `test_rewrite_by_flydsl_pipeline.py` cover the same cases plus accumulator identity across PORT and apply-back; CI covers the rest.

PR #1475, opened 2026-09-10
ci: balance six test shards and remove redundant coverage work

zengleixin-amd:
## Summary
- Resolve the rolling duration cache key once per workflow and require all shards to restore that exact immutable seed. Preserve worker-restart safety and reject invalid or incomplete timing artifacts.
- Disable per-shard coverage reports while preserving measurement. Install only coverage[toml] for aggregation and reuse one report with its native exit status.
- Preserve Python 3.10/3.11 and the strict 90% coverage gate; relaxed mode exempts only threshold failures, not report errors.
- Remove unrelated settling waits from three non-timing grid tests, retaining their assertions and the dedicated settling tests.
- Run six shards per Python version with two workers per shard.

## Review follow-up
The temporary walkthrough change from 300 to 24 reactor passes did not establish equivalent long-running coverage. It has been reverted in cbb18d541: the complete walkthrough file now matches the original version, including 300 passes for the first test and 66 for its sibling. None of the timing savings from the 24-pass experiment are counted as coverage-preserving optimization. The final configuration is being re-measured before merge.

## Validation
- Earlier local integration: 112 regressions passed under xdist with coverage and duration collection; subsequent fixed-seed changes passed 57 focused duration/workflow tests.
- Fixed-seed CI passed a complete run, a specific-shard rerun, and another complete rerun. All 12 shards logged the same exact cache key; mixed old/new artifacts combined correctly on the partial rerun without additional permissions or overwrite settings.
- Six grid/settling tests passed locally before and after their test-only changes (125.82s to 5.89s). This is subset timing, not complete CI speedup.
- Historical coverage data produced identical report output and exit status under the minimal reporting environment.
- Ruff lint/format, Python 3.10 syntax checks, actionlint, and diff checks passed. Local actionlint did not enable ShellCheck or Pyflakes integration.
- Full Linux Python 3.10/3.11 CI is running again with the restored 300-pass walkthrough. Prior short-walkthrough results are not the final performance baseline.

## Scope
- Single concern: faster non-E2E CI without reducing existing test execution depth or failure gates.
- No production behavior changes. The walkthrough shortening and temporary timing probe have both been removed.
- No linked issue; the causes are local CI and test configuration.
- No application breaking change. The test extra includes the CI extra needed by the regression tests.
- Compare queue delay, pytest execution, coverage aggregation, other checks, and runner minutes separately. Exclude main-only cache maintenance from PR feedback time.


xiaofei-zheng:
## What this PR does

Reworks the non-E2E CI path in `tests-coverage.yml` for speed without changing the 90% gate:

- **Sharding**: `total_shards` 4 → 6 per Python leg, matrix list mirrored in the workflow, still 2 xdist workers per job.
- **Duration seed**: a new `prepare-durations` job resolves one immutable cache key per workflow run (`lookup-only` prefix match) and pins it as an output; every shard restores that exact key, so partial re-runs cannot pick up a different seed. A new pytest plugin (`scripts/ci_durations.py`, loaded via `-p scripts.ci_durations`, hence the switch to `python -m pytest`) disables `--store-durations` in xdist workers so a replacement worker re-reads the original seed instead of a half-written file, and touches a `.complete` marker on the controller only after pytest-split has replaced the seed. Shards now seed `--durations-path .test_durations.shardN` from the cache copy and run with `--clean-durations`, so each shard publishes only its own disjoint timings. `update-durations` merges via `scripts/ci_durations.py`, which validates artifact completeness/disjointness instead of warning and exiting 0.
- **Coverage**: shards measure with `--cov-report=` (no per-shard rendering). The coverage job drops `pip install -e ".[test,ci]"` in favor of `coverage[toml]>=7.0`, and the two inline heredocs become `scripts/ci_coverage_report.py`, which renders the report once and passes coverage's native exit status to the gate step via a step output. Relaxed mode now exempts only status 2 (below `fail_under`), not report errors.
- **Test runtime**: `REPORT_SETTLE_SECONDS` is patched to 0 in three grid tests, plus an edit to `test_optimize_loop_walkthrough.py`.
- **Tests**: three new files under `scripts/tests/` covering the workflow contract, the durations plugin (real nested pytest+xdist runs), and the coverage reporter.

Verified while reviewing: `coverage[toml]` still pulls `tomli` on 3.10, so dropping the project install does not break `ci_coverage_report.py` on the 3.10 leg; `PytestSplitCachePlugin` collects durations from `terminalreporter.stats` on the xdist controller, so suppressing worker writes does not lose timings; `--clean-durations` replaces (not merges) the seed, so per-shard artifacts are disjoint and the strict merge is sound; `coverage report` exits 2 and writes the `Coverage failure: total of N is less than fail-under=90` line to stdout, which `run_report()` re-emits, so the deferred gate does not lose that signal. The `src/kernelforge` → `n/a` row in the per-tree table is pre-existing behavior on `main`, not introduced here.

## Blocking

**`src/hyperloom/inference_optimizer/tests/test_optimize_loop_walkthrough.py:67-68` — the loop change cuts the exercised work by 12.5x, and this is not a "settling wait" removal.**

```diff
-        for tick in range(1, 25):
-            await coord.tick(tick)
+        for _ in range(24):
+            await coord.tick(1)
```

`Coordinator.tick(n)` takes a *count of reactor passes*, not a tick index (`orchestrator/loop/coordinator.py:1374-1377`: `async def tick(self, n: int = 1)` → `for _ in range(n)`). The old loop therefore ran `1+2+...+24 = 300` reactor passes; the new one runs `24`. Every pass is a full `reconciler.run` + `_advance_phase_if_needed` + per-role `_reactor_pass` + dispatcher/framework/enablement pump cycle, so anything that only surfaces after roughly 25+ passes — phase-budget accumulation, reconciler drift, `pending_escalate_hint` handling — is no longer exercised by this test. The assertions are unchanged, so the loss is silent: the test still goes green while verifying much less.

This is not covered by the PR description, which attributes the test-file edits to "remove unrelated 30-second settling waits from three non-timing grid tests" and states "no tests skipped or assertions removed". The other three files do exactly that (`monkeypatch.setattr(gr, "REPORT_SETTLE_SECONDS", 0.0)`); this fourth one changes what the test covers.

The same call-site misuse is still present in the sibling test in the same file — `test_both_arms_dry_walks_the_rest_of_the_chain` at line 97-98 still does `for tick in range(1, 12): await coord.tick(tick)` (66 passes) — so if `tick(tick)` is being treated as a bug, the fix is only half applied and the two tests in one file now disagree on the convention.

Please either keep the exercised pass count equivalent (e.g. `await coord.tick(300)` in one call, or `for _ in range(300): await coord.tick(1)`) and measure whether that is actually a runtime cost worth removing, or state the reduction explicitly in the PR description and apply the same convention to line 97-98.

zengleixin-amd:
Thanks for catching this. The 24-pass change only proved the existing final assertions still passed; it did not prove equivalent long-running coverage. I restored the walkthrough to its original 300 reactor passes in cbb18d541 and updated the PR description so that reduction is no longer counted as a coverage-preserving optimization. I am re-running the full Python 3.10/3.11 CI now and will use that result for the final timing comparison.

zengleixin-amd:
Revalidation with the original walkthrough is complete: https://github.com/AMD-AGI/Hyperloom/actions/runs/34569787418 . All 12 Python 3.10/3.11 test shards and both coverage jobs passed; strict coverage is 90.37% / 90.36%. The walkthrough file exactly matches its original version (300 passes / 66 passes), and its 3.10 timing is 223.07s versus 235.86s in the corresponding main baseline, not the earlier 5.48s reduced-depth result.

Using the actual merge parent 888831aa and baseline run 34568254282, test-plus-coverage feedback was 9m13s before versus 7m13s now (2m00s / 21.7% lower in this observation). Test runner consumption increased from 48.2 to 51.23 minutes (+6.3%), so this trades modest extra parallel runner time for faster feedback. The six timing artifacts include every baseline nodeid, plus 122 regression cases, with no duplicates. This is a single-run comparison affected by scheduling/runtime variation, not a guaranteed percentage. The previous short-walkthrough performance claims are withdrawn.

xiaofei-zheng (approved):
Re-reviewed at `cbb18d5`. The blocking issue is resolved — `test_optimize_loop_walkthrough.py` is now byte-identical to the merge base (`git diff f51ca4d9 cbb18d541 -- src/hyperloom/inference_optimizer/tests/test_optimize_loop_walkthrough.py` is empty), so the 300 reactor passes are back and the PR no longer silently changes what that test verifies. The diff is now 10 files, and the three remaining test-file edits are exactly the `REPORT_SETTLE_SECONDS = 0.0` patches the description claims.

**No blocking issues. LGTM.**

Paths I checked, in case it is useful to know the depth of this review:

- **Duration seed plumbing.** Confirmed against the CI logs that duration-based sharding has never actually worked on `main`: every `main` shard prints `[pytest-split] No test durations found` (e.g. run 34568254282), because `--durations-path .test_durations.shardN` is what `PytestSplitPlugin.Base.__init__` reads, while the cache restored `.test_durations`. The `cp .test_durations ".test_durations.shard${SHARD}"` in this PR is what makes `least_duration` load a real DB for the first time — the shard logs on this branch no longer print that message.
- **Worker-restart fix.** `PytestSplitCachePlugin` collects durations from `terminalreporter.stats`, which on the xdist controller already holds every worker's `TestReport`, so suppressing `store_durations` in workers (`config.option.store_durations = False` under `workerinput`) loses no timings while removing the seed-clobbering race. The `tryfirst` ordering is required for this — it has to run before pytest-split's own `pytest_configure` registers the cache plugin — and it does.
- **Merge strictness.** `--clean-durations` replaces rather than merges (`self.cached_durations = dict(test_durations)`), and all shards partition from the same pinned seed, so per-shard artifacts really are disjoint and the strict duplicate/completeness checks in `merge_durations` cannot fire on well-formed input.
- **Coverage job dependency reduction.** Verified `coverage[toml]>=7.0` still resolves `tomli; python_full_version <= "3.11.0a6" and extra == "toml"`, so dropping `pip install -e ".[test,ci]"` does not break `ci_coverage_report.py`'s tomllib/tomli import on the 3.10 leg.
- **Deferred gate.** Reproduced locally with coverage 7.16: `coverage report --skip-empty` under `fail_under = 90` returns status 2 and writes `Coverage failure: total of N is less than fail-under=90` to **stdout**, which `run_report()` re-emits into the `Coverage summary` step log. So splitting render from enforcement does not lose the failure reason.
- **Pre-existing, not introduced here:** the `| src/kernelforge | n/a |` row. `_SOURCE_SEGMENTS` on `main` already lacked a bucket for it, so the new test only records existing behavior rather than causing it. Worth a follow-up, not a blocker for this PR.
- **CI evidence:** run 34569787418 on `cbb18d5` is green on all 12 shards + both coverage legs.

One data point for a follow-up, non-blocking. Comparing runs 34552480290 (24-pass version) and 34569787418 (restored 300-pass version) on the py3.10 leg, total shard execution time went 701 s -> 890 s, and shard 1 went from the fastest (53 s execution) to the slowest (239 s), which sets the critical path at 351 s versus 264 s for every other shard. The restored walkthrough test is roughly 190 s on its own, and `least_duration` cannot split a single test, so it now bounds the whole workflow. Separately, each shard job carries about 117 s of fixed cost (~32 s runner setup + `pip install`, ~85 s of full-suite collection before `--group` deselection), which does not shrink with shard count — so going past 6 shards buys little (8 shards would save ~29 s of critical path, 12 shards ~59 s, while doubling runner usage). Fixing the imbalance and the collection cost is worth more than more shards. None of this needs to happen in this PR.

PR #1476, opened 2026-09-10
feat(baremetal): bump vLLM default to 0.28.0 with glibc gate

BaoYunkai:
## Problem

1. Hyperloom still defaulted bare-metal vLLM to 0.27.1 while 0.28.0 is the current ROCm release.
2. vLLM 0.28.0+ wheels require glibc >= 2.39 (Ubuntu 24.04+), but docs still treated 22.04 and 24.04 equally and the installer did not fail early on incompatible hosts.

## Fix

1. Bump the bare-metal vLLM default to 0.28.0 (`rocm723`) and sync the pinned version across compatibility docs, install docs, example skills, and Slurm model metadata.
2. Enforce the glibc requirement inside `install_vllm_framework()` before any venv/pip work (including `--check-only` / `--dry-run`), with actionable guidance for Docker or a pre-0.28 override.
3. Document Ubuntu 24.04 as the recommended bare-metal OS; mark 22.04 as legacy for vLLM 0.28.0+.
4. Trim `hyperloom-setup` to a short UX explanation of the installer gate; remove the unconditional `VLLM_VERSION=0.27.1` from the normal vLLM command.
5. Revert the SGLang profiler-hotfix gating from this PR; it will land in a separate PR based on observed `torch.version.hip`.

## Test plan

- [x] `pytest src/hyperloom/inference_optimizer/tests/test_setup_cli.py -k "hotfix or glibc or vllm_install"`
- [x] `pytest src/hyperloom/inference_optimizer/tests/test_baremetal_doc_version_consistency.py`
- [x] `ruff format --check .`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
   **Operator note**: sessions lost no recorded knowledge, but until now none of
   it was being shown to the agent.
 
+### Changed
+
+- **Bare-metal `vllm` default bumped from `0.27.1` to `0.28.0` (still `rocm723`).**
+  `install_baremetal.sh`'s `VLLM_VERSION` default, `docs/compatibility.rst`,
+  `docs/install/install.md`, the example `SKILL.md` recipes, and
+  `assets/slurm/models.tsv` now all name `vllm==0.28.0+rocm723` /
+  `vllm/vllm-openai-rocm:v0.28.0`. Verified against the real upstream
+  `v0.28.0` tag that TraceLens' `config_vllm_v0.28.0.patch` applies cleanly
+  (`git apply --check`), so the TraceLens profiler-config patch path is
+  unaffected by the bump. `VLLM_ROCM_VARIANT` is unchanged: `wheels.vllm.ai`
+  only publishes a `rocm723` build for `0.28.0`, same as `0.27.1`. Overridable
+  via `VLLM_VERSION`/`VLLM_ROCM_VARIANT` as before.
+
 ### Removed
 
 - **The `learning/` tuning database, the tracker's scoring layer, and the
--- a/docs/compatibility.rst
+++ b/docs/compatibility.rst
@@ -146,7 +146,7 @@ uses a private registry mirror, set the registry prefix accordingly.
      - MI300X / MI325X
    * - ``lmsysorg/sglang-rocm:v0.5.18-rocm724-mi35x-20260825``
      - MI355X
-   * - ``vllm/vllm-openai-rocm:v0.27.1``
+   * - ``vllm/vllm-openai-rocm:v0.28.0``
      - MI300X / MI325X / MI355X
 
 The vLLM image entrypoint is ``vllm serve``, so override it (for example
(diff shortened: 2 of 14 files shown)


xiaofei-zheng:
## Review

**What this PR does:** bumps the bare-metal `VLLM_VERSION` default from `0.27.1` to `0.28.0` (variant stays `rocm723`) and propagates the literal to every place that names it — `docs/compatibility.rst`, `docs/install/install.md`, the four example `SKILL.md` recipes, `assets/slurm/models.tsv`, `IMAGE_HINT`, and the mirrored comments in `test_baremetal_doc_version_consistency.py`.

### Blocking

**The 0.28.0 wheels raise the glibc floor to 2.39, so bare-metal vLLM install now hard-fails on Ubuntu 22.04** — `install_baremetal.sh:80`.

The index contents differ from 0.27.1 in more than the version:

```
0.27.1: vllm-0.27.1+rocm723-cp312-cp312-manylinux_2_34_x86_64.whl
        torch-2.11.0+gitd0c8b1f-cp312-cp312-manylinux_2_35_x86_64.whl
0.28.0: vllm-0.28.0+rocm723-cp312-cp312-manylinux_2_39_x86_64.whl
        torch-2.12.0+git6bbd260-cp312-cp312-manylinux_2_39_x86_64.whl
```

`manylinux_2_39` requires glibc >= 2.39 (Ubuntu 24.04+). Ubuntu 22.04 ships glibc 2.35, which satisfied both 0.27.1 wheels and satisfies neither 0.28.0 wheel.

Failure path (isolated mode, the default for vLLM): `install_vllm` always creates a fresh venv and installs unconditionally — there is no "already present, skip" branch. It first scrapes `${VLLM_ROCM_INDEX}/torch/` and pins `torch==2.12.0+git6bbd260` (`install_baremetal.sh:900`); pip finds no wheel matching the interpreter's supported platform tags and the script hits `die "failed to install ROCm torch==... into ${VLLM_VENV_ROOT}"` at line 901. Even if torch resolution were skipped, `pip install vllm==0.28.0+rocm723` fails the same way at line 908.

This breaks two paths the repo documents as supported:

1. `docs/compatibility.rst:50` still lists `Ubuntu 22.04` as a supported OS, and `docs/compatibility.rst:187` still tells users the bare-metal route installs `vllm==0.28.0+rocm723`. On a 22.04 ROCm 7.2.3 host provisioned per the doc, the documented command now fails with `Could not find a version that satisfies the requirement`.
2. `IMAGE_HINT` (`install_baremetal.sh:190`) recommends running the installer inside `vllm/vllm-openai-rocm:v0.28.0`, and `docs/compatibility.rst:191` asserts the bare-metal stack "matches" that image. That image's base is Ubuntu 22.04 (`org.opencontainers.image.version=22.04` on the config blob for the `v0.28.0` manifest), so the installer cannot install its own matching wheel inside the very image it points at.

Note the docker path is unaffected — the image ships vLLM prebuilt. Only the bare-metal/isolated-venv path regresses.

Suggested resolution (either is fine, but the PR shouldn't land as-is):
- gate the vLLM bare-metal path on glibc >= 2.39 with an explicit, actionable error (and cover it in `--check-only` / `--dry-run`, which today print a clean preflight and exit 0 on a host where the real run will die), and update the support matrix + the "matches its container image" sentence to state that bare-metal vLLM 0.28.0 requires Ubuntu 24.04; or
- keep the 0.27.1 default until the 22.04 requirement is dropped repo-wide.

### Verification performed

- Fetched both `wheels.vllm.ai/rocm/{0.27.1,0.28.0}/rocm723/{vllm,torch}/` index listings directly; tags quoted above are the real filenames.
- Pulled the `vllm/vllm-openai-rocm:v0.28.0` manifest + config blob from Docker Hub to confirm the 22.04 base.
- Read `install_vllm` end to end on the PR head to confirm the isolated path installs unconditionally and that both pip failures are fatal (`die`), not warnings.
- Confirmed `test_baremetal_doc_version_consistency.py` only asserts the installer default against `docs/compatibility.rst`; the version/image literals in `install.md`, the four `SKILL.md` files, `models.tsv`, and `IMAGE_HINT` are unguarded — not blocking for this PR, but it is why the bump was an 8-file hand sweep.

No other blocking issues: the doc/example string replacements are complete and consistent with the new default, and the version-consistency test still passes against them.

ZhengGong-amd:
## Request changes

Please keep this PR to the vLLM default upgrade and make the installation boundary authoritative. Do not add further prompt rules or version-inference fallbacks.

1. **Fix CI first.** `ruff format --check .` currently fails in `test_setup_cli.py`.

2. **Enforce the vLLM 0.28 ABI requirement in `install_vllm_framework()`, before any venv or pip mutation.** The `0.28.0+rocm723` vLLM and torch wheels are `manylinux_2_39`; Ubuntu 22.04's glibc 2.35 cannot install them. The new `hyperloom-setup` instruction is advisory and is bypassed by direct installer/backend/automation calls; `--check-only` and `--dry-run` currently look healthy too. Check the actual glibc capability (not merely the Ubuntu label), fail with actionable Docker / pre-0.28 override guidance, and cover: glibc 2.35 rejection, 2.39 acceptance, and a permitted pre-0.28 override. Keep the prompt as a short UX explanation of that installer result, not as a second policy implementation.

3. **Remove the unconditional `export VLLM_VERSION=0.27.1` from the normal vLLM command in `hyperloom-setup/SKILL.md`.** A user on Ubuntu 24.04 who follows that command never receives the new 0.28.0 default. Put the downgrade in a separate, explicitly conditional snippet.

4. **Split out the SGLang profiler-hotfix change.** It is independent of the vLLM upgrade, and the new helper is not a valid runtime detector: `SGLANG_ROCM_EXTRA` defaults to `rocm724` and wins before the real HIP version is read, so an actual rocm720 SGLang container with that env unset skips its required overlay. `.env` does not restore this variable before the gate and the default is later written back. The current tests force the env value, so they miss the real path. In a separate PR, use the SGLang interpreter's observed `torch.version.hip` as the single source of truth; do not infer runtime state from an install-index setting or `/opt/rocm`. Verify real rocm720/rocm724 images and test unset/stale config plus the exact supported HIP boundaries.

This produces one small, deterministic vLLM-default change now; the host-library mutation can then be reviewed independently against real stack evidence.

PR #1477, opened 2026-09-10
fix(forge): recover the nested forge-loop token ledger, or say it is partial

jiaqiang-dot-liu:
## Summary

Follow-up to #1474, which reported the FlyDSL rewrite's LLM usage but lost the nested forge-loop's share in exactly the cases where it matters most.

* recover the nested forge-loop's token ledger from the experiment record it checkpoints at every session boundary, so a run that is cut off at the optimize deadline — or that never keeps anything, and therefore never writes a result file — no longer reports zero for the phase that spends the most
* pick the furthest-along snapshot of that single ledger instead of summing the result file and the record, which are two views of the same accumulator
* mark the combined `llm_usage` partial (`cost_available: false`, `cost_source: "partial"`) whenever the loop never closed its ledger, so the payload stops claiming a complete provider-priced total while a session's spend is missing, and print a warning naming the shortfall

`llm_usage_complete` on the optimize result is the new internal signal: it is set only when forge-loop exited cleanly with a result of its own, the one case where its ledger is final.

## Test plan

* `python -m pytest -q src/kernelforge/tests/test_usage.py src/kernelforge/tests/test_rewrite_by_flydsl_pipeline.py src/kernelforge/tests/test_rewrite_by_flydsl.py src/kernelforge/tests/test_tracker.py`
* `python -m pytest -q src/kernelforge/tests` (full suite green; the `test_codex_backend.py` failures on my host are a sandbox permission error on `~/.cache/kernelforge`, and pass unsandboxed)
* `python -m ruff check src/kernelforge`
* full CI suite

### DIFF
--- a/src/kernelforge/rewrite_by_flydsl/optimize.py
+++ b/src/kernelforge/rewrite_by_flydsl/optimize.py
@@ -5,6 +5,7 @@
 
 from __future__ import annotations
 
+import contextlib
 import json
 import logging
 import os
@@ -21,6 +22,7 @@ from typing import Callable
 from kernelforge.llm.git import git
 from kernelforge.config import Config
 from kernelforge.rewrite_by_flydsl.spec import RewriteSpec
+from kernelforge.tracker import ExperimentTracker
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


xiaofei-zheng (approved):
No blocking issues — LGTM.

**What this PR does**

When forge-rewrite runs the nested forge-loop as a child process, the loop's token ledger only reached the caller through `--result-json` (rewritten on a KEEP) or the stdout `__FORGE_RESULT__` sentinel (emitted only on a clean exit). A loop that kept nothing, or was cut off at the optimize deadline, therefore had its entire spend silently dropped from the run's `llm_usage`. This PR recovers that ledger from the experiment record forge-loop checkpoints at every session boundary, and when the recovered ledger is known to be truncated, publishes the run total as partial instead of as a complete provider-priced answer.

- `optimize.py`: `_experiment_llm_usage()` reads `ExperimentTracker(...).get(id).llm_usage`; `_latest_ledger()` picks the furthest-along snapshot by (calls, tokens) instead of summing; `llm_usage`/`llm_usage_complete` are attached to the returned dict, with the recovery also wired into the launch/stream-failure path.
- `usage.py`: `combine_usage_totals(..., incomplete=True)` forces `cost_available=False`.
- `runner.py`: `_total_usage()` takes `optimize_ran` and marks the combination incomplete when OPTIMIZE ran without reporting a closed ledger, plus a WARNING on that path.

**What I checked**

- Double-counting: `result_json["llm_usage"]` (cli.py:1804) and the tracker record (cli.py:2005, loop/runner.py:607 `_checkpoint_llm_usage`) are both snapshots of the *same* `self._usage` accumulator, so `_latest_ledger`'s max-not-sum is the correct combinator, and its (calls, tokens) key is monotonic for that accumulator.
- All four `_total_usage()` call sites: runner.py:192, :374, :456 all execute strictly before OPTIMIZE, so leaving `optimize_ran` at its `False` default there is right; only :604 can have nested spend and it passes the flag.
- `combine_usage_totals` behavior, executed directly: `incomplete=True` with a priced record -> `cost_available False` / `cost_source "partial"`; with an unpriced record -> `"unavailable"`; without the flag the pre-PR result is unchanged. The new keyword is keyword-only, so no existing positional caller changes meaning.
- The relaxed early return (`not result and not terminated_for_deadline and not nested_usage`) can now return `best_kernel_restored`/`terminated_for_deadline` where it previously returned `{}`; `run_optimize` has exactly one caller (runner.py:518) and it only reads `best_ms`/`llm_usage*`, so nothing downstream regresses.
- `usage_complete = bool(result) and not terminated_for_deadline and _poll_process(proc) == 0` — `_poll_process` runs after `_wait_process`, so the exit code is settled by then.
- Docstrings and the `run_optimize` contract were updated alongside the behavior change; no stale comment left behind.

Tests could not be run locally (this box is Windows; `kernelforge.tracker.experiment` imports `fcntl`), so I verified `combine_usage_totals` by loading `tracker/usage.py` standalone and exercising the three cases above, and left the rest to CI. The added coverage looks like it hits the paths that matter: ledger recovery with no result file, latest-snapshot preference, clean-exit completeness, truncated-loop partial reporting, and the OPTIMIZE-never-started case.

PR #1479, opened 2026-09-10
feat(forge): publish the run's token spend at one fixed path

jiaqiang-dot-liu:
## Summary

Reading what a run spent still required knowing something the reader may not have: a `--result-json` path the caller is free not to pass, or an experiment record named after an id you have to discover from stdout first. This adds one fixed artifact, `<experiments-dir>/llm_usage.json`, written by both `forge-loop` and `forge-rewrite-by-flydsl`.

* every producer folds **its own share** into the file after each counted call, under an `flock`-protected read-modify-write, so the FlyDSL rewrite and the forge-loop it nests — two processes sharing one experiments directory — add up to a single running total instead of overwriting each other
* the payload is flat: the eight canonical usage counters plus `updated_at`, no per-producer breakdown to interpret and nothing to sum by hand

```console
$ jq . <experiments-dir>/llm_usage.json
{
  "input_tokens": 1021230,
  "output_tokens": 77120,
  "cache_creation_input_tokens": 12900,
  "cache_read_input_tokens": 658000,
  "total_cost_usd": 12.29,
  "calls": 47,
  "cost_available": true,
  "cost_source": "provider",
  "updated_at": "2026-09-10T19:45:02"
}
```

* `UsageAccumulator` grows an optional `on_update` sink, which is what makes the file current after every call rather than at the run's next coarse checkpoint — the spend of a session that is killed mid-flight is now recorded too
* cost provenance folds the same way it does everywhere else: one contributor without complete provider pricing makes the shared answer `partial`, and it never recovers
* writes are best-effort — a failing ledger logs and is dropped, it never ends a run

Also resolves `experiments_dir` once in `run_rewrite`: the nested forge-loop runs from the workspace and resolves that path against its own working directory, so a relative one sent the two processes to different directories for every artifact they share, this file included.

The ledger is scoped to the experiments directory, not to a process: a resumed campaign keeps accumulating across segments, which is why one run should own one directory.

## Test plan

* `python -m pytest -q src/kernelforge/tests/test_usage_ledger.py` — covers the fixed name, the file existing before the first call, two producers not overwriting each other, provenance degrading and staying degraded, a corrupt file, an unwritable directory, the accumulator sink firing per call and surviving a failing sink, and a spawn-based two-process run that loses no update
* full CI suite

### DIFF
--- a/src/kernelforge/cli.py
+++ b/src/kernelforge/cli.py
@@ -1182,7 +1182,7 @@ def forge_loop(
         aggregate_regression_detail,
         warm_start_improvement_flags,
     )
-    from kernelforge.tracker import ExperimentTracker, UsageAccumulator
+    from kernelforge.tracker import ExperimentTracker, UsageAccumulator, UsageLedgerFile
     from kernelforge.orchestrator.agent import make_agent_fn
 
     iter_config = IterationConfig(
@@ -1215,7 +1215,11 @@ def forge_loop(
         commit_new_paths=commit_new_paths,
     )
     tracker = ExperimentTracker(config.experiments_dir)
-    usage = UsageAccumulator()
+    # Published at a fixed path from the first call onwards, so a caller that never asked for --result-json, or that
+    # loses this process outright, can still read what the run spent.
+    usage_ledger = UsageLedgerFile(config.experiments_dir)
+    usage = UsageAccumulator(on_update=usage_ledger.publish)
+    usage_ledger.publish(usage.totals())
 
     # The caller-owned experiment ID is an EXTERNAL recovery channel, deliberately independent of the internal
     # per-segment experiment identity (each resume segment gets a fresh ID so the campaign parent/child chain stays
--- a/src/kernelforge/rewrite_by_flydsl/runner.py
+++ b/src/kernelforge/rewrite_by_flydsl/runner.py
@@ -38,7 +38,7 @@ from kernelforge.rewrite_by_flydsl.optimize import run_optimize
 from kernelforge.rewrite_by_flydsl.port_loop import PortResult, run_port_loop
 from kernelforge.rewrite_by_flydsl.budget import DEFAULT_REWRITE_BUDGET
 from kernelforge.loop.scoring import DEFAULT_SNR_THRESHOLD_DB
-from kernelforge.tracker import UsageAccumulator, combine_usage_totals
+from kernelforge.tracker import UsageAccumulator, UsageLedgerFile, combine_usage_totals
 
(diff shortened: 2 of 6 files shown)


xiaofei-zheng (approved):
No blocking issues. LGTM.

What I checked:

**Merge semantics (`usage_ledger.py`)** — I ran the delta/fold logic standalone (with `fcntl` stubbed, since I reviewed from Windows) against the scenarios the PR claims:
- zero-call bootstrap publish writes `calls: 0 / cost_source: unavailable` and, once the file exists, a no-op republish short-circuits before taking the lock (`_delta` returns `None` → early return);
- two producers interleaving running totals fold to 425 tokens / 7 calls / $4.30 with `cost_source: provider` — neither share is overwritten;
- one unpriced contributor degrades the shared record to `cost_available: false / cost_source: partial`, and a subsequent fully-priced publish from the other producer does not lift it back — the degrade is sticky, as documented;
- a corrupt ledger falls back to this producer's own share rather than raising;
- `UsageAccumulator(on_update=...)` fires per counted call (2 calls → `calls: 2`, tokens and cost summed), and a sink that raises is swallowed while `add_usage` still returns `True` and keeps its own totals intact.

**No double counting** — `run_rewrite` feeds only its in-process accumulator to the ledger; the nested forge-loop's spend reaches the returned `llm_usage` through `_total_usage`/`combine_usage_totals` from `opt_result`, and reaches the file through the child's own `UsageLedgerFile`. The two paths do not overlap, so the file is not inflated.

**Concurrency** — `_current()` read and `atomic_write_text` both happen inside the `flock`, and the write is a rename, so an out-of-lock reader can only see a whole record. `_published` advances only after a successful write, so a failed publish re-sends the full delta on the next call instead of losing it.

**`experiments_dir` resolution** — confirmed the real bug this fixes: `forge_loop` resolves `--experiments-dir` against its own cwd (cli.py:1150) and `optimize.py` launches it with `cwd=spec.workspace`, so a relative path really did split the two processes' artifacts. Resolving at the top of `run_rewrite` before any use is the right place.

**Failure handling** — the ledger is best-effort by design and says so; the exception path logs at debug rather than swallowing silently, and `run_rewrite` still warns on an incomplete nested ledger. Comments and docstrings match the new behavior.

Notes, non-blocking: `fcntl` is imported unconditionally by `kernelforge.tracker`, which is consistent with existing POSIX-only code in the repo; and a re-run reusing an existing experiments directory keeps accumulating on top of the previous total, which the PR description calls out as intended.

I did not run the full suite locally — CI covers it.

PR #1487, opened 2026-09-11
Feat/vllm v29 patch

devalshahamd:
Adding patches for vLLM v29

- Description: what and why

### DIFF
--- a/src/hyperloom/agents/kernel/scripts/install.sh
+++ b/src/hyperloom/agents/kernel/scripts/install.sh
@@ -156,7 +156,7 @@ INFERENCEX_PATH="${INFERENCEX_PATH:-}"
 # The internal extension is used ONLY when $TRACELENS_INTERNAL_ROOT is set
 # (env / .env); leave it unset for the base-only report. No separate toggle.
 TRACELENS_REPO="https://github.com/AMD-AGI/TraceLens.git"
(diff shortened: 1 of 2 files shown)


devalshahamd:
@BaoYunkai  Can you please check this PR. This is to add v29 patches for vLLM!

xiaofei-zheng (approved):
No blocking issues. LGTM.

What this does: bumps the pinned TraceLens ref from 210fb5c8 to c74d4d2c so the kernel agent picks up the graph-capture patch for vLLM v0.29.

Checked:
- Both pin sites are updated together — `scripts/install.sh` (TRACELENS_REF) and `tools/tracelens_analysis.py` (_TRACELENS_REF_DEFAULT) — which is what the "kept in sync" comment above the Python default requires. Code search over the repo finds no third occurrence of the old SHA.
- c74d4d2c exists in AMD-AGI/TraceLens, is a descendant of the previous pin (5 commits ahead, 0 behind) and is an ancestor of TraceLens main, so the pin is fetchable and not a rewind.
- The upstream commit (TraceLens #1020) only adds the v0.29 patch plus agent/eval assets; nothing in it removes or renames paths the installer or the analysis tool depends on.
- No local test run needed for a pin bump; CI on this PR is green across all required checks.

PR #1488, opened 2026-09-11
fix(profile): recognize SGLang graph capture directories

meinali-566:
## Summary
- keep nearby capture-directory selection deterministic as `capture_traces/` → `graph_capture_profile/` → `graph_capture/`
- accept SGLang's upstream `graph_capture_profile/` directory in profile health checks instead of reporting a false missing-capture warning
- preserve the intentional capture-only preflight: capture sidecars remain auxiliary input passed with `--capture-folder` alongside a real Replay `--trace-input`
- add regression coverage for directory priority and SGLang profile health

## SGLang definition

[SGLang PR #24370](https://github.com/sgl-project/sglang/pull/24370) defines `GRAPH_CAPTURE_PROFILE_DIRNAME = \"graph_capture_profile\"` under `${SGLANG_TORCH_PROFILER_DIR}` and uses that directory for both supported capture exports:

- `SGLANG_ENABLE_CUDA_GRAPH_CAPTURE_TRACE=1` writes one combined trace per tensor-parallel rank as `cuda_graph_capture-{runner}-TP-{tp_rank}.json.gz`.
- `SGLANG_GRAPH_BATCH_CAPTURE=1` writes one trace per captured batch size and rank as `{runner}_bs_{bs}_rank{tp_rank}.json.gz`.
- If both environment variables are set, the combined `SGLANG_ENABLE_CUDA_GRAPH_CAPTURE_TRACE` mode takes precedence.

Hyperloom's shared sidecar classifier and nearby-folder discovery already understand `graph_capture_profile`, but profile health still checked only `capture_traces/`. The explicit priority keeps legacy vLLM/AMD layouts preferred when multiple recognized directories coexist, while valid upstream SGLang output is no longer reported as missing.

A `graph_capture_profile/` directory by itself is deliberately not a valid `--trace-input`: it contains auxiliary Capture sidecars, not the Replay trace required for analysis. It must be passed as `--capture-folder` beside a real Replay input.

## Test plan
- `python -m pytest src/hyperloom/agents/kernel/tests/test_tracelens_csv.py -k 'discover_capture_folder'` (3 passed)
- `python -m pytest src/hyperloom/inference_optimizer/tests/test_profile_and_kernel_handlers.py -k 'input_dims_fraction or upstream_sglang_capture_directory or zero_cpu_op'` (4 passed)
- `ruff check` and `ruff format --check` on all four modified files

### DIFF
--- a/src/hyperloom/agents/kernel/tests/test_tracelens_csv.py
+++ b/src/hyperloom/agents/kernel/tests/test_tracelens_csv.py
@@ -2199,14 +2199,6 @@ def test_discover_capture_folder_finds_the_unpatched_sglang_layout(tmp_path):
     assert tlr.discover_capture_folder(trace_dir, [real]) == trace_dir / "graph_capture_profile"
 
 
-def test_discover_capture_folder_accepts_capture_directory_as_input(tmp_path):
-    capture_dir = tmp_path / "graph_capture_profile"
-    capture_dir.mkdir()
-    capture = _rank_trace(capture_dir / "cuda_graph_capture-ModelRunner-TP-0.json.gz", kernels=4)
-
-    assert tlr.discover_capture_folder(capture_dir, [capture]) == capture_dir
-
-
 def test_discover_capture_folder_preserves_legacy_priority(tmp_path):
     trace_dir = tmp_path / "torch_trace"
     for name in ("graph_capture", "graph_capture_profile", "capture_traces"):
--- a/src/hyperloom/agents/kernel/tests/test_tracelens_csv.py
+++ b/src/hyperloom/agents/kernel/tests/test_tracelens_csv.py
@@ -2199,6 +2199,23 @@ def test_discover_capture_folder_finds_the_unpatched_sglang_layout(tmp_path):
     assert tlr.discover_capture_folder(trace_dir, [real]) == trace_dir / "graph_capture_profile"
 
 
+def test_discover_capture_folder_accepts_capture_directory_as_input(tmp_path):
+    capture_dir = tmp_path / "graph_capture_profile"
+    capture_dir.mkdir()
+    capture = _rank_trace(capture_dir / "cuda_graph_capture-ModelRunner-TP-0.json.gz", kernels=4)
+
+    assert tlr.discover_capture_folder(capture_dir, [capture]) == capture_dir
+
+
+def test_discover_capture_folder_preserves_legacy_priority(tmp_path):
+    trace_dir = tmp_path / "torch_trace"
+    for name in ("graph_capture", "graph_capture_profile", "capture_traces"):
+        (trace_dir / name).mkdir(parents=True)
+    real = _rank_trace(trace_dir / "rank_0.trace.json.gz", kernels=4)
+
+    assert tlr.discover_capture_folder(trace_dir, [real]) == trace_dir / "capture_traces"
+
+
 def test_discover_capture_folder_ignores_a_descriptive_sibling(tmp_path):
     trace_dir = tmp_path / "torch_trace"
     (trace_dir / "torch_profiler_with_graph_capture").mkdir(parents=True)
--- a/src/hyperloom/agents/kernel/tools/tracelens_skill_runner.py
+++ b/src/hyperloom/agents/kernel/tools/tracelens_skill_runner.py
@@ -301,6 +301,7 @@ def discover_capture_folder(trace_input: Path, trace_files: list[Path]) -> Path 
         Path | None: The capture folder if one exists nearby, else ``None``.
     """
 
+    capture_dir_priority = ("capture_traces", "graph_capture_profile", "graph_capture")
     search_roots: list[Path] = []
     if trace_input.is_dir():
         search_roots.append(trace_input)
@@ -311,12 +312,18 @@ def discover_capture_folder(trace_input: Path, trace_files: list[Path]) -> Path 
         if root in seen or not root.is_dir():
             continue
         seen.add(root)
+        if is_capture_dir_name(root.name):
+            return root
         try:
-            children = sorted(root.iterdir())
+            children = [child for child in root.iterdir() if child.is_dir()]
         except OSError:
             continue
-        for child in children:
-            if child.is_dir() and is_capture_dir_name(child.name):
+        children_by_name = {child.name.lower(): child for child in children}
+        for name in capture_dir_priority:
+            if child := children_by_name.get(name):
+                return child
+        for child in sorted(children):
+            if is_capture_dir_name(child.name):
                 return child
     return None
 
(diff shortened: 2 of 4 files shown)


xiaofei-zheng:
## Review

### What this PR does
- `discover_capture_folder()` now returns the search root itself when the root is a capture directory, and picks among sibling capture directories with an explicit priority `capture_traces/` → `graph_capture_profile/` → `graph_capture/` (falling back to the shape match for anything else).
- `_validate_trace_structure()` check [1] now accepts any of the three capture directory names instead of only `capture_traces/`, records `capture_dir` in the structured check row, and reports the actual directory name in the "exists but is empty" issue.
- Adds unit coverage for direct-directory discovery, legacy priority, and the SGLang health check.

### Blocking

**1. `graph_capture_profile/` passed directly as `--trace-input` still fails before the new code can run — `src/hyperloom/agents/kernel/tools/tracelens_skill_runner.py:315`, `src/hyperloom/agents/kernel/tools/tracelens_analysis.py:7161-7176`**

The first bullet of the summary ("recognize `graph_capture_profile/` when it is passed directly as the trace input") is not reachable on a real run. In `run_trace_analysis`, the capture-only preflight runs at line 7161, long before `discover_capture_folder()` is called at line 7728:

```python
capture_root = _capture_classification_root(trace_input)   # == trace_input for a directory
if all(_is_capture_fragment(p, capture_root) for p in trace_files):
    raise RuntimeError("trace_input_capture_only: ...")
```

For a real upstream SGLang capture directory every file is classified as a sidecar by filename alone (`CAPTURE_FRAGMENT_RE` matches `graph_capture` anywhere, and `bs_<n>` anchored), so the `all(...)` is true and the run aborts. Verified against this branch:

```python
from _capture_shapes import is_capture_fragment
d = tmp / "graph_capture_profile"          # cuda_graph_capture-ModelRunner-TP-0.json.gz, bs_8_rank0.json.gz
[(p.name, is_capture_fragment(p, d)) for p in d.iterdir()]
# [('bs_8_rank0.json.gz', True), ('cuda_graph_capture-ModelRunner-TP-0.json.gz', True)]
```

So `raise RuntimeError("trace_input_capture_only: ...")` fires and `discover_capture_folder()` is never reached. The new `if is_capture_dir_name(root.name): return root` branch only fires when the capture directory also holds a file whose name is not sidecar-shaped — which does not happen in the SGLang layout this PR targets. The added test exercises the helper in isolation, so it passes while the end-to-end path it stands for does not work.

Note the preflight is deliberate (its own error message says "Sidecars are a supported auxiliary input — pass them as `--capture-folder` alongside a real trace, never as `--trace-input`"). Either that decision should be revisited explicitly for the SGLang layout, or the direct-directory branch and its test should be dropped from this PR and the summary corrected — the health-check half of the change (`profile.py`) stands on its own and is correct.

### Checked, no issues found
- `capture` is no longer used after check [1] in `_validate_trace_structure()`, so `capture = next(..., None)` cannot leak a `None` into a later attribute access; check [2] is gated on `capture_files` and is unaffected.
- The reworded issue strings (`graph-capture subdirectory missing`, `f"[1] {capture.name}/ exists but is empty"`) have no other consumers in the repo (grep for the old literals returns nothing outside the modified files).
- `children_by_name` lower-cases keys and the priority tuple is lower-case, so the priority lookup keeps the case-insensitivity of `CAPTURE_DIR_RE`; `Path` truthiness makes the walrus guard safe. `child.is_dir()` inside the comprehension stays under the `try/except OSError`.
- Adding `capture_dir` to the check row is additive; `capture_traces_present` consumers (`roofline.py:964`, `roofline_event.py:340`) read only the boolean.
- Priority ordering does not regress the existing behaviour when several capture directories coexist (previously alphabetical first match).

### Tests run
- `python -m pytest src/hyperloom/agents/kernel/tests/test_tracelens_csv.py -k discover_capture_folder` → 4 passed.
- `src/hyperloom/inference_optimizer/tests/test_profile_and_kernel_handlers.py` could not be collected locally (`ModuleNotFoundError: fcntl`, Windows); left to CI.

xiaofei-zheng (approved):
## Re-review (79285f4)

The blocking issue from the previous round is resolved: the direct-capture-directory branch in `discover_capture_folder()` and its unit test are gone, so the deliberate `trace_input_capture_only` preflight is no longer contradicted, and the PR summary now states explicitly that `graph_capture_profile/` is auxiliary `--capture-folder` input rather than a valid `--trace-input`.

**No blocking issues. LGTM.**

### What the remaining change does
- `discover_capture_folder()`: deterministic sibling priority `capture_traces/` → `graph_capture_profile/` → `graph_capture/`, falling back to the existing `CAPTURE_DIR_RE` shape match in alphabetical order. Only behavioural delta vs. main is when `graph_capture/` and `graph_capture_profile/` coexist (alphabetical order previously picked `graph_capture/`); `capture_traces/` already won alphabetically, so the legacy layout is unchanged.
- `_validate_trace_structure()` check [1]: accepts any of the three directory names, records `capture_dir` in the structured row, and names the actual directory in the "exists but is empty" issue.

### Verified
Exercised `_validate_trace_structure(tmp, "sglang")` directly on this commit:

| layout | status | `capture_dir` | `[1]` issue |
|---|---|---|---|
| `graph_capture_profile/` with one `cuda_graph_capture-ModelRunner-TP-0.json.gz` | passed | `graph_capture_profile` | none |
| `graph_capture/` only | passed | `graph_capture` | none |
| `capture_traces/` + `graph_capture_profile/` | passed | `capture_traces` (priority holds) | none |
| no capture directory | failed | — | `graph-capture subdirectory missing` |

`capture_traces_present` is `True` for the upstream SGLang layout, which is what previously blocked the capture from reaching TraceLens.

Also checked:
- `capture` is not referenced after check [1], so `next(..., None)` cannot leak `None` into a later attribute access; check [2] is gated on `capture_files`.
- The reworded issue strings have no other consumers in the repo (grep for the old literals returns nothing outside these files).
- `children_by_name` lower-cases keys against a lower-case priority tuple, preserving `CAPTURE_DIR_RE`'s case-insensitivity; `Path` truthiness makes the walrus guard safe; `child.is_dir()` stays inside the `try/except OSError`.
- Adding `capture_dir` to the check row is additive — `capture_traces_present` consumers (`roofline.py:964`, `roofline_event.py:340`) read only the boolean.
- Pre-existing, not introduced here: `CAPTURE_FRAGMENT_RE` (`\Abs_\d+|graph_capture`) does not match the `{runner}_bs_{bs}_rank{tp}.json.gz` shape described in the summary by filename alone. In the SGLang layout those files sit under `graph_capture_profile/`, so `is_capture_fragment()` still classifies them correctly via the path component. Worth a separate look only if a future layout writes them beside the workload trace.

### Tests run
- `python -m pytest src/hyperloom/agents/kernel/tests/test_tracelens_csv.py -k discover_capture_folder` → 3 passed.
- `src/hyperloom/inference_optimizer/tests/test_profile_and_kernel_handlers.py` cannot be collected on Windows (`ModuleNotFoundError: fcntl`); covered the same paths with the direct invocation above and left the suite to CI.

PR #1493, opened 2026-09-11
ci: diagnose hanging tests on PR #1476 (do not merge)

zengleixin-amd:
## Purpose
Diagnostic reproduction branch forked directly from PR #1476 head 7f047044be37f7dd543c5eb795ac56dd7703f3b9. This PR intentionally includes the original PR changes so the full merge test exercises that code, not just main. Do not merge this diagnostic PR independently.

## Diagnostic change
Only pyproject.toml differs from the source PR head: add -vv and -o faulthandler_timeout=180 to CI pytest arguments. Dumps do not terminate tests. No test selection, 300-pass walkthrough depth, coverage gate, shard or worker count changed.

## Validation
- TOML structural comparison confirms only the three diagnostic argv additions.
- The identical flags were verified with pytest 9.1.1 and xdist 3.8.0: call and teardown stack dumps arrive before completion and both fixture tests pass.
- git diff --check passed.
- Linux full-suite diagnostics pending; this is not a claimed root-cause fix.

## Boundaries
Original PR #1476 branch is untouched. Native faulthandler dumps Python threads, not all suspended async tasks or nested child processes, and does not cover collection/session-exit hangs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
   **Operator note**: sessions lost no recorded knowledge, but until now none of
   it was being shown to the agent.
 
+### Changed
+
+- **Bare-metal `vllm` default bumped from `0.27.1` to `0.28.0` (still `rocm723`).**
+  `install_baremetal.sh`'s `VLLM_VERSION` default, `docs/compatibility.rst`,
+  `docs/install/install.md`, the example `SKILL.md` recipes, and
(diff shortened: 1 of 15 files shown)


zengleixin-amd:
The verified root-cause fix has been cherry-picked into PR #1476 as 397532784. Only the two-file import-hook recursion/test-isolation fix was transferred; diagnostic pytest flags were not copied. Closing this diagnostic PR as requested. Validation of the fix on this branch: https://github.com/AMD-AGI/Hyperloom/actions/runs/34604481222 (all test shards and coverage passed). The original PR will run its own CI on the transferred commit.

PR #1497, opened 2026-09-14
Rewrite layer: one speedup metric, one baseline, and KB ports admitted on evidence

jiaqiang-dot-liu:

  Two things the rewrite layer got wrong, plus one CI fix it needed to go green.

  **1. Every published speedup now means the same thing.**

  The arena grades a candidate on the equal-weight arithmetic mean of per-case speedups. The forge loop's keep/revert, its in-session gate and the loop-layer warm start all agree. The rewrite layer did not, in two independent ways.

  *Wrong aggregation.* It ordered its warm-start field by aggregate wall time and published that same ratio of aggregates as the speedup a KB record claims. A ratio of aggregates answers a different question and disagrees by up to 2x on a shape sweep spanning two orders of magnitude: over the 21-shape SIKL campaign it reads 1.06x where the arena reads 2.06x. The per-case timings needed to score were already in the driver output and already parsed — `read_driver_output` collected the case ids and dropped the values — so this keeps them, carries them through `PreflightReport`, and scores with `calculate_mean_case_speedup`, the same primitive the loop uses. Nothing new is measured.

  *Wrong denominator.* Fixing the aggregation exposed a second problem on the OPTIMIZE path. `run_optimize` launches forge-loop with `--kernel` pointing at the ported FlyDSL file, and the loop benches whatever the workspace holds to establish its anchor — `--ref-bench-mode`, the only mode that times the source, belongs to the rewrite layer's driver contract and the loop never calls it. So `mean_case_speedup` scored *best vs port* while the two wall times published beside it, and the interim result written before OPTIMIZE, scored *best vs source*. One field, two denominators. It is not recoverable after the fact either: an equal-weight mean of per-case ratios does not compose, so the port's own speedup cannot be multiplied back onto the loop's.

  forge-loop therefore takes a new **optional** `--baseline-json` naming per-case times and a wall time measured outside it. Given one, every speedup it reports divides by those times; omitted, it measures its own first bench exactly as before, so no existing caller changes behavior. The loop still benches the kernel it starts from either way — with a caller anchor that bench is the *search start and first incumbent* rather than the anchor itself. That distinction is load-bearing: the incumbent's score was hardcoded to `1.0`, which is only true when the starting kernel IS the anchor. Against a supplied anchor the port is already ahead of it, and a 1.0x bar would KEEP a candidate that loses to the kernel the run began with. The bar now starts at what the first bench actually scored, which is still exactly 1.0 when the loop anchors on itself.

  **2. Warm start stops refusing ports it would have proven.**

  Admission ran five gates in front of the trial that actually decides. Three of them rejected reusable ports:

  - `source_sha256` / `driver_sha256` are raw file digests, so a moved comment retired every record under an identity at once and the run paid for a full PORT — an LLM loop costing hours — to rebuild a port the KB was already holding.
  - The implementation signature hashes parsed kernel symbol names and paths. It survives reformatting but not a second kernel added to the same file, nor a renamed source. The loop-layer warm start already treats this signature as a ranking signal rather than a veto, so the rewrite layer was the outlier.
  - The schema version and rewrite kind are constants today: one producer writes these records and writes the same pair every time, so the check can only fire on a format that does not exist yet.

  None of them predicted anything the run does not establish itself: the current driver validates the candidate against the current source, times it, and scores it against this task's own per-case times. What remains is the builder symbol — the one admission question a trial cannot answer cheaply, since the driver imports a named factory out of the artifact and a different one cannot be imported however good the port is. It fires for real (`softmax` and `softmax_kernel` normalize to one identity but slug to different builders) and has no false positives.

  A warm-start candidate whose benchmark failed also no longer wins by default when nothing else was timed: it is dropped rather than ranked last. Correctness establishes that a kernel computes the right answer, not that the timing behind its claim can be reproduced.

  **3. One unrelated CI fix.**

  `test_agentx_state_to_external_reference_and_final_report` fails on `py3.10 shard 4` on `main` itself, not just here. The executor reads `PRECISION` ahead of the session's own precision deliberately — `test_agentx_explicit_precision_override_wins_over_stale_state` pins that as an operator override — so a test reading precision off `SharedState` has to clear the variable first, which the tests around it already do. This one cleared three others and stopped short. The variable arrives from inside the run: the optimize CLI assigns `os.environ["PRECISION"]` directly, which `monkeypatch` cannot undo. Test-only; no production behavior touched.

  None.

  Added and updated throughout.

  - Scoring: the warm-start selection and the drop of unscorable candidates, `case_ms` surviving into the preflight report, `build_result` taking the score rather than computing it.
  - Denominator: the loop's two anchor modes (its own bench vs a supplied one), the KEEP bar starting at the search start rather than 1.0, `--baseline-json` input validation (8 parametrized rejections), `--baseline-json` forwarded when source timings are given and omitted when they are not. The interim-result test spreads two cases across two orders of magnitude, where the mean of ratios reads 2.556 and the ratio of aggregates reads 1.119, so it pins the metric rather than merely a non-null value.
  - Admission: the two gates removed had no test in either direction. Three cases are now pinned — two admissions (a moved comment, a kernel added to the source) and one refusal (a builder the driver cannot import). Both admissions were confirmed to fail against the old gates before being kept.
  - The CI fix was verified by exporting `PRECISION=fp8` around the file: two failures before, the same 32 tests green after.

  ```
  python -m pytest src/kernelforge/tests/ -q          # 4048 passed
  python -m ruff check src/kernelforge/               # clean
  python -m ruff format --check src/kernelforge/      # clean
  ```

  Yes, in the value of a published field rather than in its shape. The `speedup` a KB record and a rewrite result JSON carry changes meaning from the ratio of aggregate wall times to the mean of per-case ratios, measured against the source. Records written before this PR keep the old number, so a consumer comparing across the boundary is comparing two metrics — including `store.champion_speedup`, which promotes on `speedup > champion` and will therefore compare a new score against an old-metric champion. Both wall times still travel alongside as raw evidence. Worth deciding separately whether existing champion pointers should be rebuilt.

  `--baseline-json` is additive and optional; forge-loop without it behaves exactly as before.

  Two further issues found while reviewing and deliberately **not** addressed here, both pre-existing in kind:

  - A warm-start candidate whose bench was skipped because the deadline expired is dropped by the same path as one whose bench failed, and the run then falls through to a PORT it has no budget left for.
  - The driver contract does not require `case_ms:` in `--ref-bench-mode` (`preflight_reference` only checks the aggregate). A driver that reports only an aggregate there leaves `source_case_ms` empty, so no anchor is passed and the OPTIMIZE score silently reverts to a port denominator.

  No. The scoring metric and the warm-start admission gates are two concerns in the same subsystem, and the `PRECISION` test fix is a third, unrelated one carried only because `main` is red without it and this PR cannot go green on its own. Happy to split any of them out.

  No, all in this repository.


xiaofei-zheng:
Reviewed the full non-test diff plus the surrounding scoring paths (`loop/runner.py` baseline/anchor handling, `cli.py::_build_result` / `_checkpoint_on_best_committed`, `rewrite_by_flydsl/{runner,optimize,report,kb,driver_contract}.py`) against `git merge-base origin/main`. The metric unification itself checks out: `_measure_baseline` scores `own_bench / injected_anchor` in the right argument order, `_baseline_case_times` is seeded from `ic.baseline_case_times` at `loop/runner.py:600` so the CLI injection reaches it, `pristine_baseline_wall_ms` stays the caller's wall time, and `best_wall_ms` / `mean_case_speedup` / `aggregate_regression_detail` all end up on the same source-side denominator. Warm-start ordering (`max` on `(speedup, -index)`) preserves the lowest-index tiebreak. The dropped `implementation_match` / `consumer_implementation_identity` keys are only consumed by `knowledge/experience_reader.py`'s own dicts, not by the rewrite-layer candidate dict, so removing them from `_read_top_candidates` is safe.

Two blocking issues.

**1. An empty `source_case_ms` silently flips the published `speedup` back to a port denominator**

`preflight_reference` only requires the aggregate (`driver_contract.py:318-330`); a driver that prints `median_ms:` but no `case_ms:` lines in `--ref-bench-mode` is accepted, and `runner.py:312-314` even prints `cases=... or 'unreported'` as a normal state. When that happens `source_case_ms` is `{}` and three things follow:

- `scored_speedup(candidate.case_ms, {})` returns `None` — `calculate_mean_case_speedup` returns `None` (not `CaseCoverageError`) when the baseline map is falsy — so every warm-start candidate is rejected with `unscorable_measurement` and the run pays for a full PORT even when the KB holds a usable one. That is the opposite of what concern #2 of this PR set out to achieve.
- `optimize.py:267` (`if source_ms and source_case_ms`) is false, so `--baseline-json` is never passed and forge-loop anchors on the port.
- `report.py:98-99` then publishes the loop's port-anchored `mean_case_speedup` verbatim as `speedup`, while `source_ms` and `flydsl_best_ms` next to it are still source-vs-best. One field, two denominators — the exact defect this PR exists to remove.

Verified directly:
```
scored_speedup({'c0': 0.5}, {})                      -> None
build_result(port_ok=True, source_ms=1.0,
             optimize_result={'best_ms': 0.4,
                              'mean_case_speedup': 2.5}).speedup
                                                     -> 2.5   # vs the port
                                          source_ms  -> 1.0   # vs the source
                                     flydsl_best_ms  -> 0.4
```

This is not pre-existing in consequence: before this PR `speedup` was unconditionally `source_ms / flydsl_best_ms`, so a missing `case_ms:` in ref mode changed nothing a consumer reads. Now it silently changes what the number means, and `store.champion_speedup` promotes on it. The driver is LLM-authored (the prompt at `flydsl_rewrite_driver_preparation.py:460-461` asks for `case_ms:` in ref mode but nothing enforces it), and operator-supplied drivers take the `already conforms; preparation skipped` path with no check at all. Either make `preflight_reference` fail when `--ref-bench-mode` reports no per-case timings, or refuse to publish a `speedup` at all when `source_case_ms` is empty — but it must not degrade into a differently-denominated number under the same field name.

**2. `--baseline-json` combined with `--resume` puts the KEEP bar back at 1.0x**

`_measure_baseline()` is guarded by `if not self.resume` (`loop/runner.py:4521`), so on a resume `search_start_mean_case_speedup` is never assigned and `loop/runner.py:4545` falls through to `1.0`. `_seed_and_hydrate_run_state` only restores `best_mean_case_speedup` when `should_resume(run_state, head)` holds, i.e. when a prior KEEP's commit is still HEAD. Resume a campaign that had an injected anchor but no KEEP yet and the bar is 1.0x against a source-side anchor, while the workspace incumbent is the port already sitting well above 1.0 — `beats_current_best` will then KEEP a candidate slower than the kernel the campaign started from. That is the same failure mode the PR's own `test_an_injected_anchor_starts_the_keep_bar_at_the_search_start` pins for the fresh path. Persist the search-start score in run state, or refuse `--baseline-json` alongside `--resume` the way the other immutable-per-campaign options are refused.

Checks that came back clean: `--baseline-json` validation rejects every non-dividing payload; `preflight_candidate` in the warm start relies on `calculate_mean_case_speedup`'s exact case-set equality, so a candidate covering different cases is rejected rather than mis-scored; `_best_case_times` ends up holding the loop's own bench because `_measure_baseline` runs after the `4514` seed; workspace rollback in `try_flydsl_kb_warmstart` still restores the original bytes on the new `unscorable_measurement` path. CI is green on all shards, and the `PRECISION` fix matches the existing `monkeypatch.delenv` pattern in the tests around it.

jiaqiang-dot-liu:
Thanks — both blocking issues were real and are fixed in d0b5a20. Notes on each, including one place where I did not take the suggested route.

## 1. Empty `source_case_ms` flipping the published denominator

Fixed by the first option: `preflight_reference` now requires the per-case source timings, refusing a reference mode that reports only an aggregate with a named `REF_CASE_TIMINGS_MISSING` before any budget is spent. The `or 'unreported'` in the runner's baseline log went with it — that state should no longer exist.

Your framing of this as a regression rather than a pre-existing gap is the part I had underweighted. I had spotted the hole and filed it as a caveat in the PR description on the grounds that the contract had always permitted an aggregate-only reference mode. But the old code published `source_ms / flydsl_best_ms` unconditionally, so a missing `case_ms:` changed nothing a consumer read; after this branch it silently changes what the field means, and `champion_speedup` promotes on it. A caveat was the wrong call for that.

I did not also implement your second option (refuse to publish a `speedup` when `source_case_ms` is empty). With the contract check in place that state is unreachable past preflight, and this branch has just finished removing several unreachable guards from the warm start, so adding one here seemed inconsistent. Say the word if you would rather have it as defence in depth.

## 2. `--baseline-json` with `--resume`

Fixed, but it needed two changes rather than one, because the KEEP bar and the search-start score are different quantities that happen to share a value on the fresh path:

- **The bar** is now read off the incumbent's own per-case times against the anchor (`_incumbent_mean_case_speedup`). That is correct by construction on a resume, needs no new state, and comes out at exactly `1.0` whenever the incumbent is the anchor — so every existing caller is unchanged.
- **The search-start score**, which `incremental_speedup` divides by and which no later session can re-measure, is checkpointed in `RunState.search_start_mean_case_speedup` and restored next to the per-case timings it belongs with. Without this the publication path at `runner.py:1481` had the same defect on resume, one line away from the one you found.

On your alternative of refusing `--baseline-json` alongside `--resume`: I don't think it would have worked. The anchor is already persisted as `baseline_case_times` and restored by `_restore_resume_baseline_case_times`, so a resume runs against a source-side anchor whether or not the flag is passed again, and the bar would still have fallen back to `1.0`. Refusing the flag would have closed the way the state arrives without closing the failure.

## Tests

Both resume paths were unreachable by the suite before this. New coverage: the reference mode without per-case timings is refused by name; the bar is read off the incumbent; a checkpointed search-start score survives into a later session and keeps a candidate that beats the source but loses to the port from clearing the bar. Each was confirmed to fail against the previous code before being kept. One existing test needed its driver body extended with a `case_ms:` line — it pins the deprecated *aggregate* key and was never about per-case output.

Full `src/kernelforge` suite: 4051 passed. `ruff check` and `ruff format --check` clean.

## Unrelated CI failure

`py3.10 shard 2` came back red on `test_an_active_reactor_turn_is_cancelled_at_its_total_timeout`. It sets a 30 ms internal turn timeout and wraps the whole assertion in `asyncio.wait_for(..., timeout=0.2)`, so on a loaded 2-vCPU runner with two xdist workers the outer 200 ms wall is thin enough to trip on cancellation latency — which matches the reported `asyncio.exceptions.TimeoutError`. It passes locally, though its runtime swings between 0.6 s and 3.9 s across consecutive runs. The previous commit was green on every shard. I have re-run the job rather than widen the margin here, since it is a third subsystem and this PR is already carrying one unrelated fix; happy to file it separately.

xiaofei-zheng:
Re-reviewed at `3a3cc246`. Both blocking issues from the previous round are fixed correctly, and one new blocking issue came in with the fix for the second.

**Both previous findings verified fixed**

- *Missing per-case source timings.* `preflight_reference` now refuses `REF_CASE_TIMINGS_MISSING` before any budget is spent, which is the right place: it is the only production caller path (`flydsl_rewrite_driver_preparation.py:134`), a non-ok reference report feeds `prior_failure` into the driver-authoring loop rather than aborting the run, and that file has no failure-class allowlist that would make this one unrecoverable. Both bundled example drivers already print `case_ms:` in `--ref-bench-mode` (`triton2flydsl-softmax-flydsl-rewrite/driver.py:153`, `triton2flydsl-mxfp8-grouped-gemm/driver.py:329` via the shared `_run_benchmark`), so the evidence the prompt points at stays consistent with the new requirement. Dropping the `or 'unreported'` fallback in the log line is the right follow-through.
- *KEEP bar across a resume.* Reading the bar off the incumbent instead of assuming it is the better fix — it is correct by construction on both paths rather than by remembering to carry a number. I traced the ordering in `_run_locked`, since the fix depends on it: the resume branch calls `_restore_scoring_state()` at `runner.py:4368`, well before `_incumbent_mean_case_speedup()` at `runner.py:4571`, and `_restore_resume_baseline_case_times` at `4385` lands in between, so both sides of the ratio are populated by the time the bar is computed. `_scoring_state_restored` keeps `_seed_and_hydrate_run_state` from re-restoring. `_persist_scoring_state` assigns `search_start_mean_case_speedup` after `_measure_baseline` sets it, and the `is not None` guard stops `_promote_best` from erasing it on a later checkpoint. The new tests exercise the helpers directly rather than through `_run_locked`, so the ordering above is not covered by a test — it is correct as written, just worth knowing it rests on call order.

**New blocking issue: the new `RunState` field breaks every existing checkpoint**

`search_start_mean_case_speedup` was added to `RunState` (`run_state.py:210`) without bumping `SCHEMA_VERSION` (still `19`, `run_state.py:29`) and without a migration step. `from_dict` validates the field set exactly:

```python
expected = set(cls.__dataclass_fields__)
missing = expected - set(payload)
if missing:
    raise ValueError("run state missing fields: " + ", ".join(sorted(missing)))
```

A `run_state.json` written before this branch carries `schema_version: 19`, passes the version check unchanged, and then fails the field check. Reproduced on `3a3cc246`:

```
SCHEMA_VERSION = 19
simulated pre-PR payload schema_version = 19
RESULT: ValueError -> run state missing fields: search_start_mean_case_speedup
```

`LoopStateStore.load()` (`run_state.py:726-734`) does not catch it, and `_run_locked` calls `load()` unconditionally whenever `run_state.json` exists — not only under `--resume`. So every campaign with an existing workspace dies at startup on this branch, `--resume` included. That is the one path the second fix exists to protect.

The file's own convention covers this: v13 through v18 each bump the version and backfill the new field (`payload["analysis"] = ...`, `stall.setdefault(...)`). The same shape applies here — `SCHEMA_VERSION = 20` plus `if version == 19: payload.setdefault("search_start_mean_case_speedup", None); version = SCHEMA_VERSION`. Worth a regression test that loads a v19 payload, since nothing in the current suite reads a state dict it did not just serialize with the current schema, which is why CI is green on this.

Note that the bar itself does not depend on the new field — `_incumbent_mean_case_speedup()` derives it from `_best_case_times` and `_baseline_case_times` — so a backfilled `None` is fully safe. It only feeds `incremental_speedup` reporting.

The session-budget timeout change is fine: the 0.2 s wait was a suite-hang guard standing in for a 30 ms bound, and `max(activity) - min(activity) < 1.0` asserts the property the old wall was proxying for with more headroom.

jiaqiang-dot-liu:
You're right, and thanks for catching it — fixed in 1c2a4c1.

## The `RunState` field

I added a persisted field and skipped the migration the file has performed for every one of its last six versions. `SCHEMA_VERSION` is now 20, with v19 backfilling `search_start_mean_case_speedup` as absent, in the same shape as the v18 block above it.

Backfilled to `None` rather than `1.0`, for the reason you identified: a campaign that never had a caller-supplied anchor never measured a search-start score, and nothing reads the field to decide anything — the bar is derived from the incumbent's own per-case times — so `None` states what the checkpoint actually knows and only incremental reporting sees the difference.

Your point about the blast radius is the part I want to acknowledge properly. I had been thinking of this as a resume concern, and it is not: `_run_locked` builds `LoopStateStore` and calls `load()` unconditionally, ahead of the `if self.resume` branch, and the store is keyed on the workspace rather than the campaign. So a **fresh** campaign in any workspace that had ever run forge-loop would have died at startup too, on an error naming a field nothing in that campaign asked for. The commit that was supposed to protect the resume path was the one that broke it, and it would have broken considerably more than that.

Verified the whole chain still lands rather than just the new hop: v13 through v19 all migrate to 20.

## Why CI never saw it

Worth stating plainly, because it is the more useful finding: no test in the suite loads a state dict it did not just serialise with the current schema. Every `RunState` test round-trips `RunState().to_dict()`, so the field set is self-consistent by construction and an absent-field regression is invisible. That is why this was green through two rounds of review.

The new test loads a v19 payload with the field removed, following the v13–v18 tests next to it. Confirmed it reproduces your exact `ValueError: run state missing fields: search_start_mean_case_speedup` against the un-migrated code before keeping it. I have noted the gap in the commit message; a broader fixture of frozen historical payloads would be the real answer, but that is its own change.

`src/kernelforge`: 4052 passed. `ruff check` and `ruff format --check` clean.

## On the two things you flagged as non-blocking

**Call-order dependency.** Agreed, and thanks for tracing it — `_restore_scoring_state()` at 4368, `_restore_resume_baseline_case_times` at 4385, `_incumbent_mean_case_speedup()` at 4571. Correct as written, and untested as written, since the new tests drive the helpers directly. Happy to add a test that runs the resume branch of `_run_locked` end to end and asserts the bar, if you'd like it in this PR rather than as a follow-up.

**Session-budget timeout.** Noted, thank you.

xiaofei-zheng (approved):
Re-reviewed `1c2a4c120`. The run-state blocker from the last round is fixed correctly, and nothing new is blocking.

**What I checked**

- `src/kernelforge/loop/run_state.py:29` bumps `SCHEMA_VERSION` to 20, and the v18 branch now ends with `version = 19` instead of `version = SCHEMA_VERSION`, so the chain still falls through end to end. The new v19 step backfills `search_start_mean_case_speedup` with `setdefault(..., None)`, matching the v13-v18 convention in the same function.
- Ran the exact repro that failed before: `RunState().to_dict()` with `schema_version` forced to 19 and the field popped now loads and yields `schema_version=20, search_start_mean_case_speedup=None` (previously `ValueError: run state missing fields: search_start_mean_case_speedup`). Also re-checked the full chain from v13 (loads, backfills to `None`), a current v20 round trip (preserves `1.4`), and that an unknown v21 still raises.
- Backfilling `None` rather than `1.0` is safe for every consumer: `runner.py:2643` only assigns when the restored value is not `None`; `runner.py:1481-1482`, `cli.py:1825-1829` and `cli.py:2015-2019` all terminate in `or 1.0` before any comparison or division, so the `>` comparisons at `cli.py:1881-1882` / `cli.py:2049-2050` and the `incremental_speedup` divisions never see `None`. The KEEP bar is unaffected because it comes from `_incumbent_mean_case_speedup()` over the incumbent's own per-case times.
- Ran `src/kernelforge/tests/test_long_horizon_state.py`: 54 passed, 1 failure (`test_workspace_lock_rejects_concurrent_owner_and_can_be_reacquired`) that is purely an artifact of my Windows `fcntl` shim making the lock a no-op, unrelated to this change. `test_load_v19_migrates_without_a_recorded_search_start_score` passes and does exercise `LoopStateStore.load()`, which is the path `_run_locked` takes unconditionally.
- CI: 26 pass, 1 skipping, 2 still pending at review time.

Both earlier blockers stay resolved: `preflight_reference` now fails with `ref_case_timings_missing` when the driver prints no `case_ms:` in `--ref-bench-mode` (`driver_contract.py`), so `source_case_ms` can no longer be silently empty and flip the published `speedup` to a port denominator; and the resume KEEP bar is derived from `_incumbent_mean_case_speedup()` at `runner.py:4571` rather than from `search_start_mean_case_speedup or 1.0`, which I confirmed holds against the `_run_locked` call ordering (`_restore_scoring_state` 4368, `_restore_resume_baseline_case_times` 4385, bar 4571, `_seed_and_hydrate_run_state` 4575).

No blocking issues. LGTM.

PR #1498, opened 2026-09-14
probe: isolate shard-split effect (DO NOT MERGE)

luochen-amd:
Control experiment for #1445. Adds only a test file with the same number of tests as #1445's, and no product change at all. If shard 4 fails the same target-analysis test here, that failure is a pre-existing order dependency in the suite that any added test file exposes, not something #1445 causes. Will be closed once it has answered that.


luochen-amd:
Answered its question: shard 4 fails identically here, on a branch whose only change is a test file with the same number of tests as #1445's and no product code at all. Recording the result on #1445; closing.

PR #1501, opened 2026-09-14
Fix/controller patch integration and pump liveness

lishuoshuo-amd:

  Four defects, each of which on its own stopped a Kimi-K3 / forge e2e session from
  producing a measured result.

  1. **A controller patch could never reach integration.**
     `integrate_handler` received the controller's pre-applied flag but never read it,
     so the publication's unified diff was routed into the whole-file apply contract,
     which rejects a `.patch` suffix. All four patches of the reference session failed
     with `patch_path is not a complete source file` and the session closed at
     `cumulative_gain_validated=0.00%`. Re-running integration after the fix landed
     3 KEEPs for +9.99% e2e (511.02 → 562.08 tok/s).

  2. **The supervisor killed sessions that were making progress.**
     The tick stamp only advanced at the main loop boundary, but
     `_pump_dispatcher_once()` synchronously joins dispatched work, and
     `_NOT_JOINED_KINDS` holds only `targeted_build` — a PRELUDE baseline is joined.
     A cold-cache baseline took 3924s, so the watchdog reported
     `supervisor_tick_stalled` and killed the coordinator. Two runs confirm this is
     structural rather than a threshold to tune: a 3600s limit died at 3604s and a
     7200s limit died at 7208s, both still on tick 0. The pump now refreshes the
     stamp through main's existing `Reconciler.stamp_progress()`.

  3. **The forge prepass commit had no git identity.**
     The baseline commit in the same file passes `-c user.name` / `-c user.email`
     inline while the prepass commit was bare. `-c` lasts one command and the
     workspace carries no persistent identity, so `git commit` failed with
     "unable to auto-detect email address" and all five nominated operators reported
     `task_preparation_failed`. The prepass commit now names the same identity.

  4. **The pre-applied trust could be forged from an agent payload.**
     For `kind=integrate` the router merges `params` into the handler payload verbatim
     (`intent_router.py:1126`), and `params` is declared as an unconstrained
     `{"type": "object"}` in the tool schema, so an agent could set the flag itself and
     skip patch validation entirely — sending whatever state the worktree happens to
     hold into the e2e re-baseline, and turning any favourable measurement noise into a
     KEEP whose recorded `patch_path` never took effect. The flag moved from a payload
     key to a keyword-only parameter: the router's `handler_kwargs` is a hard-coded
     `{"session_dir": ...}`, so only an in-process caller can reach it.

     Content-verifying the worktree was considered and not taken: it defends a channel
     that should not be reachable at all, and a comparison can still be satisfied when a
     previous KEEP happens to have touched the same lines — the leftover-state case the
     concern was raised about.

  None filed; all four were found while running the Kimi-K3 forge session.

  Added, 9 tests over 3 files. Each was confirmed to fail on the pre-fix code with the
  production error signature before the fix was written:

  - `test_integrate_controller_preapplied_patch.py` (4) — the publication reaches the
    re-baseline, the diff is never re-applied, a diff without the flag is still refused,
    and a payload claiming to be pre-applied is refused.
  - `test_dispatcher_pump_tick_heartbeat.py` (3) — the pump stamps the tick during a long
    join, the stamp carries the right tick, and an unwritable stamp does not break the pump.
  - `test_task_preparer_git_identity.py` (3) — the prepass commit lands without an ambient
    identity (using `user.useConfigOnly=true` to reproduce a container), is attributed to
    KernelForge, and KernelForge wins over repo config.

  Commands: `ruff check .`, `ruff format --check .`, and `pytest -n` over the touched
  suites (`test_coordinator_gemm_promote_units.py` included as the neighbouring
  `preapplied_apply_result` coverage).

  No. `integrate_handler` gains a keyword-only parameter that defaults to `False`; every
  existing caller keeps its current behaviour, and the removed payload key had exactly one
  producer, which this PR updates.

  No — four independent defects. They are grouped because each one alone was enough to stop
  the same session from reporting a gain, so they had to be fixed together to get one clean
  e2e run. Happy to split them if that is preferred for review.

  No. All four are in this repository (`orchestrator/kernel`, `orchestrator/loop`,
  `kernelforge/loop`).

---

**Known unrelated CI failure:** `test (py3.10/py3.11 shard 4/6)` fails on
`test_target_analysis_executor.py::test_agentx_state_to_external_reference_and_final_report`
with `assert 'no_match' == 'ok'`. This is pre-existing and intermittent on `main` (2 of the
last 4 `main` runs, e.g. `27fc2d91` and `f6a2296b`, differing only in which parametrization
trips), and the `coverage` jobs fail downstream of it with `shard MISSING (did not report)`.
Left untouched as out of scope.


xiaofei-zheng:
Reviewed against merge-base `27fc2d915`. Read the full `integrate_handler` apply/re-baseline/cleanup path, `_default_validator` + `integrate_controller_patches`, `apply_kernel_patch`'s aiter invalidation blocks, `_pump_dispatcher_once` / `stamp_progress` / `supervisor_store.stamp_tick`, and every `git commit` site that runs against the campaign workspace. Tests were not run locally (`kernelforge.loop.external_artifacts` imports `fcntl`).

Two blocking issues.

**1. `request_handlers.py:6244-6249` — the synthesized apply_result skips aiter JIT / cpp_itfs invalidation, so an aiter publication is re-baselined against the stale binary.**

`_maybe_apply_kernel_patch` is not just "write the file": for a target under `aiter/csrc/` it calls `_invalidate_aiter_jit_build`, and for `aiter/csrc/cpp_itfs/` it also calls `_invalidate_aiter_cpp_itfs_cache` and records `cpp_itfs_cache_backup` (`apply_kernel_patch.py:2245-2299`) — the comment there is explicit: "Refuse to benchmark against an unknown/stale binary." The `preapplied_git_patch` branch bypasses all of it and returns a three-key dict, so at `request_handlers.py:6356-6357` `force_aiter_rebuild` is always `False`, `AITER_REBUILD=1` is never set, and `verify_cpp_itfs_rebuilt` (line 6486) is skipped because `cpp_itfs_backup` is empty.

Trigger: the controller explicitly nominates `aiter` as a backend for editable aiter sources (`kernel_rewrite_controller/opportunity_agent.py:299-300, 380`), and `_default_validator` passes `target_file = repo_root / kernel_path`. For such a publication the re-baseline server serves the pre-patch compiled kernel — `new_tput` measures the unpatched code, the gain is noise, and a KEEP can commit an unvalidated rewrite. The existing `preapplied_apply_result` branch does not have this problem because it reuses the real manifest. The preapplied-git branch needs the same invalidation (target-gated), not just a status dict.

**2. `task_preparer.py:1951-1958` — the identity fix covers one commit site; the next commit in the same repo still fails on the same host.**

`prepare_task`'s `workspace` is the campaign `workspace_dir` (`campaign_setup.py:124`), the same repo the forge loop commits into. On the host this PR describes (no persistent `user.name`/`user.email`, hostname with no domain) the prepass commit now succeeds, and then:

- `loop/runner.py:1192` — `git("commit", "-m", message, cwd=self.ic.workspace_dir)` with no identity; `git()` (`llm/git.py:51`) injects nothing and runs with `check=True`, so the KEEP raises `GitError`.
- `knowledge/experience_integration.py:330` — the warm-start commit, same repo, same gap, raises `RuntimeError("git commit failed: ...")`.

`-c` is per-invocation, so the prepass commit buys nothing for either. The fix that covers all of them at once is a persistent `git config user.name/user.email` on the repo in `_ensure_agent_git_workspace` (`task_preparer.py:1218`), which is what `fusion/shadow_repo.py:139-140`, `gemm_tune/tier3/generate.py:174-175` and `rewrite_by_flydsl/runner.py:78-79` already do.

Not blocking, and not reported above: the dispatcher stamp at `dispatcher.py:427` is fine as written (`coordinator.py:1506,1532` already call it unguarded, and `stamp_tick` documents `OSError` only), and the missing `manifest_path` on the synthesized result is compensated by the only caller's own reverse-apply (`controller_patch_integration.py:411/429`).

xiaofei-zheng:
Re-reviewed at `ecd4d0a8c` (`8b152a6e1` + `ecd4d0a8c`). Both issues from the first pass were addressed; each new implementation has one blocking problem left.

**1. `llm/git.py:36-51` — `GIT_AUTHOR_*` / `GIT_COMMITTER_*` outrank `user.name`/`user.email`, so this silently rewrites the identity on every commit into a repository that already has one.**

The helper's comment says "an ambient or caller identity still wins", and that is true for the environment and for the `env=` argument — but not for config, which is what real workspaces carry. Git's precedence is env var over `.git/config` over `~/.gitconfig`, verified locally:

```
git init; git config user.name Alice; git config user.email alice@example.com
GIT_AUTHOR_NAME=KernelForge GIT_AUTHOR_EMAIL=kf@localhost ... git commit -m t
git log -1 --format='%an <%ae> | %cn <%ce>'
-> KernelForge <kf@localhost> | KernelForge <kf@localhost>
```

`git()` is the single helper behind all 24 call sites in `kernelforge`, and the campaign workspace is the operator's own repository on its own development branch (`campaign_setup.py:108-111` checks out `git_branch`). After this commit every KEEP (`loop/runner.py:1192`) and warm-start commit (`knowledge/experience_integration.py:330`) in that repository is authored and committed as `KernelForge <kernel-forge@localhost>` instead of the configured identity, and a `kernel-forge@localhost` author will be rejected by any remote that validates committer email.

The default has to be a fallback, not an override: probe once (e.g. `git var GIT_AUTHOR_IDENT` / `git config user.email`, non-zero exit meaning "nothing to auto-detect") and only then inject, or keep the repo-local `git config` write in `_ensure_agent_git_workspace` for the repositories forge itself creates — which is what `fusion/shadow_repo.py:139-140`, `gemm_tune/tier3/generate.py:174-175` and `rewrite_by_flydsl/runner.py:78-79` already do and which cannot touch a repository that already has an identity.

**2. `request_handlers.py:946-1010` — the cache invalidation is back, but the rebuild is still missing, so a compiled target is still re-baselined against the pre-patch binary.**

`_invalidate_preapplied_aiter_caches` reproduces two of the four stages the skipped apply performs. Still absent:

- `_run_strategy_rebuild` (`apply_kernel_patch.py:2302-2316`). `integrate_handler` never sets `skip_rebuild` (`request_handlers.py:829` defaults it to `False`) and `_default_validator` does not either, so the normal path always rebuilds and refuses the measurement when the rebuild fails. For the editable aiter checkout `_detect_strategy` returns `rebuild_mode=command` with `setup.py develop` (`apply_kernel_patch.py:1521-1524`), and `/sgl-workspace/sglang/sgl-kernel/` returns `pip install -e .` (1525-1530). A controller publication touching either is git-applied, its JIT cache moved aside, and then measured with no recompile of the extension that actually serves the kernel.
- The multi-node fan-out branch (`apply_kernel_patch.py:2210-2243`), which invalidates the JIT tree on every pod. `integrate_handler` itself handles `is_multi_node()` for the restart, so this is a supported configuration; on it the worker pods keep their pre-patch build.
- `_runtime_jit_invalidation_error` (1682-1700) is also not consulted, so a `runtime_jit` strategy whose pinned `jit_build_dir` was not the one inspected passes silently here while the normal path refuses.

Both stages are target-gated and would no-op for a pure-Python sglang edit, so the fix is to run the same post-apply sequence rather than to re-pick stages from it.

xiaofei-zheng (approved):
Re-reviewed at `9de87a5bd`. Both blocking issues from the previous round are resolved, and the second one was resolved by the stronger fix rather than by patching the symptom. No blocking issues — LGTM.

What I checked this round:

**Identity (`b63697678` + `e09bd57f2` + `12eb79989`).** The `GIT_AUTHOR_*` default is gone from `git()`; `ensure_commit_identity` probes `git var GIT_AUTHOR_IDENT` and writes a repo-local value only when git can resolve none, per missing key. That probe is the right oracle: `git var GIT_AUTHOR_IDENT` uses the same strict identity resolution `git commit` does, so it succeeds exactly when a commit would. Exercised it against a repo that already has an identity — returns `""` and leaves `user.name` untouched, which is the override the last round flagged. Placement covers every workspace commit: `resolve_campaign` runs ahead of the resume branch and before `IterationConfig` is built (`cli.py:1053` then `1188`, same function), so `runner.py:_git_commit` and the warm-start commit are covered; the prepass call is at the commit itself rather than behind `_ensure_agent_git_workspace`, which only runs for a backend needing a git cwd.

**Pre-applied apply (`8b152a6e1` reworked, `4f8a727fd`).** Routing through the real apply with the worktree captured as `snapshot_dir` is the right shape — it restores backup, JIT/cpp_itfs invalidation, multi-node fan-out, rebuild, post-verify and the manifest in one move instead of re-picking stages. Walked the lifecycle for the cases that worried me:

- The local backup holds the *patched* bytes (the worktree was git-applied first), so revert order matters. `reverted_commit_failed` settles the manifest before `_revert_note`'s reverse-apply, which is the correct order; pod-side backups are pristine because the pods were never git-applied.
- `reverted_e2e_failed` needs no settle because integrate already reverted on the non-KEEP path.
- Deferring finalize to the caller does not mis-grade the result: `lifecycle_complete` counts `skipped`, so KEEP still reports `cleanup_status=complete` (`patch_lifecycle.py:16,32-35`).
- `_preapplied_snapshot_payload`'s snapshot dir is `change.patch`'s parent, i.e. per-publication (`controller_publication.py:77`), so no cross-publication leakage; `discover_controller_patch_dirs` is unaffected by the extra subdirectory.
- Delete descriptors are fine without a snapshot copy: apply unlinks with `missing_ok=True` and post-verify only asserts absence.
- `_final_content_snapshot` accepts the captured dir as-is since every written path exists in it, so the worktree bytes are what get measured.

**Tests.** The reworked guard tests assert the sharper property — an unflagged or payload-claimed integrate measures `HEAD + diff`, not the smuggled worktree content — which still fails if the payload ever regained the trust. CI is green (all checks pass).

PR #1502, opened 2026-09-14
fix(ci): overwrite stale shard artifacts on reruns

zengleixin-amd:
## Description

Replace each rerun shard's coverage and duration artifact using the existing `actions/upload-artifact@v7` `overwrite: true` option. Keep the fixed artifact names, download layout, shard selection, coverage thresholds, and failure gates unchanged.

In run [34816121060](https://github.com/AMD-AGI/Hyperloom/actions/runs/34816121060), Python 3.10 shard 2 passed on rerun, but coverage still selected its previous failed artifact. The old artifact ID (`10336681353`) was larger than the new one (`10336403083`), and the downloader's same-name latest selection used the old ID. This is not a coverage-threshold failure or a concurrent-unzip overwrite.

This patch changes only the two machine-consumed upload steps and adds a regression contract test. It does not introduce attempt-suffixed names, a selector script, new permissions, or artifact-cleanup jobs.

## Linked issues

None. Incident observed while validating #1497; that PR's application code is not changed here.

## Tests

- RED: the new parametrized contract test failed for both uploads before the workflow change.
- GREEN: `python -m pytest -q scripts/tests/test_ci_workflow_contract.py scripts/tests/test_ci_durations.py scripts/tests/test_ci_coverage_report.py` — **124 passed** using an isolated temporary environment with the CI test plugins.
- Ruff lint, Ruff format check, and `git diff --check` passed.
- `actionlint` was not available locally and was not run.

### Real GitHub Actions validation

[Isolated overwrite experiment, run 34825594906](https://github.com/AMD-AGI/Hyperloom/actions/runs/34825594906), using the same upload v7/download v8 revisions with `contents: read` only:

1. Initial shard-2 failure produced failure artifacts.
2. Rerunning only failed jobs replaced shard 2 with success; its old `.failed` file disappeared.
3. Rerunning only aggregation reused the successful artifact IDs unchanged.
4. Rerunning shard 2 as a failure replaced the prior success and kept the failure gate red.

All four payload-verification steps passed. Each artifact name had exactly one artifact, the untouched shard-1 IDs stayed unchanged, and coverage/duration payloads belonged to the expected attempt. The final experiment run is intentionally red because scenario 4 verifies failure propagation.

Notably, attempt 2 deleted coverage artifact `10340645890` and replaced it with the **smaller** ID `10339728220`; the subsequent download still used the new successful payload.

## Compatibility and limits

- No application behavior change; no change to the six-shard matrix, fixed duration seed, 90% threshold, or matrix failure gate.
- Rerun replacement removes the previous same-name artifact; historical per-attempt coverage/duration bundles are not retained under that name.
- The experiment used synthetic bundles, not the complete production test matrix. It did not inject artifact-service deletion failures or test fork execution. The action's deletion remains best-effort; this PR does not claim fault-proof replacement under arbitrary service errors.
- Existing polluted runs use their original workflow revision; validate the fix in a new run.

## Scope

Single concern: replace stale machine-consumed shard artifacts during reruns. Two files changed. No upstream ticket filed; this uses the action's existing supported overwrite mechanism.


xiaofei-zheng (approved):
No blocking issues. LGTM.

Checked:
- Both machine-consumed uploads (`.github/workflows/tests-coverage.yml:361` covdata, `:395` durations) now carry `overwrite: true`. These are exactly the two artifact families consumed by the downstream jobs: `Download shard coverage data` (`pattern: covdata-py*-shard*`, `merge-multiple: true`) and `Download shard durations` (`pattern: durations-shard*`). With one artifact per name, the non-monotonic-ID selection that caused run 34816121060 to pick the stale failed shard-2 bundle can no longer happen, and the `merge-multiple` unzip-order ambiguity for duplicate names goes away too.
- `pytest-log-py*-shard*` deliberately left without `overwrite`: it is human-consumed only and never downloaded by a job, so stale copies from an earlier attempt do not affect any gate.
- Nothing else changed: shard matrix, `fail-fast: false`, duration seed resolution, `if` conditions (`always()` for coverage, `always() && py3.10 && hashFiles(...complete)` for durations), `if-no-files-found` values, retention, coverage threshold and the `Enforce shard results` gate are untouched.
- Contract test `test_reruns_replace_only_their_own_shard_artifact` parses the real workflow YAML, so `overwrite: true` loads as a bool and `is True` holds; `pytest`, `yaml` and the `step()` helper are already present in the file. The coverage-path assertion still matches the three-entry path block.
- The PR body's live-Actions experiment (run 34825594906) covers the case that matters here: replacement with a smaller artifact ID still yields the new payload downstream.

PR #1503, opened 2026-09-14
Make the bare-metal framework install work on pip-packaged ROCm

lishuoshuo-amd:
## What

`install_baremetal.sh` and `preflight.py` assumed ROCm lives under a single `/opt/rocm` prefix. When ROCm arrives as TheRock's pip wheels instead, it is split across the `_rocm_sdk_core`, `_rocm_sdk_libraries` and `_rocm_sdk_devel` namespace packages, with some libraries nested under subdirectories the dynamic loader does not search. Three consequences, all fixed here:

- **Library resolution.** `ldd` against torch ran with a search path that missed most of the install, so the Phase 1 gate failed setup while listing libraries that were in fact present. All three packages are now probed, including the `rocm_sysdeps` and `host-math` subdirectories, both at launch and in the gate.
- **Headers for source builds.** SGLang's in-tree kernels and AITER compile against hipBLAS/hipSPARSE/thrust headers, which this layout ships only in `rocm-sdk-devel`, archived until `rocm-sdk init` expands them. That wheel is now installed pinned to the `rocm-sdk-core` version already present, and expanded before a source build.
- **Toolchain root.** torch's `cpp_extension` derives `ROCM_HOME` itself when neither `ROCM_HOME` nor `ROCM_PATH` is set, and settles on `_rocm_sdk_core` — which carries no hipBLAS headers and ships its own `hipcc`, whose clang driver then adds core's include tree. The AOT kernel build failed on a missing `hipblas/hipblas.h` even with the devel wheel installed. The root reported by `rocm-sdk path --root` is now exported before a source build, so the compiler and its includes come from the same package.

All of the above is a no-op on a standard `/opt/rocm` image, where those packages are not importable and the `rocm_sdk` module does not exist.

## Also in here

Three defects found along the way that fire regardless of how ROCm is packaged:

- vLLM's ROCm torch build links system OpenMPI (`libmpi.so.40`), which container base images generally do not carry. It is installed before the framework layer, best-effort, trying both the pre- and post-`time_t` package names.
- `SGLANG_ROCM_EXTRA` defaulted to `rocm724` whatever ROCm was installed, and the wheel-versus-source branch keyed off the interpreter version alone. The target is now derived from the installed ROCm torch, and the source path is taken when no published wheel matches it. On a ROCm 7.2 stack it still resolves to `rocm724`, so behaviour there is unchanged.
- Ray's preflight smoke test ran with `HIP_VISIBLE_DEVICES` cleared, which makes Ray refuse to import on ROCm, so every run performed a redundant reinstall. The pin is also treated as a floor rather than an exact version, with the `click` ceiling applied only to 2.44.1, whose CLI is the one that breaks with `click >= 8.3.0`.

## Tests

Added: `test_derive_runtime_paths.py` (extended), `test_install_baremetal_rocm_sdk_libpath.py`, `test_install_baremetal_sglang_wheel_target.py`, `test_preflight_ray_version_floor.py`, `test_install_ray_version_resolution.py`.

Each new test was run against the pre-fix code first and confirmed failing. The no-op cases are asserted with exact equality rather than absence, so a change that stopped detecting the wheel layout — or started firing on a standard image — would fail. The backward-compatibility cases (a ROCm 7.2 stack deriving `rocm724`, Ray 2.44.1 with a compliant `click`, the pin resolving normally) pass both before and after by design.

## Checklist

- Description: what and why — above.
- Linked issue(s): none.
- Tests: added, and verified failing against the pre-fix code.
- Breaking changes: none for ROCm 7.0/7.2 stacks. On a ROCm 6.x stack the SGLang wheel branch now takes the source path instead of requesting a `rocm724` wheel it was never going to match.
- PR addresses single concern: **no** — the pip-packaged ROCm layout, plus three adjacent framework-install defects. The latter three are independent and can be split into their own PR if that reads better.
- Root cause is upstream: no.


xiaofei-zheng:
Reviewed the full diff against `git merge-base origin/main fix/rocm-sdk-ld-library-path` (97eadf5e1).

### One blocking issue

**`install_baremetal.sh:851` — a ROCm 7.0 host on any Python other than 3.10 now aborts the SGLang install, where it previously fell through to the source path.**

The `rocm700` guard was written for an operator who explicitly exported `SGLANG_ROCM_EXTRA=rocm700`. It now also sees the value the new `sglang_rocm_extra_for_torch()` derives at line 839, and it runs before both the `--check-only` and the `DRY_RUN` branches, so the whole framework phase dies:

```
if [ -z "$SGLANG_ROCM_EXTRA" ]; then
  SGLANG_ROCM_EXTRA="$(sglang_rocm_extra_for_torch "$py")"   # torch hip 7.0.x -> rocm700
fi
...
if [ "$SGLANG_ROCM_EXTRA" = "rocm700" ] && [ "$py_mm" != "3.10" ]; then
  die "SGLANG_ROCM_EXTRA=rocm700 currently supports Python 3.10 AMD wheels only; ..."
fi
```

Trigger: `torch.version.hip` starting with `7.0`, `SGLANG_ROCM_EXTRA` unset, interpreter != 3.10 — i.e. a ROCm 7.0 image on Ubuntu 24.04's Python 3.12. This is the amdgpu-6.3.x-pinned host the comment above `install_sglang_framework` names as the reason `SGLANG_ROCM_EXTRA` is overridable at all, so it is not a contrived configuration.

Before this PR, `SGLANG_ROCM_EXTRA` defaulted to `rocm724` on that host, the guard never fired, and the run took the source install with a warning. Reproduced by running lines 838-853 with `sglang_rocm_extra_for_torch` returning `rocm700` and `py_mm=3.12`: exits via `die`, whereas the merge-base code logs `SGLANG_ROCM_EXTRA=rocm724` and continues.

Suggest distinguishing the derived value from the exported one — only `die` when `rocm700` came from the environment, and otherwise fall through to the source install (or clear the derived extra) as before. The `docs/compatibility.rst` breaking-change note and the PR's "Breaking changes" line should cover the 7.0 case too, not just 6.x.

### What was checked

- `install_baremetal.sh`: the new `rocm_sdk_wheel_lib_dirs` / `rocm_sdk_wheel_include_dir` / `rocm_devel_headers_present` / `export_rocm_sdk_toolchain_root` / `ensure_rocm_devel_headers` / `ensure_openmpi_runtime` helpers and their call sites. Confirmed the `LD_LIBRARY_PATH` loop uses process substitution rather than a pipe, so the `export` lands in the caller's shell; confirmed the new `cmd && cmd && return 0` guards do not trip `set -euo pipefail` (a failing non-final element of an AND-list is exempt — verified in a scratch script); confirmed the new possibly-empty `SGLANG_ROCM_EXTRA` / `SGLANG_ROCM_PYPI_VERSION` are not the last statement of `write_runtime_dotenv`, so its exit status is unaffected; confirmed `check_rocm_toolchain_alignment` has only the one call site the new `"$py"` argument was added to.
- `preflight.py`: `_rocm_sdk_wheel_lib_dirs` prepend order in `_derive_runtime_paths` (`reversed()` preserves probe order), the Ray floor-vs-exact smoke change, the `click` ceiling narrowed to 2.44.1, `_ray_probe_env` re-indexing `HIP_VISIBLE_DEVICES` from `ROCR_VISIBLE_DEVICES`, and the `>=` install fallback. The fallback path still fails loudly through `RuntimeError` / the post-install smoke test rather than swallowing a bad resolution.
- `agents/kernel/scripts/install.sh`: `resolve_ray_version` only relaxes the built-in default, the empty `RAY_CLI_CLICK_MAX_VERSION` is guarded in the verification heredoc, and a `pip index versions` failure leaves the pin untouched.
- `_server_patcher.py`: the `_sgldev` preference and the fallback-scan tie-break. Walked both orderings of `sglang_0_5_18` / `sglang_0_5_18_sgldev` under `sorted()` — `want_branch` true picks the variant, false picks the plain dir, and a lone variant is still usable. `_subdir_version_tuple` parses the suffixed name to `(0, 5, 18)`. The `_resolve_sglang_apply_root` hoist in `_discover_sglang_plan` only changes which warning is logged when the apply root is unresolvable. No `*_sgldev` tree ships in this repo, so the second `_resolve_versioned_patches_dir` call site (the KernelForge CK plan, line 728) keeping the old two-arg form is not a behaviour change today.
- `launcher.py`: the supervisor default flip to opt-in matches `docs/reference/environment-variables.md:686` and the new test. Not mentioned in the PR description; worth adding there.

Tests run locally (targeted, not the full suite): `test_server_patcher_sglang_release_branch.py`, `test_server_patcher_sglang_version_fallback.py`, `test_baremetal_doc_version_consistency.py`, `test_supervisor_opt_in.py` — 29 passed. `test_derive_runtime_paths.py` and `test_preflight_ray_version_floor.py` do not import on Windows (`fcntl` via `cli/__init__`); CI covers them.

xiaofei-zheng (approved):
Retracting the blocking issue from my previous comment.

The `rocm700` path it depended on is unreachable for the stacks this project targets: Hyperloom requires Python 3.10+, and the supported ROCm floor is 7.2, so `sglang_rocm_extra_for_torch()` derives `rocm724` and the `install_baremetal.sh:851` guard never sees a derived `rocm700`. The guard keeps doing what it was written for — catching an operator who exports `SGLANG_ROCM_EXTRA=rocm700` explicitly. Not a real-configuration failure, so it should not have been reported as blocking.

Everything listed under "What was checked" in the previous comment stands: the `_rocm_sdk_*` probing and its `set -euo pipefail` interactions in `install_baremetal.sh`, the `LD_LIBRARY_PATH` prepend order in `preflight._derive_runtime_paths`, the Ray floor-vs-exact change and its `>=` install fallback (fails loudly rather than swallowing a bad resolution), the narrowed `click` ceiling and the empty-`RAY_CLI_CLICK_MAX_VERSION` guard in the kernel-agent verification heredoc, the `_sgldev` preference and fallback-scan tie-break in `_server_patcher.py`, and the supervisor opt-in flip against `docs/reference/environment-variables.md:686`.

Tests run locally (targeted): `test_server_patcher_sglang_release_branch.py`, `test_server_patcher_sglang_version_fallback.py`, `test_baremetal_doc_version_consistency.py`, `test_supervisor_opt_in.py` — 29 passed. `test_derive_runtime_paths.py` and `test_preflight_ray_version_floor.py` do not import on Windows (`fcntl` via `cli/__init__`); CI covers those.

Two non-blocking notes for the author's judgement: the supervisor default flip is not mentioned in the PR description, and `docs/compatibility.rst` still describes the pre-PR `rocm700` override story.

PR #1509, opened 2026-09-14
Added support for ATOM detailed profiling

devalshahamd:
**Description: what and why**
ATOM profile runs now emit TraceLens-grade traces. On `PROFILE=1` + `--framework atom`, `_workload_envs` sets `NUM_PROMPTS=CONC` (ATOM has no `delay_iterations`/`max_iterations` window, so extra prompts only grow the HTTP-bracketed trace) and injects `--mark-trace`, `ATOM_ENABLE_DETAILED_ANNOTATION=1`, `ATOM_PROFILER_MORE=1` — each only when a cached one-shot probe of the installed atom reports it (`EngineArgs --help` + `atom.utils.envs`). Fail-soft: an older argparse never sees `--mark-trace`. Never injects vLLM/SGLang `--profiler-config.*`.

**Linked issue(s):** none

**Tests:** updated `test_workload_envs.py` (probe hit/miss + caps parse/fail-soft), `test_workload_envs_branches_unit.py`, `test_profile_and_kernel_handlers.py`. Ran `pytest src/hyperloom/inference_optimizer/tests/test_workload_envs.py src/hyperloom/inference_optimizer/tests/test_workload_envs_branches_unit.py` (125 passed), `ruff check .`, `ruff format --check`.

**Breaking changes:** no

**PR addresses single concern:** yes

**Root cause is upstream:** no


BaoYunkai (approved):


## What I checked

- **Injection ordering.** The block lands at line 2070, after `extra_envs` / `replace_args` / `remove_args`, and before both `filter_untrusted_env_mapping` (2083) and `seal_server_argv` (2094). The filter only drops `BENCHMARK_SECRET_ENV_NAMES`, so the `ATOM_*` keys survive; `_sealed` only tokenizes and normalizes, so `--mark-trace` survives too. The comment's claim that the seal is the last write holds.
- **`conc_val` is in scope.** Defined at 1427, used at 1557.
- **The `NUM_PROMPTS` write is reachable for atom.** It sits under `if not _is_scriptable_profile`, and atom is not scriptable.
- **Failures are not swallowed silently.** Both the exception path and the bad-output path log a warning naming what was omitted.
- **Comments and config docs are in sync.** `profile_atom.yaml` drops the old "defer to Magpie" and "`ATOM_PROFILER_MORE` intentionally left OFF" text and states the new behaviour. A repo-wide grep for those stale claims returns nothing.
- **CHANGELOG is not applicable.** It was retired and removed from `main` in #1589, and the PR template does not ask for one.
- **PR description matches the diff** after all three commits.

PR #1510, opened 2026-09-15
chore(vllm): bump bare-metal default to 0.29.0

BaoYunkai:
## Problem

1. Hyperloom bare-metal vLLM still defaulted to 0.28.0 while upstream 0.29.0 is available and TraceLens already ships `config_vllm_v0.29.0.patch` on main (#1487).
2. Docs, example skills, and Slurm metadata still pointed at `v0.28.0` Docker tags and pip specs, so new setups would not pick up the supported stack.

## Fix

1. Bump `VLLM_VERSION` default to `0.29.0` (`rocm723`) in `install_baremetal.sh` and keep the existing glibc >= 2.39 gate for vLLM >= 0.28.0.
2. Sync `docs/compatibility.rst`, `docs/install/install.md`, example `SKILL.md` files, and `assets/slurm/models.tsv` to `vllm/vllm-openai-rocm:v0.29.0` / `vllm==0.29.0+rocm723`.
3. Update glibc-gate tests and the bare-metal doc consistency test; document the bump in `CHANGELOG.md`.

## Test plan

- [x] `pytest src/hyperloom/inference_optimizer/tests/test_baremetal_doc_version_consistency.py`
- [x] `pytest src/hyperloom/inference_optimizer/tests/test_setup_cli.py -k glibc`
- [x] `ruff check .` and `ruff format --check .`


ZhengGong-amd (approved):
Approved

PR #1511, opened 2026-09-15
fix(roofline): classify capture failures instead of retrying degraded

haishuok0525:
- **Description**: what and why

  Roofline measures an already-validated workload, so a CUDA-graph capture failure there is evidence about that workload, not a cue to change its configuration. The eager fallback did change it: on capture failure it silently re-ran with a degraded config and reported the degraded number as the roofline. The failure disappeared, its cause was never recorded, and the resulting ceiling was wrong in a way nothing downstream could detect.

  Roofline now classifies the failure and stops. Classification separates an **instrumentation** collision (profiler contending with capture, our own doing) from a **config** incompatibility (hardware or framework cannot capture this shape), which is the distinction that decides who fixes it. The evidence goes to a diagnosis artifact next to the run.

  Stopping is only useful if the report says enough to act on, and it did not: a reader could not tell a broken profile from a broken workload. So this also reports the facts that settle that question.

  Preflight, asserted before the measurement runs:
  - orphan reap results (what was left behind by a prior run, and what we removed)
  - disk headroom and stale trace artifacts (capture writes gigabytes; a full filesystem looks exactly like a capture failure)
  - an instrumentation assertion carrying patcher outcomes, so a missing patch is visible before it costs a run

  Post-flight, measured from what the run produced:
  - capture sidecar operator-metadata coverage, measured by opening the sidecars rather than inferring from filenames (up to 24 scanned, with partial-scan bounds recorded)
  - metric self-description: every ratio carries its numerator, denominator and aggregation scope, so `attributed_pct` cannot be read against the wrong basis
  - a verdict with continuous measures and a severity rank (`blocked` > `silently_wrong` > `warn` > `ok`), replacing a pass/fail that could not express "the number exists but do not trust it"
  - a server liveness probe per attempt, separating "the server died" from "the capture failed"

  Patch outcomes previously reached the session breakdown only through `trace_validate.checks`, which does not exist on a run that produced no trace — exactly the run whose patch state you need. They now ride on the per-attempt run row and are drained on every path, success or failure.

  On volume: the full certificate with its per-rank tables and candidate inventories is written to `selfcert.json` on disk (82 KB on the verification run). The session breakdown keeps the ~40 scalars and a path to it, 8.2% of the certificate's size.

  Two bugs surfaced while verifying this against real traces:
  - the cpu_op structure check searched for `{"name": "cpu_op"}`, which never appears in Kineto output — the category lives in `"cat"`. On a real sidecar the searched form matched 0 events while the correct form matched 2915, all carrying Input Dims. The check had been reporting "no cpu_op events" on fully instrumented captures for its entire lifetime, masked by fixtures that set both keys.
  - `HYPERLOOM_TRACELENS_PATCH_STATUS` was written only on failure, so "never attempted" and "attempted and failed" were indistinguishable. It is now written in every branch as `ok` / `unavailable` / `not_attempted`.

- **Linked issue(s)**: none

- **Tests**: added/updated.

  New: `test_trace_selfcert_observability.py` (8 tests — sidecar scanning, partial-scan bounds, rank-level basis transparency, verdict severity). Extended: `test_roofline_executor_units.py` (non-retryable failure still produces a run row; instrumentation draining), `test_sbd_v6_roofline_timeline.py` (scalars inlined, lists excluded), `test_profile_and_kernel_handlers.py` (instrumentation preflight; fixtures corrected from the synthetic dual-key form to real Kineto `{"cat": "cpu_op", "name": "aten::mm"}`).

  Commands run — `pytest` over the touched modules and their neighbours on the merged tree: 1775 passed, 5 skipped, 0 failed. Full-suite run before the merge showed no new failures (the 20 failures are pre-existing credential/auth cases unrelated to this path).

  Verified end to end on the cluster against sglang 0.5.19 (graphattr), a workload that actually captures. The roofline action succeeded, produced 53 capture files, and every observable above came back with real values: `orphans_reaped_count=0`, `free_pct=6.05`, `stale_trace_count=0`, `tracelens_patch_status=unavailable` with six patchers true (the mirror image already carries the patch), `capture_op_meta_coverage=1.0` over 52 files with 24 scanned, `attributed_pct=1.77` beside `attributed_gpu_ms=10.89` and its scope, `graph_launch_coverage=1.0`, `severity=ok`, liveness `{"pidfiles": 0}`. An earlier attempt on vLLM 0.24.0 could not generate captures at all (it rejects the profiler flags), which is what sent the verification to sglang.

- **Breaking changes**: no. The removed fallback only ever produced a misreported roofline; callers that relied on roofline returning a number for an uncapturable workload now get a classified failure instead, which is the intent.

- **PR addresses single concern**: yes — a capture failure in roofline should be reported rather than worked around, and reporting it requires the facts to be present.


xiaofei-zheng:
Reviewed the full diff against `git merge-base origin/main fix/haiskong/roofline-drop-eager-fallback` (7023f05). One blocking issue.

## Blocking

**`selfcert.json` is written into the trace directory under a name the trace resolver treats as a trace candidate.**

`_write_trace_certificate` (`src/hyperloom/orchestrator/actions/executors/profile.py:310-318`) writes the certificate to `selected_trace_dir / "selfcert.json"` — the same directory that is later handed to the resolver as `trace_input`. `.json` is in `_TRACE_EXTS` (`src/hyperloom/agents/kernel/tools/_bypass_trace_reader.py:38`), so `_trace_candidates()` picks the certificate up, and neither `_is_capture_fragment` nor `_role_of` filters it out — `trace_selfcert.inventory()` classifies it as `role="source"`.

Two confirmed consequences, both reproduced against this branch:

1. `require_single_rank=True` with an unknown TP stops resolving a directory that resolved fine before. `select_primary_trace` (`_trace_rank.py:46`) returns the single unranked trace only when `len(non_merged) == 1 and len(unranked) == 1`. The certificate is a second unranked non-merged candidate, so the branch falls through to `return None`. This is size-independent:

```python
d = Path(tempfile.mkdtemp())
(d / "host_123.1700000000.pt.trace.json.gz").write_bytes(b"x" * 5_000_000)
resolve_trace_file(d, require_single_rank=True)   # -> the 5 MB trace
(d / "selfcert.json").write_text("x" * 82_000)
resolve_trace_file(d, require_single_rank=True)   # -> None
```

`trace_analyze_handler` accepts a directory (`request_handlers.py:5331`, `payload.get("trace_input") or payload.get("trace_dir")`), and `--require-single-rank` is added for agentx mode while `--tensor-parallel-size` is only passed when `state.tp > 0` (`request_handlers.py:5135-5142`), so that combination is reachable. `analyze_trace` then returns `no usable trace file at <dir>`.

2. Without `require_single_rank`, `_select_trace_file` falls through to `max(candidates, key=file_size)` whenever no candidate name carries a rank and none starts with `merged-`. The 82 KB certificate then outranks a degenerate capture:

```
candidates ['selfcert.json', 'tiny.trace.json']
resolved   .../selfcert.json
```

Inside `inventory()` this also feeds `production_pick_probe`: `analyze_trace` on the certificate reports zero kernels, and `build_verdict` (`trace_selfcert.py:871-878`) emits the blocking reason "the live resolver selects ... no GPU kernels, so both consumers see an empty trace" — a false positive on the PR's own headline signal.

sglang names encode `TP-<n>`, so the ranked branch shields that path; vLLM and other rank-less layouts do not. The certificate's own directory is also re-scanned whenever the same trace dir is certified or analyzed again, so the first write is enough to affect every later pass.

Fix: give the certificate a name outside `_TRACE_EXTS`, or write it to a sibling directory. `result["trace_validate_path"]` already carries the location, so nothing depends on it sitting among the traces.

## Checked, no blocking issue found

- Classification logic in `baseline.py` is a pure `bool` -> `(category, marker)` refactor; marker sets, OOM demotion and the `±1` line radius are byte-identical to the base.
- `_server_liveness_probe`'s substring match on `task_id` is safe: task ids default to `uuid.uuid4().hex` (`state/task_registry.py:115`), and the substring form is what lets the parent roofline id match its `-profile` child's pidfiles.
- `_preflight_probe`'s stale-trace scan uses `_TRACE_SUFFIXES` (`.pt.trace.json.gz`, `.trace.json`, ...), which the certificate does not match, so `stale_trace_count` is unaffected.
- `_extract_vllm_block_fp8_profile_shapes` rglobs `*.json` over the trace dir but skips payloads with no `traceEvents`, so the certificate is a no-op there.
- Every `return`/`continue` in the attempt loop reaches `_note_profile_run` before exiting, including the new `_NON_RETRYABLE_PROFILE_ERRORS` branch; no failure path leaves the timeline event dangling.
- `HYPERLOOM_TRACELENS_PATCH_STATUS` is now written in all three branches; the `not_attempted` default is set before the `patch_attempted` check, so no branch can leave it unset.

One thing worth a second look, not blocking: the `if not disable_cuda_graph` guard that used to gate the capture classification is gone, so with `HYPERLOOM_PROFILE_DISABLE_CUDA_GRAPH=1` a stale weak-marker line in the 16 KB log tail would hard-fail the action with a message that reads "does not retry with graph capture disabled" on a run where capture was already disabled. I could not construct a realistic log that triggers it, since a run booted eager should not print the marker.

haishuok0525:
Thanks — the blocking issue is real and I reproduced both consequences verbatim before changing anything. Fixed in ce9dde9.

## The blocking issue

Confirmed exactly as described. `_TRACE_EXTS` contains a bare `".json"` and `_trace_candidates` rglobs the trace dir, so the certificate became a trace candidate for the very directory it describes. Both consequences reproduce on this branch:

```
case1 before: host_123.1700000000.pt.trace.json.gz
case1 after : None
case2 candidates: ['selfcert.json', 'tiny.trace.json']
case2 resolved  : selfcert.json
```

Worth adding to your writeup: consequence 2 is self-inflicted in a particularly bad way. The certificate winning the size fallback makes `analyze_trace` report zero kernels, which trips `build_verdict`'s blocking reason — so the artifact this PR adds to explain trace quality was capable of manufacturing a blocking verdict about itself.

I took the sibling-directory option rather than renaming outside `_TRACE_EXTS`. Two reasons. A subdirectory is not an option at all (the scan is recursive), and any name-based dodge stays one `_TRACE_EXTS` edit away from breaking again, whereas being outside the scanned root is not extension-dependent. Keeping `.json` also keeps the file usable with `jq` and editors, which matters for something whose whole purpose is being read by a human during triage.

`_write_trace_certificate` now writes `trace_dir.parent / f"{trace_dir.name}.selfcert.json"`. The parent is safe to write into: `_candidate_trace_dirs` only ever hands the resolver `workspace/torch_trace`, `workspace/capture_traces` or `workspace.parent/capture_traces`, so the parent of a selected trace dir is never itself scanned. The name carries the trace dir's so a workspace certifying both `torch_trace` and `capture_traces` does not collide on one file. You were right that nothing depends on the location — `trace_validate_path` is the only reference and the only consumer is `certificate_path` in the SBD recorder, so this is a pure move.

Regression test added (`test_trace_certificate_stays_out_of_the_resolver_namespace`), covering both of your cases plus the collision. I checked it fails against the old path and passes against the new one, so it actually pins the behaviour rather than just asserting the new layout.

## The non-blocking point

Also fixed, in the same commit. You were right that a realistic log should not trigger it — a run booted eager should not print the marker — but the trade is wrong regardless of reachability: the 16 KB server log tail can span an earlier boot, so with the override set a capture marker is not evidence about *this* run, and hard-failing the action on it means acting on untrustworthy evidence. The misleading message was the symptom, not the problem.

Both classification call sites are now gated on `disable_cuda_graph` being false. The guard is back where the old code had it, but for a different reason, and the comment says so: the old one existed to avoid arming a next attempt, this one exists because the classification is only meaningful when capture was actually on. A genuine failure on an override run still fails, just as a plain profile failure, and the run row already carries `disable_cuda_graph` and `env_disable_cuda_graph` so the override is visible to anyone reading it. Test: `test_capture_marker_is_not_classified_when_capture_was_already_off`.

## Verification

1777 passed / 5 skipped / 0 failed over the touched modules and their neighbours, ruff clean.

Two things your review checked that I want to flag as *still* true after this change, since the certificate moved: `_preflight_probe`'s stale-trace scan uses `_TRACE_SUFFIXES` and the certificate does not match it either way, and `_extract_vllm_block_fp8_profile_shapes` rglobs the trace dir, which the certificate is now outside of entirely.

Thanks for reproducing it rather than just flagging the extension overlap — the `require_single_rank` path is one I would have argued was shielded by sglang's `TP-<n>` naming, and the second case shows it is not the only way in.

PR #1512, opened 2026-09-15
KernelForge: add fuzzy KB warm-start fallback across framework versions and GPUs

meinali-566:
## Summary

Add fuzzy KernelForge KB identity matching for `forge-loop` and `rewrite-by-flydsl` warm starts.

Warm start still performs an exact six-tuple lookup first:

```text
kernel:{producer}:{kernel_name}:{framework}:{framework_version}:{backend}:{gpu}
```

When the exact identity has no candidates, KernelForge now queries the KB Store identity index through `POST /v1/kb/search` and ranks compatible donor identities client-side.

## Motivation

KernelForge previously required all six identity dimensions to match exactly. As a result, useful historical solutions could not be reused when only the framework version or GPU model changed.

For example, a solution recorded under:

```text
kernel:forge-loop:softmax:vllm:0.11.2:triton:mi300x
```

was invisible to a run targeting:

```text
kernel:forge-loop:softmax:vllm:0.11.3:triton:mi355x
```

Even though the candidate could still be useful after validation and remeasurement.

## Matching behavior

The lookup order is:

1. Query the exact six-tuple identity.
2. If the exact identity has candidates, use only those candidates.
3. If the exact identity is empty, search identities with these dimensions fixed:
   - `producer`
   - `kernel_name`
   - `framework`
   - `backend`
4. Rank the remaining identities using `framework_version` and `gpu` as soft similarity dimensions.

Framework-version preference:

1. Exact version
2. Same major/minor version
3. Same major version
4. Other valid versions

GPU preference:

1. Exact GPU identifier
2. Known GPU with the same ISA
3. Other non-empty GPU identifiers, including future GPU models not yet present in the compatibility map

Missing or unusable metadata is rejected:

- empty values
- `unknown`
- `none`
- `unspecified`
- invalid framework versions

The exact identity is explicitly excluded from fallback results because it was already queried first.

## Safety

Fuzzy identity matching only expands candidate discovery. It does not directly trust or apply a donor result.

Selected candidates still pass through the existing safeguards:

- artifact manifest and integrity validation
- patch path validation
- implementation compatibility handling
- correctness validation
- measurement on the current GPU
- existing warm-start speedup and search-budget policies

Writes remain exact and continue using the current run’s complete six-tuple identity. Only the read path is relaxed.

## Implementation

- Add client-side identity ranking in `warmstart_identity.py`.
- Add `search_identities()` to the vendored KernelForge KB Store client.
- Use `/v1/kb/search` with the four non-degradable dimensions.
- Add paginated identity discovery to the KB Store rewrite record backend.
- Keep the local record backend exact-only for the initial rollout.
- Integrate fallback lookup into:
  - `forge-loop`
  - `rewrite-by-flydsl`
- Preserve fuzzy-match metadata through the final forge-loop warm-start result:
  - `match_tier`
  - `requested_canonical_id`
  - `selected_canonical_id`

## KB Store compatibility

No KB Store server change is required.

The deployed service already:

- supports `scheme=kernel` through `/v1/kb/search`
- has a ready PostgreSQL identity index
- contains historical backfilled kernel identities
- permits the Hyperloom credential to access the `kernel` scheme

A follow-up should synchronize the new `search_identities()` method into the upstream KB Store Python client to avoid long-term divergence from the vendored KernelForge copy.

## Validation

- 165 targeted and integration tests passed
- 71 metadata-preservation regression tests passed
- Ruff passed
- IDE lint passed
- Verified the deployed KB Store identity index:
  - 249 kernel identities
  - 132 forge-loop identities
- Verified a Hyperloom credential can successfully call `POST /v1/kb/search` for the `kernel` scheme

## Follow-up validation

- Run a production-like forge-loop workload with an exact identity miss.
- Confirm `/v1/kb/search` selects a cross-version or cross-GPU donor.
- Confirm the donor is revalidated and benchmarked before adoption.
- Confirm requested and selected canonical IDs appear in the final warm-start status.


jiaqiang-dot-liu:
LGTM

jiaqiang-dot-liu (approved):

PR #1514, opened 2026-09-15
refactor(orchestrator): drop five mechanisms with writers but no readers

fengshaoyi-amd:
﻿- Description: what and why

An audit of the orchestrator found five mechanisms that had writers, a schema
column, or a test suite — everything except a production reader. This lands the
removals one batch per commit, in the order the closure required: each deletion
is what made the next one visibly dead, which is why `_action_in_phase_map`
only shows up in batch 3 rather than in the original scan.

| Commit | What goes |
|---|---|
| `e5fcf8695` | `kernel_agent` inbox subscription + `IntentType.RESPONSE`, end to end across bus / intent / policy / roles |
| `492b353d8` | the `events.priority` column — 15 writers, no reader, no index, no `ORDER BY` |
| `8c7d3ef2f` | `IntentSpec.builder` and its two unwired builders |
| `e85ad1ea0` | rewrite demoted from a phantom kernel lane to `REWRITE_RESERVE_SHARE` (**refactor, not removal**) |
| `14d6eed1f` | `PHASE_EXIT_REASONS`, `is_action_allowed_in_phase`, `phase_history_event_name`, `LIFECYCLE_STATUSES` |
| `6588a5041` | 11 stop reasons no code path can produce |
| `03f605b65` | CHANGELOG |
| `6f24b674a` | follow-up: seven raw event `INSERT`s the column removal missed |
| `46f00b210` | follow-up: collapse the `INSERT` strings that removal shortened |
| `dd22df101` | review follow-up: three archived-session explanations restored, two comments that outlived their mechanism corrected |
| `c9d7c08ad` | review follow-up: `ensure_schema` retires the column off a resumed database, so the removal no longer strands an in-flight session |

Four judgements worth review attention.

`IntentType.RESPONSE` was dispatched, policy-gated and validated end to end
while the `kernel_agent` process that would consume it is never instantiated.
The `request` / `response` topics and `kernel_agent` as a routing target are
alive and untouched — only the subscription and the intent type go.

`PHASE_EXIT_REASONS` went while its twin `STOP_REASON_VOCAB` stayed, and that
asymmetry is the point: a phase exit reason only reaches `phase_history` for a
human to read, so a typo there cannot misroute anything, whereas the stop
vocabulary gates `machine.py`, `close.py` and `set_stop_reason`.

The "legacy sentinels — kept for backward compat (resume from old sessions)"
comment did not survive checking, so it goes with the reasons it described.
`SharedState.from_dict` restores `stop_reason` as a dataclass field, bypassing
`set_stop_reason`, and `_global_terminal` returns unrecognised values verbatim
under `vocab: "unknown"` — vocabulary membership was never what let an old
session report its terminal. `SKILL.md` and `troubleshooting.md` both claimed
`policy_loop` sat in the vocabulary for that reason; both now describe
`policy_denied` streaks as what they are, advisory and stopping nothing.
`enablement_stalled` stays legal in the enablement timeline event namespace,
which is a different vocabulary and is untouched.

`_STOP_REASON_EXPLANATIONS` keeps entries for three reasons the vocabulary no
longer carries, because that table is a fallback dict rather than a mirror of
the vocabulary. Those three had real production writers in shipped versions —
`enablement_stalled` was set in four places as late as `82df24f39` (2026-08-27),
and `plateau_explore` / `framework_agent_plateau` were returned by exit rules
that mirror into `stop_reason` — so a session archived before those writers went
away still carries one, and a report over it should still say why it stopped.
The coverage test runs one way (every vocabulary member has an explanation), so
the extra keys cost nothing. The vocabulary removals themselves stand.

None.

Updated the tests asserting removed symbols. Deleted a test where it was the
only thing holding a symbol up, or where it duplicated a sibling — e.g.
`test_close_sequencer_does_not_overwrite_existing_stop_reason` asserted the
same behaviour as the `signal` case directly beneath it while carrying a stop
reason being removed, so it goes rather than being re-pointed at a live value.
Added `test_rewrite_takes_its_share_off_the_top` (pins that each lane's share
of the phase is unchanged by the reserve refactor) and
`test_response_is_not_a_valid_intent_type`.

**CI is green on the full suite** (12 shards across py3.10 / py3.11, both
coverage jobs, every static check). That run is the authoritative verification
here, because a full local run is impossible on the authoring machine: `fcntl`
is POSIX-only and the coordinator imports it, so 223 test modules fail to
collect on Windows — the same 223 on `origin/main`, unrelated to this branch.

The first CI run caught what local verification could not, and the two
follow-up commits are that fix. Seven raw `INSERT INTO events` statements name
the columns themselves instead of going through `Message.to_db_row()`, so
dropping the column left them writing one the schema no longer has
(`sqlite3.OperationalError`, 6 tests × 2 interpreters = 7 red shards). They sit
in `test_sources_local_probe.py` and `test_scaffold_smoke.py`; both files run
fine on Windows, but neither was in the set I had judged affected, and a search
for the *field's* readers does not reach a column name spelled inside a string
literal. Searching for `INSERT INTO events` instead found all seven at once.

**No, and getting there is worth a paragraph.** Dropping `events.priority` did
break `--resume`: `CREATE TABLE IF NOT EXISTS` does nothing to a database that
already has the table, so a resumed session kept a column declared `NOT NULL`
with no default and refused its first append with an opaque `IntegrityError`.
That is fail-open into a wall, which is neither shape this repo already uses
for a schema that moved. `ensure_schema` now drops the column in place and the
session continues with its event history and `seq` numbering intact.

The precedent that fits is state schema v6, migrated on load: nothing about
such a session is unresumable, only a column nobody reads is in the way. The
`EXPLORE` refusal is the other shape and does not apply — that phase named a
machine that no longer existed, so there was nothing to migrate to. Shipping
`DEFAULT 1` now and dropping the column a release later was also considered and
declined: it buys one release of resume parity at the cost of two releases
carrying a column nothing reads. `DROP COLUMN` needs SQLite 3.35; below that
the drop is refused with a message naming the session rather than the
constraint, so the migration is a strict superset of refusing outright.

One near-miss checked and cleared: `ClosePhase._derive_close_stop_reason` falls
back to `time_exhausted` when a `phase_history` reason is outside the
vocabulary, where `_global_terminal` passes unknown values through. That
asymmetry cannot bite here — the function only scans rows whose `to_phase` is
`CLOSE`, and it is reached only when `stop_reason` is blank, so none of the
removed strings is on a path a session can reach.

- PR addresses single concern: yes/no (details if no)

Yes — one audit's worth of dead-code removal. Split into reviewable commits
rather than six PRs because the batches share one argument and one closure scan,
so reviewing them apart would mean rebuilding the same context six times. Each
commit stands alone and is independently revertable.

N/A — all in-repo.


xiaofei-zheng (approved):

Paths I checked rather than took on trust:

**`events.priority` removal + the resume migration.** `git grep "INSERT INTO events"` across the branch: every remaining writer (`message_bus`, `resource_lock` ×2, `reconcile._TIMEOUT_DENY_SQL`, and the test literals) matches the new column list, and no reader of `Message.priority` or `row["priority"]` survives. No index, view or trigger names the column, so `ALTER TABLE ... DROP COLUMN` is legal. I built a pre-removal database by hand (old DDL with `priority INTEGER NOT NULL`, both `idx_events_*`, three rows) and ran `ensure_schema` over it: column gone, both indexes intact, `seq` continues at 4 rather than restarting, a subsequent 7-column INSERT succeeds, and a second `ensure_schema` is a no-op. The `sqlite_version_info < (3, 35, 0)` guard raises before the `ALTER`, so an old-SQLite host gets the named message instead of a syntax error.

**The lane refactor is share-neutral.** Compared old (`split_lanes(phase_sec, {rewrite .5, fusion .3, gemm .2})`) against new (`split_lanes(phase_sec - int(phase_sec * 0.5), {fusion .6, gemm .4})`) over `phase_sec` in `[0, 200000)`: zero cases where a lane loses a second, and at most +1s on odd `phase_sec` from the reserve rounding down. `split_lanes` now rejects `"rewrite"`, but `allocate` is only reached through `request_handlers._lane_budget`, whose two call sites pass `LANE_GEMM` and `LANE_FUSION` and never a custom `weights` mapping, so nothing can hit that error.

**The vocabulary removals have no live writers.** Grepped each of the 11 removed reasons as a string literal across non-test `src/`: the only hits are in `_STOP_REASON_EXPLANATIONS`, which is read through `.get(reason, "")` and is not asserted to mirror the vocabulary in either direction, so the deliberately-kept archived-session keys cost nothing. `enablement_stalled` appears elsewhere only as an enablement-timeline event reason, a separate namespace this PR does not touch — matches the description. `is_valid_phase_exit_reason`, `PHASE_EXIT_REASONS`, `phase_history_event_name`, `LIFECYCLE_STATUSES`, `is_action_allowed_in_phase`, `_action_in_phase_map`, `build_escalate`, `build_update_state` and `IntentSpec.builder` have zero remaining references in `src/` or the docs.

**`IntentType.RESPONSE`.** Removed in lockstep from the upstream enum, the robustness mirror, `_PAYLOAD_REQUIRED` on both sides, `_INTENT_DISPATCH`, the coordinator collaborator map and `PolicyGate.validate_intent`. `ROLE_SUBSCRIPTIONS` is only consumed by `MessageBus.replay_for`, which is called from `conversation.py` for the three LLM agents (`orchestration` / `critic` / `robustness`) — never for `kernel_agent` — so dropping that subscription changes no live inbox. An LLM emitting `"intent_type": "response"` now lands in the existing `IntentType(...)` `ValueError` branch in `mcp_emit_intent.validate_emit_intent_input` and comes back as a validation error, not a crash.

Ran `test_storage_schema_units.py` and `test_lane_budget.py` locally (60 passed). I did not run the full suite; CI is the authority there.

PR #1482, opened 2026-09-10
Migrate `hyperloom-workload-optimizer` skill to `AMD-AGI/Hyperloom` product repo

danielholanda:
# Description

This PR migrates `hyperloom-workload-optimizer` skill from `amd/skills` to `AMD-AGI/Hyperloom` product repo. The skill will continue to be part of the `amd/skills` catalog.

This enables `AMD-AGI/Hyperloom` to be the single source of truth (SSOT) for the `hyperloom-workload-optimizer` skill.

From now on `amd/skills` will simply mirror any changes made to this skill instead of requiring code to be changed at `amd/skills`.

PR #1515, opened 2026-09-15
feat(skills): add the catalog entry skill for Hyperloom bootstrap

lishuoshuo-amd:
# Description

`amd/skills` now federates every catalog skill from the product repo that owns it: the skill lives here, and the catalog imports a copy nightly. This adds `hyperloom-workload-optimizer` to this repo so that import has a source, plus the check that keeps it importable.

This is an alternative to #1482, which copies the catalog's current tree into this repo. Please take this one instead and close that.

Registering the path is [amd/skills#213](https://github.com/amd/skills/pull/213), which also brings the catalog copy in step with this one. **Merge this PR first** — the catalog's nightly importer clones `main` and raises `FileNotFoundError` on a path that is not there yet, and its own PR check reads the schema without cloning, so it would go green either way.

## What this PR adds

| Path | What it is |
| --- | --- |
| `examples/skills/hyperloom-workload-optimizer/SKILL.md` | the skill the catalog imports |
| `scripts/tests/test_federated_skill_contract.py` | four assertions holding it to the rules the catalog validates |
| `.github/workflows/federated-skill.yml` | runs that test on every pull request |

## The skill

Only the bootstrap: confirm the workspace, `pip install --target .`, run `/hyperloom-setup`, then hand the run to the skill that owns it.

Everything after setup already ships with the runtime, so it stays in step with the installed version by construction:

- `hyperloom-setup` — credentials, `USER_DATA_PATH`, run mode, Docker target host, bare-metal framework install
- the demo skills — a workload preset and the container lifecycle
- `inference_optimizer` — launcher gates, resume, monitoring

It is the agent-facing form of `examples/README.md`, which stays as the human quickstart.

### Placement

`examples/skills/hyperloom-workload-optimizer/`, beside the README it mirrors and one level up from the demo skills it hands off to.

Unlike those four demo skills, it has **no `data-files` entry** in `pyproject.toml`, and that is deliberate rather than an omission: a user follows this skill *before* the wheel exists, so shipping it in the wheel would only overwrite the copy they installed from the catalog.

### What it deliberately leaves out

The catalog copy was written for a workspace with no Hyperloom in it, so it carries its own `launch.sh`, `resume.sh`, `preflight.py` and their tests. Importing those would give this repo a second launch path that drifts from the CLI it wraps, so they are not here, and the skill says so rather than leaving it to the reader.

Also out: `skill-card.md` (the catalog synthesizes one on import, as it does for the federated TraceLens skill) and `evals/` (federation does not carry it in either direction — the datasets stay in `amd/skills`).

## The check, and why this repo needs one

The catalog validates the skill on its side, so without a check here a broken edit lands as a red bot pull request over there, where nobody on this side is watching.

Nothing already in this repo would have caught it either: the skill is markdown, and `lint.yml` and `tests-coverage.yml` both ignore `**/*.md`, so the one file that ships to the catalog was the one file no job read. `federated-skill.yml` therefore carries no `paths-ignore`, for the same reason `packaging.yml` carries none, and the job is seconds long.

It asserts what the catalog enforces: the frontmatter parses, `name` matches the directory, the description fits 1024 characters, and the body fits 500 lines. The description is the one with no room left — at 943 characters, a single added trigger sentence takes it to 1098.

The catalog's own `skillscope` harness is not reused here: its discovery step only treats a directory as a skill once an `evals/evals.json` sits beside `SKILL.md`, and federation deliberately leaves that dataset in `amd/skills`. Naming this workflow after the catalog's checks would also show a green tick for a harness that never ran, which is why it is named for the relationship it guards instead.

## Notes

- The `description` is kept byte-identical to the catalog's, because the catalog's routing cases are graded against it.
- `SKILL.md` matches the copy in amd/skills#213 apart from one line: the link out to `examples/README.md` is relative here and absolute there, which is the rewrite the importer performs.

## Test plan

- [x] `test_federated_skill_contract.py`: 4 passed, and each assertion was checked against the break it exists for — an over-long description (fails at 1098 characters), a `name` that no longer matches the directory, and a moved folder
- [x] `test_packaging_lint.py` passes (7 passed); no asset under `src/` changed, so nothing new to declare
- [x] `ruff format --check` and `ruff check` clean on the new test
- [x] Relative link to `examples/README.md` resolves from the skill's location
- [x] Structural limits: `name` matches the directory, description 943/1024 characters, body 127/500 lines
- [x] Branch merged with current `main`; 30 checks pass, the one skip is `update test durations (main)`, which only runs on `main`
- [ ] Reviewer check: the handoff names the run skills the wheel actually installs


xiaofei-zheng:
## What this PR does

Adds `examples/skills/hyperloom-workload-optimizer/SKILL.md` as the source of truth for the skill `amd/skills` federates nightly — bootstrap only (confirm workspace → `pip install --target .` → `/hyperloom-setup` → hand off to the wheel-installed run skills), with the catalog's `launch.sh`/`resume.sh`/`preflight.py` deliberately left out. It also adds a new CI gate (`.github/workflows/federated-skill.yml` + `scripts/tests/test_federated_skill_contract.py`) that checks the catalog's structural contract: file location, `name` matches the directory, description ≤ 1024 chars, body ≤ 500 lines.

## Blocking

**The PR description is out of sync with the diff.** Three points:

1. The body describes only the skill. The diff also adds `.github/workflows/federated-skill.yml` (a new CI job) and `scripts/tests/test_federated_skill_contract.py`; neither appears anywhere in the description, and the title (`add the catalog entry skill`) does not cover adding a CI gate. A reviewer reading the description reviews a different change than the one submitted.

2. Test plan: *"Markdown-only change, so `Lint` and `Tests with Coverage` are skipped by their `paths-ignore`"* — this is false. Both workflows ignore only `**/*.md`, `docs/**`, `LICENSE*`, `COPYRIGHT`, `**/CODEOWNERS`, `.gitattributes` (lint.yml, tests-coverage.yml), so the added `.py` and `.yml` make both run — and they did run on this PR (`ruff`, `pylint`, `bandit`, `test (py3.10 shard 1..6)`, `coverage (Python 3.10/3.11)` are all in the check list).

3. Placement section: *"`examples/skills/hyperloom-workload-optimizer/`, beside the README it mirrors and one level up from the demo skills it hands off to."* The README is `examples/README.md` and the demo skills are `examples/<name>/SKILL.md`, so the new file is one level **deeper** than the demo skills and not beside the README. The stated rationale describes the opposite of the actual path.

Please update the description (and the title, if the CI gate is meant to stay in this PR).

## What I checked

- Ran `pytest scripts/tests/test_federated_skill_contract.py` locally on the PR head: 4 passed.
- Re-derived the `paths-ignore` lists of `lint.yml` and `tests-coverage.yml` and cross-checked against the PR's actual check runs.
- Verified the four run skills named in the handoff (`hyperloom-qwen3-8b-3h`, `hyperloom-qwen3-14b-fp8-12h`, `hyperloom-qwen3-14b-fp8-12h-forge`, `hyperloom-custom-advanced`) each have a matching `[tool.setuptools.data-files]` entry in `pyproject.toml` — that closes the unchecked reviewer item in the test plan.
- Checked CHANGELOG convention against history: skill-markdown-only and `.github/workflows/*` additions in this repo do not carry `[Unreleased]` entries, so the absence of one here is consistent — not raising it.
- Diffed against `git merge-base origin/main` (bdbe6c10d) to confirm the above are introduced by this PR.

xiaofei-zheng (approved):
Approving. The code itself is sound — the contract test passes locally on the PR head (4 passed), the new workflow is correctly scoped, and the four run skills named in the handoff all have matching `[tool.setuptools.data-files]` entries.

Please still fix the description before merge as noted in my earlier comment: it does not mention the new CI workflow or contract test, the "Markdown-only change, so Lint and Tests with Coverage are skipped" claim is false (both ran on this PR), and the Placement rationale describes the opposite of the actual path.

PR #1516, opened 2026-09-15
fix: honor configured Ray versions during CLI preflight

iraj465:
An exported `RAY_VERSION` is honored by the kernel installer, but CLI preflight still requires Ray 2.44.1. A container with a different explicitly selected Ray release therefore gets a second incompatible install attempt during CLI startup. For example, Ray 2.44.1 has no Python 3.14 distribution even when the image already contains a working newer Ray.

Resolve the existing Ray and Click version overrides when preflight runs, and use the same values for installation, both smoke checks, and the result metadata. Unset or empty values keep the existing defaults. Exact Ray matching, the Click upper bound, and Ray CLI import checks remain enforced. This covers exported process environment variables; it does not add `.env` allowlist entries.

- Linked issue: none; reproduced during an exact-image optimizer bootstrap.
- Tests: 144 focused tests passed; an independent review ran 20 tests. The actual smoke probe passed with Python 3.14.7, Ray 2.58.0 and Click 8.2.1. Whole-tree Ruff, formatting and REUSE passed. Advisory mypy comparison has zero new errors (1,010 candidate versus 1,011 base).
- Full-suite validation: 22,186 passed, 49 skipped, 77 deselected, and five failures in the initial local environment. Three failures pass after correcting subprocess PATH/PYTHONPATH. The remaining two fail identically on base `0425bde3`: the deployed-mapper subprocess loads a stale editable install from site-packages, and the supervisor test assumes PID 123 is dead although it is a live kernel thread on this host. Both pass when the intended test isolation is supplied by a scoped external harness (no source changes). No affected production paths differ from base.
- Hosted validation: all 12 Python 3.10/3.11 test shards and both coverage gates passed; lint, typing, packaging, documentation, REUSE and security checks passed.
- Breaking changes: no; existing default versions are preserved.
- Single concern: yes, consistent Ray configuration across the two preflight paths.
- Upstream root cause: this mismatch is owned by Hyperloom; no component ticket is required.


chao-xu-spec:
The reported mismatch is real: Python 3.14 runs with Ray 2.58 attempted to reinstall 2.44.1 and failed because no wheel exists. However, this branch now conflicts with main’s Ray-version-floor implementation.

The override path must be integrated without removing main’s default floor, wheel fallback, actual-version metadata, or ROCm probe environment. Also, do not apply the default click<8.3 ceiling to newer Ray releases. Real runs with Ray 2.58 + Click 8.5 repeatedly downgraded Click to 8.2.1 and produced a dependency conflict with huggingface-hub>=8.4.2.

Please distinguish explicit exact RAY_VERSION from the default floor policy, apply the default Click ceiling only to Ray 2.44.1, and add combined regression tests after rebasing.

PR #1522, opened 2026-09-16
fix(kernel): align Forge KEEP state and breakdown attribution

haishuok0525:
## Summary
- route Forge controller KEEPs through the canonical session writeback in every benchmark mode
- preserve Forge backend provenance so stack attribution lands in the kernel/forge bucket
- cover the framework-then-kernel KEEP sequence that previously left session_breakdown behind state.json

## Test plan
- [x] Run controller integration and SBD stack ledger tests (52 passed)
- [x] Run Ruff checks on changed files
- [x] Run Ruff format check on changed files


xiaofei-zheng:
## What this PR does

`_run_kernel_rewrite_controller` used to hand `integrate_controller_patches` a `record_keep` callback only when `agentx_active(shared_state)` was true; every other benchmark mode fell back to the module-local `_record_keep`, which writes `optimization_stack` / `current_best` / `cumulative_gain_validated` directly. That local writer bypasses `_lift_to_current_best`, which is the only caller of `stack_event.record_adoption`, so a Forge controller KEEP in synthetic mode landed in `state.json` but never produced a stack-ledger adoption row — `session_breakdown` ended up behind `state.json`.

The fix drops the `agentx_active` gate so the single production call site always passes `WritebackCollaborator._record_integrate_keep`. `_keep_result` now also carries `backend: "forge"` / `engine: "kernel_rewrite_controller"`, and `_record_integrate_keep` forwards `backend`/`engine` from the result into the stack entry, so the adoption row lands in `by_source.kernel.by_backend.forge` instead of folding into `unattributed`. A new test drives a synthetic-mode framework-then-kernel KEEP sequence end to end and checks both `state.json` and the assembled ledger.

Verified by reading the code rather than re-running the suite: the `backend`/`engine` pass-through is safe for the other `_record_integrate_keep` callers (the `integrate_handler` results reaching `kernel_stack.py` and `intent_router.py` carry neither key, so the new entries are `None`, and every consumer — `_by_source` in `stack_event.py`, `phases/kernel.py:530` — coerces with `or ""`); the fusion branch still wins the key collision because it is spread last; and the anchor stays consistent across operators because `_fill_integrate_defaults_from_state` reloads `current_best` from disk after each `shared_state.save`. Tests were not run locally (`kernelforge.loop.editable_repo` imports `fcntl`, Linux-only); relying on CI.

## Blocking

1. `src/hyperloom/orchestrator/kernel/controller_patch_integration.py:282` — the docstring is now wrong:

   > `record_keep: Session-owned writeback for AgentX; other workloads use the local recorder.`

   After this PR every workload uses the session writeback; the only production call site passes it unconditionally. The `record_keep is None` branch at line 495 and the `_record_keep` helper at line 188 are unreachable outside tests. Either delete the fallback and make the parameter required, or at minimum update the docstring — as written it documents exactly the behaviour this PR removes, and a future caller that omits the argument silently gets back the no-ledger path this PR is fixing.

2. No `CHANGELOG.md` entry. This changes what `session_breakdown` reports for every non-AgentX run: the kernel bucket and its `by_backend.forge` split go from empty to populated, and `validated_total_gain_pct` / `at_head` now reflect controller KEEPs. That is user-visible output, and `docs/contributing/style-guide.md:176` asks for a note under `## [Unreleased]`.

haishuok0525:
Both blocking points are addressed in ba1050436.

**1. The docstring / unreachable fallback.** Took the stronger option rather than only fixing the prose: `record_keep` is now a required argument, and the `record_keep is None` branch plus the module-local `_record_keep` are deleted. `phases/kernel.py` is the only production caller and it already passed the writeback unconditionally, so nothing else had to change. The docstring now says what the parameter is, i.e. the session-owned writeback every KEEP is recorded through.

The tests that reached the fallback (21 of them) now call `_integrate`, the helper already in the module that binds the Coordinator's own `_record_integrate_keep`. That also makes the assertions stricter than before, since they now observe what the live loop writes: `test_multiple_patches_are_kept_and_committed_one_by_one` and `test_a_commit_that_reports_no_change_is_reverted_and_the_next_one_still_lands` are checking the stack the lift produced, not the one the local writer produced.

One thing that fell out of it: with every test now constructing a Coordinator, an unset `USER_DATA_PATH` resolves to the shared `/workspace/hyperloom` default, so the module sets it under `tmp_path` in an autouse fixture. Four tests were already doing this individually.

**2. CHANGELOG.** Added under `## [Unreleased]` → `### Fixed`, framed on the output change you named: for a non-AgentX run the kernel bucket and its `by_backend.forge` split go from empty to populated, and `validated_total_gain_pct` / `at_head` now account for controller KEEPs.

Verified in a container with the repo's own ruff config: 28 passed in `test_controller_patch_integration.py`, and 270 passed across `orchestrator/kernel/tests` plus `test_integrate_controller_preapplied_patch.py`. `ruff check` and `ruff format --check` clean on both touched files.

xiaofei-zheng (approved):
Both blocking items from the earlier round are addressed, and the follow-up commits hold up on a re-read.

`_record_keep` and the `record_keep is None` branch are gone; `record_keep` is now a required parameter and its docstring describes what it actually is. I checked every call site on the branch (`phases/kernel.py:3907` plus the two in the test module) — all pass it, so the signature change is complete. The test suite now goes through `_integrate`, which binds the Coordinator's real writeback; `WritebackCollaborator.__getattr__` delegates to the coordinator, so reassigning `coordinator.shared_state` does rebind `writeback.shared_state` and the SharedState assertions in those tests still observe what the live loop would write. Dropping the `measurement` kwarg from the `record_validation` call matches the signature on current `main`.

The CHANGELOG entry lands under `## [Unreleased]` → `### Fixed` and matches the prose style of the neighbouring entries. The duplicate `### Fixed` heading under `[Unreleased]` is pre-existing on the merge base, not something this PR introduced.

Also re-checked on the merged tree: the `backend`/`engine` pass-through is inert for the other `_record_integrate_keep` callers (`kernel_stack.py`, `intent_router.py` — `integrate_handler` sets neither key, and `_by_source` / `phases/kernel.py:530` coerce with `or ""`); the fusion branch still wins the key collision; no consumer keys on the stack entry's old `source_file` field or the `kernel_rewrite_controller:` variant-name prefix that the removed local recorder used to write; and the source-layer recipe export reads `scope` / `base_sha` / `name`, all of which the writeback entry carries.

Tests not run locally (`kernelforge.loop.editable_repo` imports `fcntl`, Linux-only) — relying on CI.

No blocking issues.

PR #1525, opened 2026-09-16
Revert "fix(kb): guard warm-start replay on the shape the canonical_id cannot express"

ZhengGong-amd:
Reverts AMD-AGI/Hyperloom#1495


rpoornac:
For the record: re-landed as #1542 now that 1.1.1 is cut, per @ZhengGong-amd's recommendation. Reverting the revert onto current main, so the content is unchanged from what was reviewed.

PR #1526, opened 2026-09-16
fix(roofline): classify capture failures instead of retrying degraded

haishuok0525:
Resubmission of #1511, which was reverted in #1524. This branch reapplies the same change on top of current `main`; the diff is identical to the reverted commit. Original description follows.

---

- **Description**: what and why

  Roofline measures an already-validated workload, so a CUDA-graph capture failure there is evidence about that workload, not a cue to change its configuration. The eager fallback did change it: on capture failure it silently re-ran with a degraded config and reported the degraded number as the roofline. The failure disappeared, its cause was never recorded, and the resulting ceiling was wrong in a way nothing downstream could detect.

  Roofline now classifies the failure and stops. Classification separates an **instrumentation** collision (profiler contending with capture, our own doing) from a **config** incompatibility (hardware or framework cannot capture this shape), which is the distinction that decides who fixes it. The evidence goes to a diagnosis artifact next to the run.

  Stopping is only useful if the report says enough to act on, and it did not: a reader could not tell a broken profile from a broken workload. So this also reports the facts that settle that question.

  Preflight, asserted before the measurement runs:
  - orphan reap results (what was left behind by a prior run, and what we removed)
  - disk headroom and stale trace artifacts (capture writes gigabytes; a full filesystem looks exactly like a capture failure)
  - an instrumentation assertion carrying patcher outcomes, so a missing patch is visible before it costs a run

  Post-flight, measured from what the run produced:
  - capture sidecar operator-metadata coverage, measured by opening the sidecars rather than inferring from filenames (up to 24 scanned, with partial-scan bounds recorded)
  - metric self-description: every ratio carries its numerator, denominator and aggregation scope, so `attributed_pct` cannot be read against the wrong basis
  - a verdict with continuous measures and a severity rank (`blocked` > `silently_wrong` > `warn` > `ok`), replacing a pass/fail that could not express "the number exists but do not trust it"
  - a server liveness probe per attempt, separating "the server died" from "the capture failed"

  Patch outcomes previously reached the session breakdown only through `trace_validate.checks`, which does not exist on a run that produced no trace — exactly the run whose patch state you need. They now ride on the per-attempt run row and are drained on every path, success or failure.

  On volume: the full certificate with its per-rank tables and candidate inventories is written to `selfcert.json` on disk (82 KB on the verification run). The session breakdown keeps the ~40 scalars and a path to it, 8.2% of the certificate's size.

  Two bugs surfaced while verifying this against real traces:
  - the cpu_op structure check searched for `{"name": "cpu_op"}`, which never appears in Kineto output — the category lives in `"cat"`. On a real sidecar the searched form matched 0 events while the correct form matched 2915, all carrying Input Dims. The check had been reporting "no cpu_op events" on fully instrumented captures for its entire lifetime, masked by fixtures that set both keys.
  - `HYPERLOOM_TRACELENS_PATCH_STATUS` was written only on failure, so "never attempted" and "attempted and failed" were indistinguishable. It is now written in every branch as `ok` / `unavailable` / `not_attempted`.

- **Linked issue(s)**: none

- **Tests**: added/updated.

  New: `test_trace_selfcert_observability.py` (8 tests — sidecar scanning, partial-scan bounds, rank-level basis transparency, verdict severity). Extended: `test_roofline_executor_units.py` (non-retryable failure still produces a run row; instrumentation draining), `test_sbd_v6_roofline_timeline.py` (scalars inlined, lists excluded), `test_profile_and_kernel_handlers.py` (instrumentation preflight; fixtures corrected from the synthetic dual-key form to real Kineto `{"cat": "cpu_op", "name": "aten::mm"}`).

  Commands run — `pytest` over the touched modules and their neighbours on the merged tree: 1775 passed, 5 skipped, 0 failed. Full-suite run before the merge showed no new failures (the 20 failures are pre-existing credential/auth cases unrelated to this path).

  Verified end to end on the cluster against sglang 0.5.19 (graphattr), a workload that actually captures. The roofline action succeeded, produced 53 capture files, and every observable above came back with real values: `orphans_reaped_count=0`, `free_pct=6.05`, `stale_trace_count=0`, `tracelens_patch_status=unavailable` with six patchers true (the mirror image already carries the patch), `capture_op_meta_coverage=1.0` over 52 files with 24 scanned, `attributed_pct=1.77` beside `attributed_gpu_ms=10.89` and its scope, `graph_launch_coverage=1.0`, `severity=ok`, liveness `{"pidfiles": 0}`. An earlier attempt on vLLM 0.24.0 could not generate captures at all (it rejects the profiler flags), which is what sent the verification to sglang.

- **Breaking changes**: no. The removed fallback only ever produced a misreported roofline; callers that relied on roofline returning a number for an uncapturable workload now get a classified failure instead, which is the intent.

- **PR addresses single concern**: yes — a capture failure in roofline should be reported rather than worked around, and reporting it requires the facts to be present.


xiaofei-zheng:
Reviewed against `git merge-base origin/main HEAD` = `0af1c0969`; the diff is identical to the reverted #1511, as the description says.

Two blocking items.

**1. No CHANGELOG entry for a user-visible behavior change.**

The repo keeps an actively maintained `## [Unreleased]` section and comparable `fix(...)` PRs all add to it. This PR removes the eager retry, adds `profile_cuda_graph_capture_config_failed` / `profile_cuda_graph_capture_instrumentation_failed`, and renames the SBD field `eager_fallback_applied` -> `graph_capture_disabled`. `grep -i "eager\|graph_capture_disabled" CHANGELOG.md` returns nothing for this change. An operator whose roofline action starts failing after upgrade has no record of why. (The rename itself is complete — no stale `eager_fallback_applied` reference remains anywhere in the tree.)

**2. `_server_liveness_probe` does an unbounded blocking filesystem walk on the event loop.**

`roofline.py:674` — `_note_profile_run` is a sync closure called from the async `_execute` (`roofline.py:514`), on every attempt, on all paths including success. `_server_liveness_probe` (`roofline.py:390`) does `sorted((session_dir / "runs").rglob("*.pid"))`, which traverses the entire `runs/` tree — the same tree that holds the multi-GB trace output this action produces, typically on network storage — plus a `/proc/<pid>` read per live pid. The loop is stalled for the whole walk, blocking every concurrently scheduled task and the bus heartbeat.

The same PR already establishes the correct pattern two dozen lines below: `roofline.py:694` wraps the comparable `_preflight_probe` in `await asyncio.to_thread(...)`. `_trace_dir_inventory` (`roofline.py:301`, called at `roofline.py:637`) has the same problem — it materializes `sorted(p for p in root.rglob("*") if p.is_file())` over the whole profile workspace to report the first 200 entries — though it only runs once on a terminal failure path.

Checked and found clean:
- `_preflight_probe`'s stale-trace glob `runs/*/<task_id>`: the profile child inherits `ctx.extra["workspace"]` from the parent roofline ctx (`_wrap_profile_ctx`, `roofline.py:1525`), which `sub_agent_runner._pre_mkdir_workspace` set to `runs/roofline/<task_id>`, so the glob does cover where traces actually land.
- `certify_trace_dir`'s capture-only early return not passing `capture_op_meta_coverage` to `build_verdict` (`trace_selfcert.py:1060`): unreachable. `selected` is empty only when `_trace_candidates` is empty, and `_select_trace_file` always returns a file for a non-empty candidate list, so the probe dict and the early return cannot both be populated. Verified by running `certify_trace_dir` on a capture-only directory: `measures.capture_op_meta_coverage = 1.0`.
- `measures.idle_pct_threshold` falling back on a truthy-but-incomplete `idle` dict: every `idle_gate` exit sets `idle_pct_threshold_effective`, including the `no_window` branch.
- `graph_capture_disabled` rename: no remaining readers of the old key.

xiaofei-zheng (approved):
Re-reviewed at `15e70e4` (merged with `main` at `1173611`). Both blocking items are addressed.

**Event-loop stalls** — `_note_profile_run` and `_fail_capture` are now coroutines; `_server_liveness_probe` (roofline.py:674) and `_trace_dir_inventory` (roofline.py:637) run under `asyncio.to_thread`, matching the `_preflight_probe` call at roofline.py:696. Checked every call site of both closures — all 13 are awaited (724, 737, 758, 783, 802, 818, 839, 863, 885, 906, 921, 1288, 1299), including the two inside `except` blocks and the two whose return value feeds `recorder.adopt_profile_run`. No un-awaited coroutine left behind, and the probe helpers themselves stay synchronous, so nothing else calling them changes shape.

**CHANGELOG** — entry added under `## [Unreleased]` / `### Changed` (CHANGELOG.md:59), describing the classification behavior, the diagnosis artifact, the `eager_fallback_applied` -> `graph_capture_disabled` rename, and the `to_thread` move. Matches the diff.

No blocking issues. Note on verification depth: I could not execute the roofline test files locally (Windows host, `recipe_kb.local_store` imports `fcntl`), so the async conversion is verified by reading every call site rather than by running the suite — CI covers the rest.

LGTM.

PR #1527, opened 2026-09-16
feat(sbd-v6): declare the axis a session graded on, and grade the conc sweep on it too

haishuok0525:
Resubmission of #1520, which was reverted in #1523. This branch reapplies that change on top of current `main`, plus two commits that close the grading path it left open -- see [Fail closed](#fail-closed-when-the-axis-pair-is-missing).

---

Reopens the work from #1480, rebased onto current `main`. That PR was closed by request with its commits kept on the branch; this is that branch, brought forward. One conflict, an import line in `conc_sweep.py`, resolved by keeping both sides.

I re-checked the three defects it claims to fix against today's `main` before reopening, because #1380 and #1445 were both invalidated by `main` moving underneath them. All three are still live -- see [Still current](#still-current) at the end.

## Why

Nothing in `session_breakdown.json` said which axis its gains were on. An AgentX replay is ranked on the slow-tail interactivity percentile (`e2e_norm_intvty_p90`) with total throughput held as a guard; a synthetic run is ranked on output throughput alone. Every throughput field in the breakdown is the output axis by construction, and `benchmark_mode` reaches `reports/final.json` but never this document.

A consumer therefore had no way to keep an AgentX session out of a synthetic leaderboard. It would sort one against the other and every number would look plausible -- on the canonical corpus the two axes differ by roughly two orders of magnitude.

## The two blocks, and why they are two

```json
"metadata": {
  "grading": {
    "benchmark_mode": "agentx",
    "objective": "e2e_norm_intvty_p90",
    "tput_guard": {"enabled": true, "noise_pct": 3.5}
  }
}
```

`metadata.grading` is what the session was **configured** to rank on. `outcome.validation.graded_on` is what the run **actually decided its last promotion on**, read off that promotion. On a session that promoted anything the two now agree, because a comparison that cannot supply the configured axis pair fails rather than settling for output.

Neither resolves the other even so, deliberately. A session can be configured for an axis and promote nothing on it. Publishing the configured axis in `outcome` would put an interactivity label on a figure no promotion produced; resolving `metadata` from `outcome` would erase what the session asked for. There is a test on each direction.

## Recorded, not re-derived

Resolution reads `HYPERLOOM_PERF_METRIC` / `HYPERLOOM_PERF_NOISE_PCT`. Every later reader of those is somewhere the environment is not evidence: a resume is a new process, a re-baseline runs in a subprocess, and CLOSE frequently drives this export from one that never inherited them. So the axis and its band are resolved once at seed and recorded in `SharedState.grading`, beside `benchmark_mode` and `agentx_epoch`, which live there for the same reason.

This fixes a live defect rather than only serving the export. `resolve_graded_comparison` re-derives on **every** comparison, and the recorded band now reaches `passes_intvty_gate` and `passes_tput_guard` -- both already accept `noise_pct`, and neither call site passes it, so a lost `HYPERLOOM_PERF_NOISE_PCT` silently widens a configured 3.5% guard band back to the 5% default. The KEEP/REVERT rule has to be the one the session started with.

The same recorded block now settles what the AgentX `workload_spec` hands to GEAK: `metric_basis` and `intvty_p90_veto_pct` come from `SharedState.grading` wherever the materializing caller holds the live state (baseline, explore, the conc sweep), instead of being re-read from the environment of whichever process rendered the recipe. Callers that cannot reach the state pass nothing and keep the environment-derived fallback.

Sessions seeded before the field derive as before and report a **null** band: the band they actually applied was never recorded, and today's default is not evidence of it.

## Fail closed when the axis pair is missing

Two commits on top of the resubmitted change, and the reason the PR touches the orchestrator's decision path rather than only the export.

Declaring the axis is worth nothing if a comparison that cannot measure it quietly promotes on another one. It could: when either side of an AgentX comparison was missing the axis pair, the lanes fell through to the output-throughput figures and a large enough output gain read as a KEEP. The breakdown would then publish `objective: e2e_norm_intvty_p90` over a stack whose rungs were taken on output -- the label the first half of this PR exists to make trustworthy, on a number it does not describe.

`resolve_graded_comparison` is the chokepoint every lane's KEEP passes through, so the rule is enforced there: a degraded pair returns **REVERT**, never KEEP. The output figures still travel on the comparison, and `degrade_reason` still names why the objective could not apply, but they are diagnostics now -- explore, `_lift_to_current_best`, stack validation and the integrate handler all refuse the promotion instead of taking it on throughput. The round-local `ANCHOR_DEGRADED` output fallback was the same fallback in miniature and is gone with it.

One consequence worth flagging for review: with no fallback left, an anchor that cannot supply the interactivity axis makes **every** subsequent variant FAIL, and today that is visible only as a `log.info` -- there is no `stop_reason` and nothing in the breakdown that names the cause. I have not established how reachable that is on a real configuration; the only evidence I have that the axis can go missing is the existence of the fallback code I removed, which is not evidence that it does. Flagging rather than fixing speculatively.

## The conc sweep, grading on the same axis

The sweep resolved its own axis through `graded_metric_key`, which diverges from the KEEP path two ways now that KEEP reads the axis recorded at seed.

A resume whose shell lost `HYPERLOOM_PERF_METRIC` would redraw the curve on output throughput while the promotions in the same session were decided on interactivity. And `graded_metric_key` has no scriptable carve-out where `intvty_serving_grading_enabled` does, so an image framework -- which reports no interactivity axis at all -- would rank every rung on a field it never measures and report the whole sweep as failed. Both call sites now go through `resolved_grading`.

Each pair also carries the throughput axis beside the objective, and whether it held within the session's band. **Reported, not enforced:** InferenceX publishes a 2-D frontier with no fixed interactivity target, so a rung that traded throughput for interactivity moved along that frontier rather than violating a constraint -- and drawing that frontier is what a sweep is for. Gating here would drop half the curve. `best_conc` stays the best rung on the objective alone, with `best_conc_guard_holds` saying whether the session's own KEEP rule would have taken it.

The pair columns are renamed off throughput accordingly: `baseline_tput` / `optimized_tput` held an interactivity percentile whenever the session graded on one. They are now `*_value`, with `result.metric` naming the axis. Also fixes `V6ConcSweepPoint` declaring `intvty_p90` where the recorded curve has always written `e2e_norm_intvty_p90`.

## The axes themselves

`outcome.baseline.perf`, `outcome.final.perf` and `outcome.validation.perf` carry the four AgentX axes, recorded **where the measurement is taken** rather than read off `state.baseline_perf` and `current_best` at export. A second source for one baseline is a second answer to the same question, and a revalidation moves the cumulative gain without re-promoting the recipe, so `current_best` can be a later measurement than the gain published beside it.

An unmeasured axis is an **explicit null**. Absent would be indistinguishable from an axis the framework failed to report, and zero reads as "measured, and it was zero", so a synthetic run publishes four nulls. TPOT p90 was already parsed out of the benchmark report and dropped at five call sites; it now lands.

## The reconciliation stays single-axis

`outcome.validation.graded_on` names the axis the stack reconciliation is on, because an attributed figure on one axis against an unattributed figure on another makes the gap meaningless.

Every adoption in that sum is on the named axis by construction, and that is a property of the fail-closed rule rather than something the export has to check: a comparison that cannot supply the configured axis pair returns REVERT, and a REVERT never reaches the adoption writeback. `validation.notes` therefore has no off-axis finding in it -- an earlier revision of this branch carried one, and it could not fire.

## Still current

Verified against `main` at `ff7c235e4` before reopening:

| Claim | State on `main` today |
| --- | --- |
| `resolve_graded_comparison` re-derives the axis from the environment on every call | Still does -- `shared_state.py` calls `intvty_serving_grading_enabled(...)` per comparison |
| The configured noise band never reaches the guards | Still true -- both call sites pass no `noise_pct`, so both fall back to `parse_intvty_noise_pct()` |
| The conc sweep resolves its axis separately, with no scriptable carve-out | Still true -- `conc_sweep.py` calls `graded_metric_key` at two sites; it has no `scriptable` parameter |

## Rollout

**One breaking rename, in the `conc_sweep` timeline rows:** `baseline_throughput` → `baseline_value` and `optimized_throughput` → `optimized_value`. No alias is kept, because an alias would preserve exactly the misreading the rename exists to stop -- on an AgentX session those keys held a latency percentile in milliseconds. An external parser of `reports/sbd_v6/` has to be updated; the CHANGELOG entry carries the same warning.

Everything else is additive. `throughput_tok_s_per_gpu`, `output_throughput` and every other existing key keep their output-axis meaning, so nothing downstream has to change to keep working.

The behavioural change is not additive, and is the point: an AgentX session that could previously land a KEEP on output throughput when the interactivity axis was unavailable will now FAIL that variant instead. Expect fewer promotions on such a session, not different ones.

Cortex has no column for the interactivity or total axes yet, so the immediate value is that an AgentX session becomes *identifiable* -- the prerequisite for keeping it off a leaderboard it is not comparable on.

## Test

`test_sbd_v6_grading.py` covers both axes at seed, the scriptable carve-out, recorded-beats-environment in both directions, the null band on a legacy session, the block surviving the recorder and the overlay end to end, the metadata/outcome separation locked in both directions, the axis fallback for a session that adopted but never validated, and the explicit nulls. Plus the conc-sweep cases for the renamed pair columns and the guard column.

The fail-closed rule is pinned on the refusal rather than on a fallback: `test_a_degraded_agentx_lift_is_refused` asserts the lift returns False and writes no adoption row, and the degraded cases in `test_report.py` and `test_sweep_phase_auto.py` now assert REVERT where they previously asserted a throughput KEEP.

`ruff check src/hyperloom` is clean. The suites import `fcntl` transitively, so they are POSIX-only; I ran the breakdown, grading, conc-sweep, workload-envs and AgentX suites on Linux (1048 + 635 passing) and CI is the check for the rest.


xiaofei-zheng:
Reviewed the full diff against `git merge-base origin/main HEAD` = `0af1c0969`. Four blocking issues.

**1. The new degrade finding can never fire on the production path** — `collectors/v6.py:585`

`_degraded_adoptions` reads `degrade_reason` off the adoption rows, but the only production writer is `writeback.py:3357`, and `writeback.py:3120` returns early on `if not graded.comparable` — `comparable` is `not degrade_reason` (`perf_metric.py:244`), and that early return predates this PR (#1447). So `record_adoption` is only ever reached with `degrade_reason == ""`, `_degraded_adoptions` always returns `{}`, and the note never appears. The one test covering it (`test_the_notes_name_adoptions_that_fell_off_the_configured_axis`) calls `record_adoption` directly with a hand-written reason, so it passes while the real path cannot produce one.

This is the same argument the PR body uses to justify *not* putting `degrade_reason` on the validation row — it applies to adoptions too.

**2. `resolved_grading` was wired into two call sites; explore still derives the axis from the environment** — `actions/executors/explore.py:876`

`explore` decides whether to maintain `running_base_perf` from `intvty_serving_grading_enabled(...)`, while `resolve_graded_comparison` now resolves the axis from the recorded block. When the two disagree — recorded objective `e2e_norm_intvty_p90`, current process with `HYPERLOOM_PERF_METRIC` explicitly set to something else — explore passes `anchor_perf=None` and the comparison falls back to `resolve_grading_anchor_perf(state)`. `current_best` does not advance inside a round (winners are written back after the action), so every variant stacking in that round is graded against the round's opening state instead of the running anchor — the over-KEEP the comments at `explore.py:1246` and `shared_state.py:134` exist to prevent. Before this PR both sites derived identically and could not disagree.

Two more readers of the same config were left on the environment and can now contradict the recorded values the guards enforce:
- `actions/executors/report.py:423` — `final.md` prints `parse_intvty_noise_pct()`, not the band the session recorded and actually enforced.
- `actions/executors/_workload_envs.py:346` — `intvty_p90_veto_pct` handed to the harness is still read from the environment.

**3. No CHANGELOG entry**

CHANGELOG.md is kept current in this repo (#1462, #1472, #1476, #1510 each added one), and this PR changes the published wire shape: new `metadata.grading`, new `outcome.*.perf` / `graded_on`, and the conc-sweep pair columns renamed `baseline_throughput`/`optimized_throughput` -> `*_value`. Nothing in the diff touches CHANGELOG.md.

**4. Two spots not synced with the change**

- `docs/reference/session-breakdown.md:449` — the prose added at line 226 says "`baseline.perf` and `final.perf` carry the four AgentX axes", and the worked example was updated for `baseline.perf` only. The `outcome.final` block in the example still has neither `perf` nor `graded_on`, while `v6.py:653` emits both unconditionally.
- `recorder/conc_sweep_event.py:278` — the `record_workload` docstring still names `intvty_p90`, which this PR renames to `e2e_norm_intvty_p90` at `schema.py:589`. The identically worded comment at `conc_sweep.py:1618` was updated, so the two now disagree.

---

What was checked: the `record_adoption` call chain end to end; `resolve_graded_comparison`'s two branches against every `anchor_perf`/`anchor_tput` combination explore passes; every remaining reader of `intvty_serving_grading_enabled` / `graded_metric_key` / `parse_intvty_noise_pct` under `src/hyperloom`; the schema renames against in-repo consumers (none outside the TypedDicts); the docs diff against what the collector emits. `conc_pair_comparison` was executed directly to confirm the new guard columns. Full suites were not run locally (they import `fcntl`) — CI is the check for those.

Two things I looked at and am *not* raising: `perf_snapshot_from_mapping` derives `total = input + output` where `graded_axes_of` does not, so a measurement graded on a derived total would publish `total_throughput: null` and `guard_holds: null` — but `agentx/mapping.py:74` already coalesces the same way, so no real parser reaches it. And `prompt_builder.py:165` still uses `graded_metric_key` with no scriptable carve-out — pre-existing, not introduced here.

Also worth noting: the branch is 10 commits behind `origin/main` (`58421b04d`), so the body's "on top of current `main`" is stale; #1522 in particular touches breakdown attribution.

xiaofei-zheng:
Re-reviewed at `0525e40ee` (fix commit `2528be1ac`, on top of the `main` merge `a4c1bea70`).

Three of the four earlier points are resolved: `explore` and the report path now read `resolved_grading` / the recorded block, GEAK axis selection is threaded through `state.grading` at both `kernel.py` call sites and in `request_handlers.py`, the docs example gained `final.perf` / `final.graded_on`, and the `conc_sweep_event` docstring was corrected.

Before the list: my earlier point 1 was that the degrade note can never fire on the production path. That was a statement of fact, not a request to make it fire, and the fix took it the other way. Correcting that now, because the rest of this comment follows from it — a session the user launched on AgentX must be graded on AgentX. A measurement that cannot supply the interactivity axis is a failed measurement, not a measurement to re-rank on a different axis. The fail-closed gate was the correct behaviour and should stay.

Five blocking issues.

**1. Opening the `comparable` gate on the lift path is the wrong direction, and it degrades the session's axis in a cascade** — `loop/writeback.py:3131`

Replacing `if not graded.comparable: return False` with a log means a winner whose own measurement lacks the interactivity axis (`degrade_reason == "candidate_axes_missing"`) now promotes on output throughput instead. In an AgentX session that is a config accepted on an axis the user did not ask to be ranked on, and it does not stay contained to one adoption. The promotion ends at `writeback.py:3399`, `current_best.update(graded_axes_of(cand_source))`, which only stamps the axes the winner actually carried — so the new `current_best` has no `e2e_norm_intvty_p90`, `resolve_grading_anchor_perf` returns `current_best_axes_missing` from then on, and the whole session grades on output throughput until some later winner happens to carry all four axes. The comment directly above that line warns about exactly this ("the session degrades to output grading permanently after the first KEEP").

Reproduced against this commit:

```
anchor before: ({'e2e_norm_intvty_p90': 30.0, 'total_throughput': 20000.0, ...}, '')
degrade_reason= 'candidate_axes_missing' objective= output_throughput ref= 1000.0 cand= 1100.0
passes lift gate?  True
anchor after : (None, 'current_best_axes_missing')
```

While degraded, `AGENTX_KEEP_THRESHOLD_FLOOR_PCT` is bypassed too: the output branch's only admission check here is `graded.candidate <= graded.reference`, so a +0.1% output move overwrites `current_best`.

The lift path is also now the only `.comparable` reader that fails open. `_update_cumulative_gain_validated` (`writeback.py:857`) still returns early on the same condition, so a degraded winner is appended to `optimization_stack` and `gain_per_stack_entry` while the validation watermark is held — the ledger gains a step the reconciliation can never account for, which is the opposite of what `validation.notes` is trying to report.

The new test covers `current_best_axes_missing`, where the winner carries all four axes and the anchor recovers on promotion — the safe branch. `candidate_axes_missing` is the one that cascades and is untested.

Suggested resolution: restore the fail-closed return, and drop `_degraded_adoptions` and its note from `collectors/v6.py` rather than keeping a field for a state the orchestrator should not reach. `metadata.grading`, `outcome.*.perf`, `graded_on` and the conc-sweep axis work all stand on their own without it.

**2. The `grading` argument added to `build_agentx_workload_spec` is not passed at its only production call site** — `actions/executors/_workload_envs.py:490`

```python
bench["workload_spec"] = build_agentx_workload_spec(bench, envs, model_path=model_path, env=_agentx_env)
```

No `grading=`, so `intvty_p90_veto_pct` still falls back to `parse_intvty_noise_pct()` and `metric_basis` to the environment-derived `intvty_grading_enabled`. That part of the fix does not take effect. (The two `geak_metric_axis` call sites in `phases/kernel.py` and `kernel/request_handlers.py` do pass it.)

**3. `ruff check src/hyperloom` fails**

```
F401 `...state.shared_state.framework_is_scriptable` imported but unused
  --> src/hyperloom/orchestrator/actions/executors/explore.py:41:5
```

Left behind by the switch to `resolved_grading`. The PR body's "`ruff check` is clean on `src/hyperloom`" no longer holds.

**4. The CHANGELOG entry covers only the degrade change, not the PR's own wire-shape change**

The new entry describes `_lift_to_current_best` — and if issue 1 is taken, it describes a change that is being reverted. The change this PR exists to make — new `metadata.grading`, `outcome.validation.graded_on`, `outcome.*.perf`, and the conc-sweep pair columns renamed `baseline_throughput`/`optimized_throughput` -> `*_value` — still has no entry. The rename in particular is a break for any external consumer of the sweep pairs; there are none left in-repo, so the changelog is the only notice.

**5. `_lift_to_current_best`'s docstring and the PR description were not updated for the gate change** — `loop/writeback.py:3121`

The docstring still says it returns "``False`` when a performance winner was not comparable or did not beat the anchor". Not-comparable no longer returns `False`. The PR description likewise does not mention that this branch now promotes degraded winners — that is a behaviour change to the KEEP path, not a recorder fix, and the "Degrade is a finding, not a field" section reads as though the gate were untouched. Moot if issue 1 is taken by reverting.

---

Checked: the full fix diff; every remaining reader of `intvty_serving_grading_enabled` / `graded_metric_key` / `parse_intvty_noise_pct`; all 18 `.comparable` call sites (the fail-closed readers in `integrate_patch.py`, `kernel_stack.py`, `integrate_performance.py` and `writeback.py:857` are unchanged, so the lift path is the only one that fails open); the `record_validation` caller, which does still gate on `comparable`, so the `stack_event` / `schema` docstrings claiming that remain accurate; `SharedState.grading` round-trips through `asdict` / `fields()` so it survives resume; `seed_grading` and `SharedState.framework` both read `os.environ["FRAMEWORK"]`, so they cannot disagree; the renamed conc-sweep pair fields against in-repo consumers (none); `ruff check src/hyperloom`. Issue 1 was reproduced by driving `resolve_graded_comparison` / `graded_axes_of` / `resolve_grading_anchor_perf` directly. Full suites still not run locally (`fcntl`).

xiaofei-zheng (approved):
No blocking issues. LGTM.

The fail-closed rework in `dbaefe4f2` / `9d8333907` addresses the earlier findings: `ANCHOR_DEGRADED` is gone, the resolver forces REVERT on any degraded pair, `_lift_to_current_best` holds on `not graded.comparable`, and the v6 collector's `_degraded_adoptions` note plus the degrade narrative in the schema, the recorder and `docs/reference/session-breakdown.md` are all replaced by the no-off-objective-promotion statement. CHANGELOG carries the fail-closed Fixed entry and the `breaking:` conc-sweep `*_throughput` -> `*_value` rename.

What I checked:

- Reproduced the resolver at `fb26eb331`: a degraded AgentX pair yields `verdict=REVERT, comparable=False`; a synthetic session is unaffected (`KEEP, comparable=True`).
- Traced every `.verdict` consumer to confirm the forced REVERT causes no collateral damage. `integrate_performance.py:63` and `integrate_patch.py:3356` both test `graded.comparable` before the verdict, and `GradedComparison.graded_on_intvty` is `self.objective == GRADED_INTVTY`, so it is False on any degraded pair and the `graded_on_intvty and verdict == "REVERT"` consumers do not fire.
- Confirmed `explore.py`'s `elif graded.degrade_reason:` branch sets a non-empty `reason`, which short-circuits the later `if outcome == "FAILED" and not reason:` block that would otherwise rewrite the outcome to KEEP.
- Confirmed `record_validation` (writeback.py:872) and `record_adoption` (writeback.py:3370) are each reachable only behind a `comparable` check, so nothing lands in the v6 adoption sums off-objective.
- Confirmed `SharedState.grading` round-trips across resume: `save` uses `asdict(self)` and `load_or_init` filters on `{f.name for f in fields(cls)}`.
- `kernel_stack.py` now tests `not graded.comparable -> NEEDS_REVIEW` before the `verdict != VERDICT_KEEP -> REVERT` branch, so a degraded pair no longer reads as a REVERT decision.
- `ruff check src/hyperloom` passes; `ruff format --check` reports 1276 files already formatted. pytest is not runnable on this host (the suite imports `fcntl`); CI covers it.

One note, not blocking: the `grading=` threading added at `explore.py:550` and `kernel/conc_sweep.py:1340` is inert in practice. `_grid_runner._build_variant_yaml` (line 423) rebuilds each variant YAML through `apply_runtime_benchmark_overrides`, and `apply_agentx_switch` unconditionally reassigns `bench["workload_spec"]`, so the seeded spec is overwritten with environment-derived values for every variant. That is harmless today because `workload_spec` is only ever read back from `state.baseline_config_path` (`phases/kernel.py:1206`), which is set from an anchor run's `materialized_config` -- grid variants never emit that key, and the baseline producer does thread grading (`baseline.py:2659`). Worth knowing if the grid path ever starts feeding the anchor.

PR #1528, opened 2026-09-16
fix(forge): merge a lane's patch against the KEEPs that landed before it

chennyiiis:

  Forge lanes run in parallel and every one of them diffs against the same
  pinned base commit, but `integrate_controller_patches` commits the winners
  one at a time. The moment KEEP #1 is committed, every diff still queued for
  a file it touched is stale. Integration only ever tried a strict
  `git apply --check`, so the second lane was recorded
  `reverted_apply_conflict` and its measured speedup was dropped — under a
  status name that reads like a performance revert, which is why this has
  been easy to misread.

  This is not a rare shape. Lanes tuning the same kernel family tend to
  insert their sweep-flag boilerplate at the same anchor line. In the
  Kimi-K3 a16w4 MoE campaign, `flydsl_moe_stage1` (1.1378×) and
  `flydsl_moe_stage2` (1.1727×) both patched `moe_kernels.py` line 14;
  stage1 landed, stage2 was thrown away.

  `patch_conflict_merge.py` rebuilds the patch instead, on a ladder that
  stops at the first rung that works:

  | rung | how | when it fires |
  |---|---|---|
  | `strict` | `git apply` | nothing landed above touched this file |
  | `three_way` | `git apply -3` | pure line drift around the hunk |
  | `union_disjoint` | union of the diff3 hunk | add/add with an empty base region |
  | `llm` | a resolver, one conflicted region at a time | overlapping edits, or a union that left one name defined twice |

  The resolver is shown a single conflicted region plus the 40 lines of
  source either side of it, and its reply is spliced back exactly where the
  markers were. It is never shown the file to rewrite: asking a model to
  reproduce thousands of unconflicted lines to decide a handful costs a
  generation budget nothing covers — measured, a whole-file call timed out at
  the gateway after 183s — and puts every one of those lines at risk.

  **What any rung is allowed to stand on.** The two rungs that reconstruct a
  file — `union_disjoint` and `llm` — are verified the same way before the
  result is accepted: no added line of this patch, **or of any patch already
  landed in this run**, may go missing; no conflict marker may survive; the
  file must parse; no module-level symbol may be redefined beyond the count
  HEAD already had. `three_way` is not put through that check and does not
  need to be — git merged the hunks itself and cannot drop a side without
  leaving the path unmerged, which routes it to the rungs above. So the LLM
  is a proposer, never an authority, and the existing benchmark and accuracy
  gates still decide what is kept. If no rung succeeds the worktree is
  restored to HEAD and the old `reverted_apply_conflict` result is recorded,
  with the ladder's steps in `reason`.

  The landed set is a sequence, not a single patch, so lane 3 merges against
  lanes 1 and 2 and lane 4 against 1–3; nothing but the ladder's own rungs
  bounds how many can land.

  The resolver runs on whichever backend `preferred_agent_backend` picks for
  the deployment -- Claude through the single-shot Anthropic transport, Codex
  through `achat_completion` -- so an OpenAI-only box, where forge's rewrite
  lane already runs Codex, reaches rung 4 like any other. When neither side is
  credentialed the ladder simply stops at rung 3.

  `PatchIntegrationResult` gains `merge_strategy` — carried into
  `summary.json` — so a KEEP that was rebuilt is distinguishable from one
  that applied verbatim. The resolver's spend is attributed to `forge`.

  **Measured, on one MI355X (whole-op speedup vs base, three routing cases):**

  | case (tokens, seed) | +stage1 | +stage2 | fused |
  |---|---|---|---|
  | (24, 11) | 1.1411× | 1.0618× | **1.2140×** |
  | (24, 31) | 1.1440× | 1.0656× | **1.2214×** |
  | (64, 17) | 1.1729× | 1.0609× | **1.2531×** |

  On the real production artifacts the LLM rung was never reached —
  `union_disjoint` was enough, and `mixed_moe_gemm_2stage.py` was pure line
  drift absorbed by `-3`. Per-stage timing confirms the two winners are
  orthogonal: in the fused tree the stage-1 kernel matches stage1-alone
  within 0.15% and the stage-2 kernel matches stage2-alone within 0.1%, and
  the fused tree compiles byte-identical FlyDSL modules to each single-stage
  arm.

  **The rebuilt tree is the hand-merged tree.** The campaign had to combine
  the two winners by hand to get an end-to-end number at all, precisely
  because integration had dropped stage2. Running the ladder over the same
  two published patches reproduces that hand-merged tree byte for byte:

  | file | ladder output vs hand-merged |
  |---|---|
  | `kernels/mixed_moe_gemm_2stage.py` | identical, sha256 `3e58a5035ae050cd...` |
  | `moe_kernels.py` | identical once the campaign's two-line ASM-registry hook - present in every arm, part of neither winner - is applied: sha256 `1bed845b1c2c0cf7...` |

  So the fused numbers above are not a separate hand-built artifact being
  measured; they are what this code produces.

  **Rung 4 against the live gateway.** Neither published winner can reach it:
  both are pure insertions into `moe_kernels.py`, so their diff3 base region
  is always empty and the union settles them. A four-lane case over the same
  real aiter file — the two winners, plus two lanes written to tune the same
  `tile_n` resolution, which is the shape a sweep produces — exercises one
  rung each against the real model:

  | # | lane | result | rung |
  |---|---|---|---|
  | 1 | `flydsl_moe_stage1` (real) | landed | 1 `strict` |
  | 2 | `flydsl_moe_stage2` (real) | landed | 3 `union_disjoint` |
  | 3 | `xcd_tile_cap` | **rejected**: dropped `tile_n = _S1_TILE_N` | 4 reached, gate held |
  | 4 | `tile_n_pow2` | landed, both sides verbatim, compiles | 4 `llm` |

  Lane 3 is the gate doing its job on a resolution that was arguably correct
  — it folded stage1's cap into a `min()` — and it is worth being explicit
  that the check is one of presence, not semantics: it rejects a reply that
  paraphrases a benchmarked line away, and it would accept a reply that keeps
  both lines verbatim where the second clobbers the first. The backstop for
  that remains the downstream E2E measure-and-revert, which is unchanged.

  None filed.

  Added `tests/test_patch_conflict_merge.py` (20 tests): strict apply,
  three-way absorbing a neighbouring edit, two lanes unioned at one anchor, a
  third lane merging on top of two already merged, the resolver being shown
  one region and its context, a dedented first line being put back at the
  markers' column, a union that duplicates a helper escalating to the
  resolver (and being rejected without one), a rejected patch leaving nothing
  of the file it created, a resolver that raises costing the lane and nothing
  else, overlap with no resolver leaving a clean worktree, and the rejection
  paths for a resolver that drops the incoming side, drops a landed KEEP,
  leaves markers, shadows a module symbol, or emits something that does not
  parse; an OpenAI-only box resolving through `achat_completion` with the
  Codex default model and the forge attribution tags, a dual-configured box
  keeping Claude, and a box credentialed for neither having no resolver.

  Added `test_two_lanes_inserting_at_one_anchor_both_land` to
  `tests/test_controller_patch_integration.py`, asserting both lanes are
  `kept` with strategies `["strict", "union_disjoint"]`.

  `pytest src/hyperloom/orchestrator/kernel/` — 203 passed, excluding the four
  files that import `fcntl` and so are not collectable on this machine
  (`test_campaign_baseline`, `test_controller_patch_integration`,
  `test_controller_publication`, `test_forge_handoff`); those are covered by
  CI. The three regression tests added for this review's findings were
  confirmed to fail on the commit before the fix.
  `ruff check` and `ruff format --check` clean on `orchestrator/kernel/`.
  `mypy` was not run: no typed-dependency environment was reachable.

  No. A patch that applied strictly before still takes the `strict` rung and
  the same code path; a patch that failed before now gets three more chances
  and, failing those, the identical `reverted_apply_conflict` result.

  Yes — serial integration of patches cut against a shared base.

  No, the root cause is in Hyperloom.


xiaofei-zheng:
Reviewed at `6d21287` (after the region-at-a-time rework), against `git merge-base origin/main HEAD`. Repros below were run in this branch's worktree.

## Blocking

**1. `give_up` leaves files the rejected patch created staged, and the next lane commits them.** `patch_conflict_merge.py:482-484`

`git apply -3` implies `--index`, so it stages everything it wrote. `_git_restore_to_head(repo, touched)` only restores `touched`, and `_patch_touched_paths` drops a *created* file: `new_exists` is false at the time it is computed and `rel_old` is `/dev/null`, so the path never enters the list. `_git_commit_kept` then runs a bare `git commit` with no pathspec (`_patch_snapshot.py:226-240`), so whatever is left in the index lands in the *next* lane's KEEP. The dirty pre-check cannot catch it either — it is scoped to the next patch's `touched` and uses `--untracked-files=no` (`controller_patch_integration.py:359-366`).

Repro — lane modifies `mod.py` at a contested anchor and creates `newfile.py`, union rung rejects:
```
A applied: False | steps: (..., 'mod.py: merge redefines module-level _sweep')
A status after give_up: 'A  newfile.py\n'
```
This is new: the old path used plain `git apply`, which never staged anything. `git reset -q --` on the unmerged/staged set, or restoring `touched | _unmerged_paths | created paths`, would fix it.

**2. A resolver failure outside the caught tuple aborts the whole integration run with the repo unmerged.** `patch_conflict_merge.py:431`

`except (LLMConfigError, httpx.HTTPError, asyncio.TimeoutError)` does not cover what the SDK transport raises (`claude_oneshot._drive` / `_load_sdk` raise `RuntimeError`), nor anything else unexpected. `integrate_controller_patches` does not wrap the `apply_patch_resolving_conflicts` call, so the exception leaves the loop.

Repro — resolver raises `RuntimeError`:
```
C ESCAPED: RuntimeError('sdk exploded')
C status: 'UU mod.py\n'
C mod.py head: ['<<<<<<< ours', 'A = 7', '||||||| base', 'A = 1']
```
`give_up` never runs, already-kept results are never written, every remaining lane is skipped, and the worktree keeps conflict markers and an unmerged index. Both existing `aanthropic_completion` callers use `except Exception` (`rca_engine.py:334`).

**3. `resolve_forge_llm_model("claude")` is called without `default=`, so the model id is empty when `CLAUDE_MODEL` is unset.** `patch_conflict_merge.py:326`

`resolve_forge_llm_model` returns `source.get("CLAUDE_MODEL") or default` with `default=""` (`llm_config.py:317-332`); verified: `resolve_forge_llm_model('claude', env={})` → `''`. `llm_resolution_available()` only checks the credential, so on a box with `ANTHROPIC_API_KEY` but no `CLAUDE_MODEL` the ladder commits to the LLM rung and then posts `{"model": ""}`. The other forge caller passes `default=DEFAULT_CLAUDE_MODEL` for exactly this reason (`request_handlers.py:4774-4779`), and `AnthropicClient` hardcodes a model default.

**4. A union-rung rejection ends the ladder instead of escalating to the resolver.** `patch_conflict_merge.py:437-439`

When `_union_disjoint` resolves every region (`overlapping == 0`, no markers) but the union duplicates a module-level name, `_reject_reason` returns at line 439 and the resolver is never called — even when one is available.

Repro — two lanes inserting the *same* `_sweep` helper plus their own distinct flag at one anchor:
```
B applied: False | steps: (..., 'mod.py: kept both sides of 1 disjoint insertion(s)', 'mod.py: merge redefines module-level _sweep')
B resolver calls: []
```
This is the shape the PR names as motivating ("each inserted its own sweep helpers at the same anchor"), and `_RESOLVER_SYSTEM` rule 4 — "If both sides define a helper identically, keep one copy" — is unreachable as written, because the only way to reach the resolver is `overlapping > 0` or surviving markers.

**5. `test_overlapping_edits_are_left_to_a_resolver` makes a live network call and fails whenever a credential is exported.** `tests/test_patch_conflict_merge.py:220`

`apply_patch_resolving_conflicts` substitutes the real resolver when `resolver is None and llm_resolution_available()` (line 495), and `llm_resolution_available()` reads `os.environ`. There is no conftest in `kernel/tests/` scrubbing those keys. Verified:
```
$ ANTHROPIC_API_KEY=sk-ant-bogus py -m pytest ... -k overlapping
E  assert 'no resolver available' in " (... mod.py: resolver call failed (...CERTIFICATE_VERIFY_FAILED...))"
1 failed
```
Clean env: `13 passed`. Needs `monkeypatch.delenv` on `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` / `CLAUDE_CODE_OAUTH_TOKEN`, or an explicit no-op resolver sentinel.

**6. No CHANGELOG entry.** `## [Unreleased] / ### Fixed` carries a paragraph for every comparable behavioral fix (baseline eval attribution, atom kernel-backend default, ATOM teardown). A KEEP that was previously dropped as `reverted_apply_conflict` now lands, and `PatchIntegrationResult` gains `merge_strategy` in `summary.json` — both belong there.

**7. The PR description no longer matches the diff.**
- "Every rung above the first is verified the same way before it is allowed to stand" — `three_way` returns at line 492, before the `_missing_additions` loop at 509-512. Only `union_disjoint` and `llm` are verified. The module docstring ("the last two steps", line 22) is the correct one.
- "Added `tests/test_patch_conflict_merge.py` (11 tests)" — the file now has 13, and the description does not mention the region-at-a-time resolution or the forge cost-attribution commit added after the first push.

## Checked, no issue found

- Ladder ordering and `PatchMergeOutcome` plumbing into `PatchIntegrationResult.merge_strategy` / `reason`.
- `_union_disjoint` base-emptiness test and `_CONFLICT_RE` diff3 parsing.
- `_missing_additions` whitespace normalization and the `len(body) >= 4` identity filter.
- `_reject_reason` shadowing check against the HEAD tally (correctly tolerates names already duplicated at HEAD).
- `_restored_first_indent` / `_strip_code_fence` splice behavior.
- `env -u ANTHROPIC_API_KEY -u ANTHROPIC_AUTH_TOKEN -u CLAUDE_CODE_OAUTH_TOKEN pytest src/hyperloom/orchestrator/kernel/tests/test_patch_conflict_merge.py` — 13 passed. `test_controller_patch_integration.py` is Linux-only (`fcntl`) and was left to CI.

xiaofei-zheng (approved):
Re-reviewed at `a175334`. All seven blocking findings are fixed; no blocking issues remain. Approving.

**Verified against the same repros used in the first pass:**

| # | Finding | Fix | Confirmed |
|---|---|---|---|
| 1 | `give_up` left a created file staged | `_abandon_attempt` unstages, removes created paths, then restores | `status --porcelain` after refusal is now `''`; `sweep.py` gone |
| 2 | Resolver exception escaped the run | `except Exception` at the call site | `RuntimeError` repro now returns `applied=False`, worktree back at HEAD, no markers |
| 3 | Empty model id without `CLAUDE_MODEL` | `default=DEFAULT_CLAUDE_MODEL` | `agent_role.DEFAULT_CLAUDE_MODEL == "claude-opus-5"` |
| 4 | Union rejection ended the ladder | rejection now routes the marked file to the resolver | repro's resolver is called once (`['mod.py']`) where it was `[]` before |
| 5 | Tests reached a live gateway | autouse fixture scrubbing `ANTHROPIC_CREDENTIAL_ENV_ORDER` | `ANTHROPIC_API_KEY=... pytest` → 17 passed (was 1 failed on an SSL error) |
| 6 | No CHANGELOG entry | added under `[Unreleased] / Fixed`, matching the surrounding format | — |
| 7 | Description overstated verification | now "the two rungs that reconstruct a file"; test count and rung table match the diff | — |

**Also checked on this commit:**

- The three regression tests were run against `6d21287`'s `patch_conflict_merge.py` with the new test file in place: all three fail, all three pass on `a175334`. The claim in the commit message holds.
- `_abandon_attempt` ordering: unlink before `git reset` (so the pathspec is still resolvable), and `survivors` excludes created paths so `checkout --force HEAD --` is never handed a path HEAD has never carried.
- The verification chain still holds after the escalation change — a resolver reply that drops both sides is caught by `_missing_additions` (`merge dropped 3 added line(s)`), not by the union step it replaced.
- `_reject_reason` is recomputed after the LLM rung, so escalating past a union rejection cannot launder a bad resolution through.
- `pytest src/hyperloom/orchestrator/kernel/tests/ -n 8` — 200 passed, excluding the four `fcntl`-importing modules (Windows), which CI covers.

The residual scope of `_staged_additions` (repository-wide rather than patch-scoped) is deliberate and correct here: the integration repo is expected clean at HEAD before each lane, and anything staged that HEAD does not carry is exactly the leftover this is meant to clear.

xiaofei-zheng:
One more blocking finding, on `a175334`. This supersedes my approval above — please treat the PR as not ready to merge until this is resolved or explicitly deferred.

## Blocking: rung 4 hardcodes Anthropic instead of following Hyperloom's backend selection

`patch_conflict_merge.py:266-270` and `:357`

The repository already owns this decision in one place — `llm_config.preferred_agent_backend()` — and forge's own other path defers to it: `_resolve_forge_agent` (`request_handlers.py:4773-4779`) resolves the backend first, then picks `DEFAULT_CODEX_MODEL` or `DEFAULT_CLAUDE_MODEL` from it. The specialists (`subprocess_.py:1186`), robustness (`config.py:316`) and the CLI (`cli/executors.py:105`) all do the same. The conflict resolver does not:

```python
def llm_resolution_available() -> bool:
    return bool(anthropic_transport_ready())          # asks only the Anthropic side

model=resolve_forge_llm_model("claude", default=DEFAULT_CLAUDE_MODEL)   # backend pinned
```

`anthropic_transport()` accepts only `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` (`ANTHROPIC_SYNTHESIZABLE_KEY_ENVS`) or `CLAUDE_CODE_OAUTH_TOKEN`. An `OPENAI_API_KEY` is never synthesized into an Anthropic credential, so on a Codex-configured box rung 4 does not exist.

Verified on this commit, with the Anthropic variables unset and `OPENAI_API_KEY` set:

```
preferred_agent_backend : codex
anthropic_transport     : ''
llm_resolution_available: False
```

`preferred_agent_backend`'s own docstring names this as the one shape that selects Codex outright: "an OpenAI-only credential is the only shape that selects Codex outright; a dual-configured deployment keeps Claude." So this is not a corner case — it is precisely the deployment the function exists to describe. On such a box every lane that reaches an overlapping region falls straight to `give_up` and is dropped, which is the behavior this PR set out to remove, while forge's rewrite lane and forge-fusion on the same box run Codex normally. Two backend policies inside one component.

The capability is already there, so this is not a question of building a second transport:

- `achat_completion(client, *, component, operation, **params)` (`llm_config.py:734`) is the async twin of `aanthropic_completion` and carries the same `component` / `operation` cost-attribution tags the forge-spend commit added.
- `get_async_openai_client` / `openai_client_kwargs` (`llm_config.py:433-533`).
- `resolve_forge_llm_model` already branches on the backend and reads `CODEX_MODEL` (`llm_config.py:330-332`); `DEFAULT_CODEX_MODEL = "gpt-5.6-sol"` sits in `agent_role.py` beside the `DEFAULT_CLAUDE_MODEL` this file already imports.

Suggested shape:

- `llm_resolution_available()` asks both sides — the Anthropic transport as today, or an OpenAI-credentialed one.
- `_anthropic_resolver()` becomes `_conflict_resolver()`, dispatching on `preferred_agent_backend()`; the Codex arm posts the same `_RESOLVER_SYSTEM` + region prompt through `achat_completion`, with `default=DEFAULT_CODEX_MODEL`.
- Everything downstream is backend-agnostic already: `_strip_code_fence`, `_restored_first_indent` and the whole verification chain operate on the reply text, so neither the splice nor the gates need to change.

If supporting Codex here is deliberately out of scope for this PR, that is a reasonable call — but then the resolver's unavailability should be stated as a backend limitation rather than a credential one. The CHANGELOG currently says "the resolver is skipped entirely when no Anthropic credential is configured", which reads as "no LLM configured" and hides that a fully-configured Codex deployment gets nothing.

Nothing in my earlier review changes: findings 1-7 are still fixed, and the 200 kernel tests still pass.

xiaofei-zheng (approved):
No blocking issues. Approving at `e6a5bff8b`.

`e6a5bff8b` closes the last open finding: the ladder's fourth rung called Anthropic unconditionally, so an OpenAI-only deployment — where forge's rewrite lane and forge-fusion already run Codex — had no resolver at all and dropped every overlapping-region lane. `_resolver_backend()` now defers to `preferred_agent_backend()` and confirms the chosen side is reachable (`openai_agent_credentialed` for Codex, `anthropic_transport_ready` for Claude); `_conflict_resolver()` dispatches to the Codex or Claude arm, and both arms share `_resolver_prompt()`, so the two backends see the same system prompt and region.

Verified the dispatch by importing the module under three credential shapes:

| exported | `preferred_agent_backend` | `_resolver_backend` | resolver built |
|---|---|---|---|
| `OPENAI_API_KEY` only | `codex` | `codex` | `_codex_resolver` |
| `ANTHROPIC_API_KEY` only | `claude` | `claude` | `_anthropic_resolver` |
| neither | — | `""` | `None`, ladder stops at rung 3 |

The first row is the inverse of what the previous head produced (`llm_resolution_available: False` on an OpenAI-only box).

Checked in this pass:
- `_codex_resolver` against the helpers it calls: `resolve_forge_llm_model` (`llm_config.py:317`), `get_async_openai_client` (`:506`), `apply_reasoning_effort` (`:643`), `achat_completion` (`:734`), `build_chat_messages` (`roles/base.py:47`). Signatures match, and the arm mirrors the established call shape at `critic_agent.py:1068-1082` (`build_chat_messages` + `max_completion_tokens` + `apply_reasoning_effort`). Spend is tagged `component="forge"`, `operation="patch_conflict_merge"`, same as the Claude arm.
- The earlier seven findings are still fixed after the rebase onto `a3612d81d`: `_abandon_attempt`/`_staged_additions` (`patch_conflict_merge.py:142-171`) unstage and delete files a rejected patch created before the next lane's `git add -A`; the union→resolver escalation re-runs `_reject_reason` after the LLM rung (`:551-578`); a resolver raising anything is caught at `:571` and costs only that lane.
- The autouse `_offline` fixture now clears `OPENAI_API_KEY` and `_ANTHROPIC_MANAGED_GATEWAY_ENVS` alongside `ANTHROPIC_CREDENTIAL_ENV_ORDER`, so the no-resolver tests cannot become live calls on a developer box configured for either side.
- CHANGELOG entry no longer describes the limitation as Anthropic-credential-shaped, and the PR description covers the Codex arm and the three new tests. Title matches the diff.

Tests run locally (Windows, `-n 8`):
- `tests/test_patch_conflict_merge.py` — 20 passed.
- `src/hyperloom/orchestrator/kernel/tests/` — 203 passed, excluding the four `fcntl`-importing modules that are not collectable on this platform (`test_campaign_baseline`, `test_controller_patch_integration`, `test_controller_publication`, `test_forge_handoff`); CI covers those.

PR #1529, opened 2026-09-16
fix(kb): address one kernel's records by one address, however its source and its environment spell it

chennyiiis:
## The defect

A forge-loop record is addressed by the operator as the campaign declared it, and the campaign declares the entry point **as the source it read spells it**. One tree spells a kernel both ways:

| where | spelling | normalized to | page |
|---|---|---|---|
| the header that declares the kernel | `KdaPackedDecodeKernel` | `kdapackeddecodekernel` | A |
| the module that binds it | `kda_packed_decode_kernel` | `kda_packed_decode` | B |

`normalize_operator_name` lowercased, replaced `.`, collapsed separators and stripped a trailing `_kernel` — but never split camel case. So the two spellings addressed two pages, and the second addresses one nothing ever wrote to.

The read is exact, so the miss surfaced as `no_prior_record`: indistinguishable from an operator nobody has ever ported. A validated **1.44x** port stayed on disk, still applying cleanly, while the campaign that should have started from it re-derived the kernel from scratch.

## Why this is the common case, not a rare one

The task contract *asks* for the camel-cased, namespace-qualified spelling:

> Give `operator_name` the entry point as the source writes it, keeping camel case and any namespace prefix — `aiter::fusedAddRmsNorm`, not `fused_add_rms_norm`. Upstream pull-request search splits that spelling into terms, and a name normalized before it arrives has no boundaries left to split on.

That instruction is right about search and was silently wrong about addressing. Every kernel whose Python binding is snake-cased has both spellings within reach of the agent that reads it.

## The change

`normalize_operator_name` splits camel-case boundaries before lowercasing.

An acronym run stays one word. Splitting on every case change would cut `MoE` into `mo_e` and `QKV` into `q_k_v` — inventing a difference between spellings of one kernel instead of removing one. The tail of a capitalized run only starts a new word when a lowercase letter follows it, and a one-character piece is given back to the word it was cut from **only when that word is itself a fragment of at most two letters** — which is what an acronym written with a lowercase letter inside looks like once the boundaries have cut it. After a whole word a capital is a word of its own.

```
KdaPackedDecodeKernel    -> kda_packed_decode
kda_packed_decode_kernel -> kda_packed_decode
aiter::fusedAddRmsNorm   -> fused_add_rms_norm
FusedMoE                 -> fused_moe          (Mo|E rejoined: Mo is a 2-letter fragment)
ChunkFwdKernelO          -> chunk_fwd_kernel_o (Kernel|O kept: Kernel is a word)
paged_attention_ll4mi_QKV_mfma16_kernel -> paged_attention_ll4mi_qkv_mfma16
unified_attention_with_output           -> unified_attention_with_output   (unchanged)
```

No other identity dimension moves, and the write and read sides share this function, so they cannot disagree.

## Validated against every real kernel name reachable from here

The hand-picked examples above are not evidence that the rule is right, so it was property-tested over three corpora rather than over examples: the **107** distinct page names in the production store and the **201** `implementation_symbols` recorded across its 144 pages (as the store stood when the corpus was taken; it has since grown to 147 pages and 108 names, which changes no conclusion below), and the **632** kernel symbols declared in this tree (`@triton.jit` defs, `__global__` entry points, `*_kernel` defs, C++ camel entry points) — **926** distinct names. For every snake-cased name in that union, both its Pascal and its camel spelling were generated and required to normalize to the same page.

**This found a real defect in the first version of the rule.** Giving a trailing capital back unconditionally kept `MoE` whole but also welded the dimension letter these kernels end in onto the word before it: `ChunkFwdKernelO` became `chunk_fwd_kernelo` while the source declaring it spells it `chunk_fwd_kernel_o`. **82 of 769 kernel names, 230 of 1538 generated cases.** The two-letter fragment bound above is the fix; the two classes are separable because an acronym cut in half leaves a fragment and a word does not.

| property | result |
|---|---|
| idempotence — read side and write side normalize independently, so a drifting rule splits a page | pass, 926/926 |
| a name already written as words is a fixed point | pass |
| all 107 live page names are their own normal form | yes |
| **camel agreement** — Pascal and camel spellings reach the snake name's page | **330 / 359 = 91.9%** |
| injectivity over live page names | no merges beyond the 2 pre-existing `_kernel` pairs below |

The 29 disagreements are all one ambiguity and it is not decidable: a digit beside a letter. `mxfp4_moe_2stage` says `2stage` is one word, `Mxfp4Moe2Stage` says it is two, and `gemm_a8w8_blockscale` says `a8w8` is one — a rule that split digits from letters would re-address that page to `gemm_a_8_w_8`, which is worse than the gap. 28 of the 29 are of this shape; the last, `recompute_w_u_fwd_kernel`, spells two adjacent single letters that Pascal writes `WU`, indistinguishable from an acronym. 9 live page names fall in this set, and each costs a page only if a later campaign supplies the camel spelling of one. `test_a_digit_run_keeps_whatever_boundary_the_spelling_gave_it` pins the gap so a change in it cannot pass unnoticed.

## Measured against the production store

Read-only scan of `producer=forge-loop`, re-run 2026-09-21: **147 pages, 108 distinct
names.** Every page's champion record was fetched and compared; nothing was written.

**This rule moves no stored page.** Running both normalizers over all 108 live names,
`origin/main`'s and this branch's, **zero stored names move**: every stored name is
already lowercase, and the new rule is the identity on a lowercase input. Boundaries
destroyed before the write cannot be recovered by reading. (The version dimension is a
different story and does require a migration — see below.)

What changes is the address a *later* campaign computes, because the contract asks that
campaign for the camel spelling:

| a later campaign reports | addressed today | addressed after this PR | what that costs |
|---|---|---|---|
| `KdaPackedDecodeKernel` | `kdapackeddecodekernel` | `kda_packed_decode` — populated | **nothing**; it stops minting a second page for a kernel that already has one |
| `TopKGatingSoftmax` | `topkgatingsoftmax` | `top_k_gating_softmax` — empty | one cold start, then correct |
| `hipBSolGemm` | `hipbsolgemm` | `hip_b_sol_gemm` — empty | one cold start, then correct |
| any lowercase name | unchanged | unchanged | — |

So the fork this PR reports stops widening; it is not retroactively closed. The four
squashed pages keep their records and stay unreachable, exactly as they are on `main`.

The store does prove the fork is one kernel rather than two that share a name — which is
what makes the redirect the right target rather than a guess:

| field | `kda_packed_decode` | `kdapackeddecodekernel` |
|---|---|---|
| `implementation_signature` | `527b02951b281dac…` | `527b02951b281dac…` |
| `implementation_symbols` | `["kda_packed_decode_kernel"]` | `["kda_packed_decode_kernel"]` |
| `source_paths` | `sglang/kernels/…/kda_packed_decode.{cuh,py}` | identical |
| champion speedup | **1.4439x** | **1.4439x** |

Two campaigns, one kernel, the same result derived twice.

An earlier revision of this section claimed `kdapackeddecodekernel` itself merges into
`kda_packed_decode`. It does not, and the claim contradicted this section's own closing
sentence; the table above is what the scan actually shows.

**No two stored pages can collide.** Every stored name being its own normal form is also
what makes the new rule injective over the 108 live names.

## What the same scan ruled out

The scan found 18 groups where one `implementation_signature` spans several names. **Only the `kda` pair is a rename.** The other 17 are distinct operators that share a source file and therefore share a file-level signature — six different operators in `vllm/_aiter_ops.py` alone, `flydsl_moe_stage1` beside `stage2`, `ck_moe_stage1` beside `stage2`, `mxfp4_moe_2stage_t16/t32/t64/t128`. That ruled out the two designs this PR previously carried and one considered after it:

- **Probing other names on a signature match** (the earlier revision of this PR, reverted here): the signature is file-level, so a new operator in a shared file would warm-start from a sibling's patch at `match_mode='exact'` and then file its result on the sibling's page. @xiaofei-zheng raised exactly this; the store says it is the majority structure, not a corner case.
- **Deriving the name from the file's kernel symbols**: the symbols are GPU kernels, while the operator a campaign targets is usually the host-side API. Re-deriving the pages that way renames `unified_attention_with_output` to `apply_softcap`, collapses `mxfp4_moe_2stage_t16/t32/t64/t128` into one name and `custom_all_reduce_tp2`/`tp4` into another.

`operator_name` names the right thing. Only its normalization was lossy.

## The same defect in the version dimension

The same scan found the address splitting a second way, for the same reason: a dimension was whatever string the campaign happened to hold rather than the fact it names.

Three code paths answer "which framework version?" and they answer in three different words.

| who answers | what it says | what it means |
|---|---|---|
| `framework_version()` via `importlib.metadata` | `0.24.0+rocm723`, `0.5.15.post1.dev20260724+g3d91a569ce` | the build the wheel was compiled as |
| a campaign reading the package tag or its image | `v0.24.0`, `v0.5.15.post1-rocm720-mi35x-20260724` | the same release, written the way a tag writes it |
| a path that observed nothing | `none`, `unspecified`, `unknown`, `""` | one absence, under four names |

A port is written against a **release**. It does not depend on the machine that compiled the framework, nor on the tag convention of whoever wrote the version down, nor on which of three code paths failed to find a version. Each of those spellings opened its own page.

### What it cost, measured

Read-only scan of the same 146 `producer=forge-loop` pages, replaying the canonicalizer this PR ships and fetching each page's champion record:

**146 pages → 135. Eleven merge, nine of them provably one kernel** by identical `implementation_signature`. Every one of those nine held a *different champion on each page*:

| kernel | the two spellings | champion on each |
|---|---|---|
| `sparse_attn_prefill_ragged` | `v0.24.0` / `0.24.0+rocm723` | **13.7293x** / 8.2320x |
| `mxfp8_grouped_gemm` | `0.5.15.post1.dev20260724+g3d91a569ce` / `v0.5.15.post1-rocm720-mi35x-20260724` | **4.1771x** / 3.3885x |
| `mhc_fused_post_pre` | `0.24.0+rocm723` / `v0.24.0` | **3.3633x** / 3.2760x |
| `unified_attention_with_output` | `unspecified` / `unknown` | **2.3303x** / 2.0238x |
| `paged_attention_rocm` | `unspecified` / `unknown` | **1.5456x** / 1.4935x |
| `rmsnorm` | `unknown` / `unspecified` | **1.4946x** / 1.1582x |
| `fused_moe` | `0.24.0+rocm723` / `v0.24.0` | **1.4650x** / 1.4435x |
| `aiter_mxfp4_moe_2stage` | `unknown` / `unspecified` | **1.3203x** / 1.3097x |
| `moe_flydsl_stage2` | `unspecified` / `unknown` | **1.0345x** / 1.0283x |

Each row is one kernel ported twice, by two campaigns that could not see each other. The remaining two groups share an operator but not a signature, and both elect the faster of the two: `dynamic_per_tensor_quant` 10.4489x over 2.8170x, `unified_attention_with_output`@`0.24.0` 11.5837x over 2.5712x.

### The change

`canonical_framework_version` resolves a version string to the release it names, and every word for not having observed one to a single word.

```
0.24.0+rocm723                              -> 0.24.0
v0.24.0                                     -> 0.24.0
0.5.15.post1.dev20260724+g3d91a569ce        -> 0.5.15.post1
v0.5.15.post1-rocm720-mi35x-20260724        -> 0.5.15.post1
0.1.dev19253+g5f76ae224.d20260727.rocm723   -> 0.1
none | unspecified | unknown | "" 	        -> unknown
not-a-version                               -> not-a-version   (unchanged)
0.5.15 / 0.5.15.post1 / 1.0.0rc1 / 1.0.0    -> distinct, as they are distinct releases
```

Parsing is PEP 440 (`packaging.version.Version`, already a hard dependency of `warmstart_identity`). An image tag joins the build to the release with the same character a pre-release uses, so it is only distinguishable by failing to parse whole — hence the one retry on the first `-`. A string naming no release is returned exactly as written, so the read side and the write side, which canonicalize independently, cannot turn it into some other page's name.

It is applied at the three points that write the dimension and once on the read side:

| | where | what it fixes |
|---|---|---|
| write | `rewrite_by_flydsl/identity.py: framework_version()` | the build suffix, and the two words this function itself used for absence — `NO_FRAMEWORK_VERSION = "none"` and `UNKNOWN_VERSION = "unspecified"`, both deleted |
| write | `kernel_identity.py: KernelRecipeIdentity.from_mapping` | the tag spelling an agent declares in `task.json` |
| write | `task_publisher.py: _normalize_agent_task_payload` | makes the *published* task.json state the address the controller went on to use, beside the `kernel_name` it already re-derives |
| read | `warmstart_identity.py: rank_fallback_identities` | resolves the *target*, which arrives raw from distribution metadata or an image tag |

This also un-strands the pages that never had a version. **55 of 146 name no release** — 42% — and the fuzzy tier used to end before it started for all of them, because `unknown` fails to parse. Under one word for absence they can reach each other.

What the ranking still refuses is unchanged: how far a *known* release sits from an unknown one is not a question the strings can answer, so that comparison stays rejected rather than being given an invented distance. `test_fuzzy_ranking_rejects_unknown_or_unparseable_dimensions` passes untouched.

### This requires a store migration, and it has to land first

An earlier revision of this PR resolved the *stored* version as well as the target's, so
a page addressed `v0.24.0` kept serving warm starts while every write went to `0.24.0`.
That is comfortable and wrong: one kernel holds two pages, one of them frozen, and
nothing reports the split because reads still succeed. At the KB team's request that
read-side fold is **removed** — a stored dimension is read as written, because it was
resolved when its page was addressed.

Read-only scan of the live store, 2026-09-21: **87 of 147 pages** carry a spelling that
is no longer the address they would be written at, together holding **605 sessions**.

| action | pages | sessions |
|---|---|---|
| `RENAME` — target address empty | 68 | 316 |
| `MERGE_INTO_EXISTING` — target already holds a page | 6 | 109 |
| `MERGE_WITH_PEER` — two or more movers land together | 13 | 180 |

**29 of the 87 stop being reachable the moment this merges** and stay dark until
migrated: their stored spelling (`unspecified` ×22, `v0.5.15.post1-rocm720-mi35x-20260724`
×6, `none` ×1) names no release at all, so it ranks as unparseable and is dropped. The
other 58 are still found by the fuzzy tier in the meantime, 40 at unchanged rank and 18
demoted by a local segment.

So the order is **migrate the store, then merge this**. The change list — 87 rows, one
per page, with `old_canonical_id`, `new_canonical_id`, session counts and the champion
speedup of each side of every merge — has been handed to the KB team, who are performing
the migration. It is a store-side change and is not performed by this code.

Two merges change which session is champion, and are worth a look before they are folded:

| kernel | page A | page B |
|---|---|---|
| `sparse_attn_prefill_ragged` @ vllm 0.24.0 | `0.24.0+rocm723` — 46 sessions, 8.232× | `v0.24.0` — 13 sessions, **13.7293×** |
| `unified_attention_with_output` @ vllm 0.24.0 | `0.24.0+rocm723` — 27 sessions, 2.5712× | `v0.24.0` — 15 sessions, **11.5837×** |

Seven of the 87 also have a non-canonical `kernel_name` (a trailing `_kernel`). That is
**pre-existing** — the normalizer has stripped that suffix since before this PR — and is
included in the change list only because the address is wrong either way.

### Where the six dimensions stand after this

| # | dimension | distinct values in store | what pins it | |
|---|---|---|---|---|
| 1 | `producer` | 1 | whitelist, plus a hard `== forge-loop` check in `task.py` | pinned |
| 2 | `kernel_name` | 106 | `normalize_operator_name`, re-derived host-side from `operator_name` | this PR |
| 3 | `framework` | 5 | folded to the override the run resolves, then checked against the task's own paths | this PR |
| 4 | `framework_version` | up to 4 spellings of one release | `canonical_framework_version` | this PR |
| 5 | `backend` | 5, all registered | membership in `KERNEL_BACKENDS`, the set campaign setup resolves against | this PR |
| 6 | `gpu` | 1 (`mi355x`) | uniformly lowercased | see below |

## The same defect in the backend dimension

`identity.backend` is one of the five dimensions the agent types into `task.json`, and it becomes `--kernel-backend` on the forge-loop command line. `campaign_setup.resolve_kernel_backend_override` then resolves it against `KERNEL_BACKENDS` and, for a name that is not there, substitutes `_FALLBACK_KERNEL_BACKEND = "flydsl"`, warns on stderr, and continues.

So a draft saying `tilelang` produces a run that **builds with flydsl and files its result on the flydsl page**, while the task directory, the `--experience-id` and the published pointer all say `tilelang`. One run, two addresses — the shape this PR exists to remove.

The comment standing at that spot in `task.py` justified not checking, and stated the reason as fact:

> forge-loop answers an unregistered one with no layer rather than an error

It does not. It answers with a different layer, silently. The rule that comment defended is what let the substitution through.

This commit checks membership in `KERNEL_BACKENDS` — the same frozen set `campaign_setup` resolves against, and the same one `opportunity_agent.py:418` already injects into the agent's prompt. The name the agent has to spell is the one it was handed.

| | |
|---|---|
| registered | `assembly`, `ck`, `flydsl`, `triton`, `gluon`, `aiter`, `hip`, `hipblaslt`, `fusion` |
| present in the store (146 forge-loop pages) | `triton` 81, `hip` 32, `flydsl` 19, `ck` 11, `aiter` 3 |
| stored values this refusal would reject | **none** — every one is registered |

The refusal costs nothing that campaigns write today. It only closes the path by which a name the agent invents becomes a page nothing writes to.

## The same defect in the framework dimension

`detect_framework(kernel_path, framework_override)` infers the owning package from the path — but **a non-empty override short-circuits the inference entirely**. Since the agent's `identity.framework` travels to forge-loop as exactly that override, the dimension reaches the knowledge-base page having never been examined by anything. Two consequences:

**1. A kernel that belongs to no package is addressed two ways.** `_NO_FRAMEWORK_SENTINELS = {"standalone", "none", "unknown"}` all fold to `unknown` inside the run, but the draft keeps its own spelling in its directory name and its published pointer. The store is the same gap seen from the far end:

| spelling | pages in store | controller fixtures declaring it |
|---|---|---|
| `standalone` | **0** | 4 |
| `unknown` | 1 | 0 |

Every campaign that declared `standalone` filed its result under `unknown` and pointed at `standalone`.

**2. A framework naming a package the source does not live in is never contradicted.** The override suppressed the one check that could have noticed.

This commit folds the override at publication the way the run will, and refuses a framework that none of the task's own paths sit under. The paths were just verified to exist at the base commit, which makes them the only witness available at that point.

### Why it refuses rather than re-derives

I first wrote this to derive `framework` from the paths and overwrite whatever the agent said. Scanning the store killed that design:

| verdict over 118 pages that recorded source paths | count |
|---|---|
| path inference agrees with the stored framework | 102 |
| no single owner — declared value kept | 13 |
| inference "disagrees" | 3 |

All 3 disagreements are the same path: `sglang/aiter/ops/flydsl/kernels/mixed_moe_gemm_2stage.py` — **under both packages at once**. `detect_framework` answers `aiter` only because `_FRAMEWORKS = ("aiter", "sglang", "vllm")` starts there. Re-deriving would have moved those three pages on the strength of a tuple's ordering.

So the check refuses a contradiction and never re-derives. A declared framework appearing anywhere in the paths stands. A path under no known package witnesses nothing and is left alone — which is what keeps the 4 pages stored under `torch` (a value `_FRAMEWORKS` does not know) addressable.

## Not covered

- **`gpu` is the one dimension of the six with an alias the write side does not fold.** `warmstart_identity._GPU_ISA` already knows `mi355x` and `amd_instinct_mi355x` are one card, and `mi300x`/`amd_instinct_mi300x` likewise — but only on the read side, for ranking. Nothing folds them when the address is written. All 146 stored pages say `mi355x`, so there is no split to repair today; the fix shape is a `canonical_gpu()` fed by that same alias table, and it is a separate concern from the three here. It is the only one of the six this branch leaves unpinned.
- **A second split-history defect, pre-existing and not from camel case.** Seven stored names end in `_kernel` — `mxfp8_quant_kernel`, `mxfp8_grouped_gemm_kernel`, `fwd_kernel`, `llgemm1_kernel`, `dynamic_per_group_scaled_quant_kernel`, `recompute_w_u_fwd_kernel`, `tp2_fused_ar_rms_kernel` — which `normalize_operator_name` has stripped since before this PR, so no campaign can address them today either. `mxfp8_quant_kernel` sits beside `mxfp8_quant` and `mxfp8_grouped_gemm_kernel` beside `mxfp8_grouped_gemm`: the same two-half-histories shape, but it implies a write path that bypasses the normalizer, which is a different defect and not traced here.
- **`match_mode` means "same files", not "same kernel".** Given the file-level signature, `implementation_match` is true for siblings in one file. It labels the reference handed to the model and gates nothing — warm start is decided by whether the patch applies cleanly and benchmarks faster — so this is a naming problem, not a behavioural one, and fixing it needs an operator selector in the signature payload.
- Pages written before `implementation_identity` was recorded (74 of 144) carry no symbols; nothing here reaches them, and nothing needs to.
- Two campaigns that target the same operator at different levels — one naming the host API, one naming the Triton kernel it calls — still address two pages. That is a contract question (`operator_name` is not required to appear in `target_functions`), separate from spelling.

## Test

- `test_the_same_kernel_spelled_either_way_is_one_operator` — the reported pair.
- `test_namespaced_camel_case_matches_its_snake_case_spelling` — the spelling the contract asks for costs no page.
- `test_an_acronym_run_is_one_word` — `MoE`, `KVCache`, `QKV`, `HGEMV_WFPerRow`.
- `test_a_capital_after_a_whole_word_is_its_own_word` — the counterpart, and the regression for the defect the validation above found.
- `test_a_digit_run_keeps_whatever_boundary_the_spelling_gave_it` — the known gap, pinned.
- `test_names_already_written_as_words_are_left_alone` — real page names from the store normalize to themselves.
- `test_distinct_operators_sharing_a_prefix_stay_distinct` — `stage1` ≠ `stage2`, `rmsNorm` ≠ `addRmsNorm`.
- `test_either_spelling_of_one_operator_reaches_the_same_record` — end to end through `resolve_loop_identity`: written under one spelling, found under all three.

Version dimension:

- `test_the_build_a_release_was_compiled_as_is_not_part_of_the_release` — `0.24.0+rocm723`, and a dev build with three suffixes.
- `test_a_release_reads_the_same_however_it_was_written_down` — the tag spelling and the image spelling reach the release.
- `test_every_word_for_not_knowing_the_version_is_the_same_word` — all four absences.
- `test_distinct_releases_stay_distinct` — `.post1`, `rc1` and a patch bump are not folded away.
- `test_a_version_naming_no_release_is_left_exactly_as_written` — idempotence, which the independently-canonicalizing read and write sides depend on.
- `test_a_stored_version_is_taken_as_written_and_ranks_below_the_release_itself` — the read side resolves the target only; a stored spelling is read as PEP 440 reads it.
- `test_a_stored_spelling_that_names_no_release_is_dropped_rather_than_resolved` — `unspecified` is not folded into `unknown` on the way in, so a page nothing writes to cannot keep answering reads.
- `test_runs_that_both_observed_no_version_can_reach_each_other` — the 42%, with the known-vs-unknown comparison still refused.

Backend dimension:

- `test_an_unregistered_backend_is_refused_rather_than_silently_substituted` — replaces `test_an_unregistered_backend_is_accepted`, which pinned the old behaviour and repeated the false premise in its docstring.
- `test_every_backend_the_agent_is_offered_is_one_it_may_declare` — every name in `KERNEL_BACKENDS` survives the check, so the refusal cannot cost an operator a name its prompt offered.

Framework dimension:

- `test_a_kernel_belonging_to_no_package_is_published_under_one_word_for_that` — `standalone` is published as `unknown`, which is where the run files its result.
- `test_a_framework_alias_is_published_as_the_package_it_names` — `aiter_meta` → `aiter`.
- `test_a_framework_none_of_the_paths_sit_under_is_refused` — `vllm` declared over a `sglang/` path leaves a rejection the agent can read.
- `test_a_path_under_two_packages_keeps_whichever_the_task_declared` — the real `sglang/aiter/ops/flydsl/...` case: declared `sglang` stands.
- `test_a_path_under_no_known_package_witnesses_nothing` — `torch`, which no path names, is left alone.

Version dimension, correcting an earlier revision of this section: the fold to one
word for absence changed the address three test files still spelled the old way,
and this branch carried those failures for four days because the comparison below
was run against the wrong tree. The baseline was a commit *of this branch*, so the
regression was red on both sides and read as pre-existing. `IDENTITY` in
`test_experience_sink.py` said `none` where its own comment above it already said
`unknown`; `CANONICAL_ID` in `test_kb_measured_evidence.py` and the address asserted
in `test_kb_warmstart_end_to_end.py` said `unspecified`. All three now say the word
the dimension folds to. They belong to the commit that introduced the fold and are
squashed into it.

## Verification

Rebased onto `main` (`dcebdcd32`). The only conflict was `CHANGELOG.md`, which that
commit deletes; the deletion is accepted in all five commits and no source hunk moved
— `git diff 089228c52 <pre-rebase> -- src/` and `git diff dcebdcd32 HEAD -- src/` are
both 1001 lines and identical.

Whole suite on Linux (Ubuntu 22.04, nothing stubbed), re-run on both sides after the
rebase, against a detached worktree at **`origin/main`** — not, as an earlier revision
of this section had it, against a commit of this branch:

| whole suite | `origin/main` `dcebdcd32` | this branch |
|---|---|---|
| passed | 4336 | **4359** (+23) |
| failed | 7 | 7 |
| skipped | 14 | 14 |
| set of failing test ids | — | **identical** |

The 7 are the same seven on both trees and none is reachable from this change:
`test_rename_completeness.py` (5), `test_lane_aiter_cache.py`, and
`test_vendored_kb_store_client.py::test_vendored_sdk_matches_upstream_git_blob`.
The five `test_rename_completeness` ones are an artifact of this harness, not of
either tree: the runs happen in a git worktree whose `.git` file records a Windows
path, so `git ls-files` under WSL aborts with `fatal: not a git repository`. They do
not arise on CI.

The controller package alone is **282 passed, 0 failed**. That includes
`test_controller_full_path`, the one test worth naming here: it is where the agent's
draft and the published address are both visible, and it hardcoded `standalone` in
the operator id. This branch changes that constant to `unknown`, because publication
now folds the sentinel the way forge-loop already did — and those five cases pass, so
the change is confirmed by the test rather than by argument.

`ruff check` clean; `ruff format --diff` reports no change.

## Review follow-up

Both blocking issues from the 2026-09-21 review are fixed and squashed into the commits
that introduced them:

| issue | fix |
|---|---|
| a non-string `identity.framework` reached `str.strip` and aborted the whole publication sweep with an `AttributeError` | the fold is now `isinstance`-guarded like the loop above it, so a bad draft is refused by name and its siblings still publish; regression test `test_a_framework_that_is_not_even_a_word_is_refused_with_the_sweep_intact` stages two drafts and asserts exactly that |
| three of the four observable changes had no `[Unreleased]` entry | moot as written: `dcebdcd32` retired `CHANGELOG.md` and `AGENTS.md` moved the gate to "every change states its observable effect in its own PR description". The entries were written, then dropped when this branch was rebased onto that commit. Each observable change is stated above instead, one section per dimension — kernel name, version, backend, framework |
| the KB team asked for the backward-compat read to be removed, and to be given the change list instead | the fold of the *stored* version is deleted (fifth commit); the 88-row change list is with them and the store migration lands before this PR — see the section above |

Guarded rather than `str(...)`-coerced on purpose: coercion would publish `"['sglang']"`
as a framework, where the guard hands the value to the parser that refuses it by name.


xiaofei-zheng:
Reviewed the alias tier end to end against the merge base (`9c1d4d757`), with a local repro on the in-memory KB store. Three blocking issues.

### 1. Two operators declared in the same source file collide on the alias tier (correctness, regression vs `main`)

`implementation_signature` hashes `{source_paths, implementation_symbols}` (`implementation_identity.py:186-208`) — every symbol in the declared files, with no selector for the operator actually under optimization. `target_functions` is not part of the payload. So two forge-loop tasks scoped to different kernels in one file produce an **identical** signature, which is exactly the equality `aliased_identity` accepts (`identity_alias.py:96`), and `alias_kernel_names` feeds the sibling symbol in as a probe (`identity_alias.py:58`).

Repro — one file declaring `fwd_grouped_kernel` and `decode_kernel`, two tasks with distinct `--operator-name` / `--target-functions` (the controller passes both per task, `forge_runner.py:135-154`):

```
A written:      kernel:forge-loop:fwd_grouped:unknown:none:triton:mi300x   speedup=2.0
B read status:  read_reason='alias_hit'  match_tier='alias'
                requested=kernel:forge-loop:decode:...
                selected =kernel:forge-loop:fwd_grouped:...
B match_mode:   ['exact']        <-- auto-apply grade
B written:      kernel:forge-loop:fwd_grouped:...  speedup=5.0, champion=True
```

Operator `decode` warm-starts from `fwd_grouped`'s patch at `match_mode='exact'`, then `recorded_identity` files `decode`'s own result onto `fwd_grouped`'s page and takes the champion slot there. `decode`'s page stays empty forever, so every later `decode` run repeats the redirect, and `fwd_grouped`'s history is now a mix of two kernels that future `fwd_grouped` runs read back as exact matches.

Same script on the merge base: `read_reason='no_prior_record'`, and B writes to `kernel:forge-loop:decode:...`. This is introduced here.

`test_a_shared_name_does_not_hand_over_another_operator_s_record` does not cover it — it uses a *different* signature, which is the easy case. The hard case is a shared signature with different operators.

### 2. A transport error on the run's own page is read as "page is empty", and the write is redirected (silent failure)

`KernelRecipeKB.list_candidates` catches every exception into `self.reason` and returns `[]` (`agent_kb.py:148-153`). `recorded_identity` (`identity_alias.py:116-118`) checks only the truthiness of that list and never inspects `kb.reason`, so a 5xx/timeout on the run's own page is indistinguishable from a cold page — and the fall-through is alias probing, i.e. the record gets filed under a different operator's name even though its own page has history.

Verified with a store that raises on the own-page `candidates()` call: the exception is dropped and `recorded_identity` proceeds past the `list_candidates` guard with no log, no status field, nothing in the returned write result.

`_read_top_solutions_impl` is careful to turn `kb.reason` into `read_error` on every other read (`experience_reader.py:303-308`, `330-333`, `350-357`); the two `list_candidates` call sites in `identity_alias.py` (`95`, `117`) are the only ones that don't.

### 3. The candidates handed back from an alias page are not signature-filtered, contrary to the PR body and CHANGELOG

`aliased_identity` only proves *some* record in the alias page's top 16 carries the signature. The reader then re-reads that page with `alias_kb.read_top_n(destination, limit=top_k)` (`experience_reader.py:329`) and returns whatever ranks highest by speedup, unfiltered. On a page shared with a genuinely different implementation that happens to normalize to the same name, the top candidates can all be foreign; they are labelled `match_mode='reference'` correctly, but they are returned under `read_reason='alias_hit'` and injected as warm-start reference material.

That contradicts:
- PR body: "accepts a record only when the candidate's `implementation_signature` equals the consumer's";
- `CHANGELOG.md`: "takes it only when a record there carries the implementation signature in hand" and "`match_mode` is unaffected: the implementation matched, so an aliased candidate is as applicable as any other".

Either filter the returned candidates by signature, or correct both texts.

---

Checked and found nothing blocking: the `alias_hit` verdict surviving `kb_warmstart`'s normalization (`experience_integration.py:1117-1119`), the alias-before-fuzzy ordering, `match_tier` / `requested_canonical_id` / `selected_canonical_id` reporting, the `MAX_ALIAS_PROBES` bound, the docs update in `autonomous-loop.md`, and the CHANGELOG entry's presence and placement. No tests were run beyond the two targeted repro scripts above; CI covers the rest.

xiaofei-zheng:
### Blocking

**`identity_alias.py` swallows KB read failures, and both call sites act on the empty result as if it were an answer.**

`KernelRecipeKB.list_candidates` reports transport failures by setting `self.reason` and returning `[]` (`agent_kb.py:144-165`). `aliased_identity` and `recorded_identity` never look at `kb.reason`, so a 5xx/timeout on a single metadata read is indistinguishable from an empty page. The exact-address read two lines away in `experience_reader.py:302-307` deliberately does the opposite ("surface it as a read error so a cold start is not mistaken for an empty store"), and so does the alias `read_top_n` at `experience_reader.py:331-334`.

Two consequences, both verified against this branch with `KernelRecipeKB` stubbed to set `reason` and return `[]`:

1. `identity_alias.py:97` (read side) — a failing probe skips that alias. The reader falls through to fuzzy/cold start and reports `read_reason='no_prior_record'` with `read_error=''`: the exact "a validated solution stops being found without anything reporting a loss" failure this module's docstring is written against.

2. `identity_alias.py:124` (write side) — `recorded_identity` treats a failed probe of the run's *own* page as "this page is empty" and goes on to resolve an alias. Repro: own page `newname` holds a record and its probe raises; alias page `oldname` holds a record with the same `implementation_signature` → `recorded_identity` returns `oldname`. The write lands on the alias page while later reads hit the populated `newname` page on the exact tier and never run the alias probe, so the improvement this run just made is stranded — the split this PR exists to close, opened by a single transient error.

Suggest propagating `kb.reason` out of `aliased_identity` / `recorded_identity` so the reader can set `read_error` as it does for the other tiers, and so the write side keeps its own address rather than relocating on an unanswered probe.

### Checked, no issue found

- Read/write alias derivation agrees: the sink is handed the pristine `implementation_signature_override` / `implementation_identity_override` (`experience_integration.py:1442-1464`), so `implementation_symbols`, `target_functions` and the kernel stem are the same inputs the reader used pre-optimization, and `alias_kernel_names` sorts each tier, so both sides probe in the same order.
- `alias_kernel_names` ordering and the `MAX_ALIAS_PROBES` bound exercised directly (`['fwd_grouped','attention','zz']`, 16 for 32 symbols).
- `read_status['read_reason']` is always populated before `experience_integration.py:1119` reads it (`_set_read_status` sets both keys on every path), and `cli.py:1529` uses `setdefault`, so `alias_hit` survives to the persisted result.
- Grepped every consumer of the warm-start `read_reason`: nothing outside `experience_reader`/`experience_integration`/`cli` branches on `"hit"`. `rewrite_by_flydsl/kb.py:157,405` is the flydsl reader's own field, untouched by this change.
- Tier ordering is alias-before-fuzzy as described, and `aliased_identity` returns `None` on an empty `implementation_signature`, so a name-only match can never hand a record over.
- CHANGELOG entry and the `read_reason` list in `docs/kernelforge/how-to/autonomous-loop.md` are both updated and match the diff.
- No import cycle: `experience_sink` and `experience_reader` both import cleanly with `identity_alias` at module scope.
- Tests could not be run here (Windows host; `kernelforge.tracker` imports `fcntl`), so the assertions above come from reading the code plus the stubbed repros.

chennyiiis:
@xiaofei-zheng thank you — issue 1 decided the design. The alias tier is gone; this PR now does something much smaller. Force-pushed `61a703bd7` → `0c2d261c6`, so the line anchors on your comments are stale.

**What your repro turned out to be a sample of.** I ran a read-only scan of the production store (`producer=forge-loop`, 144 pages, 106 distinct names) looking for one `implementation_signature` spanning several `kernel_name`s. 18 groups. **One** is a rename. The other 17 are your case — distinct operators sharing a source file and therefore a file-level signature:

| group | names sharing one signature |
|---|---|
| `vllm/_aiter_ops.py` | 6 operators (`rocm_aiter_gemm_a8w8_blockscale`, `rocm_aiter_act_mul_and_fp8_group_quant`, `rocm_aiter_fused_allreduce_rmsnorm{,_decode,_prefill}`, `rocm_aiter_triton_gemm_a8w8_blockscale`) |
| `aiter/fused_moe.py` | `moe_flydsl_stage1`, `moe_flydsl_stage2` |
| mxfp4 MoE | `mxfp4_moe_2stage_t16` / `t32` / `t64` / `t128` |
| … | `ck_moe_stage1`/`stage2`, `rmsnorm`/`add_rmsnorm`, `mxfp8_linear`/`mxfp8_quant`, … |

So the blast radius of signature-equality-as-an-alias-test is the majority structure of the store, not a corner case, and the benefit it was buying is one fork whose two pages hold the *same* 1.4439 champion — a wasted campaign budget, not a lost speedup. Issues 2 and 3 were real too; I'm not fixing them, because the code they're in is deleted.

**What the defect actually was.** The one genuine rename is `kdapackeddecodekernel` vs `kda_packed_decode` — a camelCase/snake_case pair. `normalize_operator_name` lowercased, replaced `.`, collapsed separators and stripped a trailing `_kernel`, but never split camel case. The task contract *asks* the agent for the camel spelling (`aiter::fusedAddRmsNorm`) so upstream PR search has term boundaries; that instruction was silently costing the operator its page. Splitting camel boundaries before lowercasing collapses both spellings onto one address, with no probing and no cross-operator reach.

**The third design, also rejected on the data.** Before landing this I costed "derive `kernel_name` from the file's kernel symbols instead of from prose". Re-deriving all 144 pages that way: 21 anchored, 49 drift, 74 have no recorded symbols. The drift is disqualifying — `implementation_symbols` holds GPU kernels while the operator a campaign targets is usually the host API, so `unified_attention_with_output` becomes `apply_softcap`, `mxfp4_moe_2stage_t{16,32,64,128}` collapse to one name and `custom_all_reduce_tp2`/`tp4` to another. `operator_name` names the right thing; only its normalization was lossy.

**Migration.** Of 106 distinct names, 9 contain an unbroken 11+ letter run and 4 of those are camel-squashed (`kdapackeddecodekernel`, `topkgatingsoftmax`, `hipbsolgemm`, `hipgraphlaunch_*`); the rest (`bpreshuffle`, `unquantized`) are lowercase in source and normalize to themselves. `kdapackeddecodekernel` merges into the existing `kda_packed_decode` — the reported fork, closed at no cost. The other three strand one record each.

All four CodeQL cyclic-import alerts were on the deleted imports and are gone. Suite on crsuse2-m2m-069: `4351 passed`; the 8 `*_gpu.py` failures reproduce identically on an unpatched tree (`No module named 'torch'`).

xiaofei-zheng:
Reviewed at `35a8246b8` against merge base `089228c52`.

Two blocking issues.

**1. `task_publisher.py:95` — a non-string `identity.framework` crashes the publication sweep instead of rejecting the draft.**

The loop directly above (`task_publisher.py:75-79`) exists because this is the untrusted agent boundary: it skips any dimension whose value is not a `str`. The new line bypasses that guard and hands the raw value to `detect_framework`, which does `(framework_override or "").strip()`:

```
>>> detect_framework("", framework_override=["sglang"])
AttributeError: 'list' object has no attribute 'strip'
```

`_normalize_agent_task_payload` is called at `task_publisher.py:204`, inside a `try` that catches only `(OSError, json.JSONDecodeError, ValueError)` (`task_publisher.py:217`). `AttributeError` escapes `publish_staged_task`, then `publish_complete_staged_tasks`, and reaches `opportunity_agent.py:533`, where it is outside any `except` — so one draft with `"framework": ["sglang"]`, `"framework": 1` or `"framework": true` aborts the whole publication round and takes down the analysis session, discarding every other staged task in that sweep. The `finally` at `opportunity_agent.py:563` calls the same function again and raises a second time.

Trigger: agent-authored JSON only; no valid draft hits it. But the type guard two lines up is the module's own statement that this input is not trusted to be a string. Before this PR the same payload produced a `TaskPublicationResult` rejection.

Fix: only fold when the value is a `str`, or catch `AttributeError`/coerce with `str(...)` the way `canonical_framework_version` already does at `task_publisher.py:86`.

**2. `CHANGELOG.md` — three of the four observable behaviour changes have no entry.**

`AGENTS.md` requires an `[Unreleased]` entry in the same PR for anything an operator can observe. The PR adds exactly one entry (`CHANGELOG.md:398`), covering the camel-case split in `normalize_operator_name`. Missing:

- `canonical_framework_version` — `0.24.0+rocm723`, `v0.24.0` and `0.5.15.post1.dev...` now resolve to the release, and `none`/`unspecified`/`unknown_version` fold to `unknown`. Existing pages stop being addressable under their old spelling; this is the same class of change as the one that did get an entry.
- `task.py:132` — a draft whose `identity.backend` is not in `KERNEL_BACKENDS` is now refused. Previously published; now the operator is dropped with a rejection. Also means an already-published `task.json` carrying an unregistered backend no longer loads through `load_task`.
- `task_publisher.py:95`/`:145` — the framework override is folded at publication, and a framework that none of the task's paths sit under is now refused.

The PR description covers all four; the changelog only covers one.

---

What was checked: `normalize_operator_name` and `canonical_framework_version` outputs against every case the new tests assert, plus idempotency of the version fold on unparseable input; `rank_fallback_identities` ordering with the empty `version_distance` sentinel and the removal of the early `target_version is None` return; the override semantics of `detect_framework` vs `infer_source_owner_framework` (they agree, so the publisher's fold matches what the run resolves); `NO_FRAMEWORK_VERSION`/`UNKNOWN_VERSION` have no remaining references after removal; `KERNEL_BACKENDS` is the same list `opportunity_agent.py:418` injects into the prompt, so the backend refusal is one the agent can satisfy. `test_implementation_identity.py` and `test_warmstart_identity.py` pass locally (27 passed); the controller tests need `fcntl` and were left to CI.

chennyiiis:
@xiaofei-zheng both blocking issues fixed, squashed into the commits that introduced them. `35a8246b8` → `5f0b424da`, so line anchors are stale again; sorry.

**1. Non-string `identity.framework`.** You're right that the guard two lines up is the module's own statement about this input, so the fold is now guarded the same way rather than coerced:

```python
declared_framework = identity.get("framework")
if isinstance(declared_framework, str):
    identity["framework"] = detect_framework("", framework_override=declared_framework)
```

Guarded rather than `str(...)`-coerced deliberately: a draft that wrote a list here is one `parse_task_payload` refuses by name a few lines later, and coercing would publish `"['sglang']"` as a framework instead. Regression test added — `test_a_framework_that_is_not_even_a_word_is_refused_with_the_sweep_intact` stages two drafts, breaks one, and asserts the sibling still publishes. With the `isinstance` removed it reproduces your `AttributeError: 'list' object has no attribute 'strip'` at `experience_sink.py:80`; with it, `results["broken"].published is False` and `results["intact"].published is True`.

**2. CHANGELOG.** Three entries added, each squashed into the commit that owns the behaviour, per the one-concern-per-commit rule:

| commit | entry |
|---|---|
| `ee7fc12e6` | One framework release opened a page per spelling of its version |
| `0e6ab3bee` | A backend the run cannot build with was published as if it could |
| `5f0b424da` | A kernel belonging to no package was addressed two ways at once |

---

**Correction to my 09-17 comment, which the store disagrees with.** I wrote that `kdapackeddecodekernel` "merges into the existing `kda_packed_decode` — the reported fork, closed at no cost". That is wrong about what merges. The stored name is already lowercased and squashed, and no normalization can recover boundaries that were destroyed before the write:

```
KdaPackedDecodeKernel   main=kdapackeddecodekernel   this PR=kda_packed_decode
kdapackeddecodekernel   main=kdapackeddecodekernel   this PR=kdapackeddecodekernel
```

So the squashed page does not move and is not reclaimed. What changes is where the *next* campaign lands: one reporting `KdaPackedDecodeKernel` now addresses `kda_packed_decode`, which already holds records, instead of minting a second page. The fork stops widening; it is not retroactively closed. For `TopKGatingSoftmax` and `hipBSolGemm` the new address has no page behind it today, so those pay one cold start and are correct from then on. All four strand their existing record either way — the difference is only whether the correct address is already populated.

**Read-only scan of the production store, today.** `producer=forge-loop`, 147 pages, 108 distinct names, run against this branch's normalizers versus `origin/main`'s:

| dimension | live | what this PR does to the existing pages |
|---|---|---|
| `producer` | 1 value, `forge-loop` ×147 | all pass the new equality check |
| `kernel_name` | 108 values | **0 pages move.** 7 stored names are not their own normal form (`*_kernel`), and `main` strips that suffix identically — pre-existing on both sides |
| `framework` | aiter 54, vllm 52, sglang 36, torch 4, unknown 1 | all 5 fold to themselves; 0 pages move |
| `framework_version` | 17 values | **87 of 147** are not canonical; **147 addresses → 136, 11 groups merge** |
| `backend` | triton 81, hip 32, flydsl 20, ck 11, aiter 3 | the new refusal rejects **0** stored values |
| `gpu` | 1 value, `mi355x` ×147 | untouched, out of scope here |

Nothing is rewritten and there is no migration step — the change is to the address a future campaign computes and to the values new records carry.

The 11 merging groups, 9 of which share one `implementation_signature` and are therefore provably one kernel ported twice by two campaigns that could not see each other:

| kernel | signature | stored spelling → champion speedup |
|---|---|---|
| `aiter_mxfp4_moe_2stage` | same | `unspecified` 1.310 ↔ `unknown` 1.320 |
| `moe_flydsl_stage2` | same | `unspecified` 1.035 ↔ `unknown` 1.028 |
| `paged_attention_rocm` | same | `unspecified` 1.546 ↔ `unknown` 1.494 |
| `rmsnorm` | same | `unspecified` 1.158 ↔ `unknown` **1.495** |
| `unified_attention_with_output` | same | `unspecified` **2.330** ↔ `unknown` 2.024 |
| `fused_moe` | same | `0.24.0+rocm723` **1.465** ↔ `v0.24.0` 1.444 |
| `mhc_fused_post_pre` | same | `0.24.0+rocm723` **3.363** ↔ `v0.24.0` 3.276 |
| `sparse_attn_prefill_ragged` | same | `0.24.0+rocm723` 8.232 ↔ `v0.24.0` **13.729** |
| `mxfp8_grouped_gemm` | same | `0.5.15.post1.dev20260724+g3d91a569ce` **4.177** ↔ `v0.5.15.post1-rocm720-mi35x-20260724` 3.389 |
| `dynamic_per_tensor_quant` | differs | `unspecified` 2.817 ↔ `unknown` **10.449** |
| `unified_attention_with_output` | differs | `0.24.0+rocm723` 2.571 ↔ `v0.24.0` **11.584** |

`sparse_attn_prefill_ragged` is the clearest cost of the defect: one half of the fork has been reading 8.2× while 13.7× sat on the other page.

**Verification.** Full suite on WSL Ubuntu 22.04 against merge base `089228c52`, nothing stubbed:

| | `origin/main` | this branch |
|---|---|---|
| passed | 4322 | **4344** |
| failed | 7 | 7 — the same seven ids |
| skipped | 14 | 14 |

The seven are `test_rename_completeness.py` ×5, `test_lane_aiter_cache.py::test_a_session_outside_a_lane_keeps_the_process_environment`, and `test_vendored_kb_store_client.py::test_vendored_sdk_matches_upstream_git_blob`; all seven are red on an unmodified `origin/main`. `ruff check src/` clean. The `4351 passed` figure in an earlier revision of the description came from a devbox run and is superseded by the numbers above.

chennyiiis:
Done — the backward-compat read is removed, and the change list is ready for the store migration.

## What changed in the code

`rank_fallback_identities` used to run `canonical_framework_version` over the *stored* `framework_version` of every candidate as well as the target's. That is the line you asked about, and you were right to ask: it let a page addressed `v0.24.0` keep serving warm starts while every write went to `0.24.0`. One kernel, two pages, one of them frozen — and nothing reported the split, because reads still succeeded.

It now resolves the **target only**. The target arrives raw (installed distribution metadata says `0.24.0+rocm723`, an image tag says `v0.24.0`), so it has to be resolved. A stored dimension was already resolved when its page was addressed, so it is read as written. A stored spelling that names no release ranks as unparseable and is dropped rather than folded.

One commit, `227e19f5a`, with its own `[Unreleased]` entry.

## The change list

Read-only scan of `producer=forge-loop` on 2026-09-21: **147 pages, 87 of them move, carrying 605 sessions.**

| action | pages | sessions | what it means |
|---|---|---|---|
| `RENAME` | 68 | 316 | target address is empty — move the page |
| `MERGE_INTO_EXISTING` | 6 | 109 | target address already holds a page — fold sessions in |
| `MERGE_WITH_PEER` | 13 | 180 | two or more moving pages land together — fold them |

Only `framework_version` moves. `producer`, `framework`, `backend` and `gpu` are unchanged for every one of the 147, verified by running both `origin/main`'s normalizers and this branch's over every stored address.

Seven pages also have a non-canonical `kernel_name` — a trailing `_kernel`, e.g. `mxfp8_grouped_gemm_kernel`. That is **pre-existing**, not from this PR; the normalizer has stripped that suffix since before it, so those pages were already unaddressable. They are in the list because the address is wrong either way — whether you fix them is your call.

The per-page rows, with `old_canonical_id` and `new_canonical_id` in full, session counts and champion speedups: I have them as a CSV and can attach it here or send it wherever you prefer.

## Ordering — please migrate before this merges

The version rewrites split into two kinds:

| pages | stored spelling | canonical | reachable between the merge and the migration? |
|---|---|---|---|
| 25 | `v0.26.0` | `0.26.0` | yes — PEP 440 drops a tag `v`, rank unchanged |
| 15 | `v0.24.0` | `0.24.0` | yes — same |
| 18 | `0.24.0+rocm723` and other local segments | `0.24.0` etc. | yes, demoted but still found |
| **22** | `unspecified` | `unknown` | **no** |
| **6** | `v0.5.15.post1-rocm720-mi35x-20260724` | `0.5.15.post1` | **no** |
| **1** | `none` | `unknown` | **no** |

So **29 of the 87 go dark the moment this merges** and stay dark until migrated — their spelling names no release at all, so nothing can rank it. Nothing is lost; the records sit where they are and come back the moment the address is fixed. But it is a real cold-start window, so the order should be **migrate the store, then merge this PR**.

## Two merges worth a look before you fold them

The surviving page's champion changes, and in both cases the smaller page holds the better result:

| kernel | page A | page B |
|---|---|---|
| `sparse_attn_prefill_ragged` @ vllm `0.24.0` | `0.24.0+rocm723` — 46 sessions, 8.232x | `v0.24.0` — 13 sessions, **13.7293x** |
| `unified_attention_with_output` @ vllm `0.24.0` | `0.24.0+rocm723` — 27 sessions, 2.5712x | `v0.24.0` — 15 sessions, **11.5837x** |

A third, `mxfp8_grouped_gemm` @ sglang `0.5.15.post1`, folds three pages (21 sessions / 4.1771x, 7 / 3.3885x, and the `_kernel` page at 1 / 1.6189x).

Every number above came from `search_identities` / `get_top_sessions` / `get_record`. Nothing was written to the store from here.

Suite: **4345 passed, 7 failed, 14 skipped** — the seven are the same seven that fail on `origin/main` (`test_rename_completeness` x5, `test_lane_aiter_cache`, `test_vendored_kb_store_client`) and none is reachable from this change. `ruff check` clean.

xiaofei-zheng (approved):
Re-reviewed at `227e19f5a` (force-pushed since the previous review; three commits rewritten, one added). **No blocking issues — LGTM.**

Both issues from the previous round are closed, and the new fifth commit checks out.

What I verified:

- **Non-string `identity.framework` no longer aborts the sweep.** `task_publisher.py:95-98` now guards with `isinstance` instead of coercing. Ran `_normalize_agent_task_payload` directly with `["sglang"]`, `1`, `True`, `None` and `{"a": 1}`: all pass through untouched, none raises. Downstream refusal confirmed: `TaskContractError` subclasses `ValueError` (`contracts.py:51`), so `_identity` wrapping the `_validate_identity_segment` failure lands in the `except (OSError, json.JSONDecodeError, ValueError)` at `task_publisher.py:223` and becomes a `TaskPublicationResult` rejection. `test_a_framework_that_is_not_even_a_word_is_refused_with_the_sweep_intact` stages a sibling draft and pins the sweep surviving. Guarding rather than coercing is the right call — `str(["sglang"])` would have published `"['sglang']"` as a framework.
- **Changelog now carries one entry per observable change** — version canonicalization, the stored-version read, the backend refusal, the framework fold and path check — all under `[Unreleased]`, alongside the camel-case entry that was already there.
- **Fifth commit (stored version read as written).** `warmstart_identity.py:130` drops the fold on the candidate side; `canonical_framework_version` is now applied at exactly three write points (`kernel_identity.py:62`, `rewrite_by_flydsl/identity.py:63`, `task_publisher.py:86`) plus the ranking target (`warmstart_identity.py:104`), with no stored value resolved on the way in. Traced the branch arithmetic: `unknown`/`unknown` matches on string equality (affinity 3, empty `version_distance`, which sorts ahead of `(0,0,0)`); `unspecified`/`none` fail `Version()` and are dropped; `0.11.3+rocm723` still parses and ranks at affinity 2 under the release. `v0.11.3` ties with the release because PEP 440 drops a tag `v` — the test comment says so explicitly, so the asymmetry is pinned rather than accidental.
- **The migration dependency is stated, not glossed.** 29 pages go dark on merge, 605 sessions across 87 pages need to move, and the PR says the store migration lands first. That is the PR's call to make and it is written down in both the description and the changelog.
- **PR description matches the current diff**, including the "Review follow-up" section and the updated suite numbers (4345 vs 4322).

Tests: `test_implementation_identity.py` + `test_warmstart_identity.py` pass locally (28 passed). `ruff check` clean, `ruff format --diff` reports no change over `src/kernelforge`. The controller tests import `fcntl` and cannot run on this host; left to CI, which the PR reports green at 282 passed for the controller package.

PR #1533, opened 2026-09-16
fix(fusion): salvage the campaign keepers a killed run already exported

BaoYunkai:
## Problem

`fusion_manifest.json` is the only artifact that points at a keeper, and it
landed once every campaign had returned. `on_keep` exports
`fusion_<pattern_id>.patch` and smokes it the moment a recipe is kept, so a
wrapper killed between the last keeper and the aggregate reported REVERT with
`patch: null` while proven, already-smoked work sat in the output dir.

Session `20260916T050331Z-94ee8477` lost a 5.011x fusion of qkvgate split + QK
norm + RoPE that way: correctness passed at 57dB SNR and the iteration
published at 10:26, then the 5400s timeout fired at 11:10 with the next
experiment still running.

Rebuilding the rows in the consumer cannot recover it. A diff header is
repo-relative while every consumer expects the absolute `recipe.source_file`,
and the env flag that gates the fused path is not in the exported artifacts at
all -- without it integrate boots the re-baseline server un-gated, measures the
eager path and REVERTs, one stage and ~25 min later.

## Fix

Producer-side incremental publish, not consumer-side salvage fallback.

Hoist `build_manifest` + `write_manifest` into one `publish()` and call it from
`on_keep` after the smoke passes (`_publish_partial_nomination`). The aggregate
is then never missing, only as complete as the run got, carrying
`fusion_loop.kept`, `termination_reason=in_progress`, an absolute `target_file`
and the `env_flag` -- and the existing salvage path in `forge_fusion.py` reads
it unchanged, with no Hyperloom-side change.

The end-of-run `publish()` still overwrites it with the real loop /
compile-pass / error fields before any exit, so nothing partial survives a
normal run. A sibling the serving smoke rejected has its
`fusion_<pattern_id>.patch` unlinked there, so no reader can find a patch the
run itself refused.

## Test plan

- [x] `src/kernelforge/tests/fusion/test_partial_nomination.py` (new)
- [x] `src/kernelforge/tests/fusion/` (605 passed)


ZhengGong-amd:
Reviewed at `a9a36b2db`. The gap is real: the multi-patch path never writes a salvage checkpoint (`_export_salvage_patch` / `_write_kernel_keep_checkpoint` only run inside `apply_serving_gate`, which `command.py:1421` skips when `multi_patch`), so a kill between the last campaign and `fusion_manifest.json` loses proven work. But please don't rebuild the nomination rows in the consumer — as written, this does not recover the fusion.

## It doesn't work today

Loading this PR's `forge_fusion.py` and pushing the salvaged result through the real contract (`parse_outcome`):

```
env_flag    = ''                                        # -> integrate sets fusion_env_flags = {}
target_file = 'python/sglang/srt/layers/attention.py'   # relative, not absolute
```

`env_flag` is not in the rows, and it never reaches `forge_loop_<stem>.json` in the first place. Per `NominatedPatch.env_flag` and `_recipe_patch_envelope`, integrate builds `fusion_env_flags` from it; empty means the re-baseline server boots un-gated, measures the eager path, and REVERTs. The 12.0x is still lost — one stage later, after ~25 min of serial integrate.

Four more that all follow from reconstructing instead of reading:

- **`target_file` shape.** The diff header is repo-relative (`emit.py::_rel`, which falls back to a bare filename); every consumer (`apply_kernel_patch`, `record_source_path`) expects the absolute `recipe.source_file`. The PR's own tests pin both shapes: `/fw/a.py` on the manifest path, `qk.py` on this one.
- **`improved` is not the keep gate.** `campaign.py:133` keeps at `>= keep_bar` (default 1.03); `improved` is `>1.0 and not aggregate_regression`, and `cli.py:1900` states those two "can legitimately disagree by a hair". A 1.05x keeper with `aggregate_regression=True` is silently dropped — the exact case this PR exists for.
- **Serving-crash veto bypassed.** `on_keep` exports the patch *before* the smoke and does not delete it on `serving_crash` (`command.py:1386-1393`), while `improved` stays true. This re-nominates patches the run rejected for crashing a real server boot, which is what `_clear_kernel_keep_checkpoint` exists to prevent.
- **Stale artifacts.** `fusion_*.patch` / `forge_loop_*.json` are not in `main()`'s stale sweep, and kernelforge only unlinks the loop json of the stem it is about to run. Output dirs are reused (`session_dir/runs/fusion/<task_id>`), so a run killed before a campaign starts salvages the previous run's patch as its own — precisely what that sweep's comment forbids.

Also 5 silent `continue`s in the scan, against `nomination_result.py`'s "every drop carries one of these instead of vanishing into a log line" — in code whose whole purpose is not losing proven work.

## The fix: write the manifest incrementally

The bug is "the manifest is written once at the end", so write it after every keeper. `write_manifest` is already atomic (`durable_io.atomic_write_text`). The aggregate is then never *missing*, only as complete as the run got, and the **existing** salvage path covers this case with **zero Hyperloom changes**.

In `command.py`, hoist the `build_manifest` + `write_manifest` block (~908-928) into one local `publish(patches, **final)` and thread it down through `_run_multi_patch_nomination` into `_run_fusion_autoloop`, called from `on_keep` after the smoke passes:

```python
# on_keep, right before `return patch`
kept.append(patch)                       # the RecipePatch just built
kept.sort(...)                           # strongest first, same rule as LoopResult.patches
_mirror_legacy_patch(kept[0].patch_path, out)
publish(
    [_recipe_patch_envelope(p, repo_root=repo_root) for p in kept],
    artifacts=FusionArtifacts(patch=kept[0].patch_path, repo_root=repo_root),
    loop={
        "kept": True,
        "best": {"kernel_speedup": kept[0].micro_speedup},
        "best_env_flag": kept[0].env_flag,
        "termination_reason": "in_progress",
    },
)
```

Invariants that make this the one correct shape:

1. **One writer, one mapping, one ordering rule.** The end-of-run call stays as it is and overwrites with the real `loop_manifest` / `compile_pass` / `error`; nothing partial survives a normal exit, because line 928 runs before every `SystemExit`. Do not add a second envelope builder alongside `_recipe_patch_envelope`, and do not invent a second "strongest first" rule alongside `loop.py:429-440`.
2. **The partial must be honest and salvageable.** Salvage reads exactly `fusion_loop.kept`, `fusion_loop.best.kernel_speedup`, `fusion_loop.best_env_flag`, `artifacts.patch` (must exist on disk), `artifacts.repo_root`, `fusion.source_file`, `patches[]`. Satisfy those and nothing else is needed.
3. **The crash veto stays with the writer.** `serving_crash` already returns `None` from `on_keep`, so that sibling is simply never published — and unlink its `fusion_<pattern>.patch` there so no reader can find it.
4. **Stale is already handled.** `fusion_manifest.json` and `fusion.patch` are both in `main()`'s pre-run sweep.

Free with this shape: `env_flag`, absolute `target_file`, `snapshot_dir`, `base_commit`, the real keep gate, compile-pass keepers, and a hard-crashed forge-fuse (not just a timeout) reporting its keepers instead of REVERT.

## To delete from `forge_fusion.py`

`_campaign_patches_on_disk`, `_patch_target_file`, `_DIFF_TARGET_RE`, `_SHADOW_EXPERIMENTS_MARKER`, `import re`, and the new branch in `salvage_forge_fusion_from_workspace`. Note that `kernel_repo` via `best_manifest.split("/forge_experiments/")` is only correct because `CampaignConfigStore.root == workspace/forge_experiments` *and* `shadow.root == Path(repo_root).resolve()` — two unrelated facts coupled by a substring, and silently `""` when the key is absent. The producer already has the value as `kernel_repo`.

## Tests

All three fail against the current implementation:

1. a salvaged row carries `env_flag` and an absolute `target_file`;
2. a sibling whose serving smoke crashed is never salvaged;
3. artifacts left by a previous run in the same output dir are not salvaged as this run's KEEP.

Docs: if the set of salvage sources changes, `salvage_forge_fusion_from_workspace`'s docstring ("pre-smoke checkpoint + patch") and the artifact table in `docs/kernelforge/how-to/kernel-fusion.md` both need updating. With this approach the set doesn't change — one more reason to prefer it.

haishuok0525 (commented):
## What this PR does

Producer-side incremental publish, not consumer-side salvage fallback.

`forge-fuse` folds `build_manifest`/`write_manifest` into `publish()` and calls it from `on_keep` after each sibling passes serving smoke (`_publish_partial_nomination`), so `fusion_manifest.json` exists as soon as a keeper is proved (`fusion_loop.kept`, `termination_reason=in_progress`, absolute `target_file` + `env_flag`). End-of-run `publish()` overwrites with the final fields. Smoke-rejected siblings unlink their `fusion_<id>.patch`. Salvage in `forge_fusion.py` is unchanged and can read the partial manifest.

Checked: `_publish_partial_nomination`, rejected-sibling unlink before legacy mirror, salvage shape compatibility on main, `test_partial_nomination.py`.

## Blocking issues

1. **PR body does not match the diff.** Fix section still describes falling back to per-campaign `forge_loop_<id>.json` + `fusion_<id>.patch`, keeping `improved` rows, and reading the target from the diff header. The commit and code do the opposite: keep writing the aggregate on each keep, and explicitly reject consumer rebuild from diff headers (need absolute `source_file` + `env_flag`). Tests path in the body (`src/hyperloom/agents/kernel/tests/test_forge_fusion.py`) is not what this PR adds (`src/kernelforge/tests/fusion/test_partial_nomination.py`). Session id in the body also disagrees with the commit. Please rewrite Problem/Fix/Tests to match HEAD.

2. **Missing `CHANGELOG.md` Unreleased entry.** Same habit as recent `fix(...)` merges; this changes timeout/salvage KEEP vs REVERT behavior and needs a Fixed bullet.

xiaofei-zheng (approved):
## What this PR does

Moves the fix to the producer: `build_manifest` + `write_manifest` are folded into one `publish()` closure (`command.py:673-712`), threaded through `_run_multi_patch_nomination` into `_run_fusion_autoloop`, and called from `on_keep` after the serving smoke passes (`_publish_partial_nomination`, `command.py:1495`). `fusion_manifest.json` therefore exists as soon as a keeper is proved, and the end-of-run `publish()` (`command.py:948-960`) overwrites it with the real loop / compile-pass / error fields. The consumer-side rebuild from diff headers is gone -- `src/hyperloom/` has zero changes in this diff.

## No blocking issues -- LGTM

Verified rather than skimmed:

- **Rejected siblings cannot be published.** `command.py:1473-1484`: `not_wired` / `serving_crash` unlink `arts.patch` and `return None` before `_publish_partial_nomination`, so the sibling reaches neither `smoked` nor `patches[]`. `_mirror_legacy_patch` is only called on the smoke-passed best (`command.py:1029`) and at the end-of-run pick (`command.py:807`), so a rejected patch cannot land in `out/fusion.patch` either.
- **One writer, one envelope, one ordering rule.** `publish` is the only caller of `write_manifest`; partial and final rows both go through `_recipe_patch_envelope` (`command.py:1047`); the sort at `command.py:1027` matches `loop.py:427-440` (`None` speedup sorts weakest).
- **Nothing partial survives a controlled exit.** Both `raise SystemExit` sites (`command.py:990`, `:997`) are downstream of the end-of-run `publish` at `:948`, and `click.UsageError` fires before the loop.
- **The partial manifest satisfies the salvage contract unchanged.** `fusion_loop.kept`, `best.kernel_speedup`, `best_env_flag`, `termination_reason=in_progress` (`command.py:1034-1039`), `artifacts.patch` / `artifacts.repo_root` (`:1033`, patch present on disk), `fusion.source_file` via `report.py:51`, and `patches[].target_file` = `recipe.source_file`, i.e. the absolute path `locate.resolve_framework_source_file` returns -- not the repo-relative diff header. Read against main's `salvage_forge_fusion_from_workspace:403-414` and `_normalize_manifest:222,296-338`: every field it reads is provided, `in_progress` is not in `INFRA_ABORT_REASONS`, and a `None` `nomination` / `micro_speedup` is already guarded (`forge_fusion.py:366`, `nomination_result.py:110-111,148`).
- **The CodeQL unused-`kept` finding is gone** with `forge_fusion.py` reverting to main.
- PR description and the new `CHANGELOG.md` Unreleased entry both match HEAD. `ruff check` / `ruff format --check` clean on the changed files; `test_partial_nomination.py`, `test_smoke_salvage_contract.py` and `test_loop.py` exercised locally (two Windows-only path/rename assertions fail on the reviewer's box; Linux CI is green).

PR #1536, opened 2026-09-16
fix(fusion): salvage the published best when a campaign never returns

BaoYunkai:
## Problem

`run_campaign` publishes each winning iteration to the shadow repo's
`forge_experiments/best/` and writes `forge_loop_<pattern_id>.json` pointing at
that manifest, but `on_keep` exports `fusion_<pattern_id>.patch` only after the
campaign returns and the loop gates the keeper. A wrapper timeout while the
campaign is still iterating therefore reports REVERT with `patch: null` even
though correctness passed and the best iteration is already on disk.

Session `20260916T050331Z-94ee8477` lost a 5.011x fusion of qkvgate split + QK
norm + RoPE this way: the experiment was still `running` when the 5400s
`forge-fuse` subprocess was killed.

Two things make a naive salvage lose the same win anyway:

- **The exported artifacts carry no env flag.** An authored fusion is
  env-gated, and `NominatedPatch.env_flag` -> `fusion_env_flags` ->
  `extra_envs` is the only way the flag reaches the re-baseline. Salvaged
  without it, integrate boots the server un-gated, measures the eager path and
  REVERTs -- the same win lost one stage later, after ~25 min of serial
  integrate.
- **The output dir is keyed on the fusion task**, so a rerun inherits the
  previous run's `fusion_<pattern_id>.patch` and `forge_loop_<pattern_id>.json`.
  The stale sweep named only the aggregate artifacts, so a run that times out
  before writing anything of its own could report the previous run's keeper as
  its result.

## Fix

1. Walk `forge_loop_*.json` for salvage, not only exported `fusion_*.patch`.
   When a recipe has no exported patch, read the patch path from the loop's own
   `best_manifest` under the shadow repo. Prefer an exported patch when both
   exist.

2. Fill each salvaged row's `env_flag` from `driver_<stem>.py`, which
   `run_campaign` renders beside the loop result it is already keyed on -- so a
   killed run still has the flag on disk even though the patch does not carry
   it. The top-level `env_flags` is filled from the strongest row, since this
   path has neither a checkpoint nor an aggregate to read it from. A campaign
   whose driver names no flag is skipped with a reason rather than salvaged,
   because queuing it is the failure this exists to avoid.

3. Before starting a new run, unlink `fusion_*.patch` and `forge_loop_*.json`
   in the output dir by the same globs salvage uses.

## Test plan

- [x] `src/hyperloom/agents/kernel/tests/test_forge_fusion.py` (86 passed). The
      three existing salvage tests asserted the rows without ever asserting a
      flag, which is why the gap was invisible; they now write the driver the
      producer writes, and two more pin the flag itself and the refusal.
- [x] `ruff check .`
- [x] `CHANGELOG.md` Unreleased / Fixed entry
- [x] Also carries a test-isolation fix CI surfaced on this PR:
      `test_handler_passes_non_fp8_geak_to_next_hyperloom_prereq` names the geak
      branch but the dispatch reads `KERNEL_OPT_BACKEND_ORDER`, so a neighbour in
      the same xdist worker could send it down the forge branch. Adding tests
      here moved it next to one, since pytest-split redistributes shards by
      recorded duration.


xiaofei-zheng:
Reviewed the salvage fallback and the stale sweep against the nomination contract on the consumer side. Two blocking issues.

**1. The salvaged nomination rows carry no `env_flag`, so a recovered fusion is e2e-measured un-gated and REVERTed**

`src/hyperloom/agents/kernel/tools/forge_fusion.py:439-449` builds each row with `kernel_name` / `patch_path` / `target_file` / `kernel_repo` / `micro_speedup` and no `env_flag`, and line 507 puts that list straight into `result["patches"]`. An authored fusion is env-gated: `NominatedPatch.env_flag` -> `enqueue_nominated_patch` -> `record["fusion_env_flags"]` -> `extra_envs` is the only way the flag reaches the re-baseline, and `request_handlers.py:1157-1159` states the consequence of it being unset — "the patch measures as the eager path and REVERTs a real win". `env_flags` at the top level of the salvaged result is empty on this path too (no checkpoint, no manifest), so nothing downstream can recover it.

Reproduced on this branch with the PR's own published-best scenario:

```
env_flags top-level: {}
row env_flag: ''
fusion_env_flags: {}
```

Trigger is exactly the case the PR exists for: wrapper timeout with no `kernel_keep_checkpoint.json` and no `fusion_manifest.json`, salvage falling through to `_campaign_patches_on_disk`. The 5.011x qkvgate fusion from session 20260916T050331Z-94ee8477 would be recovered, queued, measured on the eager path and rejected. The flag is recoverable from the same output dir — `run_campaign` writes `driver_<stem>.py` beside `forge_loop_<stem>.json` with the recipe's `ENV_FLAGS` — so the row can be filled from the same stem the loop result is keyed on.

**2. No CHANGELOG entry**

`CHANGELOG.md` has an `### Fixed` entry per user-visible fix and this diff adds none. The entry on the base branch describes a different mechanism ("the manifest is now published as each keeper is proved") and does not cover the per-campaign fallback or the stale sweep added here.

Checked and found clean: `_published_best_patch` resolves `patch_path` correctly — `BestResultPublisher` roots it at `<workspace>/forge_experiments` and the manifest sits at `best/manifest.json`, so `best_manifest.parent.parent / rel` lands on the published patch; `kernel_repo` split on `/forge_experiments/` yields the shadow repo, matching what `forge-loop` is given as `--workspace`. The stale sweep globs match the ones salvage reads and hit nothing else the run needs. `parse_outcome` accepts the rows and dedups by `kernel_name`. Ran `src/hyperloom/agents/kernel/tests/test_forge_fusion.py`: 83 passed, 1 failed — `test_salvage_reads_the_published_best_when_the_campaign_never_returned`, Windows-only (the `/forge_experiments/` marker never matches a `\`-separated path). CI is ubuntu-only, so not blocking.

xiaofei-zheng:
Re-reviewed at 57c2fc3. Both earlier blocking issues are resolved; one new one.

Verified fixed:

- Each salvaged row now carries `env_flag`, read back from `driver_<stem>.py` beside the loop result, and a campaign whose flag cannot be read is refused with a message on stderr rather than queued to fail at the re-baseline. Ran the published-best scenario end to end on this head: row `env_flag` `'SGLANG_FUSED_QKVGATE'`-equivalent, `result["env_flags"]` `{'...': '1'}`, `enqueue_nominated_patch` -> `fusion_env_flags` `{'...': '1'}`, and removing the driver returns `None`.
- The CHANGELOG entry is there and describes what the diff does, including the flag read-back and the stale sweep.

**Blocking: the description no longer matches the diff.**

The `## Fix` section still lists only the two items from the first push. Missing:

- The env-flag read-back (`_campaign_env_flag`, `forge_fusion.py:379-402`) and the behaviour decision that comes with it — a campaign whose driver names no flag is not salvaged at all. That is the largest part of the current diff and it changes what the fix does, not just how.
- `src/hyperloom/inference_optimizer/tests/test_kernel_request_handlers_units.py` (ad4d4bf), which is in the diff and is not about fusion salvage.

Also checked: `kept` is no longer assigned in the fallback branch, and nothing reads it after that point (`result["kept"]` is a literal `True`), so no behaviour change there. The stale sweep still does not clear `driver_*.py`, which is harmless — salvage only reaches a driver through a `forge_loop_<stem>.json` the sweep does clear. Ran `test_forge_fusion.py` (85 passed) and `TestRunGemmTuningHandler` (78 passed); the one failure, `test_salvage_reads_the_published_best_when_the_campaign_never_returned`, is the same Windows-only path-separator artifact as before and does not apply to the ubuntu CI.

xiaofei-zheng (approved):
Approving at 57c2fc3. No blocking issues in the code.

Verified on this head: each salvaged row carries the `env_flag` read back from `driver_<stem>.py`, and the published-best scenario reaches `enqueue_nominated_patch` with `fusion_env_flags` set, so a recovered fusion is re-baselined on the gated path; a campaign whose driver names no flag is refused with a reason on stderr instead of being queued to fail later. `_published_best_patch` resolves `patch_path` against the experiments root that `BestResultPublisher` writes it relative to, and `kernel_repo` split on `/forge_experiments/` yields the shadow repo forge-loop is given as `--workspace`. The stale sweep uses the same globs salvage reads by; not sweeping `driver_*.py` is harmless because salvage only reaches a driver through a `forge_loop_<stem>.json` the sweep does clear. `kept` is no longer assigned in the fallback branch and nothing reads it after that point. CHANGELOG entry matches the diff. Ran `test_forge_fusion.py` (85 passed) and `TestRunGemmTuningHandler` (78 passed); the single failure is a Windows-only path-separator artifact that does not apply to the ubuntu CI.

One thing to fix before merge, not gating the approval: the `## Fix` section still lists only the original two items. Please add the env-flag read-back and the fact that a campaign without a readable flag is not salvaged, and mention the `test_kernel_request_handlers_units.py` change (ad4d4bf).

PR #1537, opened 2026-09-16
Maidas integration

abdubey-amd:
## Summary
When a MAIDAS projection is supplied (`--maidas-projection-path`), derive Hyperloom's
analytical roofline from the projection instead of the built-in analytical model:

- **End-to-end throughput ceiling** comes from the projection's per-config decode
  latency, matched to the run's whole-server config (SoC, precision, parallelism,
  prefill/decode lengths, and concurrency == global batch).
- **Per-operator breakdown** (per-op time, FLOPs, bytes moved, arithmetic intensity,
  memory-vs-compute bound) comes from the projection's per-op sheet: the operators of
  the run's batch block, scaled by layer count, with hardware constants (HBM bandwidth,
  achievable FLOPS) read per-SoC.
- Both are stamped with provenance so the source (`maidas` vs built-in) is auditable.

## Safety / fallback
- The per-op breakdown is accepted only if it reconciles with the end-to-end ceiling
  within 10%; otherwise the built-in model is used.
- Any missing/มismatched projection data falls back to the existing analytical model,
  which is unchanged. The projection overrides only on an exact config match.

## Validation
- The two analytical views reconcile to <0.5% across concurrency 4/8/16/32.
- Apples-to-apples check on gpt-oss-120b (MI355X, MXFP4, TP1) against the built-in model:
  compute agrees closely (MoE FLOPs within 0.4%, projections exact); the difference is
  isolated to bytes moved (~1.6×), which — because decode is memory-bound — yields a
  ~2× tighter, more realistic ceiling. Measured `vllm bench serve` throughput lands at
  43–77% of the projection-derived ceiling vs 18–36% of the built-in one.
- Existing roofline and new projection unit tests pass.


tsrikris:
@chaojhou we require longer term alignment on the motivation of the PR. Having parallel paths for this and TraceLens is redundant. This path lacks a lot of the kernel shapes, launcher path, kernel path, etc. that can only be derived from the Trace. Recommend holding off any merge till alignment is reached.

zili-amd:
@tsrikris Agree. MAIDAS may be useful as an analytical prior, but it cannot replace the trace-derived kernel shapes, launcher path, and kernel path needed for actionable optimization. I also recommend holding this PR until its responsibility relative to TraceLens is agreed.

src/hyperloom/orchestrator/kernel/maidas_excel_ceiling.py
```diff
@@ -0,0 +1,364 @@
...
+    # it). Warn loudly when the xlsx workload cannot be reconciled with the served
+    # model so the mis-projection is visible rather than silent.
+    nm_served, nm_wl = _norm_model(served), _norm_model(row_workload)
+    if nm_served and nm_wl and nm_served not in nm_wl and nm_wl not in nm_served:
+        logger.warning(
+            "MAIDAS CROSS-MODEL projection: xlsx workload=%r does not match served "
+            "model=%r. The ceiling is for a DIFFERENT model (matched only on "
+            "hardware shape) and may be INVALID. Provide the MAIDAS xlsx for the "
+            "served model.", row_workload, served,
+        )
```
zili-amd: A detected model mismatch should fail closed. This PR documents a mismatched workbook producing a ceiling 3.5x below measured throughput. Logging a warning and still returning that row can directly corrupt headroom decisions. Please fall back to the native model here.

src/hyperloom/inference_optimizer/cli/parser.py
```diff
@@ -1011,6 +1011,15 @@ def _build_parser() -> argparse.ArgumentParser:
     )
 
     # Integration toggles.
+    opt.add_argument(
+        "--maidas-projection-path",
```
zili-amd: How is this XLSX expected to be produced and maintained by users of the public repo? The PR defines a reader, but not a versioned schema, producer provenance, supported model/config matrix, or distribution/update path. The current validation covers only a narrow configuration. Without this contract, the integration is difficult to reproduce and maintain.

src/hyperloom/orchestrator/kernel/conc_sweep.py
```diff
@@ -219,25 +229,50 @@ def _build_roofline_ceiling(
...
+                gpu_type=gpu_type, precision=precision, tp=num_gpus,
+                concurrency=c, isl=isl, osl=osl,
+            )
+            maidas_bd = maidas_breakdown_from_excel(maidas_path, rt_c)
+        if maidas_bd is not None and maidas_bd.peak_tok_per_sec > 0:
+            t_mem = maidas_bd.mem_tok_per_sec
+            t_cmp_c = maidas_bd.cmp_tok_per_sec
+            t_peak = maidas_bd.peak_tok_per_sec
+            bound_kind = maidas_bd.bound_kind
+            ceiling_source = "maidas"
```
zili-amd: This can mix MAIDAS and native ceilings in one sweep. Since the two models have different assumptions, the curve may contain source-induced discontinuities. Please use one source for the full sweep, or avoid comparing rungs produced by different models.

tsrikris:
@zili-amd the agreement between the teams right now is to deprecate this PR. There is a separate effort ongoing to address any variation.

For model level roofline, Hyperloom uses a custom roofline module [roofline_ceiling.py](https://github.com/AMD-AGI/Hyperloom/blob/a5257e63ca7b2cf35f1ac7187b44de6ee9476ea1/src/hyperloom/orchestrator/kernel/roofline_ceiling.py#L1127) which has severe limitations (incorrect perf models, limited coverage, etc.). Credit to [abdubey-amd](https://github.com/abdubey-amd) for identifying this issue. This module requires a complete overhaul to pull in TraceLens PerfReport's capabilities ([docs/how-to/generate-perf-report-pytorch.md](https://github.com/AMD-AGI/TraceLens/blob/main/docs/how-to/generate-perf-report-pytorch.md),  [docs/how-to/generate-perf-report-pytorch-inference.md](https://github.com/AMD-AGI/TraceLens/blob/main/docs/how-to/generate-perf-report-pytorch-inference.md)) and read out the perf CSVs, de-duplicate architecture specs (load the TL arch JSONs), remove the limited operation-level performance model coverage (2) Leverage the extensive perf model inside TraceLens.

PR #1541, opened 2026-09-17
fix(agents): route every agent CLI choice through the shared rule, on official key names only

fengshaoyi-amd:
- Description: what and why

  Three places still decided which agent CLI runs a role, and #1472 only converted one of them. This PR converts the rest, and retires the key names that made the three disagree in the first place — the cleanup that was going to be a follow-up (#1551) is folded in here, because widening a credential check and then retiring the name it was widened for are the same decision and should not be two reviews.

  **The orchestration backend never joined.** `cli/backends.py` was outside that PR's reach and still read the endpoint shape through `is_openai_only()`, so the coordinator read neither the credential nor the installed SDK and disagreed with the five roles that had been converted. One shape moves now that it falls through to `preferred_agent_backend()`: both sides configured with only the Codex extra installed runs orchestration on Codex instead of on a `ClaudeBackend` that cannot import its SDK.

  Both explicit overrides stay ahead of the shared rule, because each means the CLI has already rewritten one model id into the other's: handing `CodexBackend` a `codex_model` that now holds a Claude id would be wrong, so the rewrite drops that backend out of the running. A single-provider launch is therefore the one launch the ranked path is not consulted for, which is why no single-provider shape appears above.

  **The ranking was written twice**, in `preferred_agent_backend` and in the Forge registry's `select_default_agent_provider`, kept in step by two docstrings pointing at each other with nothing failing when they drifted. Both now sort on the new `llm_config.agent_backend_rank`; the registry's hardcoded `(1, 1)` "unrunnable" literal is derived from it, and its two `find_spec` probes delegate to `llm_config` (which retires `from importlib import util` there).

  **Deleted:** the `_official_anthropic_only` / `_official_openai_only` pass-throughs, plus their redundant local import of a module already imported at the top of the file; and the hand-rolled base-URL comparisons in `parser.py`. That last one was a narrower copy of `is_openai_only()` / `is_anthropic_only()` — it never read `OPENAI_API_KEY` — so an `ANTHROPIC_API_KEY` + `OPENAI_BASE_URL` launch defaulted `--claude-model` to a Codex model id. That is fixed as a consequence.

  **The two sides disagreed about what authenticates the OpenAI side, and it is settled downward, not upward.** `openai_agent_credentialed` recognised `OPENAI_API_KEY` while `codex_session` would also resolve `LLM_GATEWAY_KEY`, each carrying its own copy of the list. An earlier revision of this branch closed the gap by teaching the predicate the session's longer list; that was the wrong end. `_validate_credentials` has only ever admitted a run on `OPENAI_API_KEY`, so a host whose sole OpenAI-side credential is the gateway key exits at preflight and never reaches a ranking to be misrouted by — widening the predicate would have made the ranking agree with a name the front door rejects. Both name `OPENAI_API_KEY` now, and `OPENAI_AGENT_KEY_ENV_ORDER` goes with the widening it existed for.

  **`OPENAI_API_KEY` and the three Anthropic-side names are the only credentials the optimizer authenticates with.** Surveying the eight vendor and gateway aliases found a credential read path behind exactly one: `LLM_GATEWAY_KEY`, accepted by `resolve_openai_client_config` one rung above the Anthropic fallback and named by `codex_session` as the Codex key variable. `LLM_API_KEY` had one too — the robustness RCA resolver — and it went out with the agent itself when `main` retired robustness, so there is nothing left to remove for that name here. The other six (`AMD_API_KEY`, `AMD_LLM_API_KEY`, `GEAK_API_KEY`, `LLM_PROXY_API_KEY`, `CLAW_API_KEY`, the LLM-side `SAFE_API_KEY`) never authenticated anything in this repo.

  Two of the gateway key's read sites were a credential name doing a second job. `_provider_only_mode` treated a bare `LLM_GATEWAY_KEY` as proof that the host fronted both protocols, suppressing single-provider detection; that is now read off each side's own URL and key. TraceLens took the same variable as an on-its-own gateway marker, three lines above the `HYPERLOOM_STRICT_GATEWAY_MARKERS` hook that exists to express exactly that. The specialist secret allowlist also stops forwarding the name to a child that can no longer authenticate with it.

  **Operator impact.** A deployment whose `.env` carries `LLM_GATEWAY_KEY` as its only OpenAI-side credential has to rename it to `OPENAI_API_KEY`; value and endpoint are unchanged, and such a box was already being refused by `_validate_credentials` and by `install.sh`. An Anthropic-only host that also carried the gateway key now resolves its Anthropic bearer for OpenAI-protocol calls, which is the credential it was already using for everything else.

  **Nothing is mirrored outbound any more, and the reason it was kept did not hold.** An earlier revision of this branch went on leaving `LLM_API_KEY` / `AMD_LLM_API_KEY` / `LLM_GATEWAY_KEY` filled from `OPENAI_API_KEY` — by CLI preflight into its own environment, and by the kernel agent into the Ray runtime env — on the reading that they are how GEAK *receives* that credential. Checked against the component instead of assumed, that is not true. `AMD-AGI/GEAK@0e02ad3` returns no hits at all for `LLM_API_KEY`, `AMD_API_KEY` or `LLM_GATEWAY_KEY`, and one for `AMD_LLM_API_KEY`: a quick-start line invoking a `scripts/run-docker.sh` the repository no longer carries. GEAK's AMD gateway key is `GEAK_AMDKEY`, declared in `interface/runtime/engine/registry.json` with the APIM header and the `llm-api.amd.com/Unified` base URL; its `provider_autoselect` triggers are `GEAK_AMDKEY` and `OPENAI_API_KEY`, and the Anthropic side reads `ANTHROPIC_*` / `CLAUDE_CODE_OAUTH_TOKEN`. So both sites were copying a live credential into names with no reader on either side of the boundary. Both are gone, and the four names leave `SAFE_ENV_KEYS`, so an operator's own export no longer reaches a worker either.

  **Three more allowlist entries go with them.** Having established that the mirrored names had no reader, the same count was run over every remaining entry in `SAFE_ENV_KEYS` — against this repository and `AMD-AGI/GEAK@main`, the ref `install.sh` resolves by default, with the allowlist itself, the redaction tables and the `.env` strip lists excluded, since membership there is not consumption. `LLM_PROXY_BASE_URL` and `GEAK_WORK_DIR` each occurred exactly once in the whole repository: the allowlist line. Nothing wrote them, nothing read them, zero hits in GEAK — and `GEAK_WORK_DIR` carried a comment describing a harness contract GEAK does not have, under any spelling. `LLM_PROXY_API_KEY` was forwarded to every worker with nothing setting or reading it. `GEAK_API_KEY` / `GEAK_BASE_URL` stay: both are documented operator overrides, and `preflight.py` reads the URL to rewrite the endpoint in GEAK's config file, so GEAK receives that one through the yaml rather than through the environment. The allowlist comment claiming the e2e runner reads them from the environment is corrected instead.

  **What this deliberately does not touch.** `LLM_API_BASE` stays — it addresses an OpenAI-protocol endpoint rather than authenticating to one, and preflight still derives it from the resolved OpenAI-side URL. The retired names all stay in the benchmark secret, variant, external and specialist-redaction lists and in the `.env` strip lists: membership there asserts that a name holds a secret worth scrubbing, not that anything consumes it, and an operator's stale export still needs stripping from a child process. Separately, `GEAK_SCORE_TARGET`, `GEAK_SKIP_PROFILE`, `GEAK_MAX_BENCHMARK_SHAPES` and `GEAK_RUN_MODE` also return zero hits in GEAK, yet `install.sh` writes each into the dotenv with a documented meaning. They are run knobs rather than credentials, so whether they are dead belongs with whoever owns the GEAK side, not with this PR.

  **Docs follow the removal.** `docs/reference/authentication.md` told an operator to expect a credential under `LLM_API_KEY` / `AMD_LLM_API_KEY` in four places. Preflight fills neither name now; `LLM_API_BASE` is the only value it still derives, and kernel tools receive the OpenAI-side credential under `OPENAI_API_KEY`, which is the name that crosses the Ray boundary.

  `is_openai_only()` / `is_anthropic_only()` are kept for the questions that really are about endpoint shape — the critic's wire protocol, whether ProposalScorer can run, and the model defaults. Those are not "which SDK", and folding them in would break `--codex-model` on an unconfigured box.

- Linked issue(s): follow-up to #1472; supersedes #1551

  `test_llm_config.py`: the gateway-key credential case and its row in the `preferred_agent_backend` ranking table now assert the retirement (uncredentialed on both sides, so the SDK decides and Claude keeps the tie), and the error-message test no longer expects the retired name. `test_cli_backends_unit.py`: the openai-only orchestration case sets a key rather than asserting on a shape preflight refuses; the forced-`openai` protocol case now expects a gateway-only host to be rejected; and two cases cover the dual-protocol gateway, the one shape that both launches and leaves the shared rule something to decide — credentials tie so Claude wins, and with the Claude SDK absent the SDK key decides for Codex. The autouse isolation fixture pins the SDK probes so the suite's own installed extras cannot change the answer. `test_codex_session.py`, `test_tracelens_model_normalize.py`, `test_specialist_codex_backend.py` and `test_anthropic_credential_registry.py` follow the retirement at their own layers.

  Run locally on Windows: `pytest src/hyperloom/inference_optimizer/tests/test_llm_config.py src/kernelforge/tests/test_provider_registry.py src/hyperloom/agents/kernel/tests/test_codex_session.py` -> 241 passed, 4 failed; the same 4 fail unchanged on `origin/main` (POSIX mode bits and path separators). `test_cli_backends_unit.py` cannot be collected on Windows at all — `cli/kb.py` imports `fcntl` — so CI covers its cases. `ruff check` and `ruff format --check` clean.

- `docs/release-notes.md` entry under `[Unreleased]`: no — the release cut aggregates PR descriptions, so this branch leaves that file untouched and the diff is described here instead.

- Breaking changes: yes, for one operator shape. A deployment whose only OpenAI-side credential is `LLM_GATEWAY_KEY` must rename it to `OPENAI_API_KEY`. That deployment could not start before this PR either — `_validate_credentials` refuses it — so nothing that runs today stops running. Separately, one environment changes which CLI runs orchestration (both sides configured, only the Codex extra installed) and it gains a backend that can import its SDK where it previously had one that could not.

- PR addresses single concern: yes — which agent CLI runs a role, and which names may credential it, are decided in one place. #1551 was opened as a separate PR for the second half and is folded back in here: it rewrote the same two functions in the opposite direction, so reviewing them apart would have meant reviewing a decision twice and merging it once.

- Root cause is upstream (Magpie/TraceLens/GEAK/IntelliKit/AgentKernelArena), ticket filed: n/a, the root cause is in this repo.


xiaofei-zheng:
Reviewed the full diff against `git merge-base origin/main fix/shafeng/unify-agent-backend-cli` (dedea68). Paths checked: `preferred_agent_backend` / `agent_backend_rank` ranking, `openai_agent_credentialed` widening and every caller of it (`kernelforge/agent_backends/registry.py`, `specialists/subprocess_.py`, `orchestrator/kernel/request_handlers.py`, `agents/robustness/config.py`, `cli/executors.py`), the `orchestration_on_codex` rewrite in `cli/backends.py`, and the `parser.py` model defaults. Predicates exercised directly against the real module for the env shapes below.

One blocking issue.

**`CHANGELOG.md:117-120` and the PR description state a behaviour change that does not occur.**

Both claim: "a Bedrock or Vertex box carrying an OpenAI key runs Claude rather than Codex". That shape never reaches the ranked path.

`CLAUDE_CODE_USE_BEDROCK` / `CLAUDE_CODE_USE_VERTEX` are not in `_ANTHROPIC_SIDE_KEYS`, so with `{CLAUDE_CODE_USE_BEDROCK=1, OPENAI_API_KEY=sk-x}`:

```
is_openai_only()        -> True
preferred_agent_backend() -> 'claude'   # never consulted
```

`main()` (`cli/__init__.py:1629`) calls `_claude_model_should_follow_codex()` -> `is_openai_only()` -> True, sets `INFERENCE_OPTIMIZER_CLAUDE_FOLLOWS_CODEX=1` and rewrites `args.claude_model = args.codex_model`. `backends.py:103` then takes the first disjunct and sets `orchestration_on_codex=True`. Orchestration still runs on Codex, exactly as before this PR. Adding `ANTHROPIC_BASE_URL` to that box does not move it either — it was already Claude on main.

The PR text contradicts itself two paragraphs later ("Both explicit overrides stay ahead of the shared rule"), which is the correct statement. Of the two shapes listed as moving, only the dual-config / Codex-extra-only one actually does — and that one is the case `test_build_backends_dual_config_drops_claude_when_its_sdk_is_absent` covers. Please drop or correct the Bedrock/Vertex bullet in both the changelog entry and the PR description.

Non-blocking observations were dropped per review policy. Notes on things I checked and cleared:

- The `LLM_GATEWAY_KEY` widening does flip `preferred_agent_backend` to Codex for `{ANTHROPIC_BASE_URL, LLM_GATEWAY_KEY}` with no Anthropic credential (verified: 'claude' on main, 'codex' here), but that is not a misroute: `claude_sdk_env_options` synthesizes only from `ANTHROPIC_SYNTHESIZABLE_KEY_ENVS`, so a gateway-key-only box cannot authenticate the Claude CLI at all, and `resolve_openai_client_config` derives `base_url` from `ANTHROPIC_BASE_URL` (checked: `https://gw/Unified/v1`), so no gateway key is sent to official OpenAI.
- `parser.py:_default_codex_model_env` now short-circuits on `is_anthropic_only()` before reading an explicit `CODEX_MODEL`, but `_resolve_models_for_run` already forced `codex_model = claude_model` for that shape on main, so no run changes; only the parsed default shown by `--help` differs.
- No tests run locally: `test_cli_backends_unit.py` is uncollectable on this Windows host (`cli/kb.py` imports `fcntl`); CI covers it.

xiaofei-zheng:
Follow-up on the `LLM_GATEWAY_KEY` divergence this PR flags as "worth a separate look". Not a blocker for this PR — recording it here so it gets closed rather than left as a known split. After this PR the repo counts a gateway key as an OpenAI-side credential in three places and refuses it in two:

Accepts it:
- `resolve_openai_client_config` (`llm_config.py:429`) — resolves the key, and derives `base_url` from `ANTHROPIC_BASE_URL` when `OPENAI_BASE_URL` is absent.
- `codex_session._api_key_env_name` — now via the shared `OPENAI_AGENT_KEY_ENV_ORDER`.
- `openai_agent_credentialed` — new in this PR, so `preferred_agent_backend()` now ranks that side as credentialed.

Refuses it:
- `cli/credentials.py:343`: `has_key = bool(os.environ.get("OPENAI_API_KEY") or has_anthropic_side)` and the `has_usable_endpoint` pair on the next line both read the `OPENAI_API_KEY` literal, so `{OPENAI_BASE_URL, LLM_GATEWAY_KEY}` fails `_validate_credentials` and the launch exits before anything else runs.
- `install_baremetal.sh` / `agents/kernel/scripts/install.sh` mirror that refusal.

The split is observable inside this PR's own test file: `test_build_backends_forced_openai_protocol_accepts_gateway_key` (line 163) sets exactly `{OPENAI_BASE_URL, LLM_GATEWAY_KEY}` and asserts the critic resolves, with a docstring calling it "a config that would have run" — while `_validate_credentials` would have rejected that same deployment at boot. Widening `openai_agent_credentialed` makes the two sides disagree on one more predicate, not fewer.

Suggested direction for the follow-up — close it by moving `credentials.py` onto the shared tuple, not by narrowing the client side. The client genuinely can authenticate with a gateway key, and the one real hazard (a gateway key with no resolvable base URL going to official OpenAI) already has an exact test in this file, at `backends.py:62`:

```python
if resolved.base_url is None and not _any_env_set(("OPENAI_API_KEY",)):
    raise ValueError(...)
```

So `_validate_credentials` can accept `OPENAI_AGENT_KEY_ENV_ORDER` and keep the refusal scoped to the base-URL-less case, instead of rejecting the whole shape. The two install scripts follow. That leaves one definition of "the OpenAI side is credentialed" rather than the current two, which is the same consolidation this PR makes for the ranking.

Worth doing in one pass with this PR's direction still fresh — the predicate names are now public (`OPENAI_AGENT_KEY_ENV_ORDER`, `openai_agent_credentialed`), so the follow-up is mostly deleting literals.

xiaofei-zheng:
Correcting my previous comment: I suggested keeping the refusal "scoped to the base-URL-less case" via `backends.py:62`, but that line is itself a hardcoded single-key test (`_any_env_set(("OPENAI_API_KEY",))`). Hardcoding one key name is the defect, wherever it appears. The rule should be the same one this PR establishes for the agent CLI: one tuple, one predicate, no call site naming an env var.

**The hardcoding starts inside `llm_config` itself.** `_OPENAI_SIDE_KEYS` (`llm_config.py:136`) is still `("OPENAI_BASE_URL", "OPENAI_API_KEY")`, so after this PR the module answers the same question two ways:

```
env = {ANTHROPIC_BASE_URL=https://gw/anthropic, ANTHROPIC_API_KEY=k, LLM_GATEWAY_KEY=g}

has_openai_side()           -> False
openai_agent_credentialed() -> True     # widened by this PR
is_anthropic_only()         -> True
```

So on that box the OpenAI side is credentialed and simultaneously not configured. `_build_backends` reads the second answer: `provider_anthropic_only=True` pins the critic to the Anthropic protocol and `_build_proposal_scorer` returns `None`, while the ranking reads the first. Nothing errors today, but the module now disagrees with itself, and widening one predicate without the other is what created that.

**Every remaining single-key decision site**, all of which should ask `llm_config` instead of naming a variable:

| Site | Current |
|---|---|
| `llm_config.py:136` | `_OPENAI_SIDE_KEYS = ("OPENAI_BASE_URL", "OPENAI_API_KEY")` |
| `credentials.py:153` | `_has_explicit_openai_key` -> `environ.get("OPENAI_API_KEY")` |
| `credentials.py:324` | `OPENAI_BASE_URL and OPENAI_API_KEY` |
| `credentials.py:343,345` | `has_key` / `has_usable_endpoint` |
| `preflight.py:115` | `has_openai = OPENAI_BASE_URL or OPENAI_API_KEY` |
| `backends.py:62` | `_any_env_set(("OPENAI_API_KEY",))` |
| `kernelforge/llm/gateway.py:197` | `base_url and OPENAI_API_KEY` |

Suggested shape for the follow-up:

1. `_OPENAI_SIDE_KEYS = ("OPENAI_BASE_URL", *OPENAI_AGENT_KEY_ENV_ORDER)` — one list feeds both `has_openai_side` and `openai_agent_credentialed`, the way `_ANTHROPIC_SIDE_KEYS` already derives from `ANTHROPIC_CREDENTIAL_ENV_ORDER`.
2. `credentials.py` and `preflight.py` call `openai_agent_credentialed()` / `has_openai_side()` rather than reading env names. That is the same move this PR made for the ranking, applied to the credential gate.
3. For `backends.py:62`, the question is not "is `OPENAI_API_KEY` set" but "is the resolved key scoped to a gateway", which the call site cannot answer without naming the variable. Have `resolve_openai_client_config` return the env name it resolved from (e.g. `OpenAIClientConfig.api_key_env`) and test `base_url is None and api_key_env != "OPENAI_API_KEY"` inside `llm_config`, exposing it as one predicate. Then no caller names a key at all.

The `api_key_env="OPENAI_API_KEY"` default parameters (`codex_session`, `proposal_scorer`, `rca_engine`, `agent_role.DEFAULT_CODEX_API_KEY_ENV`) are fine as-is — they name a preferred variable and fall through the shared tuple. The scrub and forward lists (`env_safety`, `specialists/runner`, `specialists/subprocess_`, `preflight.py:66`, `cli/__init__.py:598`) carry both names already; they should read the tuple too so a third key name is added in one place, but they are not deciding anything today.

Net: after this PR the rule "which CLI runs a role" has exactly one owner. "Is this side credentialed" still has five. Worth closing the same way.

fengshaoyi-amd:
Confirmed and fixed in 2d8b8eb9e — the Bedrock/Vertex bullet is gone from both the changelog entry and the PR description.

You are right about the mechanism, and I reproduced the chain rather than the predicate this time:

```
{CLAUDE_CODE_USE_BEDROCK=1, OPENAI_API_KEY=sk-x}
  is_openai_only()                       -> True
  is_anthropic_only()                    -> False
  main() sets CLAUDE_FOLLOWS_CODEX=1     -> True
  preferred_agent_backend() in isolation -> claude
  ranked path actually consulted         -> False
  orchestration                          -> codex   (unchanged from main)
```

The error was mine and it was a specific one: I checked what `preferred_agent_backend()` returns for that env without checking whether it is reached, and the override I deliberately placed ahead of it is what short-circuits exactly that shape. The bullet had been dropped once already and I put it back on the strength of the isolated probe.

Only the dual-config / Codex-extra-only shape moves, which is what `test_build_backends_dual_config_drops_claude_when_its_sdk_is_absent` covers.

Also tightened the `LLM_GATEWAY_KEY` paragraph, which had swung too far the other way — it claimed no deployment moves at all. Your verification is in it now: the predicate does flip `{ANTHROPIC_BASE_URL, LLM_GATEWAY_KEY}` to Codex, `claude_sdk_env_options` synthesizes a CLI key only from `ANTHROPIC_SYNTHESIZABLE_KEY_ENVS` so that box could not have authenticated Claude, and `_validate_credentials` refuses it before any of it matters.

`origin/main` merged (was 11 behind). On the follow-up PR for the key cleanup: surveying the blast radius now, and I have one question about scope that I will raise separately rather than guess at.

fengshaoyi-amd:
Pushed 58e82083e, which retires the last thing these key names were doing here: being filled from `OPENAI_API_KEY`.

The earlier changelog entry deferred `LLM_API_KEY` / `AMD_LLM_API_KEY` with the reasoning that "they are how GEAK and the kernel tools receive that credential". I checked that against the component instead of assuming it, and it is not true.

Surveyed `AMD-AGI/GEAK` at `0e02ad3`:

| Name | Hits in GEAK |
|---|---|
| `LLM_API_KEY` | none |
| `AMD_API_KEY` | none |
| `LLM_GATEWAY_KEY` | none |
| `AMD_LLM_API_KEY` | one, `perf_knowledge/workflows/authoring_a_kernel_with_geak.md:87` — a quick-start line invoking `scripts/run-docker.sh`, which the repository no longer carries |

GEAK's AMD gateway key is `GEAK_AMDKEY`, declared in `interface/runtime/engine/registry.json` with the APIM header and `llm-api.amd.com/Unified` base URL. Its `provider_autoselect` triggers are `GEAK_AMDKEY` and `OPENAI_API_KEY`; the Anthropic side reads `ANTHROPIC_*` / `CLAUDE_CODE_OAUTH_TOKEN`. The registry note even records why the name carries the `GEAK_` prefix — so it survives this repo's `DOTENV_PREFIX_ALLOWLIST`.

So two sites were copying a live credential into names with no reader on either side of the boundary:

- `cli/preflight.py` filled `LLM_API_KEY` and `AMD_LLM_API_KEY` into its own environment.
- `agents/kernel/tools/backends/ray_runtime.py` mirrored those plus `LLM_GATEWAY_KEY` into the Ray runtime env, then aliased `AMD_API_KEY` on top.

Both are gone, and the four names leave `SAFE_ENV_KEYS`, so an operator's own export no longer reaches a worker either. Verified against both trees:

```
                                    main            this branch
mirrored from OPENAI_API_KEY        3 names         none
operator-exported, crossing Ray     4 names         none
OPENAI_API_KEY / LLM_API_BASE       forwarded       forwarded
```

`LLM_API_BASE` stays — it addresses an endpoint rather than authenticating to one, and preflight still derives it from the resolved OpenAI-side URL. The redaction and `.env`-strip lists keep all four, since a stale export is still a secret worth scrubbing.

One thing I did not act on, since it is outside what this PR decides. `GEAK_AMDKEY` has zero hits in this repository, and `SAFE_ENV_KEYS` is an exact tuple with no prefix rule — while `GEAK_ROOT` / `GEAK_E2E_RUNNER` sit in that same list specifically so a Ray worker can locate `interface/run_e2e.py`, which is the file that reads `GEAK_AMDKEY`. If anyone runs GEAK against the AMD gateway through the Ray path, that key is being stripped at the boundary. Worth a separate look by whoever knows whether that path is in use. The same survey also found no reader for `GEAK_API_KEY` / `GEAK_BASE_URL`, whose allowlist comment claims the e2e runner reads them.

fengshaoyi-amd:
Pushed d800fb119. Having established that the mirrored names had no reader, I ran the same check over every remaining name in `SAFE_ENV_KEYS` rather than stopping at the ones that prompted it. Counted against this repository and `AMD-AGI/GEAK@main` — the ref `install.sh` resolves by default — with the allowlist itself, the redaction tables and the `.env` strip lists excluded, since membership there is not consumption.

Three entries had no reader on either side and are gone:

| Name | Evidence |
|---|---|
| `LLM_PROXY_BASE_URL` | One occurrence in the whole repository: the allowlist line. Nothing writes it, nothing reads it, zero in GEAK. |
| `GEAK_WORK_DIR` | Same — one occurrence. Its comment described a "GEAK/Forge harness contract"; GEAK has no `GEAK_WORK_DIR`, and no `work_dir` under any spelling. |
| `LLM_PROXY_API_KEY` | Forwarded to every worker with nothing setting or reading it. The `env_safety` redaction entry stays, for the same reason the retired credential names keep theirs. |

Four names came back with hits I did not treat as dead:

- `GEAK_API_KEY` / `GEAK_BASE_URL` — zero hits in GEAK, but both are documented operator overrides in `.env.template` and `docs/reference/environment-variables.md`, and `preflight.py` reads `GEAK_BASE_URL` to rewrite the endpoint in GEAK's config file. So GEAK receives that one through the yaml, not the environment. The names stay; the comment claiming the e2e runner reads them from the environment is corrected instead.
- `LLM_API_BASE` — zero hits in GEAK, but preflight derives it and two operator-facing documents describe it. It addresses an endpoint rather than authenticating to one, so it is outside this change either way.

One thing I am deliberately leaving alone, and flagging rather than acting on. `GEAK_SCORE_TARGET`, `GEAK_SKIP_PROFILE`, `GEAK_MAX_BENCHMARK_SHAPES` and `GEAK_RUN_MODE` also return zero hits in GEAK, yet `install.sh` writes each into the dotenv with a documented meaning — `GEAK_SCORE_TARGET` even has a comment describing a value being lost "before reaching the actor". Either they are read somewhere this survey cannot see, or they are dead in the same way. They are run knobs rather than credentials, so deciding that belongs with whoever owns the GEAK side, not to this PR.

xiaofei-zheng:
## What this PR does

Moves the last decision of "which agent CLI runs a role" onto the shared rule: `cli/backends.py` now asks `preferred_agent_backend()` for orchestration instead of re-reading the endpoint shape, `preferred_agent_backend` and Forge's `select_default_agent_provider` both sort on the new `llm_config.agent_backend_rank`, and the registry's `find_spec` probes and hardcoded `(1, 1)` are derived from `llm_config`. It also retires `LLM_GATEWAY_KEY` as a credential (`resolve_openai_client_config`, `codex_session`, `_provider_only_mode`, the TraceLens gateway marker, the specialist allowlist) and stops mirroring the OpenAI key into `LLM_API_KEY` / `AMD_LLM_API_KEY` / `LLM_GATEWAY_KEY` in CLI preflight and in the Ray runtime env.

## Blocking

1. **The PR description no longer matches the diff.** Under *What this deliberately does not touch* it says `LLM_API_KEY`, `AMD_LLM_API_KEY` and `LLM_API_BASE` "are still filled from the OpenAI side by CLI preflight and mirrored into the Ray runtime env". The later commits on this branch (`58e82083e`, `d800fb119`) delete exactly that: the alias-export block in `cli/preflight.py` and the `env.setdefault` mirroring plus four allowlist entries in `agents/kernel/tools/backends/ray_runtime.py`. Only `LLM_API_BASE` still holds. The description also does not mention the `LLM_PROXY_BASE_URL` / `LLM_PROXY_API_KEY` / `GEAK_WORK_DIR` allowlist retirement. Since the release cut aggregates PR descriptions, this has to be brought back in step with the diff.

2. **`docs/reference/authentication.md` still documents the removed alias fill.** Line 41: "The only values still filled for you are the internal LLM aliases (`LLM_API_KEY`, `AMD_LLM_API_KEY`, `LLM_API_BASE`), which the inference optimizer CLI preflight copies from the OpenAI side." Line 106: "Kernel tools inherit the OpenAI-side credential from preflight (`LLM_API_KEY` / `AMD_LLM_API_KEY`)." After this PR neither is true — preflight fills `LLM_API_BASE` only (`cli/preflight.py:2127-2131`). An operator following this page will expect a credential that no longer arrives.

## What was checked

- `preferred_agent_backend` tie-breaking after the `min(ranks, ...)` rewrite (dict insertion order keeps Claude ahead of Codex on every tie), and `UNRUNNABLE_AGENT_RANK` against the `(1, 1)` literal it replaces in `select_default_agent_provider`.
- Every shape reaching `_build_backends`: `claude_follows_codex` sets `INFERENCE_OPTIMIZER_CLAUDE_FOLLOWS_CODEX=1` at `cli/__init__.py:1596-1599` before `_build_backends` reads it, so the OpenAI-only launch that rewrote `claude_model` still pins Codex and does not fall into the ranked path with a foreign model id.
- `_api_key_env_name` raises rather than falling back silently once `LLM_GATEWAY_KEY` is gone; `resolve_openai_client_config`'s candidate list dedupes `api_key_env` and the error message lists the same names it tried.
- Whether anything still reads the retired names: no reader for `LLM_API_KEY` / `AMD_LLM_API_KEY` / `GEAK_WORK_DIR` / `LLM_PROXY_BASE_URL` in this repo or in the GEAK tree (only a quick-start line in `perf_knowledge/`), so the removals themselves look safe; the redaction and dotenv-strip lists correctly keep the names.
- `pytest src/hyperloom/inference_optimizer/tests/test_llm_config.py src/kernelforge/tests/test_provider_registry.py src/hyperloom/agents/kernel/tests/test_codex_session.py` -> 240 passed, 4 failed, and the same 4 fail unchanged on the merge base (POSIX mode bits / path separators on Windows). `test_ray_runtime_env.py` and `test_cli_backends_unit.py` cannot be collected on Windows (`resource`, `fcntl`); CI covers them.

fengshaoyi-amd:
Fixed in 5590d5621.

**3 — `docs/release-notes.md`.** You are right about `AMD_API_KEY`: at the merge base `ray_runtime.py:336-337` reads it as a *source* for `AMD_LLM_API_KEY`, and preflight filled only the other two, so calling it a copy target described something that never happened. The sentence is gone along with the rest of the entry — the whole file is reverted to the merge base and this branch now shows zero diff against it. The release cut aggregates PR descriptions, which is where that content belongs; it is in the description now.

**2 — `docs/reference/authentication.md`.** Both lines corrected, and grepping for the claim rather than the names turned up a third site the review did not list: line 290, under *Direct upstream wiring*, "Fills the internal LLM aliases from the OpenAI-side key for child processes". Same claim, same page, different section. All of them now say what preflight actually does — `LLM_API_BASE` is the only value it derives, and it addresses an endpoint rather than authenticating to one — and the kernel-tools bullet names `OPENAI_API_KEY`, which is the name that crosses the Ray boundary. `LLM_API_KEY`, `AMD_LLM_API_KEY`, `AMD_API_KEY` and `LLM_GATEWAY_KEY` now have zero occurrences anywhere under `docs/`.

**1 — the description.** Rewritten. "What this deliberately does not touch" claimed the mirroring survives, and the reason it gave for keeping it — that GEAK receives its credential through those names — is the thing the GEAK survey disproved two commits earlier. The paragraph now carries that survey, the three allowlist retirements (`LLM_PROXY_BASE_URL`, `GEAK_WORK_DIR`, `LLM_PROXY_API_KEY`), why `GEAK_API_KEY` / `GEAK_BASE_URL` stay, and the docs correction. `LLM_API_BASE` is the only name left in that section, which is the one you said still holds.

One thing outside the diff, noted rather than acted on: `.github/workflows/auto-label.yml` reads `secrets.LLM_API_KEY` for its labeling call. That is CI infrastructure authenticating with its own repository secret, not the runtime credential path this PR touches, so the name stays there.

xiaofei-zheng (approved):
Re-reviewed at 5590d5621. Both blocking points from the previous round are resolved, and no new blocking issues. LGTM.

- The PR description now matches the diff: *What this deliberately does not touch* claims only `LLM_API_BASE`, and the removal of the preflight alias-export block and the Ray mirroring is stated rather than contradicted. The `CHANGELOG.md` checklist line is gone along with the reverted release-notes entry, so nothing in the description asserts a file the diff does not carry.
- `docs/reference/authentication.md` follows the removal in all four places. Checked each new claim against the code: preflight fills `LLM_API_BASE` only (`cli/preflight.py:2127-2145`, no key write left in that function), and `OPENAI_API_KEY` is the name that reaches a worker (`ray_runtime.py` `SAFE_ENV_KEYS`).
- `docs/release-notes.md` is byte-identical to the merge base again; the diff is confined to `src/` plus the two operator-facing docs pages.

Carried over from the first pass, re-confirmed unchanged on this head: `preferred_agent_backend`'s `min(ranks, ...)` keeps Claude ahead of Codex on every tie via dict insertion order; `UNRUNNABLE_AGENT_RANK` reproduces the `(1, 1)` literal it replaces in `select_default_agent_provider`; `claude_follows_codex` sets `INFERENCE_OPTIMIZER_CLAUDE_FOLLOWS_CODEX=1` at `cli/__init__.py:1596-1599` before `_build_backends` reads it, so an OpenAI-only launch that rewrote `claude_model` still pins Codex instead of entering the ranked path with a foreign model id; `_api_key_env_name` raises rather than falling back silently once `LLM_GATEWAY_KEY` is gone; and no reader remains for `LLM_API_KEY` / `AMD_LLM_API_KEY` / `GEAK_WORK_DIR` / `LLM_PROXY_BASE_URL` in this repo or in the GEAK tree.

Tests: `pytest src/hyperloom/inference_optimizer/tests/test_llm_config.py src/kernelforge/tests/test_provider_registry.py src/hyperloom/agents/kernel/tests/test_codex_session.py` -> 240 passed, 4 failed, and the same 4 fail unchanged on the merge base (POSIX mode bits / path separators on Windows). `test_ray_runtime_env.py` and `test_cli_backends_unit.py` cannot be collected on Windows (`resource`, `fcntl`); CI covers them.

Issue #677, opened 2026-06-23
Support AMD LLM Gateway (llm-api.amd.com) for Hyperloom Local Mode


amd-ama10002-2:
### Target release

Unsure

### Optimization domain

Both

### Problem / use case

Today, Hyperloom local mode works with LiteLLM-style gateways (e.g., Core42, please see comment here: https://github.com/AMD-AGI/Hyperloom/blob/12753ce8aca54098df3ec42256b065ef84242bfc/.env.template#L5-L11). 

However, it seems like the AMD LLM gateway uses a slightly different API than the LiteLLM style API that Hyperloom expects. Please see the AMD LLM gateway API here: https://llm.amd.com/api-explorer

Request is to enable Hyperloom local mode to work with the AMD LLM gateway (llm-api.amd.com). This request is being created on behalf of Tero, who would like this feature for his Hyperloom local users who work on a network where the Core42 URL may be blocked by a firewall.

### Proposed solution

Add support for integrating with the AMD LLM gateway ( `llm-api.amd.com` )

cc: @Li, Shuoshuo — would appreciate your input on if this possible and/or aligned with the hyperloom goals because you know a lot about the Primus‑SaFE LLM gateway design. 

### Expected impact

Improves usability for Tero's team who might only have access to the AMD LLM gateway in their hyperloom local mode setup. 

### Alternatives considered

Using LiteLLM deployed in Primus‑SaFE Global (Azure-hosted) as an LLM gateway

### Additional context

- AMD LLM gateway API: https://llm.amd.com/api-explorer
- We might currently be in a code freeze for new feature requests, so maybe this is something that we just discussed for now and backlog it
- Link to a Team's Channel where Tero and I discuss this: https://teams.microsoft.com/l/message/19:o8PPuw-j0iWVUIrtZa0E-g97ws9sJsgwf4l1YPN-RX81@thread.tacv2/1782196662638?tenantId=3dd8961f-e488-4e60-8e11-a82d994e183d&groupId=e08766d3-b279-495e-a963-d26aee11b281&parentMessageId=1782168210347&teamName=Dev%20Tools%20Community&channelName=Hyperloom%20Beta&createdTime=1782196662638&ngc=true&allowXTenantAccess=true

tkarttunen:
Hi, these are the comments from Claude Code about the required changes. Claude successfully vibe coded this to work. I'm just adding this AI generated information in here if it helps someone.

```
[...]

PR #1545, opened 2026-09-17
fix(orchestrator): decode emit_intent tool input delivered as a JSON string

### LINKED ISSUE #1543: emit_intent tool input arriving as a JSON string (litellm-style proxy) is silently dropped → autonomous loop stalls at 0%
## Summary

When Hyperloom's orchestrator talks to an OpenAI-compatible / litellm-style LLM proxy rather than the Anthropic API directly, the `emit_intent` tool call can arrive with `tool_use.input` as a **JSON string** rather than a native object (and sometimes with the nested `payload` stringified too).

`ClaudeBackend._parse_tool_use_block` discards that shape, and `decode_emit_intent_input` reports **no error**, so the orchestrator receives an empty intent. The autonomous loop then makes no progress and **stalls at 0% with no traceback and no ERROR log**.

This is the sibling of #1384. That issue taught `emit_intent` to unwrap one proxy envelope (`__unparsedToolInput.raw`, a *dict* wrapper). Proxies also emit a second shape — the input itself is a JSON string — which is still unhandled.

## Where

https://github.com/AMD-AGI/Hyperloom/blob/bbc2e04c5aa26112d2c31ded3d89a7e6d32b1692/src/hyperloom/orchestrator/roles/claude.py#L777-L780

```python
block_input = getattr(block, "input", None) or {}
if not isinstance(block_input, dict):
    block_input = {}                      # <-- a JSON-string envelope is discarded here
raw_input, decode_error = decode_emit_intent_input(block_input)
```

`decode_emit_intent_input` is typed `raw_input: dict[str, Any]` and only recognises the `__unparsedToolInput` wrapper, so `{}` decodes "successfully":

https://github.com/AMD-AGI/Hyperloom/blob/bbc2e04c5aa26112d2c31ded3d89a7e6d32b1692/src/hyperloom/orchestrator/roles/mcp_emit_intent.py#L147-L162

## Reproduction

Exercising the current `main` decoder directly:

```python
# tool_use.input as delivered by the proxy (a JSON string, not an object)
s = '{"intent_type": "run_action", "payload": {"action": "baseline"}}'

block_input = s or {}
if not isinstance(block_input, dict):
    block_input = {}                       # claude.py:778-779

decode_emit_intent_input(block_input)
# -> ({}, None)      decode_error is None: the intent is silently empty
```

Observed end-to-end: the loop authors a session, calls the model, and every turn returns an intent
with no `intent_type`; no action is ever dispatched. Progress stays at 0% until the budget expires.
Because `decode_error is None`, nothing is logged — the run looks like "the model never called the
tool".

## Why this has survived

- The direct Anthropic API path (`claude-agent-sdk`) always delivers `input` already parsed as a `dict`, so the default configuration never hits it.
- The failure is **silent** — no exception, no `ERROR`, not even the `log.info` on line 781, so there is no CI or log signal.
- Existing tests cover the native-dict and `__unparsedToolInput` paths added in #1384, but not a bare JSON-string envelope.

## Proposed fix

Make `decode_emit_intent_input` tolerant of the string envelope (and of a stringified nested `payload`), and pass the raw block input straight to it instead of pre-emptively replacing non-dicts with `{}`:

- `str` input → `json.loads`, and return a real `decode_error` if it is not valid JSON / not an object (so a genuine malformation is *loud* rather than silent).
- `payload` arriving as a string → decode it to a dict.
- Native dict and `__unparsedToolInput` behaviour unchanged.

This keeps the canonical-input-wins precedence from #1384 and turns a silent stall into either a correct decode or an explicit error.

I have this implemented and regression-tested, and will open a PR referencing this issue.

## Environment

- Hyperloom `1.1.0`; reproduced against `main` @ `bbc2e04`
- An OpenAI-compatible (litellm-style) LLM proxy in front of `claude-opus-5`

soumitra-chatterjee-tech:
Fixes #1543

## Problem

When the orchestrator runs behind an OpenAI-compatible / litellm-style LLM proxy rather than against the Anthropic API directly, `emit_intent` `tool_use.input` can arrive as a **JSON string** rather than a native object, and sometimes with the nested `payload` stringified too.

Running Hyperloom behind such a proxy is a supported configuration (see #677).

`_parse_tool_use_block` replaced any non-dict input with `{}`, and `decode_emit_intent_input` only recognised the `__unparsedToolInput` wrapper added in #1384, so `{}` decoded "successfully" with `decode_error is None`. The orchestrator received an intent with no `intent_type`, dispatched no action, and the autonomous loop **stalled at 0% with no traceback and no ERROR log**.

## Change

- `decode_emit_intent_input` now accepts the JSON-string envelope and a stringified `payload`, in addition to the native object and the `__unparsedToolInput` wrapper.
- `_parse_tool_use_block` hands the raw block input to the decoder instead of pre-coercing non-dicts to `{}`.
- A genuinely malformed envelope now returns an explicit `decode_error` — the failure becomes **loud instead of silent**.
- Canonical-input-wins precedence from #1384 is unchanged.

## Test plan

New `test_emit_intent_json_string_envelope_unit.py` (12 cases): string envelope, stringified payload, both combined, string-wrapped `__unparsedToolInput`, native-object and wrapper regression, malformed / non-object / non-dict error paths, and unparseable-payload passthrough.

```
python3 -m pytest -n auto \
  src/hyperloom/inference_optimizer/tests/test_emit_intent_json_string_envelope_unit.py \
  src/hyperloom/inference_optimizer/tests/test_claude_unparsed_tool_input_unit.py \
  src/hyperloom/inference_optimizer/tests/test_claude_backend_helpers_coverage_unit.py \
  src/hyperloom/inference_optimizer/tests/test_claude_backend.py
```

Also validated end-to-end behind such a proxy on AMD Instinct MI355X (gfx950) / ROCm 10.0: with this change the orchestration loop self-drives to completion where it previously sat at 0%.


src/hyperloom/orchestrator/roles/mcp_emit_intent.py
```diff
@@ -146,8 +146,37 @@ def build_intent_envelope_schema(allowed_intents: Iterable[IntentType]) -> dict[
...
+    1. a native object (direct Anthropic API);
+    2. Claude Code's ``__unparsedToolInput.raw`` wrapper;
+    3. a JSON *string* envelope, as emitted by OpenAI-compatible / litellm
+       style proxies, optionally with ``payload`` itself stringified.
+
+    Shape 3 previously could not reach this function: the caller replaced any
+    non-dict input with ``{}``, which decoded "successfully" to an empty
+    intent, so the orchestrator stalled with no error.
+    """
+    if isinstance(raw_input, str):
```
lishuoshuo-amd, line 163: [P1] Preserve fallback-wrapper deduplication for the new string-wrapped shape

This decoder now accepts a JSON string containing `__unparsedToolInput`, but `_invoke_and_collect()` still decides whether to deduplicate by calling `is_unparsed_tool_wrapper()` on the original `block.input`. Since that helper only accepts a dict, a string-wrapped fallback is not recognized as a fallback retry.

Two identical dict wrappers produce one intent with `deduped_fallback_intents == 1`, while two identical string-wrapped wrappers currently produce two intents with `deduped_fallback_intents == 0`. There is no generic downstream intent deduplication, so actions/messages/tasks may be dispatched twice.

Please preserve whether the normalized input came from a fallback wrapper (or extend the wrapper detection safely) and add a `ClaudeBackend.run()` regression test asserting that two identical string-wrapped wrappers produce exactly one intent.

src/hyperloom/orchestrator/roles/claude.py
```diff
@@ -774,10 +774,11 @@ def _is_tool_use_for_emit_intent(self, block: Any) -> bool:
...
         """Validate one ``emit_intent`` tool-use block into an :class:`Intent`."""
-        block_input = getattr(block, "input", None) or {}
-        if not isinstance(block_input, dict):
-            block_input = {}
-        raw_input, decode_error = decode_emit_intent_input(block_input)
+        # Hand the raw block input to the decoder: it accepts a native object,
+        # Claude Code's __unparsedToolInput wrapper, and the JSON-string
+        # envelope that OpenAI-compatible / litellm proxies emit. Coercing a
+        # non-dict to {} here would decode to an empty intent with no error.
+        raw_input, decode_error = decode_emit_intent_input(getattr(block, "input", None) or {})
```
lishuoshuo-amd, line 781: [P2] Pass the actual raw input to the decoder

The remaining `or {}` still converts an empty-string input into `{}` before decoding. This disagrees with the new comment saying the raw block input is passed through, and with the decoder test that expects `""` to report `not valid JSON`.

Through the real backend path, `""` currently becomes `{}` and is reported later as `intent_type None not in allowed set`, rather than as the intended decode error.

Please pass `getattr(block, "input", None)` directly and add a backend-level empty-string test so the production call path matches the decoder contract.

lishuoshuo-amd (commented):
The JSON-string decode path fixes the reported bare-envelope failure. I found two integration gaps in the newly supported paths:

soumitra-chatterjee-tech:
Thanks — both were real. Fixed in the latest commit.

**[P1] fallback dedup.** You're right about the mechanism: `_invoke_and_collect` keys dedup on
`is_unparsed_tool_wrapper(block.input)`, and that helper took the dict path only, so a string-wrapped
retry was never recognised as a fallback. I've taught the helper to decode the string envelope before
its existing dict checks, so wrapper detection now accepts exactly the shapes
`decode_emit_intent_input` accepts. Keeping those two in step seemed like the actual fix — the drift
between them is what let the duplicate through. I left `coerce_emit_intent_input` alone, since its
non-dict behaviour is pinned by an existing test and isn't in the path you flagged.

**[P2] raw input.** Dropped the `or {}`; the block input now goes to the decoder untouched. Worth
naming one consequence beyond the empty-string case: an `input` of `None` now reports a decode error
too, where before it became `{}` and surfaced later as `intent_type None not in allowed set`. That
looks like what you intended — the decoder contract should hold on the production path — but say the
word if you'd rather `None` keep its old behaviour and only `""` change.

**Tests**, both at `ClaudeBackend.run()` level as asked, parametrised over the two tool names like
their neighbours:

- `test_string_wrapped_fallback_retries_dedupe_to_one_intent` — two identical string-wrapped wrappers
  plus a distinct third: 3 tool blocks, 2 intents, `deduped_fallback_intents == 1`. I checked it fails
  without the P1 change.
- `test_empty_string_tool_input_reports_decode_error` — an empty-string block records
  `emit_intent tool input string is not valid JSON` and is dropped without disturbing the rest of the
  turn (following `test_run_invalid_tool_use_input_drops_block_silently`'s shape).

**Rebased onto current `main`** (we were 216 behind) — it was clean, and for what it's worth neither
`claude.py` nor `mcp_emit_intent.py` had upstream changes in that range, so nothing in the fix needed
reworking.

Locally after the rebase: 71 passed across `test_claude_backend.py`,
`test_claude_unparsed_tool_input_unit.py` and `test_emit_intent_json_string_envelope_unit.py`;
`ruff check` and `ruff format` clean. I can't run the full suite (missing optional deps in my
environment), so CI is the real check — and it still hasn't run on this branch. If it needs an
approval or a label to trigger, let me know.

PR #1546, opened 2026-09-17
feat(explore): let a scriptable adapter declare its own lever grid

### LINKED ISSUE #1544: Scriptable / custom workloads get an empty EXPLORE grid — let an adapter declare its own tunable levers
## Summary

`_default_grid_for_framework()` returns a curated seed grid for `atom` and `xdit`, and `[]` for everything else — including `custom` / scriptable workloads:

https://github.com/AMD-AGI/Hyperloom/blob/bbc2e04c5aa26112d2c31ded3d89a7e6d32b1692/src/hyperloom/orchestrator/actions/executors/explore.py#L569

```python
fw = (framework or "").strip().lower()
if fw == "atom":  return _atom_default_grid(...)
if fw == "xdit":  return _xdit_default_grid(...)
return []          # <-- every custom / scriptable workload
```

For LLM frameworks that is reasonable: the LLM can infer levers from the framework's own source and emit `default_grid` variants. For a **scriptable / bypass `custom` workload there is no framework source to read**, so the model has nothing to infer from and cannot guess the workload's real knobs (e.g. `gpu_chain_workers`, `cpu_threads`, `gpu_partitions`).

Net effect: a `custom` workload supplied through the documented bypass path (`HYPERLOOM_BENCHMARK_BACKEND=bypass` + `--benchmark-scripts-dir`) **runs and benchmarks correctly, but is not autonomously tunable** — EXPLORE has zero variants to sweep.

Observed on `main` today:

```
_default_grid_for_framework() seed variants
  atom      -> 2
  xdit      -> 4
  custom    -> 0      # nothing to sweep
```

## Proposed solution

Let the adapter **declare its own tunable knob grid**, and merge those declared variants into the explore grid for scriptable frameworks only.

- Source: `HYPERLOOM_DECLARED_LEVERS` (inline JSON) or `<benchmark-scripts-dir>/declared_levers.json`.
- Merge (not replace): declared variants are prepended and de-duplicated by `name`, so they still sweep when the LLM proposes nothing, and LLM-proposed variants are preserved.
- Gated behind `framework_is_scriptable(framework)` — no behaviour change for `atom` / `xdit` / `vllm` / `sglang`.

This makes scriptable/custom workloads first-class *tunable* rather than merely runnable, without touching the LLM-framework paths.

Example `declared_levers.json`:

```json
[{"name": "workers_2", "extra_envs": {"MINIMAP2_GPU_CHAIN_WORKERS": "2"}}]
```

## Why this matters

We have driven the autonomous loop against several non-LLM GPU workloads through the bypass path
on MI355X / ROCm 10. With the declared-lever merge, EXPLORE sweeps the knobs the adapter actually
exposes and the loop keeps only quality-gated winners; in one run the agent extrapolated past the
declared values to a setting we had not listed, which is the argument for treating the declaration
as a seed that is *merged* rather than a fixed grid that replaces the model's own proposals.

Without it, EXPLORE reports `empty_grid` for these workloads and config-sweep mode is simply
unavailable.

I have this implemented (one file, ~40 lines, 12/12 unit tests, applies clean to `main`) and will open a PR referencing this issue.

## Environment

- Hyperloom `1.1.0`; verified against `main` @ `bbc2e04`

soumitra-chatterjee-tech:
Closes #1544

## Problem

`_default_grid_for_framework()` seeds a curated grid for `atom` and `xdit` and returns `[]` for everything else. For an LLM framework that is reasonable — the model reads the framework's own source and emits `default_grid` variants.

A scriptable / bypass `custom` workload has **no framework source to read**, so the model cannot infer the workload's real knobs (`gpu_chain_workers`, `cpu_threads`, `gpu_partitions`). EXPLORE therefore ends up with an empty grid: the workload runs and benchmarks correctly through the documented bypass path, but is **not autonomously tunable**.

## Change

Let the adapter declare its own knob space, and merge it into the explore grid for scriptable frameworks only.

- Source: `HYPERLOOM_DECLARED_LEVERS` (inline JSON), else `declared_levers.json` in `HYPERLOOM_BYPASS_SCRIPTS_DIR`. Legacy aliases `HLHPC_LEVERS` / `hlhpc_levers.json` are also accepted.
- **Merge, not replace:** declared variants are prepended and de-duplicated by `name`, so they sweep even when the model proposes nothing (or its variants are filtered out), and model-proposed variants are preserved.
- Gated on `framework_is_scriptable(framework)` — `atom` / `xdit` / `vllm` / `sglang` are untouched.
- A malformed declaration logs a warning and seeds nothing rather than failing the task.

Example `declared_levers.json`:

```json
[{"name": "workers_2", "extra_envs": {"MINIMAP2_GPU_CHAIN_WORKERS": "2"}}]
```

## Why

We have driven the autonomous loop against several non-LLM GPU workloads through the bypass path
on MI355X / ROCm 10. With the declared-lever merge, EXPLORE sweeps the knobs the adapter actually
exposes and the loop keeps only quality-gated winners; in one run the agent extrapolated past the
declared values to a setting we had not listed, which is the argument for treating the declaration
as a seed that is *merged* rather than a fixed grid that replaces the model's own proposals.

Without it, EXPLORE reports `empty_grid` for these workloads and config-sweep mode is simply
unavailable.

## Test plan

New `test_declared_lever_grid_unit.py` (12 cases): inline env, both filenames, legacy env alias, env-over-file precedence, the `{levers: [...]}` object form, malformed / non-list input, unnamed-variant filtering, missing file, and provenance defaulting/preservation.

```
python3 -m pytest -n auto src/hyperloom/inference_optimizer/tests/test_declared_lever_grid_unit.py
```


ZhengGong-amd:
The gap in #1544 is real — `custom` is runnable but not tunable, and `optimize-custom-workload.md:131` already promises "leave the rest for exploration" with nothing behind it. Requesting changes on the shape, not the intent.

**P0**

1. **Rebase onto `main`.** `explore.py` conflicts, and `main` no longer imports `framework_is_scriptable` into that module — after the merge the new gate references an undefined name (ruff F821). CI has never executed on this branch (all 7 workflows are `action_required`); please get it green.
2. **Gate on `framework == "custom"`, not `framework_is_scriptable()`.** The latter is `{xdit, custom}` (`framework_registry.py:56,67`), so "xdit untouched" in the description does not hold. `explore.py:802` is the existing precedent for this predicate.
3. **Converge on one source: `<benchmark-scripts-dir>/levers.json`.** Drop `HYPERLOOM_DECLARED_LEVERS`, `HLHPC_LEVERS`, `hlhpc_levers.json`, and the `{"levers"}` / `{"variants"}` wrappers. The scripts dir is already the adapter's contract directory and is already persisted and re-exported on `--resume`, which a bare env var is not.
4. **Validate at the launch boundary** (`cli/__init__.py`, next to `_require_custom_entrypoint`) and `exit 2` on a malformed declaration; remove the warn-and-continue paths in the executor. All of them currently surface as `empty_grid`, so an operator's typo reads as "the model proposed nothing".

**P1**

5. **Add a ledger filter**: seed only declarations absent from `search["name_index"]`. As written, every EXPLORE round re-prepends the full set, and re-proposals are still benchmarked (`explore.py:861`), so they re-spend the round's wall-clock budget ahead of the model's proposals.
6. **Use `is_executable(normalize_proposal(v))`** as the validity predicate instead of "has a name" — a name-only variant is benchmarked as a duplicate of the baseline, and the tests currently assert that shape.
7. **Fold into the existing lever merge** (`explore.py:720-730`) instead of a second copy of the same prepend/dedupe block.
8. **Tests are inverted**: 12 cases on the parser, none on the behaviour. Cover (a) declarations sweep when the model proposes nothing, (b) `vllm` / `xdit` unaffected, (c) already-tested declarations are not re-seeded, (d) a name collision with a model proposal.

**P2**

9. **`provenance`**: use `default_grid`, or add the label to `producer_for_provenance` — anything else is attributed to the orchestration agent in the session breakdown. Drop the redundant `setdefault("note", ...)`.
10. **Docs**: `CHANGELOG.md` under `[Unreleased]`, `docs/how-to/optimize-custom-workload.md` (right after line 131), and `SKILL.md` so the agent knows the declaration exists and can extrapolate from it. With a single file source, `environment-variables.md` needs no change.

**Sequencing**: please land this after #1587, which dissolves `ExplorePhase` and rewrites the grid-assembly and ledger code this PR hooks into. Rebasing on top of it is far cheaper than resolving both.

Issue #1539, opened 2026-09-16
Proposal: sparse (IRT-calibrated) accuracy evaluations for the accuracy gate


KHassanQazi:
# Proposal: sparse (IRT-calibrated) accuracy evaluations for the accuracy gate

Following up on a discussion with Xiaofei — posting this as an issue so the team can
evaluate whether the approach fits Hyperloom's needs.

## What this is

The accuracy gate currently runs the full GSM8K test set (1,319 items) per gated variant.
The [tinyBenchmarks](https://arxiv.org/abs/2402.14992) approach replaces that with a
~100-item IRT-calibrated subset that **estimates the full-benchmark score** rather than
merely sampling it.

The value for Hyperloom is that **these tasks already ship in lm-evaluation-harness**
(`lm_eval/tasks/tinyBenchmarks/`: `tinyGSM8k`, `tinyMMLU`, `tinyArc`, `tinyHellaswag`,
`tinyWinogrande`, `tinyTruthfulQA_mc1/mc2`), and Hyperloom's existing `MAGPIE_EVAL_TASKS`
knob already reaches `lm_eval --tasks` on the bypass backend. There is no new evaluation
machinery to build.

Worth stating up front, since it is easy to miss: the value lm-eval reports for `tinyGSM8k`
is **not** a raw 100-item mean. The aggregation is `agg_gpirt_gsm8k`, a gp-IRT estimator of
full-GSM8K performance, so it is directly comparable to the full GSM8K number.

## Measured results

**Cost** (MI300X, TP=1, SGLang v0.5.12/ROCm 7.2.0, Qwen2.5-1.5B-Instruct, lm_eval 0.4.13,
both tasks against the same served model, prefix cache flushed between them):

| | full GSM8K | tinyGSM8k | ratio |
|---|---:|---:|---:|
| Items | 1,319 | 100 | 13.2x |
| **Wall-clock** | **39.48 s** | **10.51 s** | **3.76x** |
| Variable time (excl. 7.89 s fixed harness overhead) | 31.59 s | 2.62 s | **12.1x** |

Serving throughput is unchanged (5,567 vs 4,951 gen tok/s on the variable window); the eval
task changes how much work is done, not how fast the server does it. Note that 75% of
tinyGSM8k's wall-clock is fixed lm_eval startup cost, which caps the realizable speedup at
~4x — going further would mean amortizing harness startup, which is orthogonal to this
proposal.

[...]

PR #1548, opened 2026-09-17
Make MAGPIE_EVAL_TASKS actually usable (#1539)

chennyiiis:
Closes part of #1539.

## What was broken

`MAGPIE_EVAL_TASKS` is already plumbed end to end — `_workload_envs` persists it into `benchmark.envs`, it is part of the eval-contract fingerprint, and `bypass_runner` forwards it to `lm_eval --tasks`. But it was an **undocumented switch that crashes the run if you use it**:

1. **The estimator dependency was never installed.** lm-eval's `tiny*` tasks import `tinyBenchmarks` lazily from *inside* their aggregation function. So `MAGPIE_EVAL_TASKS=tinyGSM8k` runs the entire generation pass, then raises `ModuleNotFoundError` at scoring time; the non-zero eval rc then fails the whole benchmark. Worst possible failure shape — you pay the full eval cost and get nothing.

2. **`acc_norm,none` was not a recognised metric key**, so a task reporting only that scored as "no accuracy".

## What this changes

**Dependency (gap 1)** — both install points now ensure `tinyBenchmarks`:
- `bypass_runner._ensure_eval_deps` (single-node, runtime)
- `preflight._ensure_lm_eval_dep` (multi-node) — a third install point the issue didn't mention

Installed **only when the resolved task list actually names a `tiny*` task**, so a default `gsm8k` run takes on no new dependency.

> ⚠️ The issue's proposed one-liner `pip install tinyBenchmarks` does not work — upstream publishes **no PyPI distribution** (`https://pypi.org/simple/tinybenchmarks/` → 404). The install is from pinned source at `felipemaiapolo/tinyBenchmarks@e9a8b10`, **git first then the GitHub archive**, mirroring the existing `_LM_EVAL_PINNED_SPECS` fallback because the sandbox may not ship a git binary.

**Metric key (gap 2)** — `acc_norm,none` added *after* `acc,none`, so a task carrying both still reports plain `acc` as its headline number. The key tuple lived in two places that had to agree (`_accuracy_gate.parse_eval_results` and `bypass_analysis.summarize_eval`); it is now a single `ACCURACY_METRIC_KEYS` in a new stdlib-only `hyperloom.common.eval_tasks`, which also owns `DEFAULT_EVAL_TASKS` and the pinned specs.

**Docs** — `MAGPIE_EVAL_TASKS` was absent from `docs/reference/environment-variables.md` entirely. Now listed, plus a section on what a sparse task actually changes.

**The default is unchanged: `gsm8k`, full task set.** This PR makes the switch work; it does not flip it.

## What this PR deliberately does *not* do

The issue's headline ask — swapping the gate's default to tinyGSM8k — is not in here, because two things in the current pipeline do not survive that swap and neither is addressed by the issue:

- **`DEFAULT_ENABLEMENT_ACCURACY_FLOOR = 0.5` is an absolute judgment**, applied twice (the baseline eval-failure trigger and the enablement KEEP gate). The paired-replay evidence in the proposal shows the delta gate is safe because the subset's bias cancels on both sides — but it does **not** cancel against an absolute floor. A model genuinely sitting at 0.50–0.54 can be failed by the estimate alone.
- **There is no separate delivery eval.** `conc_sweep` hardcodes `RUN_EVAL=false`, and `report.py` renders the gate's own `baseline_accuracy`. "Gate on tiny, publish full" is not configuration — it needs new code.

Both are now written down in the docs section so anyone flipping the switch sees them. Happy to take the default change as a follow-up once those two are settled.

Also worth noting for the discussion on #1539: §2/§7's "the gate only fires on high-risk variants" motivation is stale. `_accuracy_gate.py` says *"There is deliberately no 'high accuracy risk' predicate here any more"* — every candidate is evaluated, which makes the wall-clock argument stronger than the proposal states.

## Testing

- New `src/hyperloom/common/tests/test_eval_tasks.py` — 16 tests on task splitting, `tiny*` detection, metric-key ordering, spec pinning.
- New `_ensure_eval_deps` coverage in `test_bypass_backend.py` — default task installs nothing extra, tiny task installs from the pinned git spec, archive fallback when git is unavailable, present estimator is not reinstalled.
- New `_ensure_lm_eval_dep` coverage in `test_preflight_lm_eval_dep.py` — same four cases plus the both-missing path (`lm_eval[api]` does not pull the estimator in).
- `ruff check` + `ruff format --check` clean on all touched files.
- 207 passed across the eval-related suites locally; the 1 remaining failure (`test_eval_result_dir_wiring`) is a pre-existing Windows path-separator assertion, unrelated to this change.


chennyiiis:
Closing: we are keeping the accuracy gate on full GSM8K with the 0.05 threshold unchanged (see https://github.com/AMD-AGI/Hyperloom/issues/1539#issuecomment-5770978610), so there is no longer a case for this change. Reopen if MAGPIE_EVAL_TASKS becomes needed for a non-gating use.

PR #1549, opened 2026-09-17
Feat/rocm10 e2e

lishuoshuo-amd:
## Description: what and why

Enable the ROCm 10 end-to-end path for Hyperloom across Ray startup, framework discovery, vLLM profiler preflight, and KernelForge profiling.

This branch fixes the concrete failures found while running the ROCm 10 matrix:

- Keep HIP visibility aligned with the ROCR-selected device view so torch and Ray agree on allocated GPUs.
- Preserve failed Ray CLI output when no log sink exists, so bootstrap failures carry evidence instead of `see None`.
- Move the SGLang source-install pin to a version that can solve against ROCm 10's torch constraints.
- Refresh the TraceLens pin past the xprof constraint that blocked Python 3.14 installs.
- Use the active runtime PATH when materializing benchmark configs, so ROCm 10's `/opt/python` and `/opt/venv` layouts both resolve their framework executables.
- Validate InferenceX profiler arguments with the same profiler context used at launch.
- Install and discover `rocprof-compute` on TheRock's pip-packaged ROCm layout, where the profiler lives in `_rocm_profiler` rather than under `$ROCM_PATH`.
- Pin `rocm-profiler` to the already-installed `rocm` version and extend the image's pip index instead of letting pip re-resolve or downgrade the ROCm SDK.
- Make rocprof-compute usable on pip-packaged ROCm by providing a private analyze venv, restoring the missing aqlprofile soname, and exposing the SDK's own `amdsmi`/runtime root without moving torch's environment.
- Fast-fail profiling on prebuilt ROCm 10 SGLang images whose torch runtime cannot be imported under `rocprofv3`, so Forge analysis degrades through the existing unavailable-profiling path instead of hanging.

## Linked issue(s)

No issue currently linked.

## Tests

Unit / regression tests:

- `python3 -m pytest src/hyperloom/inference_optimizer/tests/test_install_rocprof_compute.py src/kernelforge/tests/test_rocpc_profile_discovery.py -q -p no:cacheprovider` -> targeted rocprof-compute / rocpc discovery coverage passing after the ROCm 10 profiler fixes.
- `python3 -m pytest src/ -n auto -q -p no:cacheprovider` was run during validation and reached `22478 passed, 61 skipped` with the known existing environment/auth/supervisor failures still present (`21 failed, 1 error`). The touched profiler tests were green.

Spur / ROCm 10 validation:

- `rocm/pytorch:rocm10.0_ubuntu24.04_py3.12_pytorch_release_2.13.0`: rocprof-compute `profile` + `analyze` passed after the SDK runtime, amdsmi, aqlprofile soname, and analyze-venv fixes; `rocm`, `rocm-sdk-core`, and `torch` versions remained unchanged.
- `rocm/vllm:rocm10.0.0_ubuntu24.04_py3.14_pytorch_2.12.0_vllm_0.27.0`: rocprof-compute `profile` + `analyze` passed; vLLM import remained OK after the pandas/analyze dependency handling.
- `lmsysorg/sglang-rocm:{v0.5.18,v0.5.19,v0.5.20}-rocm10-mi35x-*`: minimal `rocprofv3 --hip-trace -- python3 -c "import torch"` reproduces an image-level LLVM option collision (`spirv-expand-step registered more than once`). The branch now fast-fails this path and falls back rather than hanging Forge analysis.

## `CHANGELOG.md` entry under `[Unreleased]`

Yes.

## Breaking changes

No intentional breaking changes.

Operational note: prebuilt ROCm 10 docker-SGLang images are treated as profiling-unavailable when their torch runtime cannot be imported under `rocprofv3`; they can still run non-Forge/framework paths, while Forge roofline coverage should use the baremetal SGLang route until the image-level LLVM collision is fixed upstream.

## PR addresses single concern

Yes. The concern is ROCm 10 enablement and the fixes required to make the ROCm 10 execution/profiling path usable end to end.

## Root cause is upstream (Magpie/TraceLens/GEAK/IntelliKit/AgentKernelArena), ticket filed

Partially upstream / image-level:

- The prebuilt ROCm 10 SGLang images have a torch/rocprofv3 LLVM option-registry collision before any GPU kernel runs. This PR contains the Hyperloom-side fallback so long Forge sessions do not hang, but producing roofline data from those exact images requires a rebuilt/fixed SGLang image.
- No upstream ticket is linked from this PR yet.


xiaofei-zheng:
Reviewed the full diff against `git merge-base origin/main feat/rocm10-e2e` (fa4fe164c).

Blocking:

1. **No `CHANGELOG.md` entry, but the PR description says there is one.** The diff touches 16 files and none of them is `CHANGELOG.md`, while the description answers "Yes" under "`CHANGELOG.md` entry under `[Unreleased]`". This PR changes user-visible behavior that belongs there: the SGLang source-install pin moves 0.5.18 -> 0.5.19, `ensure_ray_cluster` changes its failure message, preflight now sets `HIP_VISIBLE_DEVICES` instead of unsetting it, benchmark configs take the runtime `PATH` over the image default, rocprof-compute gains a pip install route plus a private analyze venv, and forge profiling now fast-fails when rocprofv3 cannot import torch. Add the entry under `[Unreleased]` following the existing format, or correct the description.

Paths verified while reviewing, no blocking issue found in them:

- `ray_runtime.ensure_ray_cluster`: the no-log-sink branch now captures stdout/stderr and both error messages carry the exit code; the log-sink branch still names the file.
- `preflight._normalize_hip_visible_devices`: re-indexes to `0..n-1` inside the ROCR view. Checked it does not shift the GEAK handoff — `_resolve_inner_hip_mask` picks the new HIP mask up as the inner mask and `_resolve_handoff_gpu_ids` caps it with `kept[:width]`, so `ROCR=4,5,6,7` with `tp=2` still yields `"0,1"`. `_ray_probe_env` becomes a no-op now that HIP is set, which is the same env it used to synthesize.
- SGLang pin: `00a9a81b6777` sits between `v0.5.19` and `v0.5.20` on the same convention the previous pin used, and the stated cause checks out against upstream — `python/pyproject_other.toml` at the old pin has `srt_hip` pinning `compressed-tensors==0.15.0` (`torch<2.11`), at the new pin `srt_hip` no longer pins it and `runtime_common` resolves `compressed-tensors` to 0.18.0 (`torch>=2.10`, no cap). `sglang_shape_mode` keeps the sitecustomize route for both versions, so the TraceLens patch set is not involved in the bump.
- `materialize_config_with_envs` PATH override: only replaces an existing `envs["PATH"]`, and `_grid_runner` still prepends the framework venv bin afterwards, so the isolated-venv layout is preserved.
- `_preflight_server_argv` profiler-arg injection: the drop repair rebuilds from `sealed.text`, so the injected `--profiler-config.*` tokens cannot be resealed into the config.
- `install.sh` rocprof-compute: the pip route is gated on the `rocm` distribution being present, version-pinned, and uses `--extra-index-url`; apt is kept for `/opt/rocm` images; the aqlprofile soname link and the analyze venv are both fail-soft; `_rocpc_libexec_dir` is re-resolved after each install path and matches the resolver in `rocpc_profile.py`.

No tests run locally; CI is green on this head.

xiaofei-zheng (approved):
Approving. Blocking comment above stands: please add the `[Unreleased]` CHANGELOG entry (or correct the description) before merge. The code paths I reviewed — Ray start evidence, HIP re-indexing and its effect on the GEAK handoff, the SGLang 0.5.19 pin against upstream `pyproject_other.toml`, the PATH override, the vLLM argv preflight injection, and the rocprof-compute install/discovery changes — are sound.

PR #1551, opened 2026-09-17
refactor(auth): retire the non-official LLM key names as credentials

fengshaoyi-amd:

  `OPENAI_API_KEY` plus the three Anthropic-side names (`ANTHROPIC_API_KEY`, `ANTHROPIC_AUTH_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN`) are the only credentials the optimizer authenticates with. This removes every other name from the credential read paths.

  I surveyed all eight vendor/gateway aliases first. Only two had production read paths:

  | Name | Read as a credential |
  |---|---|
  | `LLM_GATEWAY_KEY` | `resolve_openai_client_config` (one rung above the Anthropic fallback), `codex_session.api_key_env_name`, `_provider_only_mode`, `tracelens_analysis`, robustness RCA |
  | `LLM_API_KEY` | robustness RCA only |
  | `AMD_API_KEY`, `AMD_LLM_API_KEY`, `GEAK_API_KEY`, `LLM_PROXY_API_KEY`, `CLAW_API_KEY`, LLM-side `SAFE_API_KEY` | nothing — no read path anywhere |

  Two of those sites were a credential name doing a second job it had no business doing. `_provider_only_mode` read a bare `LLM_GATEWAY_KEY` as proof the host fronted both protocols, which suppressed single-provider detection; that is now decided by each side's own URL and key. `tracelens_analysis` took the same variable as an on-its-own gateway marker — `HYPERLOOM_STRICT_GATEWAY_MARKERS` exists for exactly that case three lines below the check this deletes. The specialist secret allowlist also stops forwarding `LLM_GATEWAY_KEY` to a child that can no longer authenticate with it.

  `resolve_openai_client_config` also had the accepted names written twice, once as an `or` ladder and once as the error message's list. Both now derive from one tuple, so the retirement is expressed in a single place.

  **What this deliberately does not touch.** `LLM_API_KEY`, `AMD_LLM_API_KEY` and `LLM_API_BASE` are still filled from the OpenAI side by CLI preflight and mirrored into the Ray runtime env. They are *outbound*: `docs/reference/authentication.md` documents them as values the operator never sets and that the kernel tools consume, and `ray_runtime.SAFE_ENV_KEYS` carries them for GEAK. Retiring those belongs in the owning components, not here. The eight names also stay in the benchmark secret, variant, external and specialist-redaction lists — membership there asserts that a name holds a secret worth scrubbing, not that anything reads it, and an operator's stale export still needs stripping from child processes.

- **Linked issue(s): close/fix refs** — follow-up to review feedback on #1541.

- **Tests: added/updated? commands run?** Updated. Three tests that locked the retired behavior became retirement locks instead: the Anthropic fallback now wins over a present `LLM_GATEWAY_KEY` (`test_llm_config`), a gateway-only host is rejected for `--critic-protocol=openai` (`test_cli_backends_unit`), and the specialist child no longer receives the name (`test_specialist_codex_backend`). The TraceLens test now asserts the host needs a real marker. Scenarios that merely used the gateway key as a stand-in credential (secret masking, header privacy, MCP env collision) were switched to `OPENAI_API_KEY`, which is what they were actually about.

  No commands run: this box has no Python (Store stub only, no venv), so I could not run pytest or ruff locally. CI is the gate. The affected files are `test_llm_config.py`, `test_codex_session.py`, `test_cli_backends_unit.py`, `test_specialist_codex_backend.py`, `test_anthropic_credential_registry.py`, `test_tracelens_model_normalize.py`, `test_sources_local_probe.py`.

- **`CHANGELOG.md` entry under `[Unreleased]`: yes/no** — yes, under `### Removed`.

- **Breaking changes: yes/no (details if yes)** — yes, for two shapes:
  1. A deployment whose `.env` carries `LLM_GATEWAY_KEY` as its only OpenAI-side credential must rename it to `OPENAI_API_KEY`. Value and endpoint unchanged.
  2. A host carrying `LLM_GATEWAY_KEY` alongside one configured side now classifies as single-provider, so the other side's variables are stripped as stale rather than kept on the theory that the gateway serves both.

  An Anthropic-only host that also carried the gateway key now resolves its Anthropic bearer for OpenAI-protocol calls — the credential it was already using for everything else.

- **PR addresses single concern: yes/no (details if no)** — yes: stop reading non-official names as credentials.

- **Root cause is upstream (Magpie/TraceLens/GEAK/IntelliKit/AgentKernelArena), ticket filed:** n/a for what landed here. The outbound half (`LLM_API_KEY` / `AMD_LLM_API_KEY` / `LLM_API_BASE` as GEAK's and the kernel tools' credential contract) is upstream work and is intentionally left in place.


fengshaoyi-amd:
Folding this back into #1541 rather than reviewing it separately.

Two reasons. First, it rewrites the same two functions #1541 rewrites, in the opposite direction: #1541 widened `openai_agent_credentialed` to also accept `LLM_GATEWAY_KEY` so the ranking would agree with `codex_session`, and this PR retires that name, which collapses the tuple back to `OPENAI_API_KEY` alone. Reviewed apart, the two PRs argue against each other and whichever merged second would have had to re-decide the question in a conflict resolution. `_validate_credentials` has only ever admitted a run on `OPENAI_API_KEY`, so the widening was the wrong end to fix and retiring the name is the whole answer — one decision, one review.

Second, this branch was cut from `main` rather than from #1541, so it never saw `OPENAI_AGENT_KEY_ENV_ORDER` and could not delete it; it added a second key-order constant beside it instead. The merged branch deletes the constant.

Two things changed on the way over, both because `main` retired the robustness agent (0d8dd3b2f) after this branch was cut:

- The two `agents/robustness/` hunks are dropped — the files no longer exist.
- `LLM_API_KEY` had exactly one credential read path, the RCA resolver, and it went out with the agent. The survey in #1541 says so rather than claiming a removal this no longer performs.

Everything else lands unchanged: the `resolve_openai_client_config` key order, `codex_session`, `CLAUDE_GATEWAY_SIGNAL_KEYS`, `_provider_only_mode`, the TraceLens gateway marker, the specialist forwarding allowlist, the docs and the tests.

PR #1555, opened 2026-09-17
refactor(llm): one definition of the agent default model ids

fengshaoyi-amd:

  Continues the `llm_config` consolidation from #1541 and [#1551](https://github.com/AMD-AGI/Hyperloom/pull/1551): after backend selection and credential names, the remaining duplicated decision is which model a backend runs by default.

  `DEFAULT_CLAUDE_MODEL` and `DEFAULT_CODEX_MODEL` were defined twice — `orchestrator/roles/agent_role.py` and `kernelforge/agent_backends/{claude,codex}.py` — and the same two strings were spelled as bare literals in six more places, including the Forge provider registry *in the same package as the module that defined the constant*:

  | Site | Was |
  |---|---|
  | `kernelforge/agent_backends/registry.py` | `default_model="claude-opus-5"` / `"gpt-5.6-sol"` |
  | `cli/credentials.py` | `_CLAUDE_PREFERRED_MODEL` / `_CODEX_PREFERRED_MODEL` as the allowlist heads |
  | `cli/preflight.py` | `CLAUDE_MODEL` or `"claude-opus-5"` for the GEAK default |
  | `tracelens_skill_runner.py` | `resolved_model or "claude-opus-5"` |
  | `orchestrator/roles/critic_agent.py` | `codex_model: str = "gpt-5.6-sol"` |

  They now live in `hyperloom/common/llm_config.py` beside `AGENT_BACKEND_CLAUDE` / `AGENT_BACKEND_CODEX`, the pair they are keyed by. This introduces no new layering: `kernelforge` already imports `hyperloom.common.reasoning_effort` at module level in four files, and `DEFAULT_REASONING_EFFORT` is the exact precedent — a shared default in `common` that both packages read. The comment explaining why the Codex default is `gpt-5.6-sol` (the gateway publishes a bare `gpt-5.6` with no deployment behind it, which answers 400) moved with the constant.

  **`resolve_forge_llm_model` lost its `default` parameter.** Every caller passed the chosen backend's own default. `request_handlers` reimplemented the per-backend branch the function already performs just to work out what to pass, and `patch_conflict_merge` carried a comment at each of its two call sites explaining that a default is mandatory because `CLAUDE_MODEL` is unset on OAuth-token runs and the resolver would otherwise post an empty model id. The function knows the backend; it now answers that itself, and the parameter that could be forgotten is gone.

  **What is deliberately left alone.** The narrative report model (`HYPERLOOM_REPORT_MODEL`), the robustness RCA model, the KB synthesis model and the quantization driver model are separate knobs that happen to share a value today. Each has its own override and its own reason to move, so collapsing them onto one constant would couple decisions that should stay free to diverge — that would be introducing coupling, not removing duplication. The `_CLAUDE_ALLOWED_MODELS` / `_CODEX_FALLBACK_MODELS` older rungs are also untouched; only the head, which is the default, now derives.

- **Linked issue(s): close/fix refs** — follow-up to review feedback on #1541.

- **Tests: added/updated? commands run?** Updated, no new ones. `test_llm_config` now asserts the resolver's own per-backend fallback where it used to pass `default=`. Four files had their import moved to `llm_config`. One assertion was deleted rather than moved: `test_default_claude_model_is_opus_5` asserted `_CLAUDE_PREFERRED_MODEL == DEFAULT_CLAUDE_MODEL`, a drift check between two copies — the allowlist head is now that constant by construction, so the drift is unrepresentable. The tuple assertion that locks the full ladder stays.

  No commands run: this box has no Python (Store stub only, no venv), so I could not run pytest or ruff locally. CI is the gate.

- **`CHANGELOG.md` entry under `[Unreleased]`: yes/no** — yes, under `### Changed`.

- **Breaking changes: yes/no (details if yes)** — no. Values are identical everywhere; this only changes where they are written. `resolve_forge_llm_model`'s signature change is internal and every call site is in this diff.

- **PR addresses single concern: yes/no (details if no)** — yes: one definition of the agent default model ids.

- **Root cause is upstream (Magpie/TraceLens/GEAK/IntelliKit/AgentKernelArena), ticket filed:** n/a.


xiaofei-zheng (approved):
No blocking issues. LGTM.

Checked:
- All call sites of `resolve_forge_llm_model` after the signature change: only `patch_conflict_merge` (claude + codex paths) and `request_handlers._resolve_forge_agent` remain, and no `default=` argument survives anywhere in the tree. Both previously passed the same per-backend value the function now picks itself, so the resolved id is unchanged; no caller relied on the old empty-string fallback.
- No dangling references to the moved/removed names: `_CLAUDE_PREFERRED_MODEL` / `_CODEX_PREFERRED_MODEL` and `agent_role.DEFAULT_*` / `agent_backends.{claude,codex}.DEFAULT_*` are gone repo-wide, including docs and scripts.
- Layering: `hyperloom/common/llm_config.py` imports only from `hyperloom.common.*`, so the new module-level imports from `kernelforge/agent_backends/{claude,codex,registry}.py` and from `orchestrator/roles/agent_role.py` introduce no cycle. Imported `llm_config`, `agent_role`, `kernelforge.agent_backends.{registry,claude}` and `orchestrator.kernel.request_handlers` in one interpreter to confirm.
- Ran `resolve_forge_llm_model` over every branch (explicit / CLAUDE_MODEL / CODEX_MODEL / empty env / empty backend) and `_resolve_forge_agent({})`: values match the constants and the pre-change behaviour.
- Tests: `test_llm_config.py -k forge_llm_model`, `test_patch_conflict_merge.py`, `test_agent_roles_and_policy.py` all pass (the single failure there is `import fcntl` on Windows, pre-existing and unrelated).
- CHANGELOG has an entry under `[Unreleased] > ### Changed`, and the PR description matches the current diff including the merge commit.

PR #1556, opened 2026-09-17
[DOCS] carry size, complexity and test-boundary rules in review

ZhengGong-amd:
- Description: Nothing in CI measures function length, cyclomatic complexity or module size — Ruff selects `E`/`F`/`W` (no `C901`) and Pylint runs `--errors-only`, which drops `R0912`/`R0915`. The Copilot prompt nevertheless told the reviewer these were covered, so oversized and over-branched code landed unremarked. This adds the thresholds as explicit **review triggers, not gates** (~60 lines, CC 10, ~800 lines, each just above the tree's 90th percentile), states that no linter carries them, and pairs them with a test-strategy rule: pin the exported surface — CLI flags, public functions, persisted schemas, artifact layouts — and cover internal plumbing through its entry point instead of per function.

  The style guide is the single authority for the numbers, how to measure them, and when a long unit is fine as it stands; `AGENTS.md`, the Copilot prompt and the PR template link to it rather than restating it. Counts of the existing backlog are deliberately absent — they go stale and constrain nothing.

  Two rules that were silent where an agent needs an answer now speak. `Delete, don't comment out` said dead code goes but never said what makes a replacement complete, so a new path could land beside the old one, with a wrapper kept for an internal shape, and nothing called it unfinished; as **Finish the replacement** it asks for the callers migrated, the superseded implementation and its switch, configuration and fixtures deleted, the regression assertions carried across, and compatibility justified by a named consumer rather than by a text search that found nothing. **Simplify by removing a mechanism** states what a refactor is judged on — one mechanism fewer across the whole responsibility, not the same moving parts in new positions — and records that inlining a wrapper or merging two equivalent entry points counts as a win, that caller count settles nothing, and that lookalike code answering to different contracts should stay apart. There is deliberately no net-lines-removed field: a number invites a quota and rewards formatting compression.

  Also fixes two claims that were wrong in the same way the PR set out to fix: cyclomatic complexity and unreachable code were listed as linter-covered when neither is. Unreachable statements, commented-out blocks and `# removed …` tombstones move to the Copilot prompt's *Debt growth* item, matching `AGENTS.md` § *Delete, don't comment out*.

- Linked issue(s): none
- Tests: none — docs and one commented-out `pyproject.toml` block. `ruff check --select C901 --config "lint.mccabe.max-complexity=10" src/hyperloom src/kernelforge` runs as written on Ruff 0.16.
- [Size/complexity](docs/contributing/style-guide.md#size-and-complexity) triggers crossed: none
- If this simplifies or refactors: n/a
- `CHANGELOG.md` entry under `[Unreleased]`: no — docs-only, nothing an operator can observe
- Breaking changes: no
- PR addresses single concern: yes — what review has to carry because no gate does, across the four files that tell a contributor or an agent what to check
- Root cause is upstream: no


xiaofei-zheng:
Reviewed the docs-only diff (5 files, +75/-11). Verified the tooling claims against the tree: `pyproject.toml:494` is indeed `select = ["E", "F", "W"]` and `.github/workflows/lint.yml:183` is `pylint --errors-only`, so the correction about cyclomatic complexity is right. The threshold numbers reproduce under a "production code, physical lines, tests excluded" reading (93 modules over 800 lines vs. the 92 quoted; 609 functions over CC 10 vs. 618 — same order, different snapshot), and the `ruff check --select C901 --config "lint.mccabe.max-complexity=10" ...` command in the style guide runs as written on ruff 0.15. Leaving C901 commented in `pyproject.toml` with the count and the on-demand command matches the existing `B`/`I`/`UP`/`SIM`/`RUF` precedent. Two blocking items.

**1. `.github/copilot-instructions.md:52` — "unreachable code → ruff/pylint" is not true, and it is the same class of error this PR set out to fix.**

```
- Unused variables, unreachable code → ruff/pylint. (Cyclomatic complexity is *not* ...
```

Checked both halves:

- `ruff check --select E,F,W` on `def f(): print(1); return 1; print(2)` reports nothing — Pyflakes has no unreachable-code rule, and the repo selects `E`/`F`/`W` only.
- Pylint's `W0101: unreachable` is warning-category, so CI's `--errors-only` invocation excludes it.

Net effect: the review bot is told to stay silent about something no gate covers — exactly the gap the same commit fixes for complexity. `Unused variables` alone is accurate (F841 fires). Either drop `unreachable code` from this line, or move it up alongside complexity as something review has to carry.

**2. Title and description do not match the diff.**

The title `Docs/zgong/building rules` is the branch name verbatim, and the body is still the unfilled template — `- Description: what and why`, `- Tests: ...`, and the rest are placeholders, including the `Size/complexity triggers crossed` line this PR itself adds to the template. The two commit messages already state the motivation, where the thresholds come from, and why they are review triggers rather than gates; moving that into the description is enough.

No CHANGELOG entry is needed here — the change is docs-only with nothing an operator can observe, which is what the repo rule asks for.

ZhengGong-amd:
Fixed (1) in `2e8c122a7`.

Verified both halves before changing the line: `ruff check --select ALL` reports nothing for a statement after `return` (Pyflakes has no unreachable rule), and pylint's is `W0101` — warning-category, so `--errors-only` excludes it. You were right, and it was the same swap-one-false-claim-for-another the PR set out to fix.

- `What NOT to flag` now names only what ruff genuinely reports: `F841`/`F401`.
- Unreachable statements, commented-out blocks and `# removed …` tombstones moved to the **Debt growth** review item, where `AGENTS.md` § *Delete, don't comment out* already puts them.

(2) not addressed yet.

zoroyihan7 (commented):
Reviewed at `2e8c122a`, with the goal of improving future agent-authored code and preparing for code simplification.

The 60-line / C901 > 10 / 800-line thresholds are reasonable initial review prompts. I would keep them advisory and use refactoring reviews to identify the redundant implementation, state, decision path, or forwarding layer removed.

The inline comments separate two policy gaps to fix (CI execution of replacement tests and review of changes to already oversized functions), a measurement correction, and suggestions for making the rules support complete simplification. These are policy-review findings; this PR does not itself change runtime behavior or remove tests.

For the upcoming refactoring work, start with a frequently changed responsibility whose consumers and behavior can be verified independently. Preserve supported CLI behavior, persisted state and protocol compatibility, artifact layout, and relevant failure/recovery behavior. Add missing regression assertions before removing the implementation they protect. More involved orchestration and hardware paths need a reliable test boundary before their internals are simplified.

docs/contributing/style-guide.md
```diff
@@ -116,7 +137,9 @@ CI runs `pylint --errors-only` on core packages (fatal/error severity only). Fix
 
 - `critic_agent_e2e`, `robustness_agent_e2e`, `targeted_build_e2e`
 
-**Coverage:** CI enforces **90% line coverage** on measured trees (`[tool.coverage.report] fail_under`). CLI drivers, subprocess wrappers, and hardware-only paths are omitted from the denominator — see `[tool.coverage.run] omit`. Add unit tests for logic you introduce; do not chase coverage on omitted paths.
+**What to test:** Pin the **exported surface** — CLI flags, public functions, persisted schemas, artifact layouts — with unit tests that state the contract *and* its failure modes; those are what callers outside this repo depend on. Internal functions that only thread a business flow together do not each need one: per-function tests there assert the current implementation and break on the next refactor. Cover those flows **end to end** (see the `*_e2e` markers above), and unit-test an internal helper when it carries real logic of its own.
```
zoroyihan7, line 140: **[P2] Require replacement flow tests to run in normal PR CI.**

The recommendation points contributors to the three `*_e2e` markers, but [`pytest_ci_args`](https://github.com/AMD-AGI/Hyperloom/blob/2e8c122a79dbddf2c547f2eaa69f73b3f3226e50/pyproject.toml#L455) excludes all three, and the [local development checklist](https://github.com/AMD-AGI/Hyperloom/blob/2e8c122a79dbddf2c547f2eaa69f73b3f3226e50/docs/contributing/style-guide.md#L210) excludes them too. Following this rule when replacing existing unit tests could therefore remove routinely executed regression protection while leaving the PR green.

The principle of avoiding one test per private function is useful. Please choose the test boundary by the behavior being protected: focused unit tests for deterministic logic, and integration/flow tests for orchestration that run in normal CI. Hardware/external-runtime E2E can supplement that baseline. CLI flags, persisted schemas, and artifact layouts should also be allowed to use CLI/filesystem/serialization integration tests where those protect the contract more directly.

There is already a suitable [CLI + fake service + temporary files example](https://github.com/AMD-AGI/Hyperloom/blob/2e8c122a79dbddf2c547f2eaa69f73b3f3226e50/src/hyperloom/agents/framework/tests/integration/test_e2e_execute_mock.py#L59) in the tree.

Suggested rule: “Preserve contract and failure-mode assertions, and their execution in normal PR CI, when replacing tests.” The aggregate 90% line-coverage floor cannot establish this, particularly for omitted paths. Please align the corresponding AGENTS, Copilot, and PR-template wording.

.github/copilot-instructions.md
```diff
@@ -13,6 +13,16 @@ Advisory review. Flag only what static gates can't.
   unhandled `None`, silent truncation, a branch that returns the wrong default.
 - **Duplication / parallel routes**: a second implementation of something that already
   exists. Point to the existing one and ask to extend it.
+- **Size and complexity**: a function past ~60 lines, cyclomatic complexity past 10, or a
+  module pushed past ~800 lines. No linter covers this (Ruff runs `E`/`F`/`W`; Pylint is
+  `--errors-only`), so review is the only place it surfaces. Name the seam to split on,
+  not just the line count — and treat an existing long function the change merely edits as
+  out of scope.
```
zoroyihan7, line 20: **[P2] Exempt existing debt without exempting new complexity added to it.**

“An existing long function the change merely edits” does not distinguish a small fix from appending more branches or responsibilities. A function growing from 59 to 65 lines can trigger review, while the same addition to a 150-line function can be treated as out of scope. That leaves an incremental-growth gap against the style guide's “Do not grow it.”

Please make the exemption apply to unrelated cleanup demanded solely because the function was already oversized. Still review a newly crossed threshold or a change that materially increases the complexity/responsibilities of an already oversized function, and allow an explained exception when the structure remains clearer.

For an explicitly scoped simplification PR, the existing implementation and its callers are part of the target responsibility. A normal bug fix should not be required to repay all historical debt. This distinction should also appear in the “new or rewritten code” authoring rule.

docs/contributing/style-guide.md
```diff
@@ -52,6 +53,26 @@ ruff format --check .   # or `ruff format .` to apply
...
 
+### Size and complexity
+
+Nothing enforces these today: Ruff selects `E`/`F`/`W` only (no `C901`), and CI's Pylint is `--errors-only`, which excludes `R0912`/`R0915`. They are **review triggers for new and rewritten code** — the point at which a reviewer asks for a split or for the reason the shape is right.
+
+| Unit | Trigger | Where the number comes from |
+|------|---------|-----------------------------|
+| Function length | ~60 lines | Just above the tree's 90th percentile (55 lines); past this a function is usually two |
+| Cyclomatic complexity | 10 | McCabe default; measurable on demand with `ruff check --select C901` |
+| Module length | ~800 lines | Roughly the tree's 90th percentile (914 lines) |
```
zoroyihan7, line 64: **Suggestion: use percentiles to calibrate review volume, and qualify the design inference.**

The current distribution supports using these as initial review prompts; it does not establish that a function above 60 lines “is usually two.” For example, [`build_system_prompt`](https://github.com/AMD-AGI/Hyperloom/blob/2e8c122a79dbddf2c547f2eaa69f73b3f3226e50/src/kernelforge/kernel_backends/hipblaslt/prompts.py#L12) spans 212 physical lines and has Ruff CC 1: its body returns a prompt template. Length alone does not identify a second responsibility.

C901 also has limits: [`_infera_restart_config_matches`](https://github.com/AMD-AGI/Hyperloom/blob/2e8c122a79dbddf2c547f2eaa69f73b3f3226e50/src/hyperloom/inference_optimizer/multi_node/commands/infera.py#L226) is 40 lines with CC 4, yet contains two groups of 6 and 9 field comparisons and substantial short-circuit logic. Those conditions still need semantic review.

I would retain 60 / 10 / 800, replace “usually two” with “inspect whether there is a clear responsibility boundary,” and explicitly allow justified template/data-heavy cases. Splitting should improve ownership, data flow, or testability. There is no evidence here for tightening the numbers or turning them into hard gates.

pyproject.toml
```diff
@@ -502,6 +502,14 @@ ignore = ["E501", "E741"]
 # Future-proofing -- uncomment once the current findings are at zero.
 # extend-select = ["B", "I", "UP", "SIM", "RUF"]
 
+# C901 (mccabe) is the complexity trigger in the style guide's "Size and
+# complexity" table. It stays off rather than on-and-ignored: 618 production
+# functions sit above the default max of 10 today, so selecting it would red the
+# tree instead of gating new code. Measure on demand -- no config edit needed:
+#   ruff check --select C901 --config "lint.mccabe.max-complexity=10" src/hyperloom
```
zoroyihan7, line 509: **Measurement correction: 618 includes tests, and the example scans only one package.**

At this head, using Ruff 0.16.7, the style guide's two-package command produces 618 C901 findings: **597 in production code and 21 in tests**. The production split is 449 in Hyperloom and 148 in KernelForge. This comment calls all 618 “production functions,” while its example command scans only Hyperloom.

Please state the population and use a command matching it. The line-count statistics also need a reproducible definition: commit, included/excluded paths, whether async/nested functions are counted, and whether “lines” includes comments/docstrings.

For comparison, counting tracked Python in both packages, excluding tests, conftest files, shipped KernelForge data, and the two vendored SDK copies, I get 794 modules and 9,279 functions; function P90 55 and 798 above 60; module P90 916 (nearest-rank) and 91 above 800. Function length here is the inclusive AST definition span, including async/nested definitions and internal blank/comment/docstring lines, excluding decorators; module length uses physical lines. The small differences from 801/914/92 could reflect scope or snapshot choices and do not justify changing the thresholds.

Clarify whether tests and shipped examples follow the same size-review policy as production code.

AGENTS.md
```diff
@@ -49,8 +49,17 @@ doc is the authority on where that boundary lives.
...
 - **One concern per change.** A PR fixes one issue or adds one capability. If you must
   bundle, say why in the description. Don't ride unrelated refactors in on a fix.
-- **Size budget.** Prefer reviewable diffs. A large diff is a signal to split, not to push
+- **Diff budget.** Prefer reviewable diffs. A large diff is a signal to split, not to push
   harder. Cleanup in an unrelated file is a separate PR.
+- **Size is a design signal.** A function that keeps growing, a branch tree you have to
+  scroll, a module that collects everything — that is the design telling you a boundary is
+  missing, and the answer is the split, not a bigger screen. For new or rewritten code, a
+  function past ~60 lines or cyclomatic complexity 10, or a module past ~800 lines, needs
+  a reason in the PR description or a split. These are review triggers, not gates: no
```
zoroyihan7, line 58: **Suggestion: make simplification across the full responsibility an explicit success criterion.**

For agent-authored refactoring, the rules should reward removing a redundant mechanism: duplicated behavior, an extra state/configuration source, an obsolete decision path, or a forwarding layer without an independent responsibility. Review the affected implementation together with its callers, including newly introduced helpers, interfaces, parameter plumbing, and cross-file dependencies.

Extract a helper when it names a complete responsibility or shares the same semantic rule. Inlining thin wrappers and merging equivalent entry points should be equally valid outcomes. A single-caller helper can be useful; caller count alone should not decide this. Similarly shaped code with different contracts should not be forced into a generic function with extra modes.

The existing “Derive over hardcode” rule could explicitly cover duplicated mutable state and decisions: identify the authoritative owner, derive what can be derived reliably, and state the consistency invariant for any necessary cache or replica.

For a PR whose stated goal is code reduction, report net production-code changes over the affected scope, separately from tests/docs/data, together with the mechanism removed. Avoid a fixed deletion quota or credit for formatting compression; new features may legitimately add code.

AGENTS.md
```diff
@@ -90,9 +107,12 @@ doc is the authority on where that boundary lives.
...
   docstrings are a separate requirement; see the style guide.
-- **Clean design.** One boundary rule per concern, owned by one module. Derive over
-  hardcode — a single computed source beats duplicated constants. Minimal typed
-  interfaces.
+- **Clean design.** One boundary rule per concern, owned by one module. A module should
+  read as one job: cohesive inside, a minimal typed interface outward, and dependencies
+  pointing one way down the layers — no cycles, and no reaching around the layer that owns
+  a thing to touch what is behind it. Derive over hardcode — a single computed source
+  beats duplicated constants. A second copy of a behaviour is a bug you will later fix
+  once and miss elsewhere; extend the existing one, or lift the shared part out.
```
zoroyihan7, line 115: **Suggestion: complete replacements and make compatibility serve identified consumers.**

Please extend the deletion/reuse rule to cover redundant paths that are still reachable. A replacement should migrate its callers and remove the superseded implementation, obsolete switches/configuration, and tests or fixtures tied only to the old structure, while preserving required regression assertions.

For callers that can migrate together, update the interface and callers in the same change. Avoid leaving wrappers, aliases, dual-format parsing, or fallback routes solely to preserve an old internal shape. Compatibility remains necessary for supported public APIs, CLI behavior, persisted data, plugins, and independently deployed consumers; identify that boundary before deleting anything. Repository location or an empty text search is not sufficient evidence that a consumer can migrate or that an entry point is unused.

If migration must be staged, name the remaining consumer and the retirement condition for the temporary path. Keep the adapter small and avoid making it a new general-purpose entry point. This would give agents a concrete completion condition for simplification and can be folded into the existing deletion/duplication rules.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,6 +1,7 @@
 - Description: what and why
 - Linked issue(s): close/fix refs
-- Tests: added/updated? commands run?
+- Tests: added/updated? commands run? (exported behaviour pinned by UT; business flow covered e2e)
+- Size/complexity triggers crossed (function >60 lines, CC >10, module >800 lines): split, or why the shape is right:
```
zoroyihan7, line 4: **Suggestion: keep one authoritative definition and ask for a short simplification rationale.**

These policies now appear in AGENTS, the style guide, the Copilot prompt, and this template. Keep the thresholds, measurement scope, and exceptions authoritative in the style guide; retain concise authoring principles in AGENTS, evidence-oriented review guidance in Copilot, and links plus a few prompts here. Useful reminders can remain without copying the full policy into four places.

For a dedicated simplification PR, three facts are enough: **what mechanism was removed, what contract was preserved, and which tests demonstrate that preservation**. Define the scope around one responsibility and its callers; a coherent migration can span files without bundling unrelated work.

Later, a lightweight advisory report of before/after metrics for changed units could make the thresholds consistent using Ruff and the agreed line-count definition. There is no need to make a repository-wide debt platform or a fixed percentage-reduction target a prerequisite for this PR.

xiaofei-zheng (approved):
No blocking issues. Both items from the earlier round are addressed: the title and description now match the diff, and `unreachable code` moved out of *What NOT to flag* into **Debt growth** with the reason stated (`W0101` is warning-category, so `pylint --errors-only` never reports it) — the remaining `F841`/`F401` claim there is accurate.

What I checked on the current head (`b4908b9`):

- Tooling claims against the tree: `pyproject.toml:494` is `select = ["E", "F", "W"]` and `.github/workflows/lint.yml:183` is `pylint --errors-only`, so nothing measures `C901`/`R0912`/`R0915`, as the style guide and the Copilot prompt now say.
- `ruff check --select C901 --config "lint.mccabe.max-complexity=10" src/hyperloom src/kernelforge` runs as written and reports the backlog the commented `[tool.ruff.lint.mccabe]` block describes; leaving it off follows the `B`/`I`/`UP`/`SIM`/`RUF` precedent directly above it.
- The example cited under *What to test*, `src/hyperloom/agents/framework/tests/integration/test_e2e_execute_mock.py`, exists and matches the description — it drives `fa explore --execute` against a fake PR Monitor in a `tmp_path`. It carries no marker, so it is inside `addopts`' default selection (`pyproject.toml:453`), which is what the "keep them in the default CI selection" sentence relies on.
- Renames are consistent: no reference to `Size budget` or `Delete, don't comment out` survives anywhere in the branch, and the `#size-and-complexity` / `#tests-pytest` anchors resolve to the headings added or present in the style guide.
- Docs-only with no operator-visible change, so no `CHANGELOG.md` entry is required.

docs/contributing/style-guide.md
zoroyihan7: Rechecked at `b4908b925`. The normal-PR-CI regression gap is fixed: the style guide now preserves contract/failure-mode assertions and their default CI execution, treats the excluded E2E markers as supplementary, and permits integration tests at the appropriate boundary. The Copilot rule and PR template also match.

One part of the requested alignment remains: [`AGENTS.md:76-78`](https://github.com/AMD-AGI/Hyperloom/blob/b4908b9250e1bc3161f47af515d0729a3465d3b8/AGENTS.md#L76) still says the exported surface "is pinned by unit tests", while the style guide explicitly permits CLI/filesystem/serialization integration tests. Please change that to "is pinned by tests" and refer to the style guide for the boundary choice. This avoids making an agent add a redundant unit test for a contract already protected by an integration test.

Leaving this thread open only for that wording alignment; the original CI-execution gap has been addressed.

.github/copilot-instructions.md
zoroyihan7: Verified at `b4908b925`. The Copilot rule now explicitly reviews newly crossed thresholds and branches/responsibilities added to an already oversized unit. AGENTS and the style guide distinguish that increment from unrelated historical cleanup and allow an explained exception when the current structure is clearer. This addresses the incremental-growth gap. Resolving.

docs/contributing/style-guide.md
zoroyihan7: Verified at `b4908b925`. The unsupported "usually two" inference is gone. The style guide now explains both the long-template and short-but-semantic-complexity cases, accepts a justified single-template exception, and judges a split by ownership, data flow, or testability. The thresholds remain advisory. This addresses the suggestion. Resolving.

pyproject.toml
zoroyihan7: Rechecked at `b4908b925`. Removing the stale backlog counts is reasonable, and the pyproject example now scans the same two packages as the style guide. That fixes the incorrect "618 production functions" claim and the command mismatch; there is no need to restore those counts.

The remaining part of this comment is the definition of the thresholds that are still present. [`Size and complexity`](https://github.com/AMD-AGI/Hyperloom/blob/b4908b9250e1bc3161f47af515d0729a3465d3b8/docs/contributing/style-guide.md#L56) still does not define physical versus logical lines, what a function's span includes, or whether tests, shipped examples/data, and vendored code follow the same scope. AGENTS now says this section owns "how to measure them", but it only supplies the C901 command.

Please add a short line-count/scope definition there. For example, specify an inclusive function definition span (including async/nested definitions and whether decorators/docstrings/comments count), physical module lines, and the included/excluded code categories. A paragraph is enough; no historical-count table or new metrics tooling is needed. Leaving this thread open for that clarification.

AGENTS.md
zoroyihan7: Verified at `b4908b925`. "Simplify by removing a mechanism" now assesses the complete responsibility and its callers, including new helpers/interfaces/parameter plumbing; it also allows inlining/merging and distinguishes shared semantics from similar-looking code. "Clean design" covers ownership and invariants for duplicated state/decisions.

I am also comfortable with omitting a mandatory net-lines field: the mechanism removed, contract preserved, and test evidence provide the useful review criteria without introducing a deletion quota. The substance of this suggestion is addressed. Resolving.

AGENTS.md
zoroyihan7: Verified at `b4908b925`. "Finish the replacement" covers caller migration, removal of the superseded implementation/switch/configuration/fixtures, and preservation of regression assertions. It ties compatibility to identified consumers, rejects an empty text search as proof of no consumers, and requires a retirement condition for staged migration. The Copilot guidance now asks about consumers of a retained old path as well. This addresses the suggestion. Resolving.

.github/PULL_REQUEST_TEMPLATE.md
zoroyihan7: Verified at `b4908b925`. The numeric thresholds and exceptions now live in the style guide, with the author/reviewer guidance referring to it. The template links to the test/size rules and asks the three useful simplification questions: mechanism removed, contract preserved, and supporting tests. This addresses the organization/template suggestion. Resolving this thread; the remaining measurement-definition clarification stays in the separate [measurement thread](https://github.com/AMD-AGI/Hyperloom/pull/1556#discussion_r4039240670).

docs/contributing/style-guide.md
zoroyihan7: Verified at `a0a337685`. AGENTS now says the exported contract is pinned by tests and explicitly allows a unit test or a CLI/filesystem test, whichever protects it more directly. That aligns with the style guide while retaining the requirement to preserve assertions and their execution in default PR CI. The remaining wording mismatch is fixed; this thread is fully addressed. Resolving.

pyproject.toml
zoroyihan7: Verified at `a0a337685`. The style guide now defines the function span, decorator/comment/blank/docstring treatment, nested/async definitions, physical module lines, and the two-package scope using Ruff's existing exclusions. It explicitly exempts tests from the size triggers while noting that the C901 command still reports them. The referenced exclusions include the vendored SDK copies and shipped examples, and both command examples scan the same two packages. This completes the measurement clarification; no additional metrics tooling or historical-count table is needed. Resolving.

PR #1559, opened 2026-09-17
fix(breakdown): retain rows with opaque natural identifiers

iraj465:
Session Breakdown drops attempt, acceptance, and integration rows when an external identifier contains `:`, such as `aiter:paged_attention_ragged`. The recorder treats the identifier as part of its key grammar and rejects the row even though the benchmark continues.

Keep these identifiers opaque: hash the ordered identifier tuple into a reserved fragment-key namespace, preserving existing valid keys and the original IDs in row payloads. Recording, resumed updates, existence checks, event filtering, and late rejection now use the same stable key. Event IDs, row types, and empty natural IDs retain their existing validation.

- Tests: 183 distinct focused/compatibility cases passed, including tuple-boundary collisions, percent escapes, Unicode, resumed updates, and the full GEAK acceptance/rebench/integration/rejection path. Full local suite: 22,718 passed, 55 skipped, 4 failed; all four failures also reproduce on unchanged base `2cc382a8` (three robustness tests and one supervisor report test). No new failures.
- Ruff, format, REUSE, and diff checks passed. Full mypy output exactly matches the unchanged base: 920 existing errors in 156 files, with no new diagnostics.
- `CHANGELOG.md`: updated under Unreleased.
- Breaking changes: none for previously valid keys or row payloads.
- Single concern: retaining recorder rows with external identifiers; launch settings and acceptance gates are unchanged.
- Linked issue / upstream ticket: none; the root cause is in Hyperloom's recorder.


chao-xu-spec:
Please rebase (drop the retired CHANGELOG.md) and update the comments that still say a : drops the row: _record_geak_integration in kernel.py, and the predispatch run_id comment in instrument.py. After this change those comments are false. Expanding the fragment_key docstring to state the two key forms would be enough documentation.

PR #1561, opened 2026-09-17
fix(kb): stop storing what_worked/what_failed rows as empty strings

rpoornac:
## What breaks

`what_worked` / `what_failed` rows reach the KB with every value replaced by `""`.

The Coordinator is the only producer of these rows and stamps them with `{name, extra_server_args, extra_envs, gain_pct, source}` (`writeback._build_recipe_attrs_from_state`). The store then normalised each row down to the arbor pair — `_normalise_str_dicts(what_worked, ("description", "measured_impact"))` — a vocabulary nothing in this repo writes or reads on these two columns. Nothing translated between the two, so every value was dropped:

```
sent:   {"name": "mtp_on", "extra_server_args": "--speculative-num-steps 3",
         "extra_envs": {...}, "gain_pct": 4.2, "source": "https://pr/123"}
stored: {"description": "", "measured_impact": ""}
```

This is not a missing field. The row count survived, so the KB looked populated while carrying nothing.

## Observable effect

An operator resuming work on a model that Hyperloom has already optimised gets the benefit of the earlier sessions. Before this, the scout's `already_proven` list only ever held the *current* session's accepted variants: `prelude._warm_recipe_proven_items` reads `name` and skips rows without one, so it always returned `[]`. Anything proven in an earlier session was invisible and got re-researched from scratch — the knowledge base's central promise, not paying out. Recipes written before this change still load; their variant name is recovered from the old `description` key.

## The fix

These two columns have one producer and two readers (`_warm_recipe_proven_items`, `_inject_warm_recipe_history_into_ledger`), and all three speak the producer's keys. The arbor projection sitting between them had no participant, so it is deleted rather than taught exceptions.

`Finding` and `Failure` now share an `Experience` base holding `{name, extra_server_args, extra_envs, gain_pct, source}`, with one `from_mapping` / `to_dict` pair each (`Failure` adds `reason`). `put_recipe` coerces to dicts and hands them to `Recipe.from_dict` like every other field. One vocabulary, one ser/de: a field the Coordinator adds next arrives without a second schema to teach.

Rows written under the old projection kept the variant in `description`, so `from_mapping` falls back to it for `name`. There is no write-side mirror back to `description` — it has no reader, and `measured_impact` was never populated, so emitting it would rebuild half of the shape that caused this.

`put_recipe` is the boundary, so a bad type now raises there rather than landing a row with the value silently gone: a non-numeric `gain_pct` or a non-mapping `extra_envs` used to vanish while the row persisted.

## Why no test caught it

Both halves of the seam were already covered: one test asserts writeback stamps `name`, another asserts prelude reads it. But the prelude test hand-injected the row, so nothing exercised the persistence step *between* them — exactly where the data died. `test_item9_on_disk_json_uses_arbor_field_names` pinned `{"description", "measured_impact"}` as the on-disk shape, so it asserted the defect.

Added: the round trip through `LocalRecipeStore`, the end-to-end path into `already_proven`, `from_dict` → `to_dict` inverseness, a legacy `description`-only row, and the boundary raising on bad types. The negative `gain_pct` a REVERT carries is asserted explicitly, since a falsy check would drop exactly the rows that column exists to record.

## Scope

The explore dedup pre-fill stays inert: it fingerprints from `extra_server_args` / `extra_envs`, which the Coordinator records on neither the failure rows nor the reverted rows. Storage no longer stands in the way; that fix is producer-side and belongs to its own PR.

## Verification

- `630 passed` across the `recipe_kb` / sedimentation / local-store / warm-start suites.
- `ruff check src/hyperloom` clean; formatting clean.
- The whole-repo suite hangs at ~99% with zero failures on this branch. Clean `main` at the same base hangs identically under the same command, so it is the known pre-existing flake; the CI shards are the authoritative signal.
- No `CHANGELOG.md` entry: #1589 retired the file, and the observable effect is stated above per the updated AGENTS.md rule.


ZhengGong-amd:
The defect is real and well diagnosed. The fix is the wrong shape — this needs a substantial cut, not more field plumbing.

## The structural problem

These two columns have one producer (`writeback._build_recipe_attrs_from_state`) and two readers (`prelude._warm_recipe_proven_items`, `_inject_warm_recipe_history_into_ledger`). All three speak `{name, extra_server_args, extra_envs, gain_pct, source}`. Nothing in this repo writes or reads the arbor pair `{description, measured_impact}` on `what_worked` / `what_failed`. The projection in `put_recipe` is not a compatibility layer — it is a second schema with no participant, and it is what ate the data.

The PR keeps that projection and teaches it five exceptions. The same field list now lives in four places: the dataclass fields, `_experience_extras`, `_experience_kwargs`, `_normalise_experience`. The next producer field is dropped silently until all four are edited again.

Delete the projection for these two columns instead.

## What to change

1. **On-disk row shape = the producer's keys.** One `from_mapping` / `to_dict` on `Finding` / `Failure` (or a single `Experience` type). `put_recipe` coerces to dict and hands it to `Recipe.from_dict`, like every other field.
2. **Drop `_normalise_experience`, `_experience_extras`, `_experience_kwargs`** once (1) covers them. Keep `_normalise_str_dicts` for `remaining_gaps` / `pitfalls` — their producer does write arbor keys.
3. **Legacy rows: one read-side fallback**, `name = d.get("name") or d.get("description")`. No write-side `description = name`: it is a mirror column with no reader, and `measured_impact` is still `""`, so the arbor story is half-built either way.
4. **`test_item9_on_disk_json_uses_arbor_field_names` pins `set(on_disk["what_worked"][0]) == {"description", "measured_impact"}`** — that assertion is the one that should have failed. Repoint it at the producer keys; the top-level key names it also guards are the part that matters.

## Also

- `Pitfall.severity` is not the precedent claimed. It is an arbor-shape optional field that `to_dict` always emits; these superset fields are conditionally omitted. The hyperloom-superset precedents in this file are `kernel_optimizations` and `extras`.
- New silent drops, same shape as the bug being fixed: a non-`dict` `extra_envs`, a non-numeric `gain_pct` (`"4.2"`), and any key outside the allowlist (`fingerprint`, `error_class`) vanish with the row still present. `put_recipe` is the boundary — let a bad type fail, and trust `writeback`, which already passes `float | None`.
- Tests: the `what_failed` assertions check only `name` / `reason`, and `reason` is an arbor key that survives on `main`; the negative `gain_pct`, which is what the new code is for, is unasserted. There is no `Recipe.from_dict` → `to_dict` case, and the "byte-identical for arbor-only rows" invariant is claimed but not locked. `test_local_store.test_normalise_findings_and_failures` still pins the old behaviour.
- Docstrings: `Failure` says the explore dedup gate is pre-filled from these rows, while the PR body says that path is still inert. `_normalise_experience`'s docstring narrates the bug and the change (AGENTS.md: comments don't narrate the diff). `Recipe` still says "arbor payload (verbatim shape)".
- CHANGELOG landed under `## [v1.0.0a3] - 2026-08-05` rather than `[Unreleased]`, and carries an HTML `<br/>`.

Done this way the diff should be smaller than the current one: no translation step, one vocabulary, one ser/de. Scope is right — leave the producer-side explore-dedup gap to its own PR, just don't describe it as served.

rpoornac:
You're right, and the cut is made. Reworked to your design in `9ddec9dc`. I verified each point against the code rather than taking them on faith; notes below where I found something worth reporting.

## The structural problem — confirmed, and it is exactly as you describe

I traced all three participants:

- **Producer** `writeback._build_recipe_attrs_from_state` writes `what_worked` rows as `{name, extra_server_args, extra_envs, gain_pct}` plus `source` when provenance resolves, and `what_failed` rows as `{name, reason}`. `_collect_attempt_provenance`'s reverted rows add `{name, reason, gain_pct, source?}`.
- **Readers** `_warm_recipe_proven_items` reads `name` / `source`; `_inject_warm_recipe_history_into_ledger` reads `name` / `extra_server_args` / `extra_envs`.
- **`description` / `measured_impact`**: no writer and no reader on either column. The only other `measured_impact` in non-test code belongs to `lessons`, which has a real producer at `writeback.py:1727`.

I also checked the remaining consumers before deleting anything — `values.py`, `recipe_kb_t0.py`, `proposals.py` all pass these rows through opaquely as `list(... or [])` and never index either key. So nothing external depended on the projection.

## What changed

1. **On-disk shape is the producer's keys.** `Finding` and `Failure` now share an `Experience` base carrying `{name, extra_server_args, extra_envs, gain_pct, source}`, with one `from_mapping` / `to_dict` pair each; `Failure` adds `reason`. `put_recipe` coerces to dicts via `_coerce_dicts` and hands them to `Recipe.from_dict` like every other field.
2. **`_normalise_experience`, `_experience_extras`, `_experience_kwargs` are gone.** `_normalise_str_dicts` stays for `remaining_gaps` / `pitfalls`.
3. **One read-side fallback**, `name = d.get("name") or d.get("description")`. No write-side mirror, for the reason you gave.
4. **`test_item9` repointed** at the producer keys; the top-level arbor names it also guards are untouched.

Net production diff is now +98/−31 against +120/−8 before — it deletes the translation layer instead of extending it.

## On your other points

- **`Pitfall.severity`** — you're right, `to_dict` emits it unconditionally, so it was never the precedent for conditional omission. Moot now: `to_dict` emits a fixed key set with no conditionals.
- **Silent drops** — fixed at the boundary. A non-numeric `gain_pct` and a non-mapping `extra_envs` now raise rather than persisting a row with the value gone. Since the whole row dict reaches `Recipe.from_dict` unprojected, there is no allowlist left for `fingerprint` / `error_class` to fall outside of.
- **Tests** — the negative `gain_pct` is now asserted (a REVERT's gain is negative, so a falsy check would drop exactly the rows the column exists for), and there are new cases for `from_dict` → `to_dict` inverseness, a legacy `description`-only row, and the boundary raising on bad types. `test_local_store.test_normalise_findings_and_failures` is repointed at `_coerce_dicts`. `test_item8` used the old shape as sample data; updated. I left `test_put_get_history_roundtrip` on the old shape deliberately — it now exercises the legacy read path.
- **Docstrings** — `Failure` now says the dedup gate stays inert until the producer records args/envs, which matches what I found: neither the failure rows nor the reverted rows carry them, and the reader `continue`s without them. Diff narration removed; `Recipe`'s "verbatim shape" comment now excepts these two columns.
- **CHANGELOG** — resolved by #1589 retiring the file. The entry is dropped and the observable effect is in the PR description per the new AGENTS.md rule.

Scope held: the producer-side explore-dedup gap is left to its own PR and is no longer described as served.

ZhengGong-amd:
Blocking issues: 1

1. [X2] Delete the field list and store the mapping. [verified]
   Problem: `Experience._base_kwargs` / `_base_dict` (`schema.py:106-123`) are a second schema. A key outside `{name, extra_server_args, extra_envs, gain_pct, source, reason}` is dropped and the row is still saved. `local_store.py:331` says the next Coordinator field arrives without one. Today's writer only emits keys inside that list, so current rows survive; the next key does not.
   Impact: the same silent drop this PR removes, one field list later.
   Action: Delete `Experience`, `_base_kwargs`, `_base_dict`, `_experience_envs`, and `_experience_gain`. Persist the coerced dict through `Recipe.from_dict` / `to_dict`. Keep only: if `name` is absent, use `description`; a present non-numeric `gain_pct` raises; a present non-mapping `extra_envs` raises. Do not add another translator.

   Also correct the description: the old projection stored `{description: "", measured_impact: ""}` for Coordinator rows, so `schema.py:109` does not restore those names. Say the fallback applies only when `description` was actually stored. Do not add recovery code.

Checked: schema.py, local_store.py put_recipe, writeback._build_recipe_attrs_from_state, prelude readers | Ran: round trip plus bad `gain_pct` at head; pytest 50 passed on the three touched test modules | Base: f6a17b3330863383516f525e65fc199b49dd180d | Head: 562c7853e6d32e081ad7936e8940f8618846e305

rpoornac:
Both points confirmed against the code and fixed in `e24357d7`. The field list is gone.

## [X2] — reproduced, then deleted

Your diagnosis was exact, and it was live rather than hypothetical. `fingerprint` passed into `put_recipe` at the old head:

```
what_worked: [{"extra_envs": {"A": "1"}, "extra_server_args": "", "gain_pct": 4.2, "name": "TP=8 helped", "source": "explore"}]
```

The key is gone and the row saved — the same failure mode this PR opened against, one field later, exactly as you said. The drop was on the write path: `put_recipe` builds through `Recipe.from_dict` → `to_dict` (`local_store.py:351`), so the five-field `_base_kwargs` was the boundary that ate it.

`Experience`, `Finding`, `Failure`, `_base_kwargs`, `_base_dict`, `_experience_envs` and `_experience_gain` are deleted. `Recipe.what_worked` / `what_failed` are `list[dict[str, Any]]`, `from_dict` / `to_dict` carry the row whole, and `_experience_row` is the only thing left on the path: two type checks and the `name` fallback, no field list. Same input now:

```
what_worked: [{"extra_envs": {"A": "1"}, "fingerprint": "abc123", "gain_pct": 4.2, "name": "TP=8 helped", "source": "explore"}]
```

`Finding` / `Failure` came out of `schema.__all__` with them. Nothing outside `schema.py` used them — every other `Finding` in the tree is `agents.framework.models.Finding` — and the one test import is repointed.

Production diff is +44/−41, against +98/−31 for the cut you reviewed.

## The description was wrong, and it was wrong in the direction you said

I checked what `main` actually persisted rather than reasoning from the projection:

```
MAIN what_worked: [{"description": "", "measured_impact": ""}]
MAIN what_failed: [{"description": "", "reason": "OOM"}]
```

`_normalise_str_dicts(what_worked, ("description", "measured_impact"))` is `{k: str(d.get(k) or "")}`, so a Coordinator row's `name` never became `description` — it became `description: ""`. The variant was destroyed at write time, and no read-side fallback can recover it. My comment claimed those rows "have the variant under `description`", which is false for precisely the rows it named.

The fallback now applies only where a `description` was actually stored, which is the arbor-sourced rows, and the comment says so. No recovery code added. The test is split accordingly: a row that stored a description reads back named, and a Coordinator-projected row is asserted to stay nameless.

## On `gain_pct`

Taken at the strict reading, since you listed `"4.2"` as non-numeric in the first pass: a present `gain_pct` that is not `int` / `float` raises, so the stringified number fails with the unparseable one. `writeback` passes `float | None`, and a store that quietly parses `"4.2"` would take `"not-a-number"` just as quietly.

## Tests

The property the field list could not express is now pinned directly: `test_an_experience_row_keeps_a_key_this_module_never_heard_of` round-trips `fingerprint` and `error_class` through the store and asserts row equality, so the next field list to appear here fails a test instead of a session. The boundary case is parametrised over all three bad types, and the `from_dict` → `to_dict` inverse and negative `gain_pct` cases are unchanged.

1614 tests pass across the recipe_kb, sedimentation, local_store, warm-start, prelude and writeback suites. Rebased onto `09d1ff56`.

ZhengGong-amd:
## PR #1561 -- re-review at e24357d7

The code is now the short form: `Experience`, `Finding`, `Failure` and the field list are gone, rows are stored as written, and `_experience_row` keeps only the `name` <- `description` fallback plus the `extra_envs` / `gain_pct` type checks. Verified: unknown keys (`fingerprint`, `error_class`) round-trip, `0` / `-1.0` / `None` gains survive, bad types raise and leave the live row unchanged, and blank legacy rows stay nameless. Nothing left in the code.

Blocking issues: 1

1. [X2] The PR description still describes the previous head. [verified]
   Problem: `body` still says "`Finding` and `Failure` now share an `Experience` base ... `from_mapping` / `to_dict`", "Rows written under the old projection kept the variant in `description`", and "Recipes written before this change still load; their variant name is recovered from the old `description` key". None of this is true at `e24357d7`. The release notes are built from this text.
   Action: Update the description to match the code: rows are stored as written with no field list; the name is recovered only from a row that stored a non-empty `description`; Coordinator rows already on disk stay nameless; a present non-numeric `gain_pct` (including `"4.2"`) or non-mapping `extra_envs` raises at `put_recipe`; refresh the test count.

Ready to merge once the description is updated.

Checked: schema.py `_experience_row` / `Recipe.from_dict` / `to_dict`, local_store.py `put_recipe`, `gain_pct` producers (`gain_math.gain_pct`, explore outcomes) | Ran: round trip with unknown keys, blank legacy row, bad types; pytest 213 passed (sedimentation, local_store, local_kb_requirements, warm_replay, coordinator_kb_writes, recipe_kb_t0_anchor) | CI: 31 success, 2 skipped at head | Base: 09d1ff5649f187b6c9d0322bf8e6441fde5ff0cb | Head: e24357d7b670ad2dba84d03414bf37a743ea68d1

rpoornac:
PR description rewritten against `e24357d7`. The substantive corrections, rather than just wording:

- **"The fix"** described the `Experience` base and its `from_mapping` / `to_dict` pair. It now describes what the code does — rows stored as written, `Recipe.from_dict` / `to_dict` carrying them whole, no field list on the path — and names the deleted cut explicitly so the section is not read as the current design.
- **The observable effect was wrong in the operator's favour.** It claimed recipes written before this change "still load; their variant name is recovered from the old `description` key". As you pointed out, the old projection stored `description: ""` for Coordinator rows, so their name was destroyed at write time and nothing recovers it. Since this description is what the release cut publishes, that would have promised operators a repair that does not happen. It now says rows already on disk are not repaired, that those rows stay nameless, and that the benefit starts from the next session that writes them — with the arbor-sourced rows called out as the case that does read back.
- **The boundary section** now states that a stringified number raises too, which was not true of the cut the old text described.
- **Tests and verification** are listed by name against what is actually in the file, and the counts are the current ones (`1614 passed`, production diff +44/−41).

ZhengGong-amd (approved):
## PR #1561 -- re-review at d04c4e28

What it does: `put_recipe` stops projecting `what_worked` / `what_failed` onto `{description, measured_impact}`. Rows are stored as the Coordinator writes them, so a later session's `already_proven` sees earlier variants. `_experience_row` keeps only the `name` <- non-empty `description` fallback and rejects a present non-mapping `extra_envs` or non-numeric `gain_pct`. Rows already on disk are not repaired, and the description now says so.

Blocking issues: none

The description now matches the code. The PR patch is byte-identical to `e24357d7`; this head is a rebase.

Checked: schema.py `_experience_row` / `Recipe.from_dict` / `to_dict`, local_store.py `put_recipe`, the five tests the description names, `test_item9` row equality, no `Finding` / `Failure` / `Experience` left in recipe_kb | Ran: pytest 219 passed (sedimentation, local_store, local_kb_requirements, warm_replay, coordinator_kb_writes, recipe_kb_t0_anchor); ruff clean on touched files | CI: 30 success, 2 skipped, `ruff` red on `src/kernelforge/gemm_tune/tier3/dispatch.py:659` (RUF100 from #1609, fixed on main by 152d10080, not in this diff; rebase to clear it) | Base: 6ecdff3bb8fe2ca616e2559ac45b268a6f6ce190 | Head: d04c4e2834cf30b6dcb6cec67cfac04b148be474

ZhengGong-amd (approved):
Apporved

PR #1563, opened 2026-09-18
fix(warm-replay): replay a KB recipe into a pip-installed framework

BaoYunkai:
## Problem

A Recipe whose patch timeline is non-empty replays as *required*, and two guards
on that path refuse a tree a pip-installed framework always presents.

1. `_apply_warm_patches` refused any required timeline whose tree had no git
   HEAD. A framework installed from a wheel never has one, so every code-level
   optimization a session published became unreplayable the moment it reached
   the KB: warm replay failed with `missing_git_head` before booting a server.

2. An overlay the tree already carries is applied as a satisfied no-op, writing
   nothing and recording neither a snapshot manifest nor nogit backups.
   Promotion and rollback both read those absent artifacts as "cannot unwind
   this tree", so a replay that had already reproduced its gain was rejected and
   then failed with `stop_reason=warm_replay_rollback_failed`.
   `_rollback_combined_warm` also only ever restored from a git snapshot
   manifest, so a nogit tree it *had* written to could not be unwound either.

## Fix

What promotion needs is a way to unwind the tree if the replay is rejected, not
a sha. A git checkout answers with a pre_sha and a snapshot manifest; a
pip-installed framework answers with the backups nogit wrote as it applied. The
`missing_git_head` refusal is removed and promotion accepts either channel.

Absent artifacts cannot carry "nothing was written" on their own -- a no-op
apply and an apply whose artifacts were lost record the same nothing, and only
the first is safe to leave standing. Each tree now states whether the round
wrote to it. A no-op tree promotes (promotion is reached only once every
required overlay applied, so it already holds what is being promoted) and is
skipped by the rollback; a tree that was written to still has to answer with a
channel. A record predating the flag is read as written-to, so a resume that
lost its manifest fails exactly as before.

`_rollback_combined_warm` now routes through `_revert_warm_patch_state`, which
picks the channel the apply recorded.

## Test plan

- [x] `test_warm_patch_apply.py`, `test_warm_replay.py` (101 passed)
- [x] `orchestrator/actions/executors/tests`, `orchestrator/tests`,
      `test_coordinator_async_batch2_unit.py` (690 passed, 1 skipped; 2 failures
      also present on unmodified main)
- [x] Pod: MiniMax-M3-MXFP4 vLLM session reproduced +23.88% from the KB recipe
      and closed normally (`stop_reason=sweep_done`, `crash_count=0`)


xiaofei-zheng:
Reviewed against `git merge-base origin/main` (2cc382a83). Three blocking issues.

**1. Crash-recovery rollback now skips a tree it may have patched — `phases/prelude.py:1676`**

`_persist_recipe_snapshot` writes `pending["recipe_patch_trees"]` *before* any patch is applied (`executors/baseline.py:1243` → `:2804`), and those records come from `_warm_tree_records`, which since this PR carries `"mutated": False` (`baseline.py:1154`). `mutated` only flips inside the apply loop (`baseline.py:1452`), and the flipped records reach pending only after `_apply_warm_patches` returns (`baseline.py:2905`).

So for a git tree on a required timeline, a session that dies between the first write and that return leaves pending holding `mutated=False`. Resume calls `_rollback_combined_warm({}, None)` (`loop/writeback.py:5673`), which falls back to `pending["recipe_patch_trees"]`, hits the new `continue`, restores nothing, and returns `{"ok": True}` — the patched checkout is silently kept, and the snapshot manifest that would have unwound it is ignored. Before this PR that path called `_revert_patches` with the manifest and restored the tree.

The pre-mutation record is exactly the record that cannot know whether the round wrote to the tree; it must read as written-to. Either omit `mutated` from the record `_persist_recipe_snapshot` writes, or stamp it `True` there.

**2. `applied_nogit` is not accepted as a replayed overlay — `phases/prelude.py:2343-2348`**

Under `required_patch_timeline`, `replayed_patch_refs` only counts `applied` / `applied_3way` / `present_in_dirty_worktree`. On main `applied_nogit` was unreachable on this path (the `missing_git_head` refusal), so the set was complete; removing that refusal makes it the *only* status a pip-installed framework produces, and it is filtered out.

Consequences on the exact tree this PR is for:
- `replayed_patch_refs` is empty, so `_adopt_replayed_prior` (`knowledge/remote_recipe/values.py:775`) returns early and the reproduced overlay is dropped from the published stack.
- For a recipe whose only content is the patch timeline (no `extra_server_args`/`extra_envs`/kernel plan), `prelude.py:2360` then reports `reproduced_but_no_params` and pushes nothing — the replay reproduces the gain and is still discarded.

`applied_nogit` needs to join that set.

**3. No CHANGELOG entry**

`CHANGELOG.md` `## [Unreleased] / ### Fixed` is maintained per behavior fix in this repo, and this changes replay behavior for every non-git framework tree. Please add an entry in the existing format.

Checked and found correct: `method = "applied_nogit" if backups else "already_present"` — `_apply_patch_no_git` returns success with an empty backup list only on the clean-reverse-dry-run no-op (`_nogit_patch.py:234-239`), so an empty set does mean "already carried". `_resolve_promoted_recipe_checkout`'s `manifest_target = target_path` when not snapshotted correctly neutralizes the mismatch check for a nogit tree, and `tree.get("mutated", True)` keeps pre-flag records failing closed. Tests were not run locally (POSIX-only deps on this box); CI covers them.

xiaofei-zheng:
Re-reviewed at a0ed3c65d (base 484559ea2). The three items from the previous round are fixed and verified:

- `_warm_tree_records(..., before_mutation=True)` stamps the pre-apply record written-to, so the resume path in `_rollback_combined_warm` no longer skips a tree the round may have patched (`baseline.py:1234-1238`).
- `applied_nogit` now counts in `replayed_patch_refs` (`prelude.py:2347`).
- CHANGELOG entry added.

The merge of main is clean — the diff against the new base is still only these five files.

One blocking issue remains, the other half of the first one.

**A crash mid-apply on a pure nogit tree leaves no record to recover from — `baseline.py:1233`**

`_persist_recipe_snapshot` is only reached inside `if any(tree["snapshot_manifest"] for tree in trees.values())`. A pip-installed framework produces no snapshot manifest for any tree, so on the very configuration this PR exists for, the callback never fires and nothing about the trees reaches `warm_replay_pending` until `_apply_warm_patches` returns (`baseline.py:2817`). Same for `params["_warm_patch_trees"]`, gated identically at `:1462`.

A session killed between the first `_apply_patch_no_git` write and that return therefore resumes with pending holding no `recipe_patch_trees`. `_rollback_combined_warm({}, None)` (`loop/writeback.py:5673`) falls back to the single-tree dict, whose `root` is `None`, hits `if not target: continue`, and returns `{"ok": True}` — so `_resume_recover_pending_warm_replay` reports the set rolled back and clears pending while the installed framework is still patched, and every measurement after that runs on a dirty tree. Before this PR the same window did not exist: `required_timeline and not pre_sha` was refused before any tree was written to.

The window is real — one `patch` CLI invocation per overlay, with a 60s timeout each — and the resume path exists precisely to handle being killed there.

The pieces to close it are already in the repo: `_apply_patch_no_git` writes a backup ledger under `backup_root` as it goes, and `_revert_patches_no_git(backups, backup_root=...)` merges that ledger in exactly so "a mutation whose record never reached the caller is still reverted" (`_nogit_patch.py:513-516`) — `integrate_patch.py:4235` already passes it. Warm replay never persists `backup_root`, and `_revert_warm_patch_state` neither accepts one nor reaches the nogit channel at all when `nogit_backups` is empty, which is exactly the crash case. Persisting the tree records (with the backup root) before the first write, rather than only when a snapshot manifest exists, would make the resume able to unwind the tree.

Everything else on the new commits reads correct. `params["_warm_patch_trees"]` being stamped `mutated=True` at `:1234` is harmless — its only consumer, `_revert_legacy_warm_patch_trees` → `_revert_warm_patch_trees`, dispatches on backups/manifest and never reads the flag — and `:1462` overwrites it with the real post-apply records. Tests not run locally (POSIX-only imports on this machine); CI covers them.

xiaofei-zheng (approved):
Retracting the remaining item from my last comment — it is not introduced by this PR.

I checked the gating against the merge base instead of only against the previous commit: `if any(tree["snapshot_manifest"] ...)` around `_persist_recipe_snapshot` is identical at 484559ea2 (`baseline.py:1233`). And a combined warm replay already reaches `_apply_warm_patches` with a non-empty `patches` list while `required_patch_timeline` is false — `prelude.py:1326-1330` only sets the flag for a current remote timeline or a recipe that asks for it, and `warm_replay_pending` is written unconditionally at enqueue. So on the base, a non-required replay of a nogit tree already had the same window: killed mid-apply, resume finds no `recipe_patch_trees`, `_rollback_combined_warm` falls through `if not target: continue` and reports ok with the tree still patched.

This PR widens which replays reach that path; it does not create it. Out of scope here.

What I checked:
- The two regressions from the first round, against the base: `_warm_tree_records(..., before_mutation=True)` restores the pre-apply record's written-to meaning, so the resume rollback no longer skips a tree the round may have patched; `applied_nogit` now counts in `replayed_patch_refs`, so a timeline-only recipe on a pip-installed framework is no longer dropped as `reproduced_but_no_params` and its overlay reaches `_adopt_replayed_prior`.
- `method = "applied_nogit" if backups else "already_present"`: `_apply_patch_no_git` returns success with an empty backup list only via the clean reverse dry-run no-op (`_nogit_patch.py:234-239`), so the inference holds.
- `_resolve_promoted_recipe_checkout`: `manifest_target = target_path` when not snapshotted correctly neutralizes the mismatch check for a nogit tree, and `tree.get("mutated", True)` keeps pre-flag records failing closed.
- `params["_warm_patch_trees"]` stamped `mutated=True` at `:1234` is harmless — its only consumer dispatches on backups/manifest — and `:1462` overwrites it with the real post-apply records.
- The merge of main is clean: the diff against 484559ea2 is still only the five files.
- CHANGELOG entry present and in the existing format.

Tests were not run locally (POSIX-only imports on this machine); CI covers the suites listed in the description.

PR #1565, opened 2026-09-18
fix(profile): keep the profiler bounds out of a preflight rejection, and let the launcher own the trace dir

BaoYunkai:
## Problem

The argv preflight probe sees only `EXTRA_VLLM_ARGS`. Magpie's launcher sets
`--profiler-config.profiler torch` and a `torch_profiler_dir` of its own, but
only in the flags it appends *after* `EXTRA_VLLM_ARGS` at actual launch.

vLLM's `ProfilerConfig` validator refuses `capture_torch_profiler` without
`profiler == 'torch'`, and refuses `profiler == 'torch'` without a
`torch_profiler_dir`. So the probe checked this layer's
`delay_iterations` / `max_iterations` / `capture_torch_profiler` against a
config that had neither, and rejected an argv that is valid once the launcher's
flags are appended -- `VALUE_REJECTED`, terminal, no repair. The profile round
then ran with no bound on the capture window.

## Fix

Assert both flags in this same injection, so the fragment Hyperloom controls is
self-consistent on its own. An operator-set profiler flag is left untouched.

The trace directory is a placeholder and depends on losing vLLM's last-wins
dotted-flag merge to whatever the launcher computes. Magpie appends its flags
after `EXTRA_VLLM_ARGS`, so it wins there; `bypass_engine` emitted its pair
*before* `extra_args`, so the placeholder would have won instead and sent the
trace under the round directory. Trace discovery probes `workspace/torch_trace`,
`workspace/capture_traces` and `workspace.parent/capture_traces`, none of which
is where the steady-state trace would then be, so the round would have reported
no trace files. The bypass backend now orders its profiler flags the way Magpie
already does.

## Test plan

- [x] The backend's own trace dir is the last `torch_profiler_dir` on the argv
      (new; fails against the previous ordering)
- [x] `test_bypass_backend.py`, `test_bypass_engine.py` (88 passed)
- [x] `test_profile_and_kernel_handlers.py` (1 failure,
      `test_trace_certificate_stays_out_of_the_resolver_namespace`, also fails
      on unmodified main with `ModuleNotFoundError`)
- [x] `ruff check .`


xiaofei-zheng:
## Blocking

**1. The injected `torch_profiler_dir` overrides the bypass backend's trace dir, so single-node bypass profiles lose their traces.**

`bypass_engine.build_server_command` emits its own `--profiler-config.torch_profiler_dir <profile_dir>` and then appends `EXTRA_VLLM_ARGS` *last* (`bypass_engine.py:88-96` + `return cmd + list(extra_args)` at line 96), the opposite of the Magpie ordering this change assumes. With the new injection, the last `torch_profiler_dir` on the argv is `materialize_config_with_envs`'s `output_dir`, not `workspace/torch_trace`.

Reproduced against this branch (materialize a vLLM profile YAML, tokenize as `bypass_runner._tokenize_extra_args`, build the server cmd):

```
--profiler-config.profiler torch --profiler-config.torch_profiler_dir <ws>/torch_trace \
--profiler-config.delay_iterations 6080 --profiler-config.max_iterations 128 \
--profiler-config.profiler torch --profiler-config.torch_profiler_dir <output_dir> \
--profiler-config.ignore_frontend True
# last wins -> <output_dir>
```

`<output_dir>` is the round dir; the bypass workspace is `<output_dir>/benchmark_vllm_<ts>` (`bypass_report.create_workspace`). Trace discovery only probes `workspace/torch_trace`, `workspace/capture_traces`, `workspace.parent/capture_traces` (`profile.py:694-700`), so the traces land somewhere none of the candidates cover and the round reports no trace files. On the merge base `EXTRA_VLLM_ARGS` carries no `torch_profiler_dir`, so the bypass path is correct today — this is introduced by the PR. `HYPERLOOM_BENCHMARK_BACKEND=bypass` is a supported configuration (`benchmark_backend.py:15`).

Asserting `--profiler-config.profiler torch` alone is enough to satisfy the preflight validator, if the validator's requirement is the `profiler` value. If a trace dir is genuinely required by the validator too, it needs to be a value the launcher actually overrides in both backends, not the materialization dir.

**2. Title and description do not match the diff.**

Both say the change asserts `--profiler-config.profiler torch`; the diff also injects `--profiler-config.torch_profiler_dir <output_dir>`, which is the part that changes where traces are written. Please update the title/description (and the commit body) to cover it.

## Checked

- `_profiler_flag_value` / `_PROFILER_FLAG_RE` exact-name matching: `profiler` does not match `torch_profiler_dir`, so the two guards are independent — correct.
- Restore path (`_workload_envs.py:2042-2068`): the two new sentinels flow through `pending_vllm_profiler_flags`, and `_profiler_bound_holds` treats presence as the contract for both — consistent.
- Multi-node vLLM (`launch_multinode._build_vllm_cmd`) also appends extra args last, but sets no profiler dir of its own, so no override there.
- `test_workload_envs_golden_lock.py` pins no profiler flags; not affected.

xiaofei-zheng (approved):
Re-reviewed at 775717e3c. Both earlier blocking points are resolved; no blocking issues remain — LGTM.

- Trace-dir ordering: re-ran the end-to-end repro on this head (materialize a vLLM profile YAML → `bypass_runner._tokenize_extra_args` → `bypass_engine.build_server_command`). The argv now ends with the backend's own `--profiler-config.torch_profiler_dir <workspace>/torch_trace`, so vLLM's last-wins merge resolves to the dir trace discovery probes; the `EXTRA_VLLM_ARGS` placeholder (`<round-dir>`) is overridden. `profile_dir=None` takes the early return and is unchanged from before.
- No other framework is affected: `EXTRA_SGLANG_ARGS` gets only `--watchdog-timeout`, `EXTRA_ATOM_ARGS` is unset, and the atom branch injects no `--profiler-config.*`.
- Multi-node vLLM (`launch_multinode._build_vllm_cmd`) appends extra args last but sets no profiler dir of its own, so nothing is overridden there.
- Flag guards: `_PROFILER_FLAG_RE` matches on the exact dotted name, so the `profiler` / `torch_profiler_dir` guards stay independent and an operator-set value (`--flag v` or `--flag=v`) suppresses the injection; both new sentinels flow through `pending_vllm_profiler_flags` into the end-of-function re-assertion consistently.
- Title/description now match the diff, and the CHANGELOG entry covers both halves of the change.
- Tests: `test_bypass_backend.py` + `test_bypass_engine.py` — the profiler-ordering and server-command cases pass; the remaining failures in that file are Windows-only (`os.getpgid`, path separators) and reproduce on the merge base. `test_profile_and_kernel_handlers.py` cannot be collected on Windows (`ModuleNotFoundError: fcntl`), so its new cases were verified through the repro above instead; CI covers the rest.

PR #1567, opened 2026-09-18
fix(kernel-patch): stop SGLang editable reinstall for in-tree JIT sources

BaoYunkai:
## Problem

Patching a native source under SGLang's in-tree JIT tree fell through to the generic `/sgl-workspace/sglang/` strategy and ran `pip install -e python`.

The KDA patch hit this: it edits `python/sglang/kernels/jit/csrc/attention/kda_packed_decode.cuh` alongside a Python wrapper, and the `.cuh` suffix classified the target as compiled. Other kernels in the same round edited only `.py` files, which the Python branch forces back to source-only, so KDA was the one target that reinstalled the framework.

Two consequences:

1. The reinstall replaces the distribution metadata and the editable import hook mid-session. `pip` returned 0, but the candidate server then died on `ModuleNotFoundError: No module named 'sglang.srt.server_args'` before model load, so the patch was reverted for a re-baseline failure that had nothing to do with the kernel.
2. It also rewrites the dynamic version and import layout between baseline and candidate (`0.5.17` to `0.5.18.dev1+gc326cc496`), so the measured comparison is no longer limited to the kernel patch.

The reinstall was never the thing that made the patch take effect. These sources are compiled by tvm-ffi at import time: `load_jit` derives the module name from `_local_jit_source_hash()` over the source and its include closure, and caches the result under `$TVM_FFI_CACHE_DIR/<module_name>__arch_*__tvmffi_*`. A patched source therefore builds under a new cache key on its own, and an editable reinstall neither compiles it nor is needed to pick it up.

Separately, apply treated a rebuild returncode of 0 as proof of success for every framework, so an install left in this state was only discovered after the re-baseline server had already spent a start-up on it.

## Fix

1. Classify native sources under `sglang/kernels/jit/` (editable checkout and installed wheel) under a new `content_addressed_jit` rebuild mode: no rebuild command, no artifact roots, compilation deferred to the runtime, installation untouched. `sgl-kernel/`, `kernels/aot/csrc/`, vLLM and AITER strategies keep the rebuild they had.

2. Key the multi-file rebuild dedup in `_multi_root_strategies` on the root *and* the rebuild mode. `/sgl-workspace/sglang` now yields two modes -- `kernels/jit` defers, `kernels/aot/csrc` still needs the editable install -- so deduplicating on the root alone would keep whichever file the patch listed first. With the JIT source first the AOT `.cu` would never be compiled while apply still reported `ok` with a `deferred` rebuild, benchmarking the candidate against an unpatched AOT kernel. A root that yields one strategy across several files still rebuilds once.

3. Strategies whose rebuild reinstalls a framework now declare the modules the server loads first (`sglang` + `sglang.srt.server_args`, `sgl_kernel`, `vllm`, `aiter`). After the rebuild command succeeds, apply resolves them in a fresh interpreter -- the one that ran the rebuild -- and fails the rebuild when any is unresolvable, which reverts the patch through the existing path. The probe records each module's `origin` and the probed `sys.path` in the manifest.

`test_apply_kernel_patch_sglang_jit.py` covers all three. The JIT strategy tests fail before fix 1 with `unexpected rebuild subprocess: ['/opt/venv/bin/python', '-m', 'pip', 'install', '-e', 'python']`; the mixed jit-`.cuh` + aot-`.cu` cases fail before fix 2 in both file orders, keeping only the `command` strategy.


xiaofei-zheng (commented):
## Blocking

**`_multi_root_strategies` dedups by root, but the SGLang rebuild mode is now per-file** — `src/hyperloom/agents/kernel/tools/apply_kernel_patch.py:2177`

`key = strat["root"] or str(p.parent)`. Both `python/sglang/kernels/jit/...` and `python/sglang/kernels/aot/csrc/...` resolve to root `/sgl-workspace/sglang`, so only the first strategy in descriptor order survives. Before this PR both produced `rebuild_mode == command` with the same command, so the dedup was harmless; now the two differ and the surviving one depends on file order.

Reproduced against this branch vs. `merge-base origin/main`:

```
# this branch
['kda.cuh', 'dsv4.cu'] -> [('content_addressed_jit', [])]
['dsv4.cu', 'kda.cuh'] -> [('command', ['/opt/venv/bin/python', '-m', 'pip', 'install', '-e', 'python'])]

# origin/main
['kda.cuh', 'dsv4.cu'] -> [('command', [... 'pip', 'install', '-e', 'python'])]
['dsv4.cu', 'kda.cuh'] -> [('command', [... 'pip', 'install', '-e', 'python'])]
```

(paths: `/sgl-workspace/sglang/python/sglang/kernels/jit/csrc/attention/kda.cuh` and `/sgl-workspace/sglang/python/sglang/kernels/aot/csrc/elementwise/dsv4_norm_rope.cu`)

With the JIT file first, the AOT `.cu` is never compiled or installed, apply still reports `status: ok` with a `deferred` rebuild, and the candidate is benchmarked against the unpatched AOT kernel — a silent wrong-result path, not a failure. `python/sglang/kernels/aot/csrc/` holds 204 compiled sources under that same root, so the mixed patch set is reachable, and it contradicts the PR description's "`kernels/aot/csrc/` ... strategies are unchanged".

`test_kda_patch_set_drives_no_editable_reinstall` does not cover this: its second file is a `.py` wrapper, which `_detect_strategy` forces to `compiled=False` and `_multi_root_strategies` drops before the dedup. A jit-`.cuh` + aot-`.cu` case in both orders is the test that fails today.

Suggested fix: key the dedup on `(root, rebuild_mode)` — or on the rebuild command — so a root that yields two distinct strategies keeps both.

## Checked and clear

- `_detect_strategy` branch ordering: `sgl-kernel/`, vLLM, AITER editable/installed, FlyDSL and the Python-source override all unchanged by the marker test; verified by running each branch.
- Import-probe wiring: `_run_strategy_rebuild` only probes after a returncode-0 command rebuild, an `ok` rebuild with a failed probe flips to `failed`, and `_rebuild_ok_to_proceed` then reverts. Matches the tests.
- `content_addressed_jit` deferral reaches `_rebuild_ok_to_proceed` as `True` and carries no `jit_build_dir`, so the AITER JIT-invalidation gate stays out of the way.
- Tests: `src/hyperloom/agents/kernel/tests/test_apply_kernel_patch_sglang_jit.py` (run locally; the failures there are Windows path-separator artifacts, not branch failures — CI runs Linux).

xiaofei-zheng (commented):
Re-reviewed at `8177b6316`. The dedup fix is correct — verified against the new head:

```
[kda.cuh, dsv4_norm_rope.cu] -> [content_addressed_jit(no cmd), command(pip install -e python)]
[dsv4_norm_rope.cu, kda.cuh] -> [command(pip install -e python), content_addressed_jit(no cmd)]
two aot .cu under one root -> [command(pip install -e python)]   # still collapses
jit + aot + sgl-kernel/csrc -> 3 strategies, one pip per distinct (root, mode)
```

Also checked the rebuild loop at `apply_kernel_patch.py:2859`: the extra strategy costs one extra `_run_strategy_rebuild` call that returns `deferred` without a subprocess, so no duplicate pip run.

## Blocking

**The PR description does not match the current diff.** It still describes two changes and closes with "`test_apply_kernel_patch_sglang_jit.py` covers both", while the branch now carries a third commit that changes `_multi_root_strategies`' dedup key from the root to `(root, rebuild_mode)` and adds `test_jit_and_aot_sources_keep_separate_strategies` / `test_same_root_and_mode_still_rebuilds_once`. That change is also what makes the description's existing claim — "`kernels/aot/csrc/` ... strategies are unchanged" — true for a mixed patch set; as written, a reader cannot tell the per-file mode split required a dedup change to be safe.

xiaofei-zheng (approved):
Approving at `8177b6316`. No blocking issues in the code.

What I checked:

- `_multi_root_strategies` dedup: ran both orders of a `kernels/jit/*.cuh` + `kernels/aot/csrc/*.cu` patch set against this head and against `merge-base origin/main` — the order dependence is gone and the AOT `pip install -e python` survives either way. Two AOT sources under one root still collapse to a single rebuild; a jit + aot + `sgl-kernel/csrc` set yields three strategies, one per distinct `(root, mode)`.
- Rebuild loop at `apply_kernel_patch.py:2859`: the extra strategy is a `deferred` record with no subprocess, so no duplicate pip run.
- `_detect_strategy` branches: `sgl-kernel/`, vLLM, AITER editable/installed, FlyDSL and the Python-source override are all unaffected by the new `sglang/kernels/jit/` marker.
- Import probe wiring: probes run only after a returncode-0 command rebuild, a failed probe flips the rebuild to `failed`, and `_rebuild_ok_to_proceed` then drives the existing revert.
- `content_addressed_jit` deferral reaches `_rebuild_ok_to_proceed` as `True` and carries no `jit_build_dir`, so the AITER JIT-invalidation gate stays out of the way.
- Tests: `src/hyperloom/agents/kernel/tests/test_apply_kernel_patch_sglang_jit.py` (local failures there are Windows path-separator artifacts; CI runs Linux).

Please refresh the PR description for the third commit before merge.

PR #1568, opened 2026-09-18
fix(agentx): unblock baseline admission using profiling error rates

zengleixin-amd:
## Summary
- Derive missing AIPerf request error rates from consistent profiling counters. Infer zero only with positive successful requests and an explicitly empty error summary; unknown or invalid evidence remains ungradable.
- Keep the packaged mapper and standalone fallback consistent. Never mix warmup errors into profiling error rates.
- Run AgentX request-error quality parsing even when serving lm-eval is disabled. Preserve ordinary serving and scriptable behavior, thresholds, and admission gates.

## Validation
- Added failing regressions before implementation for missing zero-error metrics and the baseline parsing skip.
- Local focused tests: 189 passed, 7 deselected. Mapper and accuracy-gate modules independently passed (106 and 66 tests).
- Six new real baseline-path tests now pass quality-result assertions locally, but their final Coordinator admission assertions require Linux because Windows lacks fcntl. One existing wiring test also depends on POSIX paths. Full validation is pending Linux CI; no test skips were added.
- Ruff lint/format checks on changed Python files and git diff --check passed.

## Scope
- Request error rate is the AgentX quality proxy, not semantic model accuracy; this does not introduce lm-eval or relax quality checks.
- Does not repair old session artifacts, alter cluster resources, or change other optimization action paths.


xiaofei-zheng (approved):
No blocking issues. LGTM.

**What this PR does**

When AIPerf omits `request_error_rate`, `map_aiperf` now derives it from the profiling request counters instead of emitting `None`, and the baseline executor keeps running the AgentX request-error quality parse when serving lm-eval is disabled (`RUN_EVAL=false`), so a valid zero-error AgentX baseline is no longer blocked from anchoring for lack of a quality signal. Unknown or invalid evidence still resolves to `None` and fails closed.

**What I checked**

- Derivation vs. upstream semantics. Downloaded `aiperf==0.12.0` and read the metric definitions. `RequestCountMetric` = valid (successful) requests, `ErrorRequestCountMetric` = errored requests, `CompletedRequestCountMetric._derive_value` = successes + errors, and `RequestErrorRateMetric._derive_value` = `100.0 * errors / (successes + errors)`. `_request_error_rate` reproduces that formula exactly, including the `completed != success + errors` consistency rejection, so the derived value cannot disagree with the metric it substitutes for.
- Warmup isolation. `MetricsJsonExporter` builds the top-level `error_summary` from `self._results.error_summary` (profiling phase; `ErrorTracker.get_error_summary_for_phase` returns `[]` with no errors), while warmup travels separately under `warmup_metrics`. The `error_summary == []` precondition for inferring zero errors therefore reads profiling-phase evidence only, as the comment claims.
- Fail-closed paths. Non-numeric, bool, negative, NaN/inf, out-of-range (>100) and inconsistent counter combinations all return `None`; `parse_agentx_error_rate` -> `parse_eval_results` then yields `accuracy=0.0`, keeping the result unpromotable via `Coordinator._is_promotable_result` rather than admitting an ungraded baseline.
- Package/standalone parity. The new `error_rate_mapper` fixture exercises both `agentx/mapping.py` and the `assets/agentx/map_aiperf.py` fallback over the same 23 cases; the fallback's `_stat` None-coalescing is now aligned with the packaged `stat`, which the derivation depends on for `{"avg": None}` sub-keys.
- `shared_state` resolution in `baseline.py:4117`. `ctx.extra["shared_state"] or self.shared_state` matches the pattern used elsewhere in the file, and both spellings are covered by the new `state_in_context` parametrization. Only one `run_eval_disabled` quality gate exists in the executor, so there is no second call site left unmirrored.
- Compared against `git merge-base origin/main` to confirm the `request_error_rate` coalescing and gate behavior changes are introduced by this PR and not pre-existing.

**Tests run** (Windows, focused scope only)

- `test_map_aiperf.py` + `test_accuracy_gate_units.py`: 172 passed.
- `test_eval_result_dir_wiring.py`: 18 passed, 6 failed. All six failures are the new `test_agentx_baseline_grades_requests_with_run_eval_off` cases, and they fail only at the trailing `from ...coordinator import Coordinator` (POSIX-only `fcntl`/`signal.SIGHUP`). With those modules stubbed, every quality assertion ahead of the import — `status`, `run_eval_disabled`, `accuracy`, `accuracy_task`, `baseline_eval_failed` — passes, so the behavior under test is verified; Linux CI covers the admission assertions.
- `ruff check` and `ruff format --check` on the changed Python files: clean.

Changelog entry is present under Unreleased/Fixed, and the title and description match the diff.

PR #1570, opened 2026-09-18
fix(agentx): let an agentic recipe satisfy the server-phase pid contract

luochen-amd:
`aiperf_client.sh` delegates the server boot to `$AGENTX_SERVER_SCRIPT` with
`MAGPIE_RUN_PHASE=server`, expects a pid file back, and refuses to continue
without one so it never benchmarks a server it cannot tear down.

A flat framework script honours that contract. An agentic recipe does not: it
ignores `MAGPIE_RUN_PHASE` and owns the entire lifecycle - boot, replay,
aggregate, teardown - so it returns with the result already written and no
server left to track. The pid file is therefore never created, and the run
died on

    [aiperf_client] ERROR: builtin server phase wrote no pid to ...; refusing
    to run (would risk a GPU leak)

with `exit 3`, discarding a replay that had in fact completed with
`replay_rc=0` and a full `inferencex_result.json`.

Recognise that case and return success instead of demanding a pid: when the
builtin is an agentic recipe and the result file already exists, the work is
done and continuing would drive a second replay against a server that is
already gone.

One more step is needed before returning. The agentic writer emits only the
nested InferenceX leaderboard schema, while Magpie reads flat top-level keys,
so returning as-is makes a healthy run read as 0.00 req/s downstream. Derive
the flat keys with the same `map_aiperf.py` the non-agentic path uses and union
them in, with existing (nested) keys always winning so the leaderboard schema
is untouched; the union is asserted before the file is replaced atomically. If
the export cannot be found or mapped, the run is left alone with a warning
rather than half-rewritten.

Credit where due: the flat-metrics half of this hunk was authored by
Hyperloom's own enablement specialist during a live session (it diagnosed the
nested-vs-flat mismatch unaided and patched the deployed copy); it is carried
here because the deployed copy is overwritten from these assets on every run.


xiaofei-zheng (approved):
No blocking issues. LGTM.

Paths I checked:

- **Branch selectivity.** `*/agentic/*` cannot match a flat builtin name (`vllm_mi300x.sh`, `sglang_mi300x.sh`), so the existing pid contract is untouched for every non-agentic recipe. `test_aiperf_client_sh.py:429-431` (`write_pid=False` asserting `returncode == 3`) still describes the unchanged path.
- **Shell + merge semantics.** Ran a minimal local repro of the heredoc hunk: the `if cmd \ && python3 - a b <<'PYMERGE' ... PYMERGE / then` form parses, and the merged file carries the mapped flat keys alongside the agentic nested ones with agentic winning on conflict, as described.
- **Teardown.** `exit 0` still runs the already-installed `trap cleanup EXIT`, so the `fuser -k ${PORT}/tcp` port fallback is unaffected.
- **Mapping failure is not swallowed.** On a missing export or a failed map, no flat keys are injected, so `is_valid_measurement` (benchmark_result.py:770-780) rejects the measurement on the absent `output_throughput` rather than recording a 0.00 success. Fail-closed.
- **PROFILE=1.** Returning before the capture window means no capture-status file; profile.py:1296-1303 records `capture_status_missing` and logs an error, so a profiling round cannot pass silently without a trace.
- **CHANGELOG.** Prior changes to this asset directory (9a3aa6f1e, f51ca4d9c, cb69b8d6a, ac0be3672) carry no entry and the repo does not require one, so its absence is not raised here.

PR #1571, opened 2026-09-18
fix(slurm): pass the workload shape to the optimizer CLI as flags

luochen-amd:
## The bug

The optimizer resolves the workload shape as **flag > persisted state > default** and deliberately never reads it from the environment. `_incontainer.sh.in` only ever exported it. A `models.tsv` row is therefore a suggestion the optimizer never sees:

| declared in the row | what actually got benchmarked |
|---|---|
| `tp=4` | `tp=1` — and sglang's rank math then divides by zero |
| `conc=12` | the CLI default |
| `isl=8192 osl=512` | `1024/1024` |
| `prec=fp8` | whatever the checkpoint sniffer returned |
| `ep` | nothing: the column did not exist |

All of it silent. `_resolve_workload_knobs` fills `args.isl/osl/precision` from flag-or-default and then *overwrites* the exports at `cli/__init__.py:1956-1962`, so even the env the recipes later read no longer matches the row.

## The fix

**Pass the whole declared shape as flags**, in both the python and the claude backend: `--tp`, `--conc`, `--ep`, `--isl`, `--osl`, `--precision`. They all already exist on the `optimize` subparser.

**Keep the exports, and say why they are still there.** The framework recipes downstream *do* read `TP` / `EP` / `CONC` / `ISL` / `OSL` / `PRECISION` from the environment — that half is load-bearing. The prompt text used to tell the carrier not to omit "the env exports", which pointed it at the mechanism the optimizer ignores; it now states which half feeds the optimizer and which half feeds the recipes.

**Give `EP_SIZE` a producer.** It had none anywhere in the repo, so `--ep` was pinned to 1 no matter what the operator wrote, and an operator-set value would have been dropped by the docker backend's `-e` allowlist while enroot's `--export=ALL` kept it — the two runtimes silently disagreeing on the shape. It is now a real column:

```
models.tsv (new trailing `ep_size` column)
  -> run_hyperloom.sbatch `read` + `${EP_SIZE:-1}` default
  -> export
  -> docker `-e` allowlist
  -> `--ep` on the CLI
```

It trails `target_gain` so the existing 13-column rows still parse, and an absent value keeps expert parallelism at 1. The variable keeps the name `EP_SIZE` rather than `EP` because that is the name the InferenceX agentic recipes validate with `check_env_vars`.

## Evidence

Found while a `tp=4` row kept materialising `tp=1` baselines on MI355X. After the change the container log shows the shape arriving intact (`[hl] session=... tp=4`) and the materialised baseline config agrees with the row.

## What is in each commit

- `1a10b3ca` — the first pass: `--tp` / `--conc` / `--ep` passed as flags.
- `0405a501` — after review: `--isl` / `--osl` / `--precision` as well, the `ep_size`
  column and its full `read` → export → allowlist wiring, the corrected prompt
  text, and the `CHANGELOG.md` entry.
- `d05c7a0e`, `8d603adf` — @xiaofei-zheng merged `main` twice, resolved the
  `CHANGELOG.md` conflict with #1572 by keeping both `### Fixed` entries, and
  added the missing `ep_size` row to the column list in `docs/install/slurm.md`.

## Notes

- Both shell assets are `bash -n` clean.
- Backward compatibility of the new trailing column was checked against the
  `IFS=$'\t' read` line: a 13-column row leaves `EP_SIZE` empty so `${EP_SIZE:-1}`
  applies, a 14-column row carries the declared value.
- The `EP_SIZE` wiring in `run_hyperloom.sbatch` and #1572's container-reclaim
  hunk are independent parts of the same file and both survived the merge.


xiaofei-zheng:
Reviewed the diff against `git merge-base origin/main fix/slurm-optimizer-cli-workload-flags` (fa4fe164c), and read the resolution path it targets: `_resolve_workload_knobs` / `_export_workload_envs_for_optimize` in `src/hyperloom/inference_optimizer/cli/__init__.py` (L1099-1144, L1951-1962) plus the flag definitions in `cli/parser.py` (L320-366). The `--tp` / `--conc` half of the change is correct: those flags exist, they are the top of the ladder, and the env-only path really was dropped on the floor.

Blocking issues:

1. `_incontainer.sh.in:79`, `:102` — ISL / OSL / PRECISION are still env-only, so the fix is incomplete for the rest of the declared shape. `_resolve_workload_knobs` never reads `os.environ`; on a fresh launch it fills `args.isl/osl/precision` from flag > default and then overwrites the exports at `cli/__init__.py:1956-1962`. A `models.tsv` row declaring `isl=8192 osl=512 prec=fp8` therefore benchmarks 1024/1024 and whatever precision the checkpoint sniffer returns, silently — the same class of mismatch this PR fixes for `tp`. `--isl`, `--osl` and `--precision` exist on the same subparser (`parser.py:476-477`, `:508`); they should be passed as flags alongside `--tp`/`--conc` in both backends. The prompt text at `:76` ("Do not omit the env exports, model-class, precision") also still points the carrier at the ignored mechanism.

2. `_incontainer.sh.in:79`, `:88`, `:102`, `:111` — `EP_SIZE` has no producer, so `--ep` is still hardcoded to 1. `grep -rn EP_SIZE` over the repo returns nothing: `models.tsv` has no `ep` column, `run_hyperloom.sbatch:25` does not parse one, and `run_hyperloom.sbatch:42` does not export one. Even if an operator sets `EP_SIZE` in the submit environment, the docker backend's explicit `-e` allowlist (`run_hyperloom.sbatch:113-125`) drops it, so enroot (`--export=ALL`) and docker would disagree on the shape. The PR description states expert-parallel workloads now get the requested EP; as written they cannot. Either add `ep` to the row schema (tsv column -> `read` -> export -> docker allowlist) or drop the `EP_SIZE` indirection and say EP stays pinned.

3. No CHANGELOG entry. This is a user-visible behaviour fix, and comparable slurm-path fixes (#1510, #1476, #1442) each added one under `[Unreleased]`.

xiaofei-zheng:
Re-reviewed at 0405a501. All three earlier points are resolved, and I verified each:

- ISL / OSL / PRECISION are now flags in both backends (`_incontainer.sh.in:91-93`, `:118-120`), matching `--isl` / `--osl` / `--precision` in `cli/parser.py:476-477,508`. The prompt text and the python-branch comment now state the export-vs-flag split correctly.
- `ep_size` has a real producer end to end: `models.tsv` column -> `run_hyperloom.sbatch:29` `read` -> `:46` export -> `:124` docker `-e` allowlist, so enroot and docker now agree. I ran the `IFS=$'\t' read` line against a 13-column and a 14-column row: the old row yields `TARGET_GAIN=10 EP_SIZE=1` (trailing variable stays empty, `:-1` applies) and the new one yields `TARGET_GAIN=10 EP_SIZE=8`. Backward compatibility holds.
- CHANGELOG entry added under `### Fixed`, matching the surrounding format.

Two blocking issues remain:

1. The PR description still describes only the first commit. It covers `tp` / `conc` / `ep` and says EP is taken "from `EP_SIZE`", but 0405a501 also passes `isl` / `osl` / `precision` as flags, adds a `models.tsv` column plus its `run_hyperloom.sbatch` parse/export/allowlist wiring, and adds the CHANGELOG entry. As written it reads as if `EP_SIZE` were pre-existing, which was the gap the second commit closed. Please update it to match the current diff.

2. `docs/install/slurm.md:37` still enumerates the `models.tsv` columns as "key / repo / framework / image / TP / precision / ISL / OSL / concurrency / model-class / max-hours / target-gain". `ep_size` is now a column of that table and is not listed, so the only doc describing the row schema is out of date with the file this PR ships.

xiaofei-zheng:
Pushed two merge commits to this branch (d05c7a0e, 8d603adf): merged `main` twice and resolved the `CHANGELOG.md` conflicts by keeping both `### Fixed` entries. The second merge also picked up #1572; its container-reclaim hunk and the `EP_SIZE` row wiring in `run_hyperloom.sbatch` are independent, and I checked `EP_SIZE` survived in the `read`, the `export`, and the docker `-e` allowlist. `bash -n` passes on the sbatch.

Also fixed blocking issue 2 in the same push: `docs/install/slurm.md:37` now lists the `ep_size` column and notes that it is optional.

Blocking issue 1 is still open and is yours to make: the PR description only covers the first commit.

PR #1572, opened 2026-09-18
fix(slurm): reclaim GPUs from orphaned containers and run docker with --init

luochen-amd:
A cancelled or NODE_FAIL'd job kills the host-side `docker run --rm` client but
leaves its container running, still holding GPU memory, and `--rm` never fires.
The next job on that node then fails in one of two ways depending on framework,
both of them baffling until the residue is found:

  * sglang trips `wait_for_amd_gpu_clean`, which maxes VRAM% over every GPU on
    the box and hard-fails after fifteen minutes of
    `waiting for prior-job GPU memory reclaim`;
  * vLLM is refused at startup with `Free memory on device cuda:N
    (120.52/287.98 GiB) ... is less than desired GPU memory utilization`.

Observed on two nodes at once: a GLM sglang container from a job cancelled
three hours earlier still held ~168 GiB on each of GPU 0-3, which is exactly
where the next job's tp=4 server wanted to land - the container gets no ROCR
mask, so every framework starts from GPU 0. Both jobs then burned their whole
budget in ENABLEMENT with `baseline_tput=0`.

Reclaim such containers before launching. Ownership is decided by liveness, not
by name or image: the `docker run` command line carries `-e CLAW_SESSION_ID=`,
so a container whose session id has no live client is ours and orphaned, while a
co-tenant job on the same node still has its client and is left untouched. This
was verified read-only against nodes hosting live jobs before being armed -
live containers KEEP, third-party containers (an unrelated lmcache service)
SKIP, only ownerless hyperloom containers reclaimed.

Add `--init` to `docker run`. Without an init the container's pid 1 becomes an
unreapable zombie once the client is killed, and then even the daemon gives up:

    cannot kill container <id>: PID <n> is zombie and can not be killed.
    Use the --init option ...

which is why one orphan above could not be removed with `docker kill` at all.
The reclaim keeps a cgroup-level SIGKILL fallback for containers already in
that state. Add `--name` so a container is traceable back to its slurm job.

Every branch in the new code is spelled as `if`, not `cmd && ...`: this file
runs under `set -e`, where a failing AND-list would abort the job. That is not
hypothetical - the first draft did exactly that.


xiaofei-zheng:
Reviewed the full diff against `git merge-base origin/main fix/slurm-orphan-gpu-reclaim` (fa4fe164).

**What this PR does:** before launching the job container, `run_hyperloom.sbatch` now scans running docker containers, treats one as ours-and-orphaned when it carries a `CLAW_SESSION_ID` that (a) is not this job's and (b) has no live `docker run` client, kills/removes it (with a cgroup SIGKILL fallback for zombie inits), and then polls `rocm-smi` VRAM% for up to 180s. `docker run` also gains `--init` and `--name hl-<key>-<jobid>`.

**Blocking**

1. `CHANGELOG.md` — no entry for this change. This is a user-visible behavior change in a shipped asset (the Slurm launcher now kills containers on the node before starting, and containers are named and run under an init), and the launcher's behavior changes are tracked in this file already (see the `HL_KERNEL_BACKEND` entry). Add an item under `[Unreleased] / ### Fixed`, matching the existing entry style.

Checked and found no other blocking issues:
- ran the new preflight block under `set -euo pipefail` with stubbed `docker`/`pgrep`/`rocm-smi`: a container with a live client is kept, a third-party container with no `CLAW_SESSION_ID` is skipped, only the ownerless one is reclaimed, and the function returns 0 on the 180s-timeout path so the job continues;
- `bash -n` clean; no new `cmd && ...` under `set -e`; every `local` is declared and `set -u` is respected (`${CLAW_SESSION_ID:-}`, `${vram:-0}`);
- `CLAW_SESSION_ID` is unconditionally set at line 45 and forwarded as the first `-e` of `docker run`, so the ownership probe has something to match on every container this launcher starts;
- `hl_vram_max` matches the `GPU Memory Allocated (VRAM%): N` form of `rocm-smi --showmemuse` and degrades to 0 when rocm-smi is absent;
- `--name` cannot collide in practice: the key comes from `models.tsv` and the suffix is `SLURM_JOB_ID`, and `--rm` removes the container at exit.

xiaofei-zheng (approved):
Re-reviewed after b9226c9. No blocking issues — LGTM.

The follow-up commit only adds the `CHANGELOG.md` entry (`git diff e20ab04b0 b9226c929` touches that file alone; `run_hyperloom.sbatch` is unchanged from the revision I reviewed). The entry sits at the top of `[Unreleased] / ### Fixed`, matches the surrounding entry style, and its description matches the diff: liveness-based ownership via `CLAW_SESSION_ID`, `--init` for the zombie-pid-1 case with the cgroup SIGKILL fallback, and `--name hl-<key>-<jobid>`.

Carried over from the first pass, which cleared the script itself:
- ran the preflight block under `set -euo pipefail` with stubbed `docker`/`pgrep`/`rocm-smi`: live container KEEP, third-party container with no `CLAW_SESSION_ID` SKIP, only the ownerless one reclaimed, and the 180s-timeout path returns 0 so the job proceeds;
- `bash -n` clean; no new `cmd && ...` under `set -e`; `set -u` respected throughout;
- `CLAW_SESSION_ID` is always set before launch and forwarded as the first `-e` of `docker run`, so every container this launcher starts is classifiable;
- `hl_vram_max` matches the `GPU Memory Allocated (VRAM%): N` form of `rocm-smi --showmemuse` and degrades to 0 when rocm-smi is absent.

PR #1573, opened 2026-09-18
fix(profile): find the steady-state trace the injected profiler dir redirected

BaoYunkai:
## Problem

Magpie's launcher emits its own `--profiler-config.torch_profiler_dir
<workspace>/torch_trace` **before** `EXTRA_VLLM_ARGS`, so the placeholder
`_workload_envs` appends -- the task root -- is the value vLLM resolves under
its dotted-flag last-wins merge. Observed argv order on a real round:

```
14  --profiler-config.profiler torch                       <- Magpie
16  --profiler-config.torch_profiler_dir <ws>/torch_trace  <- Magpie
...
34  --profiler-config.torch_profiler_dir <task root>       <- EXTRA_VLLM_ARGS, wins
```

and the server confirms it: `ProfilerConfig(torch_profiler_dir='<task root>')`.

The server therefore writes its steady-state traces straight into the task root
and its graph-capture sidecars into that root's `capture_traces/`, while
Magpie's own `torch_trace/` stays empty and says so
(`[PROFILE] No trace found for relay under: .../torch_trace`).

`_candidate_trace_dirs` probed the workspace and the root's `capture_traces`,
so it saw only sidecars and fell back to capture-only. Roofline rejects a
capture-only trace, so every attempt failed with `profile_capture_only_failed`
while the traces it wanted sat one directory up.

Session `20260917T191938Z-1ad81699` lost both roofline tasks that way, 3 of 3
attempts each, against 24 complete annotated steady-state traces (8 ranks x 3
attempts, ~250MB each) already on disk. The knock-on is the whole kernel phase:
GEMM tuning had no shapes to derive (`no candidates to E2E validate`) and
forge-fusion was skipped outright (`no decode trace yet`), leaving KB replay as
the session's only source of gain.

## Fix

Probe the task root too. Which side wins that merge is not ours to rely on --
Magpie's launcher decides it -- so both destinations are answered for rather
than betting on the current order. The workspace keeps priority, so a round
whose traces did land there is unaffected. Capture sidecars stay excluded
either way, since `is_capture_fragment` matches both the `graph_capture_*`
filename and the `capture_traces/` directory, so the capture-only verdict still
means what it says.

The task root is shared between attempts, and `_trace_files_for_dir` rglobs, so
a trace found there has to be proved to be this attempt's. `task_started_unix`
is taken per executor call, which is per attempt, and the agentx path already
gates on it; the task-root candidate now does too. That also rules out the
sibling workspaces the rglob reaches.

## Test plan

- [x] Two new tests in `test_profile_and_kernel_handlers.py`: a trace the
      injected dir redirected to the task root is found rather than falling
      back to sidecars, and a previous attempt's trace in that shared root is
      not adopted. Both fail without the respective change.
- [x] `test_profile_and_kernel_handlers.py` (177 passed; 1 failure,
      `test_trace_certificate_stays_out_of_the_resolver_namespace`, also fails
      on unmodified main with `ModuleNotFoundError`)
- [x] `src/hyperloom/orchestrator/` (1297 passed, 1 skipped)
- [x] Replayed against the session above: both roofline tasks go from
      "capture-only fallback" to 24 steady-state traces, and with the watermark
      applied the newest attempt admits exactly its own 8
- [x] `ruff check` + `ruff format --check` on the changed files
- [x] CHANGELOG Unreleased entry


ZhengGong-amd:
Reconsider this as a write-side fix. The task-root search should not exist.

Magpie emits `--profiler-config.torch_profiler_dir <workspace>/torch_trace` first. `_workload_envs` then appends the same flag pointed at the task root, and vLLM last-wins keeps the task root. That directory is shared by every attempt. Probing `workspace.parent` and filtering `trace_files` by mtime does not scope the attempt: on the roofline path `_preferred_main_trace_path` returns the directory as `main_trace_path`, and `discover_trace_inputs` rglobs it with no watermark. From attempt 2 the analysis includes earlier traces.

Drop the placeholder from the launched `EXTRA_VLLM_ARGS` unless the operator already set one. `baseline.py` already injects `<output_dir>/torch_trace` on the probe argv only; `bypass_engine` already puts the real dir after extra args. Leave discovery on `<workspace>/torch_trace`. Delete `workspace.parent`, the `scoped` watermark, and the two new tests. Leave the `_workload_envs` comment — it is true once the launched flag stops overriding Magpie.

BaoYunkai:
Adopted the write-side approach from your comment.

- `_workload_envs` no longer appends a task-root `torch_profiler_dir` placeholder to `EXTRA_VLLM_ARGS` (operator-set values unchanged). Magpie’s `<workspace>/torch_trace` flag now wins last-wins merge at launch.
- Reverted task-root probing in `_candidate_trace_dirs`, the `scoped` mtime watermark on `workspace.parent`, and the two executor tests that encoded the read-side fix.
- Updated the vLLM profile materialization test to expect `profiler=torch` without a launched `torch_profiler_dir`; argv preflight still uses the probe-only injection in `baseline.py` / `bypass_engine.py`.
- Kept the expanded `_workload_envs` comment, updated to describe the new behavior.

Pushed on `bugfix/yunkai/profile-steady-state-trace` (commit message: keep torch traces under workspace/torch_trace).

ZhengGong-amd:
Rebase onto main; this head conflicts, so Lint and Tests with Coverage never started and must pass on the rebased commit.

Rewrite the PR body: it still describes probing the task root and the mtime watermark, but the change now stops `EXTRA_VLLM_ARGS` from overriding Magpie so traces stay in `<workspace>/torch_trace`.

Correct the new comment and test note: Magpie emits `torch_profiler_dir` before `EXTRA_VLLM_ARGS`, and the fix works because that flag is no longer written into the launched args, not because Magpie's copy comes last.

ZhengGong-amd (approved):
Approved

PR #1577, opened 2026-09-19
fix(mcp): prevent inline action waits from blocking the coordinator

### LINKED ISSUE #1552: critic/robustness agents (protocol=anthropic) appear to lack the idle timeout ClaudeBackend has — a stalled round wedges the coordinator
## Summary

The orchestration backend (`ClaudeBackend`) bounds each LLM call with an idle timeout — `call_timeout_s`, default 120 s via `parse_call_timeout_env("INFERENCE_OPTIMIZER_CLAUDE_CALL_TIMEOUT_SEC")` — so a slow or stalled gateway round is aborted and retried.

The **critic** and **robustness** agents are configured as `protocol=anthropic` (direct HTTP) rather than going through `ClaudeBackend`, and do not appear to share that idle bound. When one of those rounds stalls, and because the step runs synchronously on the coordinator thread, the **whole coordinator wedges with no upper bound** — no timeout, no retry, no error.

## Symptom

The loop self-drives through preflight → `target_analysis` → `baseline_executor`, then blocks indefinitely *before* dispatching the benchmark: no child process, GPU idle, 25+ minutes with no progress and no log line after `run_action_now: invoked on the coordinator loop thread`.

## Evidence — single-variable A/B

| Run | Result |
|---|---|
| default (real critic + robustness) | wedges at `baseline_executor`, 25+ min, no progress |
| `--critic-mock --robustness-mock` | loop advances past that point and proceeds normally |

Flipping only the mock flags localizes the stall to the critic + robustness agents. The run banner shows both as `protocol=anthropic` (e.g. `critic=critic-agent(... protocol=anthropic ...)`), i.e. the direct-HTTP transport, distinct from the `claude-agent-sdk` path used for orchestration.

## What I could NOT confirm

I could not capture a stack to prove which call is blocked: `py-spy` needs ptrace, and ptrace is denied inside the container on the node I had. So the mechanism below is a **hypothesis consistent with the A/B**, not a demonstrated fact — I am filing it because the localization is solid and the subsystem is yours.

**Hypothesis:** the direct-HTTP `protocol=anthropic` agent runtimes lack the per-call idle timeout that `ClaudeBackend` has, so a stalled read blocks forever; because these steps execute on the coordinator thread, the stall is not contained to the agent.

## Suggested direction

Give the direct-HTTP agent transport the same idle-timeout + bounded-retry treatment `ClaudeBackend` already has (`call_timeout_s`, `RetryPolicy`), and/or move those calls off the coordinator thread so a stalled agent degrades that agent rather than the whole loop.

## Environment

- Hyperloom `1.1.0`; paths re-checked against `main`
- Orchestration via an OpenAI-compatible (litellm-style) proxy, `claude-opus-5`

Happy to test a patch against this setup — the A/B reproduces reliably here.

zengleixin-amd:
## Summary
Related to #1552. Fix the independently reproduced scheduling defect behind the `run_action_now: invoked on the coordinator loop thread` warning without inventing another watchdog or changing critic HTTP defaults.

The in-process MCP handler awaited a synchronous provider that submitted work to the coordinator loop, then called Future.result on that same thread. The loop could not start the action, advance timers, or handle cancellation until the inline wait expired.

The MCP action callback is now explicitly asynchronous. Sync and async entrypoints share admission, scheduling and result formatting; async callers await the submitted Future without occupying the default thread pool. Direct synchronous calls on the coordinator loop are rejected before coroutine creation. Read-only context projections remain synchronous.

## Design and preserved contracts
- No change to inline eligibility, policy, session budgets, registered execution ownership, or cleanup/shield behavior.
- Caller timeout/cancellation does not cancel the already submitted action; it can complete and publish its single delegated_result. Late failures are consumed rather than becoming unhandled Future exceptions.
- The synchronous cross-thread bridge remains available with its existing timeout/result strings.
- A first to_thread approach was rejected after independent review reproduced default-executor starvation with real SQLite and concurrent SDK calls. The submitted version does not hold executor workers while waiting for the coordinator.
- Two small commits: two MCP interface/test files, then four bridge/wiring/test/changelog files. No temporary documents or environment files are committed.

## Validation
- Test-first: original handler regressions failed on thread identity, timer progress and cancellation; new asynchronous contract regressions failed before the final implementation.
- Local real SDK 0.2.157 / MCP 2.2 request-handler tests: 47 passed, 3 POSIX coordinator cases skipped on Windows. No query(), Claude CLI or LLM network calls.
- Ten independent-process repetitions of async MCP/SDK/saturated-executor tests passed with bounded subprocess deadlines.
- Independent adversarial check: 1-worker/8-call, default-20-worker/20-call and 32-worker/32-call real SQLite runs all completed with zero timeouts; source cancellation, caller cancellation and late-error cases preserved semantics with no loop exception reports.
- Existing dispatcher ownership suite: 42 passed using the explicitly documented Windows AST facade isolation. Same-loop rejection and single-worker real SQLite regression bodies also passed with original production methods AST-isolated.
- Ruff lint/format, Python 3.10 grammar and git diff checks passed. Final-head Linux validation PASSED on ddf058447: https://github.com/AMD-AGI/Hyperloom/actions/runs/35421285114 . All 12 Python 3.10/3.11 shards, both strict coverage thresholds, CodeQL and other checks succeeded. No failed or pending checks remain. This includes the native POSIX coordinator/SQLite regressions that were skipped locally.

## Scope and remaining investigation
Current critic HTTP already has connect10/read300 inactivity timeouts and an outer reactor/session budget, default1800 seconds per turn. Those defaults and RetryPolicy are not changed. SDK idle120 and inline wait120 remain independently timed; this PR does not promise a slow action result arrives before SDK idle.

This proves and fixes the scheduling defect, not the entire reported 25-minute incident. The original action name, installed revision and transport/configuration still need field confirmation; baseline is not inline-eligible by default. No paid LLM, GPU or cluster test was run. Do not automatically close #1552 pending reporter retest.

## Checklist
- [x] Regression-first implementation and independent review.
- [x] Offline real-SDK and saturated-executor validation.
- [x] Latest-head Linux tests and strict coverage.
- [x] Latest-head static checks and CodeQL.
- [ ] Reporter confirmation of the original incident.


BaoYunkai:
Bugbot review on ddf0584 (branch changes vs origin/main): no bugs found.

PR #1578, opened 2026-09-19
fix(sbd): give every timeline failure one shape and one place to fail

haishuok0525:

  Three failures of the same kind: a run could end badly and the breakdown would not say so, or would say so in a field no reader could find.

  **One shape.** The canonical failure row was keyed by `phase` while every caller passed a step — a profiling substep, a baseline round, a phase entry — and the one consumer that surfaces it reads a stage. Renamed to `stage` across the six recorders that build it, and given `V6Failure` so the shape is declared once rather than restated per event. `failure_row` now takes the exception itself, because the callers that had one were dropping its message and passing only a class name.

  **conc_sweep lost its failures outright.** An arm that raised mid-ladder left `finish_arm` with only the rungs that had already succeeded and no trace of what stopped it, so a sweep that died halfway read as a short sweep. The arm now captures and re-raises, and the three recoverable steps that used to fail silently — lifecycle resolve, server teardown, report write — record a fault naming the step and the arm. The teardown case matters most: it leaves a server alive past the arm that owned it, and a `log.debug` was the only evidence. The CONC coercion guard is gone with its test; a non-numeric CONC is a defect, and defaulting the row to `0` turned it into a plausible measurement.

  **One place to fail.** The recorder subpackage caught `Exception` at 79 of its 100 catch sites, which is not a safety net so much as a blindfold: a bug in the recorder and a spool that was never written produced the same silence. The write side already had a boundary in `EventSink`; the read side had none, so every caller that consults its own rows carried a catch of its own. Adds `recorded_section`/`recorded_rows` as that boundary and narrows both to `RECORDING_ERRORS` — the spool is JSON files on disk, individually bad fragments are already skipped by the assembler, and the only other way in is with no session bound. A `TypeError` projecting a row is a defect and now says so. With both boundaries in place the 37 wrappers in `phase_event`, `enablement_event` and `stack_event` were guarding nothing, and are gone. Broad catches drop from 79/100 to 42/67.

  Also: `V6WarmStartExt` declared `requested` where the wire has always written `request`, so a reader following the schema found nothing there.

  None.

  Added: conc-sweep wiring coverage for the report-write, lifecycle-resolve and lost-rung faults, and for a ladder cancelled mid-way; conc-sweep timeline coverage for a decline that names its error, for a stop reason recorded as runtime rather than failure, and a block-by-block check against the declared schema.

  Updated: `test_sbd_v6_stack_wiring.py`'s recording-failure test now fails at the write itself (`OSError` out of the recorder) rather than by making the writer raise `RuntimeError`, since the sink is where recording is allowed to fail quietly and a fault anywhere else is meant to surface. `phase` → `stage` assertions across the roofline, kernel and warm-start suites. `test_point_from_variant_defaults_conc_zero_on_bad_env` is deleted with the guard it covered.

  ```
  ruff check src/hyperloom && ruff format --check src/hyperloom
  pytest src/hyperloom -p no:randomly
  ```

  17029 passed. The 20 remaining failures are pre-existing on `main` and unrelated: 12 need the `openai` SDK, which is not installed here, and 8 are `/tmp` test-isolation failures in `test_external_multi_node.py` and `test_supervisor.py`. None has a recorder frame in its traceback. `test_failure_digest_properties.py` is skipped because `hypothesis` is not installed.

  Yes, within the V6 breakdown wire format. The failure block's `phase` key is now `stage`, and `V6WarmStartExt.requested` is now `request`. Both are read only inside this repo, and every reader is updated here.

  Yes — making a failure that happened reach the timeline. The three parts are the shape it is written in, the conc_sweep failures that never reached it at all, and the catches that were swallowing the rest.

  This is the first of two PRs splitting a larger SBD change. The second carries the completions that build on this shape: inline sub-step attribution, the kernel event rework, baseline eval-failure recording, and the remaining schema typing.

  No, all in-repo.


BaoYunkai:
Local review on f2e1fa5 (branch changes vs `origin/main`) found one issue.

**`record_exit` can now raise out of the phase transition path** — `src/hyperloom/inference_optimizer/breakdown/recorder/phase_event.py:552`

The docstring still states "Never raises", but the `except Exception` wrapper around the body was removed while `_finish` continues to call `event_parts` unguarded. Two reachable paths escape:

1. `_finish` calls `parse_event_id(event)`, and `event` may come from the `event_id` field that `_open_segment` reads back off the spool. `parse_event_id` raises `ValueError` when the id does not have three segments or when `macro_cycle` is not an integer, so a malformed on-disk id propagates. `EventSink.record` swallows that same `ValueError` through `RECORDING_ERRORS` one line earlier, so the write goes quiet and only `_finish` surfaces it.
2. `event_parts` -> `assemble_parts` -> `d.glob()` can raise `OSError` at the directory level. `_load` only guards per-fragment reads.

Both were caught by the removed wrapper before this change, so the PR description's "were guarding nothing" does not hold for this path. It also leaves the file internally inconsistent: `_open_segment` goes through the new guarded boundary `recorded_section` (which narrows to `RECORDING_ERRORS`), while `_finish` uses raw `event_parts`.

Routing `_finish`'s read through the same guarded boundary, or restoring a narrow `RECORDING_ERRORS` catch in `record_exit`, would keep the stated contract.

For context, unguarded `event_parts` on the write path is a pre-existing pattern in `kernel_event`, `stack_event` and `baseline_event` as well; `phase_event` is the module where this PR removes the outer catch that used to cover it.

haishuok0525:
@BaoYunkai Agreed — that was a real hole left by the unwrap.

`_finish` now catches `RECORDING_ERRORS` around the parse + `event_parts` close, so an `OSError` from the spool (or a `ValueError` from `parse_event_id`) is noted and dropped instead of escaping into the phase transition. Finalize still recovers an event left open as `interrupted`.

Separately, `_open_segment` now skips a segment whose `event_id` will not parse. That was the other half of the same path: a malformed on-disk id used to be handed to `_sink` / `make_sink`, which raises the same `ValueError` the sink already swallows on write, before `_finish` even ran.

Pinned both with `test_an_unreadable_spool_on_exit_does_not_raise` and `test_a_malformed_spool_event_id_does_not_raise_out_of_exit` in a39181632.

BaoYunkai:
Re-reviewed on a391816. The `_finish` fix looks right: the `RECORDING_ERRORS` boundary plus skipping unparsable `event_id`s in `_open_segment` closes both paths I raised, and the added tests cover them.

One instance of the same pattern is still open.

**`enablement_event.finish` can raise out of lane teardown** — `src/hyperloom/inference_optimizer/breakdown/recorder/enablement_event.py:510`

`finish` calls `event_parts` / `assemble_enablement_ext` without the `RECORDING_ERRORS` boundary, and this PR removed the `except Exception` that used to wrap the body on `main`. An unreadable spool at close therefore propagates. Neither caller guards it:

- `writeback.py:1169` `_close_enablement_lane` calls it directly with no catch.
- `lane.py:744` `_settle_enablement_round` likewise.

At `lane.py:88` that call happens after `stop_reason` is set and `state.save` has run, so a raise there skips the `log.warning` below it and unwinds into the enablement pump, leaving the lane half torn down.

`phase_event._finish` now notes this and returns quietly. Routing this one through the same boundary would make the two agree.

Worth noting this one is more ambiguous than the `phase_event` case: the docstring's "Never raises" was dropped here in the same commit that dropped the wrapper, so this may be an intentional narrowing rather than an oversight. If it is intentional, the callers are the places that need to decide what a failed close means.

haishuok0525:
@BaoYunkai Agreed — same hole, same fix.

`enablement_event.finish` now catches `RECORDING_ERRORS` around the close-time `event_parts` / `finish_event` path, matching `phase_event._finish`. The result fragment is still written best-effort through the sink first; if the spool cannot be read for assembly, the close is noted and dropped so lane teardown (after `stop_reason` / `state.save`) cannot unwind half-done, and finalize recovers the open event as `interrupted`.

Pinned with `test_an_unreadable_spool_on_finish_does_not_raise` in 7e21b4333.

Dropping the docstring's \"Never raises\" in the unwrap was incidental, not an intentional narrowing of the caller contract — the callers still assume `finish` cannot raise.

BaoYunkai:
Re-reviewed on 7e21b43. Both issues I raised are closed, and this round found nothing new.

`enablement_event.finish` now catches `RECORDING_ERRORS` around the close-time read and notes the failure, matching `phase_event._finish`, and `test_an_unreadable_spool_on_finish_does_not_raise` covers the `OSError` path through lane teardown.

One thing I looked at and decided is not a problem, recorded here so it does not get re-raised: `finish` on the conc-sweep event does not carry `self._faults` into the terminal action, but no fault is lost by that. `_record_action` goes through `record_upsert_item`, which merges into the existing fragment payload, so the `faults` field written by `record_fault` survives the close-time upsert untouched. `record_fault` also writes the full list each time rather than appending, so a dropped upsert is repaired by the next one. The only remaining window is the last fault of a sweep being dropped by the sink, which `EventSink.record` already reports as the documented best-effort behavior.

PR #1579, opened 2026-09-19
feat(explore): capture launch_config.json to skip baseline-noop Agent…

nehmathe2:
**Description: what and why**
Explore was benchmarking variants that only restated flags/env already on the live baseline (often missing from Magpie YAML). AgentX now writes launch_config.json from /proc after server-up; explore drops exact restatements before Magpie runs.

No new test file in this PR. Validated on DSV4 AgentX, conc 128, using live launch_config.json (source=proc).

| Variant | Proposal | Baseline already had | Result | Log |
| --- | --- | --- | --- | --- |
| `noop-chunked-prefill-restatement` | `--chunked-prefill-size 65536` | `--chunked-prefill-size 65536` | dropped | `skipping baseline-noop variant … (delta already active in the observed server launch)` |
| `sys-hicache-io-backend-kernel` | `--hicache-io-backend kernel` | not that flag | queued | `grid_runner: variant 1/1 name=sys-hicache-io-backend-kernel` |
Dedup: `payload=2 → runnable=1 (baseline_noop=1)`. Filter used observed launch from baseline (`59` flags, `35` env keys).

**CHANGELOG.md [Unreleased]**
No, no public API change; visible only as skipped explore variants / baseline_noop in logs.

**Breaking changes**
No. If launch_config.json is absent, explore behavior is unchanged and every variant still runs

**PR addresses single concern**
Yes. Capture + baseline-noop filter only.


ZhengGong-amd:
## PR #1579 -- feat(explore): capture launch_config.json to skip baseline-noop AgentX variants

What it does: snapshots the AgentX server's argv/env from `/proc` into `launch_config.json`, and drops explore variants whose delta already matches that launch (or the YAML stack) before Magpie runs.

The goal is right, but the route builds a second observed-launch mechanism beside the one the repo already owns, and it drops variants wrongly and silently. Please rescope rather than patch.

Blocking issues: 5

1. [D1] Parallel mechanism. `build_launch_evidence` already records `observed_server_launch_flags` on every baseline/grid measurement, warm-reuse included, and `writeback._measurement_launch_flags` resolves it. This PR adds `/proc` capture, a sidecar JSON with its own reader, a dead Python writer (`capture_launch_config_via_proc` and `write_launch_config` have no callers), and `_explore_launch_signature`, a hand copy of `_build_variant_yaml`'s merge that already omits `overlay_pythonpath`, `runtime_override` and `BLOCKED_EXTERNAL_ENV_NAMES`.
2. [D8] Wrong comparison base. The observed check ignores `base_extra_args`, and `live_launch_state_from_state` silently falls back to `last_baseline` when `current_best` has no file. Warm-up grids (the default) measure client-only, so capture does not run there. Reproduced at head: with the stack at `--chunked-prefill-size 32768`, a variant proposing `65536` is dropped as a noop.
3. [S3] Silent drops. Noop drops are only logged: they never reach `skipped_dup`/`per_variant_outcomes`, so the agent sees nothing and re-proposes them. An all-noop grid returns `empty_grid: params.grid contains no valid variants`.
4. [X2] Description vs diff. "If launch_config.json is absent, explore behavior is unchanged" is false: the YAML-signature branch drops variants with no file present (reproduced). Not declared: two new env toggles (absent from `environment-variables.md`), and `resume_stack_revalidate` now also bypassing `filter_operator_pinned_envs`. The new gate has no tests.
5. [V5] `CONFLICTING` with main: #1587 rewrote the dedup log line and unified the attempt ledger this PR edits.

Requested scope:
- Compare the variant against the launch of the measurement that *is* the current stack (`current_best`, else baseline), using `launch_evidence.observed_server_launch_flags`. Filter only when that evidence is observed. Without it, filter nothing, and never fall back to a different measurement.
- Compute the variant side through the real `_build_variant_yaml`/`seal_server_argv` merge, not a copy.
- If env matters, add `observed_server_env` to `build_launch_evidence`: one schema, one owner.
- Record drops in the #1587 ledger as `TS_SKIPPED_DEDUP` with `reason=baseline_noop`.
- Delete the `aiperf_client.sh` capture, `launch_config.json` and its reader/writer, both env toggles, the YAML-signature detector, and the `_CONFIG_REPLAY_PROVENANCE` change.
- Add tests that pin three cases: a noop is dropped and recorded; a flag the stack changed is not dropped; without observed evidence nothing is dropped.

Checked: `explore.py`, `_launch_evidence.py`, `aiperf_client.sh`, `_grid_runner._build_variant_yaml`/`_attach_grid_launch_evidence`, `baseline.py` warm-up evidence copy, `writeback._measurement_launch_flags` | Ran: `filter_baseline_noop_variants` at head on 4 constructed cases | Base: 4de5135898d8 | Head: 1ffeb82e17e5

PR #1583, opened 2026-09-21
Fix KernelForge patch hygiene and integration ordering

lishuoshuo-amd:
## Summary

- Force approved KB warm-start pathspecs so tracked-but-ignored files cannot reject the commit.
- Stage only tracked edits and preparation-created files, preventing pre-existing runtime/JIT artifacts from leaking into exported patches.
- Persist each forge-loop's stdout and stderr alongside its controller task result.
- Integrate controller patches by `task.json.priority` instead of encoded directory-name order.

## Motivation

KernelForge campaigns exposed several integration issues:

- Warm-start commits could fail when repositories tracked files matched by their own `.gitignore`.
- Task preparation could accidentally commit pre-existing JIT caches and generated artifacts, producing oversized patches that failed to apply.
- Forge-loop failures lost their full console output after workspace cleanup.
- Patch integration ignored the Controller's priority ordering, changing the cumulative patch and E2E validation sequence.

These changes preserve the existing Controller contract and complement the conflict-merge support already on `main`.

## Test plan

- Added regression coverage for forced approved pathspecs.
- Added regression coverage for excluding pre-existing untracked files from task preparation commits.
- Added coverage for persisted forge-loop stdout/stderr.
- Added coverage for Controller priority ordering.
- Ran focused test suites: `122 passed`.
- Verified lint and `git diff --check`.


xiaofei-zheng:
## What this PR does

Four independent KernelForge integration fixes:

1. `experience_integration._git_commit_all` now runs `git add -A -f -- <allowed_paths>`, so a tracked-but-`.gitignore`-matched file no longer makes the whole KB warm-start commit fail. The `staged_paths.issubset(allowed_paths)` guard downstream still bounds what `-f` can pull in.
2. `task_preparer._finish_success` replaces `git add -A -- . :(exclude)forge_experiments` with `_stage_preparation_changes()`: `git add -u -- .` for tracked edits plus an explicit `git add` of only the untracked paths that appeared after the `pre_untracked` snapshot. A staging failure now rolls back and returns a failed `PrepareResult` instead of silently continuing.
3. `dispatcher._persist_forge_loop_output` writes `forge-loop.stdout.log` / `forge-loop.stderr.log` into the task dir right after `run_forge_loop` returns, so output survives worktree release.
4. `controller_patch_integration` sorts publications by `(task.priority, task.operator_id)` read from `<output>/controller/tasks/<dir>` instead of patch-directory name order.

## Blocking

**1. `controller_patch_integration.py:240` — docstring still states the old ordering.**

```
"""Apply and E2E-validate every complete Controller patch in filename order.
```

The loop at line 269 now iterates `_priority_ordered_patch_dirs(...)`, i.e. controller priority order. Since ordering is exactly the contract this function's callers depend on (cumulative patch sequence and E2E validation sequence), a docstring asserting the opposite is worse than none. Please update it.

**2. No CHANGELOG entry.**

`CHANGELOG.md` has an active `## [Unreleased]` / `### Fixed` section, and the directly adjacent change this PR says it complements — the conflict-merge support for controller patch integration — is documented there ("A measured Controller patch was dropped because the patch kept before it had moved its context"). Changing the integration order changes the cumulative patch and E2E validation result for the same set of publications, which is observable; so are patches that previously failed to apply because of oversized preparation commits. Please add an entry in the existing format.

## What I checked

- `_priority_ordered_patch_dirs` path derivation: `patches_root` comes from `controller_submit.py:112` (`<output>/result/patches`), so `root.parent.parent / "controller" / "tasks"` resolves to `ControllerLayout.tasks_root`, and `patch_dir.name` / `task_dir` name both come from `operator_directory_name` — they match.
- `load_task` failure modes: `parse_task_payload` and `safe_relative_path` / `safe_operator_id` raise only `TaskContractError`, which `load_task` catches and turns into `task is None`, so a malformed task dir falls back to the `(2**31-1, name)` key rather than aborting the whole integration.
- `pre_untracked` is captured at `task_preparer.py:1739`, before the agent runs, and `_finish_success` is a closure over it — the scope is correct.
- `_git_untracked` uses `--exclude-standard`, so the new `created` set never widens what the old `add -A` would have staged; the change is strictly narrowing. `_git` passes `cwd=workspace`, so the `.` pathspec and the `ls-files` output are relative to the same root.
- Dropping `:(exclude)forge_experiments` from the tracked-edit stage is safe: `forge_experiments` is the loop's own workspace output dir (`worktree.FORGE_LOOP_OUTPUT_DIRNAME`), never tracked in the target repo, so `add -u` cannot reach it; the created-file filter still excludes it explicitly.
- The new failure branch in `_finish_success` matches the shape of the sibling `spec_indexed` / commit-did-not-land branches (`_rollback()` then `rolled_back=True`).
- The two new log files land in the published task dir; `is_published_task_dir` and `discover_task_dirs` key off `task.json` / `driver.py` only, and the `copytree` in `dispatcher.py:159` copies the workspace audit dir, not the task dir — nothing downstream enumerates the task dir strictly.
- Reviewed against `git merge-base origin/main HEAD` to confirm all four issues are introduced here, not pre-existing.
- Did not run the suite locally; CI covers it.

lishuoshuo-amd:
Addressed both blocking items in b50a07b84: updated the integration docstring to describe task-priority ordering and added an Unreleased/Fixed changelog entry covering the user-visible KernelForge fixes. The controller integration suite passes (30 tests).

xiaofei-zheng (approved):
Re-reviewed at 7002d30af. Both blocking items from the previous round are resolved — no blocking issues, LGTM.

## Resolved

- `controller_patch_integration.py:240` now reads "Apply and E2E-validate complete Controller patches in task-priority order", matching the `_priority_ordered_patch_dirs` loop at line 269.
- `CHANGELOG.md` gains an `Unreleased / Fixed` entry covering all four fixes, in the existing format.

## What I checked this round

- Re-diffed against `git merge-base origin/main HEAD`: all four functional changes survived the merge intact — `_priority_ordered_patch_dirs` (`controller_patch_integration.py:83`, used at 269), `git add -A -f` (`experience_integration.py:325`), `_stage_preparation_changes` (`task_preparer.py:917`, called at 1939), `_persist_forge_loop_output` (`dispatcher.py:60`, called at 226). No source change was lost or reverted.
- The `origin/main` merge at 7002d30af had one conflict, in `CHANGELOG.md`. Inspected the combined diff: both unreleased entries are preserved — this PR's KernelForge entry and main's inline-MCP-actions entry — and neither was truncated or reordered into the wrong section.
- The rest of the merge (`loop/dispatcher.py`, `loop/conversation.py`, `roles/mcp_context_tools.py` and their tests) came in clean from main and does not touch any KernelForge path, so there is nothing from main that needs mirroring into this branch.
- PR title and description still match the diff; the four bullets under Summary describe exactly what the current tree does.

Carried over from the first-round review: verified the `<output>/result/patches` -> `<output>/controller/tasks` path derivation against `controller_submit.py:112` and `ControllerLayout.tasks_root`, `load_task`'s `TaskContractError`-only failure mode and its `task is None` fallback key, the `pre_untracked` closure scope in `_finish_success`, and that `_git_untracked --exclude-standard` makes the new staging strictly narrower than the old `add -A`.

Did not run the suite locally; CI covers it.

PR #1584, opened 2026-09-21
fix(grid): keep a measurement that served its whole protocol

### LINKED ISSUE #1478: Valid completed benchmarks are discarded when the wrapper script exits non-zero (magpie_nonzero_after_valid_measurement)
### Target release

Unsure

Local

### Optimization domain

Inference

### Issue type

Wrong result or regression

### Which phase failed?

Optimization loop

### Short Summary

A benchmark that finishes and writes a valid report gets thrown away when the wrapper script's bash exits non-zero for an unrelated reason.

```shell
python3 -m hyperloom.inference_optimizer.cli optimize \
  --model /path/to/gpt-oss-120b --framework sglang --gpu-type mi355x \
  --isl 8192 --osl 1024 --conc 64 --tp 1 --ep 1 --max-model-len 13312
```

### Expected behavior

If the benchmark ran to completion and wrote a valid, protocol-comparable result, I'd expect Hyperloom to use that number, or at least retry. A failure in the shell wrapper *after* the measurement is done shouldn't cost me the measurement.

### Actual behavior

On this box the InferenceX checkout lives on NFS. Every so often bash loses its handle on the benchmark script while it's still executing it, and dies at the very end:

```
benchmarks/sglang_mi355x.sh: error reading input file: Stale file handle
```

The important part is *when* it happens. The benchmark has already finished successfully by then:

```
Magpie.modes.benchmark.result - INFO - Parsed benchmark result: 3.33 req/s
Magpie.modes.benchmark.workspace - INFO - Saved benchmark report: .../benchmark_report.json
Magpie.modes.benchmark.benchmarker - INFO - Benchmark completed in 153.00s
Magpie.main - ERROR - Benchmark failed: ['Benchmark process failed with code 2', ...]
```

The report is on disk and the result is parsed. Then bash exits 2 reading the tail of its own script, Magpie reports the whole thing as failed, and Hyperloom records `error_class=magpie_nonzero_after_valid_measurement` and discards the number.

The error class is named for exactly this situation, so the condition is already recognised somewhere — it just ends in a discard rather than a keep or a retry.

What it cost me in one session:

- 2 of 8 points in the post-optimization concurrency sweep (conc=64 and conc=256, so the curve is missing its most interesting points)
- 4 of 4 variants in one optimization leg, which meant that leg explored nothing at all

Hyperloom's own specialist worked this out during the run, which I thought was worth quoting since it's an independent read of the same evidence:

> The stale-handle line 'benchmarks/sglang_mi355x.sh: error reading input file: Stale file handle' IS present in both, exactly one occurrence each, in the single warmup benchmark_stderr.log, immediately after '+ local eval_exit=0' / '+ set +x'. Both abort_reason.json carry error_class=magpie_nonzero_after_valid_measurement with 'Benchmark completed in 330.7s' and 'Parsed benchmark result: 4.21 req/s' logged BEFORE the failure. Both inferencex_result.json are protocol-comparable: completed=192, num_prompts=192, total_output_tokens=196608, total_input_tokens=1572864, max_concurrency=64 -- identical protocol to the run that produced the 4319.4 incumbent.

So the discarded results were complete and comparable to the one that did get promoted.

Two things that would help, either separately or together:

1. When `benchmark_report.json` exists and parses, and the request counts match what was asked for, keep the measurement instead of discarding it on the wrapper's exit code. Flag it as degraded if that's safer than trusting it outright.
2. Retry the variant once on this error class. It's transient — the same script runs fine moments later.

Copying the InferenceX checkout off NFS to local disk stopped it completely (44 consecutive benchmarks clean afterwards, versus 15 failures before), so the trigger is environmental. But any site running with the checkout on NFS will silently lose measurements this way, and the loss is invisible unless you go read the stderr logs.

### Steps to reproduce

Not reliably reproducible on demand, since it depends on the NFS handle going stale. It showed up 15 times across ~3 hours of benchmarking with `INFERENCEX_PATH` on an NFS mount, and zero times in 44 benchmarks after moving the checkout to local disk.

### Logs

```shell
# tail of a failed variant's benchmark_stderr.log
+ local eval_exit=0
+ set +x
benchmarks/sglang_mi355x.sh: error reading input file: Stale file handle

# ...and what Hyperloom made of it
grid_runner: variant baseline_conc64 aborted: magpie_nonzero_after_valid_measurement (rc=1)
```

- Session `gpt-oss-120b_20260909T191958Z_484ecf65`, local run, sglang 0.5.18.dev20260825+g0c7ff19e3b, ROCm 7.2.4, MI355X, TP=1
- Hyperloom at `074382e57`
- Related: #1469 (found in the same session)

### Data handling

- [X] I confirmed the attached logs/artifacts contain no sensitive information

BaoYunkai:
Fixes #1478.

## The problem

A benchmark finishes, parses its result and writes `benchmark_report.json`. Then the bash wrapper exits non-zero for a reason that has nothing to do with the measurement, and the whole round is thrown away as `magpie_nonzero_after_valid_measurement`.

In the reported session this cost 2 of 8 concurrency-sweep points and all 4 variants of one optimization leg. The trigger was the InferenceX checkout sitting on NFS: bash reads the benchmark script off that checkout for the whole round, and a stale handle at the tail exits 2 after the benchmark is already done.

## Where it comes from

`run_grid` asks one question about the round: was the exit code zero. That was deliberate, added on 2026-08-24 to stop a server that died mid-round, but still emitted a parseable report, from becoming the number the session compares against. That protection is worth keeping.

The defect is that the exit code cannot distinguish those two cases. A server that died mid-round and a wrapper that failed on its way out both exit non-zero, so both land in the same branch.

The two cases differ in the evidence: a round that died early served fewer requests than it was asked for, and a round that finished served all of them. That is what the reporter verified on the discarded results, and it is the question the code was not asking.

## The fix

Ask it directly, from the run's own artifact.

InferenceX records what the client actually requested as `num_prompts`, and `extract_benchmark_measurement` already opens that artifact every round through `_merge_raw_result`. That value now lands on the measurement as `requested_requests`, and `served_complete_protocol` compares it against `completed_requests`, taking the measurement alone.

Deriving it from the artifact rather than from the env stack is what makes the fix reach every caller. `NUM_PROMPTS` is only written into `extra_envs` by `conc_sweep`; `explore`, `rebench` and `integrate_patch` let it come from the base YAML, so an env-based predicate would have stayed false for them and left the optimization leg in the issue unfixed. The env is also only what we asked for: without the `benchmark_lib.sh` patch the client serves `max_concurrency` instead, and the grid path never applies that patch.

Behaviour:

- A round that served at least the requests its own result records as requested keeps its measurement. It is recorded as succeeded, carries a `nonzero_rc_after_complete_protocol:<rc>` warning and the exit cause in `error`, and writes no abort marker.
- A round that served short still fails as `magpie_nonzero_after_valid_measurement` and still writes its abort marker. The original protection is unchanged.
- A round whose result records no requested count is unchanged. It did not get far enough to state its protocol, so it cannot be judged complete. Scriptable workloads drive their own iteration count and never record one.

`baseline.py` asks the same predicate instead of holding a second copy of the rule.

Two follow-on changes the above forces:

- The report read now settles however the process exited. It used to skip settling on `rc != 0` on the premise that such rounds are discarded anyway; that premise no longer holds, and the keep/discard verdict is taken from this one read.
- `preflight.py`'s network-filesystem warning no longer says a mount flap discards a completed measurement.

## Why not fix this upstream

Magpie sets `result.success` straight from the process return code while holding the parsed result, which is why the `reported_success` Hyperloom already has cannot answer this. Splitting "process exited non-zero" from "measurement incomplete" in Magpie would need no Hyperloom change at all, and would be the better place for it. This PR does not block that: `served_complete_protocol` reads the artifact, so it stays correct if Magpie later reports the distinction itself.

## Tests

Three regressions, written against the broken behaviour first:

- `test_run_grid_keeps_a_measurement_that_served_every_request` reproduces the report: 192 of 192 requests, exit code 2, stale-handle stderr. The variant carries no `NUM_PROMPTS` of its own, which is the shape every caller but `conc_sweep` uses.
- `test_run_grid_fails_a_measurement_that_served_short` pins the protection: 120 of 192 still fails.
- `test_run_grid_fails_when_the_run_recorded_no_request_count` pins the unpatched-checkout case.

`test_grid_runner_kill_and_branches_unit.py`, `test_grid_runner.py`, `test_grid_runner_behavior_lock.py`, `test_conc_sweep.py`, `test_benchmark_result.py`, `test_benchmark_result_branches_unit.py`: 420 passed, 4 skipped.

`ruff check .` and `ruff format --check .` both pass.


ZhengGong-amd:
Reproduced the gap first, then the shape I think this should take.

## 1. The fix does not fire for half of #1478

`requested_requests=variant.extra_envs.get("NUM_PROMPTS")` reads the *last* layer of the env
stack. `NUM_PROMPTS` normally comes from the base YAML's `benchmark.envs`, derived at
`_workload_envs.py:1695`; `_build_variant_yaml:462-477` is what merges the two.

| caller | `NUM_PROMPTS` in `variant.extra_envs` | fixed |
|---|---|---|
| `conc_sweep.py:452-462` | yes, written explicitly | yes |
| `explore.py` (optimization legs, default grid, LLM proposals) | no | no |
| `specialists/rebench.py:69` | no | no |
| `integrate_patch.py:5479` | no | no |

The issue reports two losses; "4 of 4 variants in one optimization leg" goes through
`explore.py`. On this branch, with `NUM_PROMPTS: 192` in the base YAML, a flag-only variant,
192/192 completed and rc=2:

```
STATUS: failed  ERROR_CLASS: magpie_nonzero_after_valid_measurement  WARNINGS: []
```

The new tests pass because `_write_base_yaml` carries only `TP/CONC/ISL/OSL` and `NUM_PROMPTS`
arrives via `GridVariant` — they cover exactly the one path that works.

## 2. The env is also not what the client requested

`NUM_PROMPTS` only reaches the client if InferenceX `benchmark_lib.sh` carries the
`num_prompts="${NUM_PROMPTS:-$max_concurrency}"` patch. `ensure_benchmark_lib_patched` has one
caller: `profile.py:1000`. The grid path never applies it. On an unpatched checkout the client
runs `max_concurrency` prompts, so `completed=CONC < NUM_PROMPTS` and the predicate is
permanently false — silently.

## 3. The shape that removes both, and the defenses with them

The requested count is in the artifact you already read. InferenceX writes it
(`benchmark_serving.py:853  result_json["num_prompts"]`), it is what the client actually
requested, and `extract_benchmark_measurement` opens that file every round via
`_merge_raw_result`.

```python
# extract_benchmark_measurement
measurement["requested_requests"] = first_int(raw.get("num_prompts"))

# served_complete_protocol(measurement) — no second argument

# run_grid
if rc != 0 and not served_complete_protocol(measurement):
    ...existing protection...
```

One owner (`benchmark_result.py` already owns measurement semantics), one derived source, no
cross-layer plumbing. §1 and §2 both disappear, all four callers are covered, and
`baseline.py:4216` can ask the same predicate instead of holding a second copy of the rule.

## 4. Defenses to delete

`served_complete_protocol:816-824`:

- `is_valid_measurement(result)` — the call site is already past
  `if not measurement.get("valid_measurement"): continue` (`_grid_runner.py:2002`), and that
  flag *is* `is_valid_measurement(measurement)` (`benchmark_result.py:683`, `:712`). Re-asking
  an answered question, and not cheap: three function-level imports inside.
- `not isinstance(result, dict)` — dead. Short-circuit means `is_valid_measurement` already
  returned False for a non-dict, and `measurement` is never None at the call site.
- `requested <= 0` — `NUM_PROMPTS` is derived by us (`max(conc*factor, conc)`), not caller input.
- `_is_scriptable_measurement` — the only one carrying meaning, and in the §3 shape it stops
  needing its own branch: no requested count, no complete protocol.

Control flow: `complete_protocol` is computed on the `rc == 0` path too, and one decision is
split across `if rc != 0 and X` / `if rc != 0 and not X`. That is `if rc != 0:` plus an
`if/else`.

## 5. Recording

The kept result carries no `error`, so the only trace of *why* the round exited non-zero is a
log line truncated to 200 chars (every other branch uses 2000) — the ledger shows "non-zero but
kept" without the cause. Same note as #1456 B6. Conversely, when the predicate is false for the
mechanical reasons in §1/§2, nothing is logged at all: the mechanism silently does not
participate.

## 6. Two questions worth answering in the description

- **Why not upstream?** `Magpie benchmarker.py:1321-1327` sets `result.success` straight from
  `process.returncode` while holding the parsed result — which is why the `reported_success`
  Hyperloom already has cannot help here. Splitting "process exited non-zero" from "measurement
  incomplete" in Magpie needs no Hyperloom change at all.
- **Is `rc != 0` meant to stay in this form?** It is a second, weaker validity rule layered on
  `is_valid_measurement`, answering "is this round comparable" through a proxy that cannot
  answer it — which is what made this bug possible. This PR adds an exception to the proxy; the
  next non-ESTALE cause will want another. Under §3, `rc` goes back to being evidence on the
  record rather than a veto, and `magpie_nonzero_after_valid_measurement` should be renamed for
  what it would then mean: served short.

## Smaller

- `preflight.py:2382-2388` still states that a mount flap is recorded as
  `magpie_nonzero_after_valid_measurement`.
- `_settled_measurement` gets `settle_seconds=0` when `rc != 0` (`_grid_runner.py:1990-1994`) on
  the premise that such rounds are discarded anyway. This PR overturns that premise but leaves
  both the settle and the comment at `:549-551` unchanged, so the keep/discard decision now
  rests on a single unsettled read.

BaoYunkai:
Thanks — §1 and §2 are both real, and I reproduced §1 before changing anything. Reworked to the shape in §3.

**§1, §2 — confirmed.** `grep NUM_PROMPTS` over `explore.py`, `specialists/rebench.py` and `integrate_patch.py` returns nothing; only `conc_sweep.py:425` writes it into `extra_envs`. The predicate was false for exactly the callers that carry the optimization leg the issue reports, and my tests passed because `_write_base_yaml` made `GridVariant` the only source. That is the half of #1478 the first shape did not fix.

**§3 — taken.** `extract_benchmark_measurement` now derives `requested_requests` from the artifact (`num_prompts`, both from the report's throughput block and from the raw InferenceX result in `_merge_raw_result`), and `served_complete_protocol(measurement)` takes no second argument. All four callers are covered, `baseline.py` asks the same predicate instead of holding a second copy, and the env stack is out of the path entirely, so §2 goes with it.

**§4 — deleted.** `is_valid_measurement`, the `isinstance` guard, `requested <= 0` and the `_is_scriptable_measurement` branch are all gone; a scriptable workload records no requested count, so it falls out of the same check. The two `if rc != 0 and X` branches are now one `if rc != 0:` with an `if/else`, and nothing is computed on the `rc == 0` path.

**§5 — fixed both directions.** The kept result carries the same 2000-char `error` every other branch uses, and the log line is no longer truncated to 200. The failing branch now logs served-vs-requested, so a round rejected for serving short says so with numbers rather than just `rc=N`.

**Smaller — both fixed.** `preflight.py` no longer claims a flap discards a completed measurement. `_settled_measurement` now settles however the process exited, with the comment rewritten to say why: the premise that non-zero rounds are discarded anyway is what this PR overturns, and the verdict is taken from that one read. Three tests that relied on the old `settle_seconds=0` path now pin `REPORT_SETTLE_SECONDS` to 0 themselves, matching what `test_rc0_invalid_measurement_continues_despite_keep_going_false` already did.

**§6 — in the description.** Added a "Why not fix this upstream" section making the Magpie point and noting that reading the artifact keeps this correct if Magpie later reports the distinction itself. On the second question: I agree `rc != 0` is a proxy that cannot answer "is this round comparable", but renaming `magpie_nonzero_after_valid_measurement` to what it now means touches recorded error classes in existing sessions, so I would rather do that separately than fold it into a bug fix.

New test for the case the rework opens up: `test_run_grid_fails_when_the_run_recorded_no_request_count`, for the unpatched checkout where no requested count is recorded. The keep test now uses a flag-only variant with no `NUM_PROMPTS`, so it covers the path that was broken rather than the one that already worked.

Two commits: the merge that resolves the CHANGELOG deletion, then the rework.

ZhengGong-amd (approved):
Approved

PR #1585, opened 2026-09-21
feat(sbd): complete the timeline events and pin their wire

haishuok0525:
- Description: what and why

  The timeline events landed with gaps that made a bad run readable as a good
  one. This closes them, and pins the wire they are read over.

  - **KERNEL verdict is derived, not narrated.** The visit took its verdict
    from the phase's own exit story, so a visit that kept nothing still closed
    clean. It is now derived from the instruments that ruled on each candidate,
    and a measured win outranks a survived fault rather than being buried by it.
  - **Sub-steps are attributed, not orphaned.** Baseline rounds and roofline
    probes opened standalone events that orphaned their measurement; they can
    now be routed into the parent event through `sbd_event_id`.
  - **Facts that were never recorded now are.** Warm replay says what became of
    each apply item; baseline carries the anchoring eval behind its quality
    claim; FRAMEWORK records its configuration-arm runs and the faults it
    survived; prelude marks dropped arms.
  - **Phase-spanning events get their exceptions named.** A KERNEL visit, a
    FRAMEWORK entry and the enablement lane outlive any executor frame, so an
    exception reached their event nowhere and left it closing clean for a phase
    that had blown up. The coordinator and the lane now name it.
  - **The wire is pinned.** The `V6Framework*` / `V6Kernel*` / `V6BaselineExt` /
    `V6WarmReplayExt` / `V6RooflineExt` TypedDicts plus a schema-contract test
    make a renamed key fail a test instead of silently emptying a report field.

  Also closes the last of the recorder boundary holes from #1578: several
  `finish` / `_close` helpers still read the spool unguarded while their callers
  (baseline executor, warm-replay settle, conc-sweep, the Never-raises stack
  ledger) assumed close could not fail, so an `OSError` at assembly time escaped
  into phase teardown or an action return. Those closes now catch
  `RECORDING_ERRORS` and leave finalize to recover the event as `interrupted` --
  the contract `phase_event` and `enablement_event` already keep.

  Follow-up to #1578 (same recorder boundary work); no tracker issue.

  Added `test_sbd_v6_ext_schema_contract.py` (pins every event `ext` against its
  TypedDict) and per-recorder regressions that an unreadable spool at `finish`
  does not raise (baseline, warm replay, stack). Updated the existing v6
  timeline/wiring suites for the new fields and the derived verdict.

  ```
  pytest src/hyperloom/inference_optimizer/tests/ -k "sbd or timeline or schema or breakdown"   # 878 passed, 6 skipped
  pytest src/hyperloom/inference_optimizer/tests/test_sbd_v6_{framework_wiring,enablement_timeline,framework_timeline,enablement_wiring,specialist_round_recording,kernel_timeline}.py   # 206 passed
  pytest src/hyperloom/orchestrator/actions/executors/tests/test_{integrate_patch_helpers_unit,warm_patch_apply}.py   # 127 passed
  ```

  The full local suite is not included: it contains tests slow enough that the
  run did not finish locally, so it is left to CI.

- `CHANGELOG.md` entry under `[Unreleased]`: yes/no (if no, why the change is unobservable):

  Yes -- one entry under `### Changed` for what the events now name, and one
  under `### Fixed` for the close paths that could raise into the loop.

  No. The event `ext` shapes gain fields; nothing an existing reader depends on
  is renamed or removed.

  Yes, with one caveat: the close-path guards are strictly the remainder of
  #1578's boundary narrowing, kept here because the new close paths this PR adds
  would otherwise land with the same hole.

  No, entirely in-repo.


haishuok0525:
### Folded in the guard cleanup that was #1586

Both branches rewrote the same close and re-publish paths, so #1586 lands here
instead of leaving one of the two to rebase onto the other. It removes the
guards in this subpackage that could only ever catch a recorder bug, measured
against `EventSink`'s own stated contract -- *"a recorder that raises anything
else is broken, and is left to say so"*:

| Site | What the outer clause could catch |
|---|---|
| `framework_event.record_proposal_review_evidence` | only a bug; `.record()` is guarded |
| `kernel_event.record_integrate_verdict` | only a bug; `has_row`/`record` guarded |
| `kernel_event.record_trace_analyze_request` | same, plus a one-line sink write |
| `baseline_event.record_action_decision` | same |
| `event_timeline._opened_sequence` | hand-rolled the catch `recorded_section` exists to remove |
| `framework_event._resume_gate_ordinals` | read side; also read 7 sections to use 1 |
| `instrument.snapshot_state_sections` | unreachable -- `recorder_for` joins a path and fills a dataclass |
| `phase_event` (x2) | `isinstance` on rows `event_parts` had already filtered |
| `event_finalize._finalize_type` (x2) | narrowed to `RECORDING_ERRORS` rather than deleted |

**One guard was genuinely missing.** `_republish_closed_event` had five callers:
three with none at all, which would take an unreadable spool straight into the
phase that recorded the row, and two wrapped in `except Exception`. The helper
now owns that boundary itself. Together with the eight close paths already in
this PR, every never-raises contract in the subpackage is now kept in one place
per path rather than at each call site.

Broad catches in `recorder/` go from 40 to 31. The rest are real boundaries
(subprocess probes in `tool_versions.py`, the `trace.py` and `note_failure` meta
layers, report rendering) and are deliberately left.

#### Conflicts resolved

Three, all where this branch had already rewritten the lines the cleanup
wrapped:

- `kernel_event._republish_closed_event` keeps `status=derived`. The cleanup had
  only added a guard around the old line, which still read `closed_status` off
  the header -- and not trusting the phase's own exit story is the point of this
  PR.
- `baseline_event._republish_closed_event` keeps `_event_header`, which replaced
  the hand-rolled first-row lookup.
- `CHANGELOG.md` keeps all three entries, including the inline-MCP one this
  branch predates.

#### Verification

```
pytest src/hyperloom/inference_optimizer/tests/ src/hyperloom/orchestrator/actions/executors/tests/ \
  -k "sbd or timeline or schema or breakdown or geak or integrate or warm or baseline or roofline or conc or stack or kernel"
# 4552 passed, 10 skipped
```

One unrelated failure locally, `test_trace_analyze_handler_dry_run_returns_structured_result`:
it spawns a subprocess that does `import hyperloom`, and this machine has no
editable install, so the import fails before any recorder code runs. Not
reachable from this diff.

The PR now carries two concerns rather than one -- the event completions and the
guard removal. Reviewing the second commit (`e1ea60c2`) on its own gives the
cleanup in isolation if that reads more easily.

xiaofei-zheng:
## What this PR does

Closes gaps in the SBD v6 timeline events: the KERNEL verdict is now derived from the instrument rows instead of the phase's self-reported exit, sub-steps can be attributed to a parent event via `sbd_event_id`, warm replay / baseline / FRAMEWORK / prelude record facts they previously dropped, and the phase-spanning events (KERNEL visit, FRAMEWORK entry, enablement lane) get coordinator-side exceptions named on them. The `ext` shapes are pinned by `V6*` TypedDicts plus a schema-contract test. It also finishes #1578's recorder boundary work: the remaining `finish` / `_close` / re-publish paths now catch `RECORDING_ERRORS` instead of letting an `OSError` escape into phase teardown.

## Blocking

### 1. A KERNEL visit that crashed can now close as `succeeded`

`kernel_event.py:1838` `finish_failed` (and `finish_crashed` through it) dropped the `status="failed"` it used to pass to `finish()`; the status now comes from `_derive_verdict`, which returns `VERDICT_IMPROVED` at `kernel_event.py:2081` **before** it ever looks at `failure`. So a visit that adopted a candidate with a positive gain and then raised closes `succeeded` / `improved`.

The "a measured win outranks a recorded fault" rationale in `_derive_verdict` is about `record_fault` — a fault the visit survived. `finish_failed` is the opposite case: the phase naming the stage it could not get past. Its own docstring still says "Close the event as failed" and "Outranks any fault recorded earlier", both of which are now false. The sibling FRAMEWORK recorder gets this right (`framework_event.py:933`, `if _as_dict(header.get("failure")): return "failed"`).

Reproduced on `9f0ec12` (passes on the merge base, which forced `status="failed"`):

```python
def test_crash_after_a_measured_win_still_closes_failed(tmp_path):
    r = _forge_recorder()
    r.enter_stage("forge_fusion")
    r.record_kernel_rewrite(run_id="attempt-7", kernel_id="k001", status="success",
                            micro_decision="keep", speedup=1.4)
    r.finish_crashed(RuntimeError("boom"))
    event = [e for e in read_timeline_events(tmp_path) if e.get("type") == "kernel"][0]
    assert event["status"] == "failed"   # got 'succeeded', verdict 'improved'
```

This is the failure mode the PR sets out to remove, in the other direction: a crashed phase readable as a good one. `_republish_closed_event` (`kernel_event.py:664`) now re-derives unconditionally too, so a late fragment can flip an already-exported `failed` event to `succeeded`.

### 2. Any coordinator-side exception marks the enablement lane failed

`coordinator.py:1473` calls `enablement_event.record_fault(...)` unconditionally for every exception `_record_coordinator_exception` sees — `reactor:*`, `advance_phase`, dispatcher pumps, `intent_router`, `phase_entered`. For KERNEL and FRAMEWORK the fault at least lands on an event that was open for that stage; the enablement lane spans the whole session and is not the thing that raised.

It is not just an extra row: `enablement_event.py:749-750` forces `status = STATUS_FAILED` whenever a failure row exists, overriding the lane's real terminal.

```python
def test_an_unrelated_coordinator_fault_does_not_fail_a_lane_that_succeeded(tmp_path):
    enablement_event.record_trigger(origin=enablement_event.ORIGIN_EVAL, mode="all",
                                    kind="accuracy_below_floor")
    enablement_event.record_fault(stage="reactor:optimizer", exc=RuntimeError("boom"))
    enablement_event.finish(outcome=enablement_event.OUTCOME_SUCCEEDED,
                            reason="revalidation promoted")
    event = [e for e in read_timeline_events(tmp_path) if e.get("type") == "enablement"][0]
    assert event["status"] == "succeeded"   # got 'failed', stage 'reactor:optimizer'
```

One reactor turn raising anywhere in the session is enough to report a working enablement lane as broken, with an `error_class` from a subsystem it never touched. The same call also stamps that exception on whatever KERNEL visit happens to be open.

## Checked

Kernel/framework/enablement/baseline status derivation, the eight guarded close paths and the re-publish helper against `git merge-base origin/main HEAD` (`bdbe6c1`), the `V6*` TypedDicts against `assemble_*_ext`, and the CHANGELOG entries (present under `Changed` and `Fixed`, matching the diff). Ran the two repros above plus `test_sbd_v6_kernel_timeline.py` and `test_sbd_v6_enablement_timeline.py`.

xiaofei-zheng (approved):
Re-reviewed at `02062f8`. Both blocking findings are fixed; no blocking issues remain. LGTM.

**Conclusion**

`bc160f8` separates a KERNEL visit's status from its verdict: `finish(failed=True)` records `closed_status` on the close row, and `assemble_kernel_ext` takes it over `_STATUS_BY_VERDICT`, so a visit that delivered a measured win and then raised closes `failed` while the verdict keeps the win. Because the marker is recorded rather than passed, `_republish_closed_event` reaches the same status on a late fragment. A fault recorded through `record_fault` is unaffected — that visit survived and is still judged on its instruments, which is the distinction the earlier code collapsed.

The enablement lane no longer answers for faults it did not raise: `_fault_open_phase_event` speaks only for KERNEL and FRAMEWORK, whose recorders are absent unless their phase is running, and `EnablementLane._pump_enablement_safely` records its own pump exceptions. `enablement_pump:*` is the only place the lane's exceptions were being swallowed, so nothing is lost by the narrowing.

**Checked**

- Re-ran both repros from the previous comment against `02062f8`: the crashed-visit case now closes `failed` (verdict still `improved`), and a succeeded lane with an unrelated `reactor:*` exception now closes `succeeded`. Added a third case confirming a *survived* fault plus a measured win still closes `succeeded`, i.e. the fix did not overshoot into the `record_fault` path.
- `test_sbd_v6_kernel_timeline.py` and `test_sbd_v6_enablement_timeline.py`: 116 passed. The one failure (`test_the_same_patch_is_named_the_same_way_everywhere_in_the_recipe`) is a Windows path-separator artifact of my local box, not this diff; `test_sbd_v6_enablement_wiring.py` cannot be collected on Windows (`fcntl`). Both are covered by CI.
- Verified `_pump_enablement_safely` is the only enablement exception sink, and that `active_kernel_recorder()` / `_framework_timeline()` are phase-scoped.
- Merge of `origin/main` resolved one conflict (`CHANGELOG.md`, retired by #1589) by taking the deletion; no other resolution in the merge.

PR #1586, opened 2026-09-21
refactor(sbd): drop the guards that could only catch a recorder bug

haishuok0525:
- Description: what and why

  `EventSink` states the write side's contract in its own docstring: a row that
  cannot be written is dropped with a warning, and *"a recorder that raises
  anything else is broken, and is left to say so."* Several callers wrapped
  those already-guarded writes in `except Exception` anyway. The only
  exceptions that layer could ever see were the defects the contract says to
  surface -- and it logged them at warning level and carried on.

  This deletes those clauses. Nothing they reported is lost: the layer beneath
  them already reports it through `note_failure`.

  | Site | What the outer clause could catch |
  |---|---|
  | `framework_event.record_proposal_review_evidence` | only a bug; `.record()` is guarded |
  | `kernel_event.record_integrate_verdict` | only a bug; `has_row`/`record` guarded |
  | `kernel_event.record_trace_analyze_request` | same, plus a one-line sink write |
  | `baseline_event.record_action_decision` | same |
  | `event_timeline._opened_sequence` | hand-rolled the catch `recorded_section` exists to remove |
  | `framework_event._resume_gate_ordinals` | read side; also read 7 sections to use 1 |
  | `instrument.snapshot_state_sections` | unreachable -- `recorder_for` joins a path and fills a dataclass |
  | `phase_event` (x2) | `isinstance` on rows `event_parts` had already filtered |
  | `event_finalize._finalize_type` (x2) | narrowed to `RECORDING_ERRORS` rather than deleted |

  **One guard was genuinely missing.** `_republish_closed_event` had five
  callers: three with no guard at all, which would take an unreadable spool
  straight into the phase that recorded the row, and two wrapped too widely.
  The helper now owns that boundary itself and catches `RECORDING_ERRORS` --
  the contract the rest of the subpackage already keeps -- so no caller needs
  its own and all five are treated alike.

  Context: a code-health scan flagged this subpackage's broad-catch ratio as
  well above the repo baseline. Two of the three files it named had already
  been fixed by #1578; this covers the write/read paths that narrowing missed.
  Broad catches in `recorder/` go from 40 to 31. The remainder are real
  boundaries (subprocess probes in `tool_versions.py`, the `trace.py` and
  `note_failure` meta layers, report rendering) and are deliberately left.

  Follows #1578; no tracker issue.

  No new tests: this deletes unreachable handlers and moves one guard down a
  level, so the behaviour under test is unchanged. The re-publish contract is
  already exercised by the existing late-settlement and adoption-ledger suites.

  ```
  pytest src/hyperloom/inference_optimizer/tests/ -k \"sbd or timeline or schema or breakdown or geak\"
  # 1387 passed, 6 skipped
  ```

  The full local suite contains tests slow enough that it does not finish on
  this machine, so it is left to CI.

- \`CHANGELOG.md\` entry under \`[Unreleased]\`: yes/no (if no, why the change is unobservable):

  Yes, under \`### Fixed\` -- for the three unguarded re-publish callers, which
  is the one observable behaviour change. The deletions themselves are not
  observable: they could only ever fire on a defect.

  No.

  Yes -- remove guards that cannot catch anything but a bug, and put the one
  guard that was missing where it belongs.

  No, entirely in-repo.


haishuok0525:
Folded into #1585 instead: both branches rewrote the same close and re-publish paths, so landing them together avoids one rebasing onto the other. The commit is preserved as `e1ea60c2` there, reviewable on its own. See https://github.com/AMD-AGI/Hyperloom/pull/1585#issuecomment-5756403577 for the conflict resolutions.

PR #1587, opened 2026-09-21
[FEAT] dissolve Explore Phase, unify the attempt ledger, judge dryness per lever

ZhengGong-amd:
# refactor(orchestrator): dissolve ExplorePhase, unify the attempt ledger, judge dryness per lever

  `ExplorePhase` had grown to 30 methods, only 2 of which concerned the explore
  action, and the phase machine judged search exhaustion from two per-arm plateau
  predicates reading two differently-shaped ledgers — so they could disagree with
  each other and with what the model was told.

  - **`phases/explore.py` is deleted.** Its methods move to modules named after
    their actual concern — `phases/macro_cycle.py`, `loop/cycle_memory.py`,
    `specialists/dispatch.py`, `state/gaps.py` — and the deliverable-routing ones
    move into `FrameworkPhase`. The `local_explore` pseudo-candidate machinery
    goes with it; the pump dispatches the specialist directly.
  - **One attempt ledger.** `SharedState.attempts` records every resolved attempt,
    lever-annotated, one writer per lever. `explore_search` no longer round-trips
    through the executor: it reports the round it benched and the state layer
    accumulates.
  - **Per-lever dryness.** `per_lever_dryness` replaces the
    `source_arm_plateaued` / `compute_plateau_explore` pair; the config lever is
    judged per benched round, the patch levers per settled candidate, and
    `exit_normal_optimize` and the model advisory read the same predicate.
  - **One outcome vocabulary.** `TS_*` constants and `is_kept()` live in
    `_grid_base.py`, so the grid's `KEEP` and `integrate_patch`'s `kept` cannot
    drift apart; both recorders derive `adopted` from it.
  - **Two coupled-deliverable fixes.** An enablement round that writes a patch no
    longer loses the specialist's `setup_commands`, and the companion-lever rule
    #1414 settled is the only one that decides whether a lever rides with a patch.

- **Linked issue(s)**: closes audit finding F0144 (live-bug half); completes F0130.

- **Tests**: updated. `pytest src/hyperloom/inference_optimizer/tests` over the
  affected modules (plateau, phase exit, explore executor, framework authoring and
  dispatch-failure, switch manifest, loop walkthrough, longrun 0/2/3, shared-state
  persistence and evolution). Three tests were replaced, contracts carried across:
  the plateau suites now seed `state.attempts` through `record_config_attempt`
  rather than the retired per-arm fields; the dispatch-failure suite pins that a
  failed dispatch leaves *no* ledger row instead of pinning a `dispatch_failed`
  outcome value nothing constructs; and the ledger test that asserted a row for a
  dispatch without the authoring flag is gone with that behaviour — its two live
  contracts (a lever derived from the deliverable, a write that does not need an
  open timeline recorder) are pinned by the ledger test beside it and by the new
  settled-candidate test, which runs on a stub with no recorder.

- **Size/complexity triggers crossed**: yes, and the shape is not yet right.
  `FrameworkPhase` goes from 47 methods / 2176 lines to 50 / 2732 by absorbing the
  deliverable-routing methods. Those are a translator — `specialist_done` payload
  in, proposal payload out, no phase state read or written — so their home is
  beside `specialists/dispatch.py`, not inside a phase handler. Splitting that out
  here would mean moving the same methods twice in one review; it is the next PR,
  and the duplicated enablement-forwarding block inside it (inherited, not
  introduced here) is the extraction that split should make.

- **If this simplifies or refactors: which mechanism was removed**
  - The second plateau predicate, and with it `compute_plateau_explore`,
    `source_arm_plateaued`, `framework_agent_consecutive_no_keep` and the
    `cycle_boundary` marker that existed only to stop its streak walk.
  - The executor as a co-writer of `explore_search`, and `config_arm_round_count`,
    which carried a second definition of what a round is.
  - The second writer of a patch attempt: the ledger write moves behind the gate
    the authored-outcome bridge already applies, and the dispatcher seam goes.
  - `phases/explore.py` itself (1894 lines), plus the `local_explore`
    pseudo-candidate ids and the dead `gv.framework_lever*` attributes.

  Contracts preserved: the dryness evidence keys (`source_arm_plateaued`,
  `config_arm_plateaued`) still appear for the timeline recorder; legacy
  `plateau_overrides` still resolve; a dispatch failure is still not evidence a
  lever is exhausted.

- **Observable effect**

  1. **The configuration arm now declines after five benched rounds without a win,
     not five variants.** The streak floor was calibrated when the unit was a
     round; moving the judgment onto a per-variant ledger meant one eight-variant
     grid that reverted everything crossed it alone, so a run could leave
     `FRAMEWORK_AGENT` after a single unlucky grid and switch bottleneck or close
     with most of the search unexplored. The streak also walked backwards from the
     newest row, so a grid whose KEEP was not its last variant could be read as dry
     immediately after banking a win. Operators tuning `plateau_overrides` should
     read `explore_empty_streak` as rounds; the name and default are unchanged.

  2. **A resumed or long-running session no longer forgets which variants it has
     already tried.** The round's writes were replacing `tested`, `rejected` and
     `name_index` instead of accumulating onto them, so the durable history never
     went back further than the last grid: the dedup gate re-offered variants
     measured two rounds earlier, the acceptance-threshold advisory had almost no
     prior results to weigh, and the warm recipe's `what_failed` rows — injected at
     PRELUDE precisely so the gate denies re-tests — were erased by the session's
     first round. Relatedly `explore_search.cursor` counted the round's variants
     rather than the rounds, which pinned every round id at `explore-009` for an
     eight-variant grid, froze the stalled-domain specialist's idempotency key
     after its first dispatch, and stopped the research scout's every-K-rounds
     re-dispatch from ever coming due.

  3. **`FRAMEWORK_AGENT` no longer leaves on apply-conflict noise.** A candidate
     burning the three-deep apply-retry ladder used to put four rows in front of
     the patch lever's five-deep streak floor, so two candidates that never reached
     a bench could end the phase.

  4. **An enablement round keeps the flags an earlier round established.** A patch
     shipping with a companion lever was launched from whatever the latest
     deliverable restated, dropping established `accepted_config` args and envs and
     sending the round back to a wall an earlier round had already cleared.

- **Breaking changes**: no.

- **PR addresses single concern**: one cutover — the phase decomposition and the
  ledger the dryness judgment reads — landed in reviewable steps. Two unrelated
  changes ride along and are called out rather than hidden: the origin-independent
  grid-variant guards in `integrate_patch.py`, and the removal of the dead
  `resume_reverify_best` revalidation source.


xiaofei-zheng:
Reviewed the full diff against `git merge-base origin/main feat/zgong/code-reunion` (f2f603c6c). The method moves out of `phases/explore.py` into `macro_cycle.py` / `cycle_memory.py` / `specialists/dispatch.py` / `state/gaps.py` / `FrameworkPhase` are faithful (body-for-body identical except the relative-import fixups and the now-unneeded `cycle_boundary` marker), and the `Coordinator` delegation map covers every moved method. Two blocking issues, both in the new behaviour rather than the moves.

**1. Non-measured `integrate_patch` outcomes now drive the patch-lever plateau, and retries are counted individually**

`loop/dispatcher.py:1347-1365` writes an `attempts` row for *every* `integrate_patch` result that carries a `status`, and `phases/machine_state.py:_patch_lever_dry` counts those rows with `_trailing_no_keep`. `_trailing_no_keep` documents the opposite at `machine_state.py:1621` ("Only measured attempts reach the ledger, so a specialist that never ran leaves nothing here to plateau on"), but `apply_failed`, `no_patches`, `skipped` and the new `filtered_before_bench` skip all reach it.

Two changes relative to the merge base, which both loosen the exit rule:

* `framework.py:_record_framework_agent_authored_outcome` returns early for `apply_failed` results carrying a `lane` (`perf_framework` / `perf_explore`) precisely so the retry ladder is not read as exhaustion. The new ledger write has no such guard.
* The progress ledger is deduped per candidate (`_stamp_framework_progress` keys on `candidate_id`, and the authored-outcome bridge replaces a candidate's prior row); the attempts ledger is one row per task, so a candidate that burns the `_AUTHORED_LANE_MAX_ATTEMPTS=3` apply-retry ladder contributes 4 rows instead of 1.

Repro — two upstream candidates, each exhausting the apply-failure ladder, nothing ever benched:

```
PR:   8 rows -> _patch_lever_dry  = (True,  {'source_consecutive_no_keep': 8, 'source_threshold': 5, ...})
base: 2 rows -> source_arm_plateaued = (False, {'source_consecutive_no_keep': 2, 'source_threshold': 5, ...})
```

With the config lever also dry, `exit_normal_optimize` returns `optimize_no_more_leverage` and FRAMEWORK_AGENT leaves on pure apply-conflict noise. Either filter non-measured statuses at the dispatcher seam (or in `_trailing_no_keep`) the way `writeback.py` filters `_NON_ATTEMPT_OUTCOMES` on the config side, or drop the claim in the docstring and recalibrate the threshold.

**2. The coupled-deliverable block overwrites the ENABLEMENT config inheritance it sits under**

`phases/framework.py:2816-2820`, new in this PR:

```python
config_levers = _framework_config_levers_from_done(done_payload)
if config_levers:
    integrate_params["extra_server_args"] = str(config_levers.get("extra_server_args") or "")
    integrate_params["extra_envs"] = dict(config_levers.get("extra_envs") or {})
```

Plain assignment, landing after `framework.py:2795-2812`, where an enablement round merges `shared_state.enablement.accepted_config` into `round_args` / `round_envs` — the block whose own comment says an omitted flag "sends the round back to the wall an earlier round already cleared".

The two calls do not agree: the block above uses `levers_ride_with_patches=True` (all proposals in an enablement round), the new one uses the default `False`, which with `patches_written` non-empty filters down to `atomic is True` and returns only the first usable proposal. So when an enablement round ships a patch plus an `atomic: true` lever and `accepted_config` is non-empty, the established args/envs are dropped and a multi-lever companion set is narrowed to one proposal. Merging into `integrate_params` rather than assigning over it would fix both.

Also checked and found clean: `apply_explore_search_update` accumulation (`tested` / `rejected` / `name_index`) and the cursor becoming a round ordinal against its three readers (`proposals.py:453`, `phases/internal.py:135`, `specialists/dispatch.py:538`); `is_kept()` against both spellings (`VERDICT_KEEP` and integrate_patch's `"kept"`); no dangling references to the deleted `compute_plateau_explore` / `source_arm_plateaued` / `framework_agent_consecutive_no_keep` / `_next_local_explore_candidate_id`; the `resume_reverify_best` removal in `writeback.py` is dead code on the base branch; the new `_grid_variant_filter` guards in `integrate_patch.py` cannot index an empty dropped list. `pytest test_phase_state_plateau.py test_optimize_phase_exit.py test_explore_executor.py` — 147 passed, 1 failure is a Windows-only `import fcntl` in `recipe_kb/local_store.py`, unrelated to this PR.

xiaofei-zheng:
Re-reviewed at `db58708`. Both blocking findings from the previous round are fixed; one blocking issue remains, in the description.

**Verified fixed**

1. Non-measured `integrate_patch` outcomes no longer drive the patch-lever plateau. `db58708` deletes the unconditional `record_patch_attempt` seam in `dispatcher.py` and moves the write into `framework.py:2166`, behind `_stamp_framework_progress`'s `if not recorded: return`. I checked every patch dispatch path sets `framework_agent_authoring` (`framework.py:728`, `:1318`, `:1589`, `:2834`, `:2995`, `enablement/params.py:308`), so the bridge is the only entry, and it early-returns on `apply_failed` in the `perf_framework` / `perf_explore` lanes. The repro that previously gave `(True, source_consecutive_no_keep=8)` on the branch vs `(False, 2)` on the merge base now matches the base.
2. The coupled-deliverable block that overwrote the ENABLEMENT `accepted_config` inheritance is gone in `f7883d9`. `_maybe_autosubmit_specialist_patches` now differs from `merge-base` only by the try/except simplification and the `setup_commands` merge.

**Blocking: the description no longer matches the diff**

`f7883d9` and `db58708` changed what the PR does, but the body is unchanged:

- *"**Coupled deliverables bench together.** A specialist that writes a patch *and* proposes a flag no longer loses the flag: config levers merge into the same `integrate_patch` proposal"* — `f7883d9` deleted that code. On the current head, a non-ENABLEMENT round where the specialist writes a patch and proposes a non-atomic flag still loses the flag: `_maybe_autosubmit_specialist_patches` no longer forwards config levers, and `_maybe_autosubmit_framework_config` returns early at `framework.py:2938-2940` whenever `patches_written` is non-empty. Only the `setup_commands` half of that bullet survives.
- *"**One attempt ledger.** `SharedState.attempts` records every measured attempt ... with writeback as the single writer."* — there are two writers on the current head: `writeback.py:1560` (config lever) and `framework.py:2166` (patch lever). And by design after `db58708` the patch lever records only framework-authored outcomes, not every measured attempt.
- The "Review follow-ups, as their own commits:" paragraph lists three follow-ups and omits `f7883d9` and `db58708`, which are the two that changed behaviour relative to what a reviewer read first.

Please update the body to the current diff. No other blocking issues.

**What I checked**: the `phases/explore.py` split is a faithful move — 27/32 extracted function bodies are byte-identical to `merge-base`, the 5 that differ do so only by relative-import fixups, `Task` → `"Task"` quoting, and the deliberate `cycle_boundary` removal; the `Coordinator._COLLABORATORS` delegation map covers every moved method and nothing references the deleted symbols. `per_lever_dryness` / `_fold_config_rounds` / `_trailing_no_keep` exercised directly against both branch and merge-base ledgers. `apply_explore_search_update`'s accumulate semantics checked for `tested` / `rejected` / `name_index` / `cursor`. Ran `pytest src/hyperloom/inference_optimizer/tests` on the branch: 147 passed, 1 failed — `test_actual_explore_axis_rejection_cannot_be_revived_by_geak_fallback`, from `import fcntl` in `knowledge/recipe_kb/local_store.py` on Windows, environmental and unrelated to this PR.

xiaofei-zheng (approved):
Approving. Both blocking findings from the earlier rounds are verified fixed at `db58708`. The remaining item is the description/diff mismatch noted above (the deleted coupled-deliverable path, the "writeback as the single writer" claim, and the missing follow-up commits) — that is being handled separately, so it does not block the code.

PR #1589, opened 2026-09-21
chore: retire CHANGELOG.md, and make every README link absolute

lishuoshuo-amd:
- Description: what and why

**`CHANGELOG.md` is retired in favour of GitHub Releases.** For the tagged
sections the file duplicated release bodies GitHub already publishes, while its
authoring gate made every PR maintain both. The project URL, the upgrade guide
and the release notes now point at Releases, and the gate moves to the PR
description that the release cut already aggregates -- the rule keeps its
force, it changes carrier.

The `[Unreleased]` section is the part that was not duplicated anywhere: 631
lines and 39 entries that no release contains because they have not shipped
yet, including the ENABLEMENT phase and several removed flags. They are not
recoverable from merged PR descriptions either, since the retired PR template
asked whether a changelog entry existed and never for its text. Those entries
move verbatim into `docs/release-notes.md` under an `Unreleased` heading, so an
operator tracking `main` between tags keeps the upgrade guidance.

Two retired spellings also lived only in the changelog: the `kernel-agents`
console script, and `FORGE_DISABLE_COMPILED_FELLOWS`, which is forwarded then
ignored and so silently re-enables compiled kernel backends for an operator who
had switched them off. Both are now named in the release notes, and the rename
guard's exemptions move there with them, narrowed from `(?i)fellow` to the one
literal that needs exempting.

**The README is made absolute.** PyPI uses it as the project description but
does not resolve repository-relative paths, so on
https://pypi.org/project/hyperloom-inference-optimizer/ the architecture image
never rendered and all 23 in-repo links were dead. One was dead on GitHub too:
the phase table pointed at a standalone `AMD-AGI/KernelForge` repository, which
404s now that KernelForge is vendored in-tree, so it points at
`docs/kernelforge/what-is-kernelforge.md`.

Known and accepted: the published v1.1.1 and v1.1.0 release bodies each link to
`blob/main/CHANGELOG.md` and will 404 after merge.

The commit list carries a revert and a reapply of the retirement commit, from
working through the `[Unreleased]` question above. The net diff is what matters
and is listed under Files changed.

None.

No new tests; the existing rename guard covered this and was used as the gate.
`pytest src/kernelforge/tests/test_rename_completeness.py` fails on
`test_every_allowlist_entry_still_exempts_something` with only the file deleted,
naming exactly the three dead exemptions, and fails again on
`test_no_stray_kernel_agents_references` / `test_no_stray_fellow_references`
once the spellings are inlined into the release notes; 6 passed after the
exemptions are retargeted. `ruff check` clean. Every URL in the README returns
200 under `curl -sIL -w '%{http_code}'`, including the image and the new
KernelForge target. `docs/index.rst` does not pull the README into the Sphinx
build, so absolutising its links does not turn a docs-site cross-reference into
an external one.

- Size/complexity triggers crossed: split, or why the shape is right

The 2315-line deletion is the changelog, and 631 of the 704 insertions are its
`[Unreleased]` section landing in the release notes. The remaining ~73 lines
span 10 files and do not stand alone: the exemption move is only correct
because the release notes inherited the content, and the gate rewrite is only
correct because the file it pointed at is gone.

- If this simplifies or refactors: which mechanism was removed, which contract was preserved, and the tests that show it

Removed: a hand-maintained changelog and the per-PR obligation to update it.
Preserved: every observable change is still stated in writing before it ships,
and every pending migration note and retired spelling stays reachable. The
rename guard shows the second half -- it fails if either spelling loses its
home.

- Observable effect: what an operator will now see, and what the old behaviour cost them (if none, why the change is unobservable)

The PyPI project page renders the architecture diagram and its links resolve;
a reader previously got a broken image and dead links. `CHANGELOG.md` no longer
exists in the tree, the PyPI "Changelog" sidebar link goes to Releases, and the
pending migration notes are on the release-notes page instead. A contributor
fills an "Observable effect" field in the PR template rather than adding a
changelog entry.

No runtime or interface change. Documentation, packaging metadata and the
contribution process only.

- PR addresses single concern: yes/no (details if no)

No -- two, kept together at the requester's direction: the changelog retirement
and the README link fix. They are separate commits and can be split.

No.


xiaofei-zheng (approved):
No blocking issues. LGTM.

Checked:
- `git grep -i changelog` over the PR tree: no reference to `CHANGELOG.md` survives outside `src/kernelforge/data/` (those point at ROCm/composable_kernel, unrelated). No workflow, packaging manifest or test depended on the file, so deleting it does not orphan a gate.
- Rename guard: `pytest src/kernelforge/tests/test_rename_completeness.py` -> 6 passed. Confirmed the retargeted exemptions actually match: `docs/release-notes.md` carries `kernel-agents` (lines 685, 751) and `FORGE_DISABLE_COMPILED_FELLOWS` (line 845), so `test_every_allowlist_entry_still_exempts_something` is not silently passing on dead entries. Dropping the `_COLLAPSE_ALLOWED` `CHANGELOG.md` entry is safe: the only `forge_llm` spelling in the release notes is `resolve_forge_llm_model`, which `\bforge_llm\b` does not match.
- README: every link and the architecture image are now absolute; no relative `](...)` or `src="..."` remains. All 18 in-repo targets exist at the paths the absolute URLs point to, including the new `docs/kernelforge/what-is-kernelforge.md` replacing the 404'd `AMD-AGI/KernelForge` repo link.
- Gate rewrite is consistent across all five carriers (PR template, copilot-instructions, AGENTS.md, CONTRIBUTING.md, style-guide) -- no doc still asks for a `[Unreleased]` entry.
- `ruff check` clean on the one Python file touched.
- Docs build: `docs.yml` runs `sphinx-build` without `-W`, so the new `Unreleased` section does not gate on warnings.

PR #1590, opened 2026-09-21
Add the shared review-pr skill

xiaofei-zheng:
- Description: PR review here was each reviewer's own checklist, so the same classes of defect were caught on some PRs and missed on others. This adds `.claude/skills/review-pr/`, one procedure anyone can run.

  `rules.md` holds 52 rules across 9 families, each mined from a real review on this repo and citing the PR it was learned from. A 14-row index at the top of the file maps diff shapes to rule ids, so a reviewer reads the subset this diff can trigger rather than all 52. `SKILL.md` is the procedure: fetch, select, answer five semantic questions, tier the core files, adjudicate the rules, run the AI-code diagnostic, refute every finding twice, publish. `fetch.sh` collects the PR into a scratch directory, resolving the base through the compare API's true merge base and exiting non-zero on any artifact it cannot collect. Where GitHub answers partially it says so rather than reading as complete: `mergeable` is re-read until it settles, and the commit list and the other PRs' file lists carry a marker naming what was cut. The backbone tier table lives in `references/tiers.md`.

  `.gitignore` needed the negation ahead of the exclude: `*.claude` matched the `.claude` directory itself, and git does not descend into an excluded directory to apply a negation to its contents. `AGENTS.md` gets one table row pointing at the skill.

  Rebased on main after #1589 retired `CHANGELOG.md`: the rule that checked for a changelog entry now checks the PR description for the observable effect, matching `AGENTS.md` and the current template. Two further rules that went stale with the same merge move with it ΓÇö T2 cited the style guide as silent on replacement-test depth, which it now mandates, and D5 described a `default=` parameter #1555 removed from `resolve_forge_llm_model`.

- Linked issue(s): none
- Tests: no test suite for a skill; verified by running it. `fetch.sh` end to end on #1583 and #1590 with all 13 artifacts collected and no errors, `base.txt` equal to the compare API merge base. All 52 rule ids resolve to bodies; V1-V6 and X2 are applied globally, and the remaining 45 ids are covered by index rows. `docfiles.txt` lists 4 doc paths on this PR and 1 on #1583. Every `$WORK` file cited in `rules.md` and `SKILL.md` is one `fetch.sh` writes or one a review step writes. shellcheck clean over `fetch.sh`. Truncation paths exercised on #1276: 250 of 1178 commits with the head subject named, and three open PRs flagged partial.
- [Size/complexity](../docs/contributing/style-guide.md#size-and-complexity) triggers crossed: one new self-contained directory plus two one-line edits to tracked files; splitting the procedure from the rules it runs would ship a skill that cannot execute.
- If this simplifies or refactors: n/a
- Observable effect: reviewers get `review-pr` as a shared procedure; nothing in the optimizer's runtime behaviour changes, so an operator running `inference_optimizer optimize` sees no difference.
- Breaking changes: no
- PR addresses single concern: yes
- Root cause is upstream (Magpie/TraceLens/GEAK/IntelliKit/AgentKernelArena), ticket filed: n/a

≡ƒñû Generated with [Claude Code](https://claude.com/claude-code)


chaojhou (approved):

ZhengGong-amd:
Two review axes are structurally unreportable by this skill, and they are the two that keep a tree from growing: **is there a simpler way that reuses what the repo already owns**, and **should this mechanism exist at all**.

The cause is the output contract. Blocking is defined as wrong behaviour or desync, with the bar "if this is not changed, is the current behaviour wrong?". Neither axis passes it, and non-blocking observations are deleted rather than softened. D1 is the only rule touching the first axis and it is `advisory`, so a `FIRE` on it ends as a `-- not reported` line in `verdicts.txt` the author never sees. X5 defers the second axis to `AGENTS.md` — but the rule there (*Review feedback is a hypothesis*: "ask what you would build if this code did not exist yet, and whether the mechanism under discussion should exist at all") addresses the author receiving a comment, not the reviewer writing one. Nobody on the review side owns the question.

Testable on this PR: 1380 added lines, a Step 5 diagnostic whose six checks restate C3, C2/D4, X2/V3, S1–S7, T1/T4 and P2/M3, and a `commits.txt` truncation marker spending two API calls on a case the head sha in `meta.txt` already answers. None of that is wrong behaviour or a desync, so the skill approves itself.

Requested — two edits, not a new family:

1. A third row in the blocking table, **Redundant mechanism**: the change could route through a path the repo already owns, or the mechanism it adds need not exist. Promote D1 from `advisory` to `blocking`; its escalation clause already says when a parallel path is wrong, and the plain case ("route through the existing owner") is exactly what has to reach the author.
2. A question in Step 6: "If this code did not exist, what would you build, and does the mechanism this diff adds need to exist?" Held to the same two refutation passes as a rule finding, so it cannot ship as taste.

Also reconcile `rules.md` ("advisory ... reported only if the reviewer was asked for more than blocking issues") with SKILL.md ("Non-blocking observations are deleted, not softened") — the two files answer the same question differently, and which one holds decides whether item 1 is even necessary.

xiaofei-zheng:
Addressed in #1598. Both axes were unreportable for the reason you gave, and the `advisory` wording in `rules.md` was the thing deciding whether the gap existed -- it said advisory findings ship when asked for, `SKILL.md` said they are deleted. They now say the same thing: adjudicated in `verdicts.txt`, not published.

Third blocking kind added, but scoped tighter than "redundant mechanism". The bar is a **documented** decision: the finding cites the `AGENTS.md` bullet the diff breaks and names the module that already owns the concern. "This could be simpler" and "I would have put it elsewhere" still go -- without the bullet and the owner there is nothing separating them from taste, and a third kind that admits taste stops the output being two parts.

Under that bar `D1` is blocking. Added `D10` for placement, which your framing implies but `D1` does not cover: a constant, precedence list, parser or client constructor defined outside the module that exports it. Its citation is #1338 -- `kernelforge/llm/gateway.py:29 _ANTHROPIC_KEY_ENVS` beside the identical `llm_config.py:63 ANTHROPIC_SYNTHESIZABLE_KEY_ENVS`, a second `parse_custom_headers` at `gateway.py:81` beside `llm_config.py:382`, while four `kernelforge` modules already import `hyperloom.common.llm_config`, so no layering barrier explains the copy.

Step 6 gains the `BUILD:` line, worded as you put it and held to the same two refutation passes.

On the `commits.txt` marker: the `total_commits` lookup folded into the commits call, so one request. The head-commit lookup stays and only runs when the list was cut -- `meta.txt` carries the head sha, not its subject, and `X2` reads subjects. It is the newest commit that makes a description stale, and that is the one a 250-of-1178 cut drops.

Not taking the Step 5 point. The six checks do restate `C3`, `C2`/`D4`, `X2`/`V3`, `S1`-`S7`, `T1`/`T4` and `P2`/`M3`, but Step 5 runs unconditionally while those rules reach the reviewer only if an index row matched. It is the net for a diff shape no row anticipated; removing it makes coverage conditional on the index being complete, which is the weaker of the two guarantees.

One thing your comment surfaced indirectly: answering it required re-reviewing a merged PR, and `fetch.sh` could not. `mergeable` stays `UNKNOWN` after merge so the settle-retry exhausted and exited, and the merge base resolved from the base branch name is the head itself once the branch contains the commits, so the diff came back empty. Both fixed in the same PR.

Issue #722, opened 2026-06-25
Triaging 21 June 025 June: 54 sessions: valid roofline trace, but the kernel-agent/TraceLens install was not found in the run environment


devalshahamd:
### Target release

None

### What kind of issue is this?

Suggestion / Feature idea

### Related area

None

### Description

**Root cause:** The kernel-agent / TraceLens installation was missing from the environment of these runs — `TRACELENS_ROOT` and `HYPERLOOM_KERNEL_AGENT_ROOT` were unset, or pointed at install directories that did not exist in the container (e.g. `/tmp/hyperloom/open-source-repos/TraceLens`, `/tmp/Hyperloom-src/kernel-agent`, `/kernel-agent`). The kernel-agent `install.sh` / env-bootstrap step that is supposed to install those components and write `kernel-agent.env.sh` (`$KERNEL_AGENT_ENV`) either did not run or its target path (under `/tmp`) was wiped before the analyze step, so the analysis entrypoint could not locate its own code and exited immediately on a path/`FileNotFoundError` check — before reading the trace.

```

**Sessions (54):**
- `/wekafs/users/d2d10ce6177d46ed841e1f205b74b86f/Abdullah-Taha-UTN-Qwen3-0.6B-LoRA-merged/20260625T021329Z`
- `/wekafs/users/d2d10ce6177d46ed841e1f205b74b86f/Amarnathjv-Qwen2.5-0.5B-Instruct-Gensyn-Swarm-flapping_tricky_toad/20260625T030519Z`
- `/wekafs/users/d2d10ce6177d46ed841e1f205b74b86f/ArliAI-QwQ-32B-ArliAI-RpR-v3/20260624T040353Z`
- `/wekafs/users/d2d10ce6177d46ed841e1f205b74b86f/Ayansk11-qwen3-4b-financial-sentiment-grpo/20260623T075036Z`
- `/wekafs/users/d2d10ce6177d46ed841e1f205b74b86f/ChaoticNeutrals-Poppy_Porpoise-v0.4-L3-8B/20260624T041139Z`
- `/wekafs/users/d2d10ce6177d46ed841e1f205b74b86f/Concyclics-PeoplesDaily-Qwen3-4B-Base/20260622T084251Z`
- `/wekafs/users/d2d10ce6177d46ed841e1f205b74b86f/Entropicengine-Trifecta-L3-8b/20260625T021622Z`
[...]

PR #1591, opened 2026-09-21
fix(kernel): preserve operator GEAK checkouts

### LINKED ISSUE #1481: install.sh: a user-supplied GEAK_ROOT working checkout is overwritten
## Summary

If a user points `GEAK_ROOT` at their own GEAK working checkout, `install.sh` treats it as an
installer-managed clone and realigns it to `GEAK_REF`:

```sh
# src/hyperloom/agents/kernel/scripts/install.sh L1303-1304 (ensure_geak)
run git -C "${GEAK_ROOT}" fetch --depth 1 origin "$GEAK_REF"
run git -C "${GEAK_ROOT}" checkout -q --force FETCH_HEAD
```

On the next `install.sh` run that checkout is taken off its branch onto a detached
`FETCH_HEAD`, **all uncommitted work in it is destroyed**, and a previously-full clone is left
shallow. No warning, no prompt, no stash.

`ensure_tracelens` already refuses to do this to an overridden `TRACELENS_ROOT` (L975-983);
`ensure_geak` has no equivalent guard.

## Reproducer

Self-contained: uses a local stand-in for the GEAK remote, so no access to AMD-AGI/GEAK is
needed, and it never touches the network. Runs in a few seconds.

```bash
#!/usr/bin/env bash
set -uo pipefail
HL="${HL:?set HL to a Hyperloom checkout}"
W=/tmp/geak-clobber
rm -rf "$W"; mkdir -p "$W"

# 1. stand-in for the GEAK remote
git init -q -b main "$W/geak-remote"
git -C "$W/geak-remote" -c user.email=r@e -c user.name=r commit -q --allow-empty -m v1
echo upstream-v2 > "$W/geak-remote/kernel.py"
git -C "$W/geak-remote" add kernel.py
git -C "$W/geak-remote" -c user.email=r@e -c user.name=r commit -qm v2

# 2. the user's working checkout: own branch, own commit, uncommitted edit
git clone -q "$W/geak-remote" "$W/my-geak"
git -C "$W/my-geak" checkout -q -b my-experiment
echo "my committed experiment" > "$W/my-geak/kernel.py"
git -C "$W/my-geak" add kernel.py
git -C "$W/my-geak" -c user.email=o@p -c user.name=o commit -qm "my experiment"
echo "my UNCOMMITTED edit" >> "$W/my-geak/kernel.py"

echo "=== BEFORE ==="; git -C "$W/my-geak" status --short --branch
echo "kernel.py: $(tr '\n' '|' < "$W/my-geak/kernel.py")"
test -e "$W/my-geak/.git/shallow" && echo "shallow: yes" || echo "shallow: no"

# 3. run the shipped ensure_geak. Stripping the trailing `main "$@"` so the file can be
#    sourced and one function called is the same technique install.sh's own tests use
#    (src/hyperloom/agents/kernel/tests/test_install_env_credentials.py).
sed 's/^main "\$@"$//' "$HL/src/hyperloom/agents/kernel/scripts/install.sh" > "$W/installer.sh"
export REPO_ROOT="$W" USER_DATA_PATH="$W/userdata"
export GEAK_ROOT="$W/my-geak" GEAK_REPO="$W/geak-remote" GEAK_REF=main
export PIP_NO_INDEX=1   # keep the trailing pip step off the network
. "$W/installer.sh"
DRY_RUN=0; CHECK_ONLY=0
ensure_geak

echo "=== AFTER ==="; git -C "$W/my-geak" status --short --branch
echo "kernel.py: $(tr '\n' '|' < "$W/my-geak/kernel.py")"
test -e "$W/my-geak/.git/shallow" && echo "shallow: yes" || echo "shallow: no"
git -C "$W/my-geak" branch -v
```

## Expected behavior

`GEAK_ROOT` is a user override. The installer should leave that checkout alone — the rule
`ensure_tracelens` already applies to `TRACELENS_ROOT`.

## Actual behavior

```
=== BEFORE ===
## my-experiment
 M kernel.py
kernel.py: my committed experiment|my UNCOMMITTED edit|
shallow: no
=== RUN ===
[kernel-agent] ensuring e2e optimizer geak (GEAK@main, formerly PerfSkills)
[kernel-agent] e2e optimizer checkout already present: /tmp/geak-clobber/my-geak
[kernel-agent] git -C /tmp/geak-clobber/my-geak fetch --depth 1 origin main
[kernel-agent] git -C /tmp/geak-clobber/my-geak checkout -q --force FETCH_HEAD
=== AFTER ===
## HEAD (no branch)
kernel.py: upstream-v2|
shallow: yes
* (HEAD detached at FETCH_HEAD) d9f8a43 v2
  main                          d9f8a43 v2
  my-experiment                 d5f8d2c my experiment
```

Precisely what is lost:

- **Uncommitted changes: gone, unrecoverable.** `--force` discards them; nothing is stashed.
- **HEAD: detached** off `my-experiment`.
- **Clone: now shallow.** `.git/shallow` is created on a repo that was a full clone
  (recoverable with `git fetch --unshallow`).
- Committed work survives — the `my-experiment` ref still points at `d5f8d2c`, so it is
  recoverable with `git checkout my-experiment`. The uncommitted edit is not.

## Root cause

`ensure_geak` (L1288-1307) branches only on whether `${GEAK_ROOT}/.git` exists. It has no
notion of the root being user-supplied, so a user's checkout takes the same "realign to
`GEAK_REF`" path as the installer's own dep-cache clone. The realign itself is wanted —
without it a `GEAK_REF` bump leaves the runtime on stale e2e code — it just must not apply to
a root the user provided.

`TRACELENS_ROOT` records the override at parse time (L160-165, L230-234) and
`ensure_tracelens` gates both the clone and the realign on it (L975, L1020).

## Related

- #722 — the `TRACELENS_ROOT` version of the "compare the path, not env presence" rule.

## Environment

- Reproduced on `main` @ `e3a0ef325`
- Ubuntu 24.04, git 2.43, bash 5.2 — no GPU or ROCm involvement; this is installer logic only

lishuoshuo-amd:
## Summary
- distinguish operator-supplied GEAK roots from installer-managed `GEAK@*` cache checkouts
- leave valid operator checkouts and worktrees unchanged, and fail clearly for missing or invalid roots
- retain clone and ref realignment behavior for installer-managed GEAK caches

Fixes #1481

## Test plan
- [x] Verify the regression tests fail on the destructive pre-fix behavior
- [x] Run `python3 -m pytest -q src/hyperloom/agents/kernel/tests/test_install_geak_pip.py`
- [x] Run the kernel-agent test suite in parallel with a clean credential environment (`1378 passed, 23 skipped`)
- [x] Run Bash syntax, Ruff, IDE lint, and `git diff --check`


xiaofei-zheng (approved):
No blocking issues. LGTM.

What I checked:

- **Ownership classification** (`install.sh` L322-330). `GEAK_ROOT` is treated as installer-managed only when its canonical parent equals the canonical `_open_source_root` *and* its basename matches `GEAK@*`; everything else is an operator override. Verified the four relevant cases by running the helpers standalone: `<cache>/GEAK@<sha>` (existing) → managed, `<cache>/GEAK@<sha>` (not yet cloned, so `readlink -f` resolves only the parent) → managed, trailing-slash form → managed, path outside the cache → override. This keeps the re-exported default self-healing across reruns (the #722 failure mode) while the operator path stops being rewritten.
- **`_is_git_checkout_root`** (L175-179) accepts a plain clone and a linked worktree (whose `.git` is a file, so the old `-d .git` test would have rejected it), and rejects a bare `.git` directory, a missing path, and a subdirectory of some *other* repo — so a stray `GEAK_ROOT` inside an unrelated checkout cannot be mistaken for a valid root.
- **`ensure_geak` branch coverage** (L1372-1406): operator override + valid root → no git mutation, falls through to the pip step unchanged; operator override + invalid root → `die` on a real run, `warn` under `--dry-run`/`--check-only`, mirroring the `ensure_tracelens` guard at L1066-1070; installer-managed → unchanged clone/`git init`+fetch and `fetch`/`checkout --force FETCH_HEAD` realign. The `mkdir -p "${GEAK_ROOT}"` moved into the clone branch only, so a missing operator root is no longer created before failing — matching the new test.
- **`report_status`** (L1541) switching from `-d .git` to `_is_git_checkout_root` does not regress the managed path: a missing or half-created cache dir still lands on the "checkout missing" warning, and a worktree root now reports its ref instead of warning.
- **Adjacent code that still writes to `GEAK_ROOT`**: `_prune_dep_cache` only globs `${_open_source_root}/GEAK@*`, so it is consistent with the new classification and never touches an operator root. `write_env_file`/`upsert_dotenv_var` re-export the operator path verbatim, which round-trips back to "override" on the next run.
- **Comment/doc sync**: the `GEAK_REF` ls-remote comment at L331-334 still describes the code, and `docs/install/install.md` L234-235 only lists `GEAK_ROOT` as an exported path, so nothing there goes stale. `CHANGELOG.md` was retired on main in #1589, so no entry is due.
- **Tests**: the five new cases cover operator clone untouched, operator worktree untouched, missing root not created, invalid root rejected without mutating the directory, and installer-managed realignment still working — which is exactly the reproducer from #1481. All CI checks on this PR are green, including the shards containing `test_install_geak_pip.py`; I did not re-run the suite locally.

PR title and description match the diff (single commit, no later additions).

PR #1594, opened 2026-09-21
fix(profile): follow InferenceX's move of benchmark_serving.py under infx/

luochen-amd:
## Symptom

Every AgentX session on a current InferenceX checkout fails `roofline`, and the failure is identical each time:

```
all 3 profile attempts failed; last: profile requires InferenceX to honour NUM_PROMPTS and
PROFILE_EXTRA_BODY, but the checkout Magpie will use is not patched:
inferencex_path=.../InferenceX, benchmark_lib_ok=True/True, benchmark_serving_ok=False/False
```

Four Crusoe sessions, three models, all `benchmark_mode=agentx`:

| session | model | roofline attempts | `roofline_snapshots` | `last_roofline_tput` |
|---|---|---|---|---|
| `20260918T142112Z-38ecea51` | GLM-5.2-MXFP4 | 2 | `[]` | 0.0 |
| `20260918T161416Z-2cc36717` | MiniMax-M3-MXFP4 | 4 | `[]` | 0.0 |
| `20260920T011545Z-a51530c5` | DeepSeek-V4.1-Flash | 2 | `[]` | 0.0 |
| `20260920T021112Z-e5d97936` | GLM-5.2-MXFP4 | 3 | `[]` | 0.0 |

All four also end with `last_trace_analyze={}`, `last_profile_trace=""` and `runs/profile=0`: no trace was ever captured, so TraceLens never produced an `analysis.md`, the orchestration prompt never received roofline evidence, and the KERNEL phase went on to search with no profile data behind it.

## Cause

[SemiAnalysisAI/InferenceX#3022](https://github.com/SemiAnalysisAI/InferenceX/pull/3022) (`refactor: migrate remaining Python tools into infx`) moved the implementation to `infx/bench_serving/benchmark_serving.py` and left `utils/bench_serving/benchmark_serving.py` as a 433-byte forwarding shim:

```python
"""Compatibility entrypoint for :mod:`infx.bench_serving.benchmark_serving`."""
...
sys.modules[__name__] = importlib.import_module("infx.bench_serving.benchmark_serving")
```

Two places addressed the old path literally:

- `_inferencex_patcher._resolve_benchmark_serving_paths` — so `ensure_benchmark_serving_patched` found the shim, could not locate `extra_body=`, and returned `False`.
- `ProfileExecutor` — `serving_path = ix_root / "utils" / "bench_serving" / "benchmark_serving.py"`, so the sentinel read-back inspected the shim, which cannot carry `PROFILE_EXTRA_BODY` however well the patch landed.

The anchor text itself did not change: it still matches **byte-for-byte, exactly once**, at the new location. This is purely a path migration.

Preflight does surface it, but only as a warning:

```
Preflight: WARNING — 2 of 4 InferenceX patch anchors no longer match; those patches will not be applied:
  profile_extra_body: ANCHOR MISSING — upstream text changed (.../utils/bench_serving/benchmark_serving.py)
```

so a session runs its whole baseline and then burns three profile attempts before the problem becomes visible.

## Fix

Resolve `benchmark_serving.py` across both layouts, newest first, and pick whichever copy actually carries the sentinel or the anchor:

```python
_BENCH_SERVING_REL_PARTS: tuple[tuple[str, ...], ...] = (
    ("infx", "bench_serving", "benchmark_serving.py"),
    ("utils", "bench_serving", "benchmark_serving.py"),
)
```

`_verify_anchors` now resolves per contract name rather than per fixed path, so both the union view (`verify_patch_anchors` / `failed_patch_anchors`) and the single-root gate (`failed_patch_anchors_in`) follow the move.

Patching and gating are deliberately scoped differently:

- **Patching** covers every root `_discover_inferencex_roots` reaches, which is what `_ensure_patched` is for.
- **The gate** is one root. `ProfileExecutor` reads the sentinel back through `benchmark_serving_path_in(ix_root)`, matching `lib_path` on the line above, so both sentinels in the error message speak for the checkout Magpie will actually run. Resolving the read-back through the union instead would let a patched unrelated checkout green-light an unpatched configured one — the same silent no-trace failure this PR is fixing. `failed_patch_anchors_in` already documents that split.

A tree carrying neither copy falls back to the legacy path, so the failure payload keeps reporting where the patch belongs rather than an empty string.

`_ANCHOR_CONTRACT` keeps recording the pinned revision's layout, so `test_inferencex_anchor_contract.py` still verifies the anchors against `INFERENCEX_REF` exactly as before — only resolution became layout-aware, not the contract.

## Tests

Five regression tests in `test_inferencex_patcher.py`:

- `test_benchmark_serving_patch_follows_the_infx_move` — a post-#3022 checkout is patched at the implementation, and the shim is left alone.
- `test_anchor_check_reads_the_infx_copy_not_the_shim` — `verify_patch_anchors` / `failed_patch_anchors` / `failed_patch_anchors_in` all report the contract intact on that layout.
- `test_benchmark_serving_legacy_layout_is_still_resolved` — a pre-move checkout keeps being patched at the old path.
- `test_gate_path_stays_inside_the_named_root` — a patched checkout exported via `$INFERENCEX_PATH` does not vouch for an unpatched configured root.
- `test_gate_path_falls_back_to_the_legacy_location` — a tree with neither layout still names where the patch belongs.

Verified on MI355X (Crusoe), `TMPDIR` writable: **201 passed** across `test_inferencex_patcher.py`, `test_inferencex_anchor_contract.py`, `test_eval_probe.py`, `test_bypass_backend.py`. `ruff format --check .` and `ruff check .` clean.

## Not included

- `eval_dest` is the other anchor this checkout reports as missing. Upstream replaced `append_lm_eval_summary`'s `mv -f "$jf" ./` with a rewritten `_stage_lm_eval_artifacts()` that still stages cwd-relative (`target="./${stem}_conc${eval_conc}${extension}"`), so that patch needs re-anchoring too — but against rewritten logic rather than a moved file, and it only affects `RUN_EVAL` runs, which AgentX disables. Left for a separate change.
- `bypass_engine.py` also names `utils/bench_serving/benchmark_serving.py`, but it *executes* the file rather than patching it, and the shim forwards correctly, so it needs no change.


xiaofei-zheng:
The path-migration fix itself is right: `_BENCH_SERVING_REL_PARTS` + content-based `_pick_benchmark_serving` resolves the post-#3022 layout, `_verify_anchors` resolving per contract name carries that into both the union view and `failed_patch_anchors_in`, and the three new tests pass locally.

One blocking issue, both symptoms in the same spot.

**`profile.py:1019` — the read-back gate widened from one root to every discovered root**

Pre-PR the gate read exactly `ix_root/utils/bench_serving/benchmark_serving.py`. Now it is `any(_contains(p, "PROFILE_EXTRA_BODY") for p in benchmark_serving_paths(ix_root))`, and `benchmark_serving_paths` -> `_resolve_benchmark_serving_paths` -> `_discover_inferencex_roots` also adds `$INFERENCEX_PATH` and `$MAGPIE_PATH/InferenceX`. So a patched *unrelated* checkout green-lights an unpatched configured one.

Reproduced against this branch:

```
A/utils/bench_serving/benchmark_serving.py   ->  "extra_body=None,"        (configured benchmark.inferencex_path, unpatched)
B/infx/bench_serving/benchmark_serving.py    ->  "PROFILE_EXTRA_BODY ..."  ($INFERENCEX_PATH, patched)

benchmark_serving_paths(A) = [A/utils/.../benchmark_serving.py, B/infx/.../benchmark_serving.py]
serving_valid              = True          # pre-PR: False
benchmark_serving_path_in(A) = A/utils/.../benchmark_serving.py
```

`ensure_benchmark_serving_patched` already returns `any_patched`, so with `serving_valid` now also true the `profile_inferencex_patch_failed` gate passes and profile runs against a checkout that ignores `PROFILE_EXTRA_BODY` — the same silent no-trace failure this PR is fixing. `failed_patch_anchors_in`'s own docstring calls the union view the thing a gate must not use. `lib_valid` on the line above is still single-root, so the two sentinels printed side by side in the error message no longer mean the same thing.

The PR already added `benchmark_serving_path_in(root)` for exactly this case; it has no production caller. Using it here fixes both symptoms.

**Same line, second symptom: the "here is where the patch belongs" path is lost**

With a single root carrying neither layout, `_resolve_benchmark_serving_paths` appends nothing (the legacy fallback lives only in the unused `benchmark_serving_path_in`), so `serving_paths == []` and the failure payload reports `benchmark_serving=""` where it previously reported `<ix_root>/utils/bench_serving/benchmark_serving.py`:

```
paths = []
error field benchmark_serving = ""
benchmark_serving_path_in(C) = C/utils/bench_serving/benchmark_serving.py
```

The description says "A tree carrying neither copy still falls back to the legacy path, so the existing 'here is where the patch belongs' diagnostics are unchanged" — that holds for `benchmark_serving_path_in`, not for the path the executor actually uses.

luochen-amd:
Good catch, and the reproduction is exactly right — fixed in 3007296b.

Both symptoms came from the same mistake: I routed the read-back through `benchmark_serving_paths`, which resolves the *union* of `_discover_inferencex_roots`. That silently widened a gate that was single-root before this PR, and it also lost the legacy fallback, because the fallback only lived in `benchmark_serving_path_in`.

`ProfileExecutor` now uses `benchmark_serving_path_in(ix_root)`:

```python
serving_path = benchmark_serving_path_in(ix_root)
...
lib_valid = _contains(lib_path, "${NUM_PROMPTS:-$max_concurrency}")
serving_valid = _contains(serving_path, "PROFILE_EXTRA_BODY")
```

One root, same layout awareness, and the legacy path when a tree carries neither copy — so the two sentinels sitting side by side in the error message speak for the same checkout again, and `benchmark_serving=` in the failure payload is a real path rather than `""`.

Against your two scenarios on the updated branch:

```
A (configured, unpatched utils/) + B (patched infx/, exported as $INFERENCEX_PATH)
  benchmark_serving_path_in(A) = A/utils/bench_serving/benchmark_serving.py
  serving_valid                = False      # was True

C (neither layout)
  benchmark_serving_path_in(C) = C/utils/bench_serving/benchmark_serving.py
  error field benchmark_serving = that path  # was ""
```

Patching itself still covers every discovered root — that is what `_ensure_patched` is for — so only the gate became single-root, which is the split `failed_patch_anchors_in` already documents.

`benchmark_serving_paths` is gone rather than left unused; `benchmark_serving_path_in` now has its production caller.

Two tests added for the scenarios above:

- `test_gate_path_stays_inside_the_named_root` — a patched checkout exported via `$INFERENCEX_PATH` does not vouch for an unpatched configured root.
- `test_gate_path_falls_back_to_the_legacy_location` — a tree with neither layout still names where the patch belongs.

`ruff format --check .` and `ruff check .` clean; the five tests covering this change pass locally.

xiaofei-zheng:
Re-checked 3007296b against both scenarios from my last comment, on a fresh checkout of the branch:

```
A (configured, unpatched utils/) + B (patched infx/, exported as $INFERENCEX_PATH)
  benchmark_serving_path_in(A) = A/utils/bench_serving/benchmark_serving.py
  serving_valid                = False        # was True

C (neither layout)
  benchmark_serving_path_in(C) = C/utils/bench_serving/benchmark_serving.py
  error field benchmark_serving = that path   # was ""

hasattr(_inferencex_patcher, "benchmark_serving_paths") = False
```

Both symptoms are gone, `lib_valid` and `serving_valid` are scoped to the same root again, and the union resolver is removed rather than left unused. The five tests covering this change pass locally (`test_benchmark_serving_patch_follows_the_infx_move`, `test_anchor_check_reads_the_infx_copy_not_the_shim`, `test_benchmark_serving_legacy_layout_is_still_resolved`, `test_gate_path_stays_inside_the_named_root`, `test_gate_path_falls_back_to_the_legacy_location`).

One blocking issue left, and it is the description rather than the code.

**The PR description still describes the pre-3007296b version**

> `ProfileExecutor` reads back the same resolved path the patcher wrote, via the new `benchmark_serving_paths()`.

`benchmark_serving_paths` no longer exists on this branch — 3007296b deleted it and dropped it from `__all__`. The executor now uses `benchmark_serving_path_in(ix_root)`, and the difference is the whole point of the fix: single-root, not the union of `_discover_inferencex_roots`. A reader following the description would go looking for the function that caused the bug.

The Tests section also still says "Three regression tests" and lists three; there are five now.

Please refresh those two spots so the description matches the diff.

xiaofei-zheng (approved):
No blocking issues in the code. LGTM.

What I checked:

- The gate fix in 3007296b, re-run on a fresh checkout of the branch: `benchmark_serving_path_in(A)` stays inside the configured root, so a patched checkout exported via `$INFERENCEX_PATH` no longer makes `serving_valid` true for an unpatched one (`False`, was `True`); a tree with neither layout reports the legacy path in the failure payload instead of `""`.
- `lib_valid` and `serving_valid` now resolve against the same root, so the two sentinels in the error message speak for one checkout.
- `benchmark_serving_paths` is deleted and out of `__all__`; `benchmark_serving_path_in` has its production caller in `ProfileExecutor`.
- `_pick_benchmark_serving` picks by content (sentinel or anchor), so a post-#3022 checkout is patched at `infx/bench_serving/benchmark_serving.py` and the `utils/` shim is left alone, while a pre-move checkout still resolves to the legacy path.
- `_verify_anchors` resolving per contract name carries the layout awareness into both the union view and `failed_patch_anchors_in`.

Tests run locally, all passing: `test_benchmark_serving_patch_follows_the_infx_move`, `test_anchor_check_reads_the_infx_copy_not_the_shim`, `test_benchmark_serving_legacy_layout_is_still_resolved`, `test_gate_path_stays_inside_the_named_root`, `test_gate_path_falls_back_to_the_legacy_location`.

Please still refresh the description before merging: it names `benchmark_serving_paths()` as the executor's resolver (deleted in 3007296b) and says "Three regression tests" where there are now five.

PR #1595, opened 2026-09-21
Reclaim resources from holders that ended, and bound the Ray round wait

zoroyihan7:
Two resource-leak defects that together deadlocked a live session on 2026-09-21. They are in one PR because they are two links of the same chain.

## What happened

A specialist raised `ExecutionCleanupUnconfirmed`; the dispatcher deliberately retained capacity rather than releasing. From there:

1. Its six lane leases (`benchmark_lane`, `server_lifecycle`, `gpu_research_lane`, `profile_lane`, `research_lane`, `bringup_round`) were never released. Occupancy is decided by `SELECT lane, holder_id FROM leases` with no regard for whether anybody is still using the row, so they were still held two hours after their `expires_at`.
2. Ray likewise kept the GPU reservations: `ray status` reported `8.0/8.0 GPU` in use while the physical GPUs sat at 0% with no serving processes.
3. Two `integrate_patch` warmups requesting `{'CPU': 1.0, 'GPU': 4.0, 'serving_slot': 1.0}` sat in `PENDING_NODE_ASSIGNMENT` forever. `_await_or_cancel` polls `ray.wait` inside a `while True` with no wall-clock deadline, so both worker threads parked there for over an hour.
4. Each parked thread held its pool slot, so its task stayed `running`, its lane leases lapsed unreclaimed, and 19 `integrate_patch` tasks starved in `queued`. The coordinator's main log went silent for 64 minutes with zero child processes.

Deleting the lease rows by hand let a task run within 90 seconds — twice.

## The fixes

**Lane leases** — reclamation keys on the holder's task being *terminal*, not on TTL expiry. A TTL is a static per-action budget that nothing enforces (`explore` budgets 7200s and holds `server_lifecycle` + `benchmark_lane` across a benchmark that can exceed it), so a lapse says only that a long run outlived its estimate; the terminal state is the holder's own account of having stopped. Reading no timestamp keeps clock skew out of the predicate entirely. Two guards carry over verbatim from `reap_dead_holders`: `owner_scope` must match this boot/PID namespace, and a task with a live `gpu_leases` row is left alone. The sweep shares the caller's cursor so it and the capacity read that follows sit in one `BEGIN IMMEDIATE`, and it also runs as its own pass — the dispatcher's lane gate reads `lane_holders` *before* any acquire, so a leaked row starves the queue without an acquire ever running to reclaim it.

Running ahead of the round rules, this also settles rounds that a leaked lane row used to wedge open indefinitely — by hand-off, or by expiry on the terminal-holder cap. Both are pinned, as is the bound: a holder whose cleanup never confirmed gets its lanes back *without* its round being advanced.

**Ray round wait** — the deadline is the round's own cap plus slack for what actually happens outside it, configurable via `INFERENCE_OPTIMIZER_RAY_ROUND_WAIT_SEC`. Bailing out goes through the cooperative cancel sequence the cancel path already uses (ask the actor, escalate only after the grace period), because `ray.kill` runs neither `__ray_terminate__` nor `atexit` and would strand the served process tree of a round that is genuinely still running.

## Verification

- Every new test was checked to fail with the behavioural block surgically removed (helpers and constants kept, so the assertion fails rather than an import), then pass restored.
- Both lint gates, ruff 0.16.2, whole tree: `ruff check .` → All checks passed; `ruff format --check .` → 1730 files already formatted.
- Regression sweep across `bringup/`, `loop/`, `actions/`: 632 passed, 1 skipped. The single failure (`test_an_unreachable_framework_is_unavailable_not_invalid`) reproduces identically on an unmodified `origin/main` checkout and is unrelated.

## Known, not fixed here

A lease row acquired while `/proc` is unreadable is stored with `owner_scope=''`, and no reaper can take it back — `holder_is_dead` rejects it on the scope check, and so does this pass. It predates this change; closing it means giving ownership a fallback identity for the case where the kernel will not name one, which changes what a scope asserts. Documented at the guard that inherits it.


ZhengGong-amd:
## PR #1595 -- Reclaim resources from holders that ended, and bound the Ray round wait

What it does: Two independent defects — `ray.wait` looping with no wall-clock deadline, and lane-release logic that required proving a served process was gone — combined on 2026-09-21 to park pool threads for over an hour and starve 19 queued tasks. The PR bounds the wait with a ceiling derived from the round's own cap plus an hourly slack (overridable via `INFERENCE_OPTIMIZER_RAY_ROUND_WAIT_SEC`), quarantines timed-out leases rather than killing the actor so their GPUs stay reserved, removes all liveness-inspection branches from lane reclamation, and adds a per-lane diagnostic that logs a paste-ready `sqlite3 DELETE` once per `(lane, holder)` alongside a new `leases_unverifiable` / `cleanup_unconfirmed` count in the maintenance summary.

Blocking issues: 3

1. [X5] `INFERENCE_OPTIMIZER_RAY_ROUND_WAIT_SEC` is absent from the reference docs [verified]
   Problem: `_ray_serving.py` adds `ROUND_WAIT_TIMEOUT_ENV = "INFERENCE_OPTIMIZER_RAY_ROUND_WAIT_SEC"` and the PR body calls it out as operator-overridable, but `docs/reference/environment-variables.md` and `.env.template` at head `aa80a2b` contain no entry for it.
   Impact: An operator who needs to widen the ceiling (e.g. on a slow head node, or to reproduce the 2026-09-21 stall deliberately) has no reference to find the knob; the env var that exists to prevent an outage is invisible unless the operator reads the source.
   Action: Add a row for `INFERENCE_OPTIMIZER_RAY_ROUND_WAIT_SEC` to `docs/reference/environment-variables.md` and `.env.template` following the existing `INFERENCE_OPTIMIZER_RAY_*` entries; state the unset default (`round_timeout + 3600s`), the `<= 0` disabling semantics, and that non-finite values are rejected.

2. [X3] `_unverifiable_holders` docstring contains a broken duplicate fragment [verified]
   Problem: `resource_lock.py:288-289` reads "Nothing reclaims these rows any more, so everything it / so everything it returns is a row that survived reclamation." — two sentence halves merged into one garbled line, present verbatim in the shipped file at head `aa80a2b`.
   Impact: The docstring is the entire operator story for a retained lane (`_report_unverifiable` logging the remedy IS the interface, as stated in the docstring itself); a garbled sentence in it undermines a function the PR explicitly puts forward as the operator's only path to a remedy.
   Action: Fix line 288 in `resource_lock.py` — the intended text is "Nothing reclaims these rows any more, so everything it returns is a row that survived reclamation."

3. [S1] New broad `except Exception: log.debug` in `maintenance.py` violates AGENTS.md and silently drops the `cleanup_unconfirmed` stat [verified]
   Problem: `maintenance.py:50-52` adds a nested `except Exception: log.debug(... exc_info=True)` around the `host.reconciler._locks.cleanup_confirmation_rate()` call. AGENTS.md prohibits new broad `except Exception`. The catch demotes any failure — including an `AttributeError` if `_locks` is ever renamed on `Reconciler` — to a debug-level log invisible in production, silently omitting `cleanup_unconfirmed` from the maintenance summary with no operator signal.
   Impact: `cleanup_unconfirmed` is the new measurement the PR introduces to decide whether an automatic lane-release mechanism is worth building. If the call breaks silently, operators see a summary that looks correct but is missing the one ratio that matters for that decision.
   Action: Remove the nested `try/except`. The `cleanup_confirmation_rate` call is a pure read against the same `_locks` object the reconciler already exercises every tick; there is no error class it can raise that would not also indicate a defect worth surfacing. If a tick-safe pattern is required, demote only a specific exception (e.g. `sqlite3.Error`) or route through `Reconciler` rather than reaching its private `_locks` directly.

Checked: `src/hyperloom/orchestrator/actions/executors/_ray_serving.py`, `src/hyperloom/orchestrator/bus/resource_lock.py`, `src/hyperloom/orchestrator/loop/maintenance.py`, `src/hyperloom/orchestrator/bringup/reconcile.py`, `src/hyperloom/orchestrator/loop/sub_agent_runner.py`, `src/hyperloom/orchestrator/state/task_registry.py`, `src/hyperloom/orchestrator/state/task_states.py`, `docs/reference/environment-variables.md`, `.env.template`, all seven test files in the diff | Ran: `gh api` for head-tree contents, diff parsing, CI status | Base: `b279e2fcbc6a9e9202278e47de3316c55501bab7` | Head: `aa80a2ba504b9d08cbdfa25a01ff57f08f6a0642`

zoroyihan7:
Thanks — all three verified against the tree and fixed in `6121997f0`.

**[X5] env var undocumented.** Correct, it existed only in source. Added to both `docs/reference/environment-variables.md` and `.env.template`. I took the semantics from `_round_wait_timeout_sec` rather than from my own PR text: unset means the round's own timeout plus `ROUND_WAIT_SLACK_SEC` (3600s), `<= 0` disables the ceiling and is also what a round with no cap of its own gets, and a non-finite value is *rejected* — worth stating explicitly, since `nan > 0` is false and one ops typo would otherwise switch off the ceiling that exists to prevent the stall. The entry also notes that a timeout abandons the round but deliberately keeps the actor alive, so an operator understands why those GPUs do not come back.

**[X3] garbled docstring.** Correct, and it was mine — left by a text replacement during the rework. Fixed, and I took the opportunity to make the sentence say the useful thing rather than the tautological one: it now reads "everything it returns is a lane that will stay held until an operator acts on it."

**[S1] broad `except`.** Correct on both counts, and your second point is the sharper one: reaching into `reconciler._locks` meant a rename would have been swallowed as a debug line, leaving a summary that looks complete while missing the one ratio the whole retention decision rests on. Removed the `try/except` entirely and added `Reconciler.cleanup_confirmation_rate()` so the call goes through a public entry point. Removing the swallow immediately broke two maintenance tests whose reconciler double lacked the method — which is the behaviour you were asking for, so I fixed the double rather than restoring the catch.

Gates after the change: both ruff gates clean whole-tree, 310 passed across `loop/`, `bringup/`, `test_resource_lanes.py` and `test_ray_backend_unit.py`. The one failure, `test_an_unreachable_framework_is_unavailable_not_invalid`, reproduces on an unmodified `origin/main` checkout and is unrelated.

Separately, one finding from this investigation is relevant to #1604 — I have left the detail there.

ZhengGong-amd (approved):
## PR #1595 -- Reclaim resources from holders that ended, and bound the Ray round wait (re-review at 6121997)

What it does: Bounds the previously unbounded `ray.wait` loop in `_await_or_cancel` with a wall-clock ceiling derived from the round's own cap plus an hour of slack, and on timeout quarantines the lease — deliberately leaving the actor alive, with its handle parked at session scope, so its GPUs stay reserved rather than being handed back while a served tree may still map them. On the lane side it abandons every attempt to prove a lane free, since each candidate proof is a proxy a `setsid`-ing server escapes, and replaces reclamation-by-inspection with a once-per-`(lane, holder)` diagnostic carrying a paste-ready `sqlite3` statement, plus `leases_unverifiable` and `cleanup_unconfirmed` counts on the maintenance summary. The task state machine also moves to a storage-free `state/task_states`, closing the import cycle CodeQL flagged rather than deferring around it.

Blocking issues: none

All three findings from the previous review are resolved at this head, each verified against the head tree rather than the commit message:

- X5 — `INFERENCE_OPTIMIZER_RAY_ROUND_WAIT_SEC` now has an entry in both `docs/reference/environment-variables.md` and `.env.template`. The documented contract matches `_round_wait_timeout_sec`: unset yields `timeout + ROUND_WAIT_SLACK_SEC` (3600.0), `<= 0` disables the ceiling and is also what an uncapped round gets, and a non-finite override raises rather than silently switching the ceiling off.
- X3 — the garbled duplicate in the `_unverifiable_holders` docstring is gone; `resource_lock.py:288-289` now reads as one sentence.
- S1 — the broad `except Exception` around the statistic is gone. `cleanup_confirmation_rate` is now a public method on `Reconciler` and `maintenance.py` calls it there instead of reaching into `._locks`, which is the better half of the fix: the private-attribute coupling that motivated the swallow no longer exists. Both call sites of `run_lease_and_db_reclaim` (`loop/maintenance._run_maintenance` and `phases/macro_cycle`) resolve `reconciler` through `CoordinatorCollaborator.__getattr__` to the cached `Reconciler` built in `loop/coordinator.py:1185`, so no caller can reach a host lacking the new method. The test stub gained the method and states why the call site is deliberately unguarded.

CI note: `test (py3.10 shard 5/6)` and both coverage jobs failed on an infrastructure flake, not on this diff. The shard itself reported `3867 passed, 10 skipped`, then `Failed to CreateArtifact: Unable to make request: ECONNRESET` while uploading its coverage data; the py3.10 coverage job then reported `Missing shard(s) ... Coverage was intentionally NOT evaluated on a partial suite`, and py3.11 measured `TOTAL 90.63%` against a `fail_under = 90` floor before failing on the same matrix-completeness gate. The failed jobs have been rerun.

Checked: `_ray_serving.py`, `bus/resource_lock.py`, `loop/maintenance.py`, `bringup/reconcile.py`, `loop/sub_agent_runner.py`, `state/task_registry.py`, `state/task_states.py`, `orchestrator/collaborator.py`, `loop/coordinator.py`, `phases/macro_cycle.py`, `docs/reference/environment-variables.md`, `.env.template`, `pyproject.toml`, and the seven changed test files | Ran: head-tree reads at `6121997`, `aa80a2b...6121997` comparison, failing-job log retrieval, coverage-floor lookup, rerun of the failed jobs | Base: `b279e2fcbc6a9e9202278e47de3316c55501bab7` | Head: `6121997f0e37698f4e669d4985e4fffe0311c7a0`

PR #1596, opened 2026-09-21
Let the rewrite pipeline's caller own its task shape, its files and its stages

jiaqiang-dot-liu:
Four fixes to what the rewrite pipeline and the loop assume about the caller that drives them, and about the agent that works inside them. Together they are **+421 / −281** over 30 files.

## The engine judged keeps by reading one consumer's task format

`canonical_correctness` reimplemented a consumer's v1 evaluator inside the engine, as its own docstrings said: _"The arena's Step 1"_, _"The arena's Step 2"_, _"The arena's absent-command branch returns a failure, not a skip"_. That last one decided keeps.

A consumer that has since moved to declaring evaluation differently cleared correctness and the performance threshold, then failed acceptance with `config.yaml declares no 'compile_command'` and was reverted. **No keep was reachable at all.** Observed on MI355X: two iterations measured 1.396x and 1.337x against a 1.26x requirement, both discarded for that reason.

The asymmetry made it worse. A workspace with no config was accepted unchecked, so shipping a task config was the thing that broke keeps.

Neither caller needed the second verdict. The loop judged the candidate with the driver in Step 4 and measured through it for three bench suites after; the KB warm-start judged every candidate with `_correctness_once` and `_bench_once` before ranking, and adoption re-applies the same patch to the same base, reproducing what was measured. Both then asked the driver the same question again through that file.

The loop keeps the step for assembly alone, the one backend that needs it: its `numerical_validation` contract carries evidence for `validate_source_independence` that an exit status cannot. That contract is the engine's own schema, unlike the mirrored command keys.

This gives up a real guarantee: a task could previously reject a candidate its driver had accepted. A caller whose driver is authored for it rather than verified by it should weigh that before relying on a keep.

## The gate an agent was told about was not the gate it was judged by

`CANONICAL_GATE_PROMPT` is rendered into all nine kernel-backend prompts, and it told each of them that a KEEP is decided by the task's `compile_command` and then its `correctness_command`, _"which forge runs on every candidate it would otherwise accept, whether the loop kept it or a warm start adopted it from the knowledge base"_. After the change above the loop runs that for assembly alone, and the warm start for no one, so the paragraph was false for eight of the nine and its warm-start clause for all nine.

An agent optimizes against the criterion it is given. It was told to discount the suite that now decides its candidate, to spend a step on commands with no bearing on the outcome, and to read the error raised by a command that never ran as _"the only thing that tells you what to fix"_ — a diagnostic that does not exist. The comment above the constant named the drift it existed to prevent, and was itself the drift.

The mapping from backend to gate now lives in one place. `scoring.py` grows `runs_task_suite_acceptance()`, which the loop's Step 7 and `canonical_gate_prompt()` both read, so a future move of the gate carries the description with it. `base.py` resolves the backend already and is the only caller of `build_system_prompt`, so it renders the right paragraph there; a prompts module naming its own backend would restore the second copy this removes.

Assembly keeps its paragraph less the warm-start clause. The other eight get a shorter one stating what actually decides — the driver's suite clearing SNR, and a mean gain over the incumbent wider than this candidate's own spread — and keeps the shape lesson as what it now is: a consumer may build at a shape forge never measures, so clearing this gate is not the same as being accepted downstream.

Five sentences repeated the claim outside the shared paragraph and are corrected with it, `GATE` as a name for the performance target becomes `performance target` where the word now has a defined meaning beside it, and `hipblaslt` gains the paragraph it had never rendered although three of its stop conditions rest on one. Six places in the docs, the CLI help and the gluon knowledge cards are corrected the same way, and `experience_integration.py` drops a comment that outlived its constant.

## The port committed one file out of the implementation it validated

The port commit named a single path, the FlyDSL entry point, while the driver validated the attempt directory as it stood on disk. A port that puts part of its implementation in a module beside the entry point produced a commit no stage had measured, and a consumer reading that commit gets an entry point whose import is not there.

Nothing stops a port from doing this: the attempt directory is typically ignored by the caller's repository, so a new module in it is invisible to the session guard, which looks for new files with `git ls-files --exclude-standard`.

The commit now takes the attempt directory. It holds only what the producer put there, so this adds nothing the caller protects, and it needs no declaration of how many files an implementation spans.

## The caller could not decline the framework patch

The pipeline ends by patching the operator back into the framework repository, and folds that patch into its own `success`. A consumer that only wants the standalone kernel could not say so: the need was inferred from the workspace having a resolvable Git HEAD, which the agent sessions require, so it was always true.

Three consequences, none asked for. The run reserved its last 20 minutes for work the caller would discard, exited nonzero when that work failed, and reported `best_commit` as an apply-back commit whose tree holds framework edits rather than the kernel. One campaign scored a correct 1.361x candidate as 0 because the apply-back agent could not start.

`--applyback/--no-applyback`, **on by default**, so every existing invocation is unchanged. Declining it skips the stage, drops `applyback_required`, and returns the reserve via `RewriteBudgetPolicy.without_applyback()`. `build_result` now reports the policy the run applied rather than the module default, because a reader comparing the search window against a reserve the run never held is reading a number about a different campaign. `capabilities()` gains `applyback_optional` so a consumer can detect support instead of probing the CLI.

## Tests

Ten added, and `test_loop_runner.py`'s canonical block reworked to follow the gate rather than pin the behaviour it removed:

- `test_a_non_assembly_keep_rests_on_the_driver_not_on_the_task_config` — both parametrizations fail if the backend condition is removed
- `test_snr_pass_with_failing_assembly_acceptance_is_reverted` and `test_assembly_acceptance_is_skipped_for_a_candidate_that_is_not_faster` — the rejection path where the suite still runs
- `TestAcceptanceGateMatchesTheLoop` — two directions, for every registered backend
- `test_without_applyback_returns_only_that_reserve`, `test_applyback_is_required_by_default`, `test_declined_applyback_skips_the_stage_and_returns_its_reserve`
- `test_ensure_git_committed_takes_a_whole_attempt_directory` — an entry point importing a sibling, with caches at two depths

Three cases are deleted rather than moved. `test_canonical_suite_output_reporting_failure_reverts` and `test_workspace_declaring_no_correctness_command_cannot_keep` differ only in how acceptance rejects, which `test_reported_failure_never_hides_numerical_evidence_or_authorizes_acceptance` and `test_config_without_a_correctness_command_fails_the_gate` already cover against the suite itself; `test_canonical_suite_passing_keeps_the_faster_candidate` is the weaker half of the new contract test. `test_a_warm_start_failing_the_task_suite_is_not_adopted_or_published` goes with the guarantee it pinned.

The nine rendered-prompt hashes are re-snapshotted, diffed line by line against the previous rendering, with the reason recorded beside them as that file requires.

## Verification

CI is the authority and runs the full suite; it was green on all twelve shards at `f55e71d5e`, before the prompt and docs commits. Locally, on the current head: the loop, acceptance, numerical-contract, resume, warm-start and rewrite files (493 passed), the prompt-contract and backend files (160 passed), and the CLI-docs and campaign-config files (363 passed). `ruff format --check` and `ruff check` are clean over `src/kernelforge`.

No GPU run on this branch. No change to measurement, scoring, budgets or artifact retention beyond the reserve a declined stage no longer holds.

One detail worth flagging for review: `git add -f -- <dir> ':(exclude)**/__pycache__/**'` silently stages an **empty set** rather than excluding the cache, and `**/` alone requires an intervening directory, missing the cache next to the entry point. The exclusion is therefore anchored under the path being added and spelled twice.

## Test plan

- CPU test suite on CI for every touched area
- A campaign on GPU that earns a keep, which this is the precondition for and which no run has produced yet
- A rewrite with `--no-applyback` end to end, confirming the returned reserve reaches the search

Made with Cursor


xiaofei-zheng:
Reviewed against `git merge-base origin/main feat/rewrite-caller-contract` (060a1e4b6).

## Blocking

**1. The Step 7 backend gate breaks four existing tests that the diff does not update.**

`loop/runner.py:3789` now requires `self.ic.kernel_backend == "assembly"`. The shared fixture `_measurement_loop` (`tests/test_loop_runner.py:2325`) pins `kernel_backend=""`, and `test_loop_runner.py` is not in this diff. Four tests built on that fixture assert the behavior the gate removes:

- `test_snr_pass_with_failing_canonical_suite_is_reverted` (:2418) — asserts `kept is False`, `validation_outcome == "canonical_correctness_failure"`, `"0.02468" in error_output`, `"[canonical] FAIL"` on stdout. Step 7 is now skipped, so `kept` is `improved`, i.e. `True`.
- `test_canonical_suite_output_reporting_failure_reverts` (:2474) — asserts `kept is False`.
- `test_workspace_declaring_no_correctness_command_cannot_keep` (:2485) — asserts `kept is False` and `"declares no 'correctness_command'" in validation_summary`.
- `test_workspace_without_a_config_keeps_on_the_snr_verdict_alone` (:2496) — asserts `"[canonical] UNVERIFIED"` on stdout; that line is printed from `loop/canonical_correctness.py:214`, inside the call that no longer runs.

This contradicts the description's "the failure set matches `origin/main` exactly with none added". Either these tests are deleted along with the guarantee, as the description says was done for the warm-start test, or the gate is wrong.

**2. `CANONICAL_GATE_PROMPT` still tells every non-assembly agent that forge runs the gate.**

`loop/scoring.py:44-53` is rendered verbatim into the prompts of aiter, ck, flydsl, fusion, gluon, hip and triton — every backend except the one that keeps the step. It still says a KEEP is decided by the task's `compile_command` and `correctness_command`, "which forge runs on every candidate it would otherwise accept, whether the loop kept it or a warm start adopted it from the knowledge base". After this PR forge runs neither, for any of those seven. The comment directly above the constant states its purpose: "so an agent's self-check and forge's acceptance decision cannot drift apart".

**3. Docs and CLI help still document the removed behavior.**

- `docs/kernelforge/conceptual/architecture.md:29` — acceptance "run on any candidate about to become the incumbent — a kept iteration or a knowledge-base warm start alike".
- `docs/kernelforge/conceptual/optimization-loop.md:22-27` — gate step 5, including "A knowledge-base warm start is accepted by the same step before it can become the starting point".
- `docs/kernelforge/reference/cli.md:93` and `cli.py:671` — "A KEEP is decided by the task's own `correctness_command`, not by this value".
- `data/local_knowledge/.../gluon_levers/forge_integration.md:33` and `:189`.
- `knowledge/experience_integration.py:48-49` — the comment describing `_WARMSTART_CANONICAL_TIMEOUT_CAP_SEC` is left behind after the constant is deleted.

**4. The description claims tests that are not in the diff.**

"New tests cover acceptance against a differently-shaped config, exit status over output text, assembly still demanding its contract, an entry point importing a sibling with caches at two depths, and the declined stage returning its reserve." The diff adds exactly four tests: `test_without_applyback_returns_only_that_reserve`, `test_ensure_git_committed_takes_a_whole_attempt_directory`, `test_applyback_is_required_by_default`, `test_declined_applyback_skips_the_stage_and_returns_its_reserve`. `test_canonical_correctness.py` and `test_numerical_contract.py` are untouched, and nothing in this diff covers the Step 7 gate or the warm-start removal — the two changes that give up the most.

## Checked, not blocking

`without_applyback()` / budget plumbing, the `--applyback/--no-applyback` wiring through `cli.py` → `run_rewrite` → `build_result`, `applyback_required` derivation, `capabilities()["applyback_optional"]`, the `_PYCACHE_EXCLUDES` pathspec anchoring, and `_ensure_git_committed` taking a directory. Test execution was not possible locally (`loop/run_state.py` imports `fcntl`); the four failures above are read off the fixture and the new guard directly.

jiaqiang-dot-liu:
Thanks — all four were real. Addressed in `f55e71d5e`, `918ebbb2d` and `2d8654c56`. Two of them I resolved differently from the suggestion, so those are spelled out below.

## 1. The four broken tests

Confirmed exactly as read: `4 failed` on the branch against `9 passed` on `origin/main` for that block, and the shard failures matched — py3.10 shard 1 carried three of them.

You offered a binary: delete the tests along with the guarantee, or the gate is wrong. I took neither. The rejection path is still real where the suite still runs, so it moved there rather than being dropped:

- `test_snr_pass_with_failing_assembly_acceptance_is_reverted` carries the mla-decode story onto assembly, which needs a declared `numerical_validation` contract before the suite is reached. It still pins what the old case pinned: the revert, the outcome label, the broken tolerance reaching `error_output`, and the printed `FAIL`. The not-faster case moved with it, where skipping the expensive check still means something.
- Three are deleted. `test_canonical_suite_output_reporting_failure_reverts` and `test_workspace_declaring_no_correctness_command_cannot_keep` differ only in *how* acceptance rejects, and `test_reported_failure_never_hides_numerical_evidence_or_authorizes_acceptance[text_only_failure]` and `test_config_without_a_correctness_command_fails_the_gate` already cover both against the suite itself. `test_canonical_suite_passing_keeps_the_faster_candidate` became the weaker half of the new contract test.
- `test_a_non_assembly_keep_rests_on_the_driver_not_on_the_task_config` is what replaces them, and it is the case nothing covered: a workspace may ship a task config whose suite would reject and still earn a keep, with no canonical verdict in the operator's log. Both parametrizations fail if the backend condition is removed — I checked by removing it.

CI is green on all twelve shards for that commit.

## 2. CANONICAL_GATE_PROMPT

The most important of the four, and the sweep turned up more than the seven backends: nine render it, and six further sentences repeated the claim outside the shared paragraph — step 4 in aiter/ck/hip/triton and flydsl, ck's iron rule, fusion's stop condition, and gluon's `compile_command` trap, which asserted a candidate "will fail acceptance" at a step gluon no longer reaches. Deleting the paragraph would not have fixed it either: `Gate met → STOP` in triton/gluon/flydsl and two references in `prompt_utils.py` rest on its definition.

Rather than split the constant in two and let each prompts module import the one that describes it, the backend-to-gate mapping is now a single predicate. `scoring.py` grows `runs_task_suite_acceptance()`, which the loop's Step 7 and `canonical_gate_prompt()` both read; `base.py` resolves the backend and is the only caller of `build_system_prompt`, so it renders the right paragraph there. Splitting the constant would have left the pairing restated in nine files — a second copy is what produced this bug, and the comment you quoted asks for the opposite.

`TestAcceptanceGateMatchesTheLoop` pins both directions for every registered backend. The nine prompt hashes are re-snapshotted, diffed line by line, with the reason recorded beside them.

Two things found while there, both pre-existing: `hipblaslt` never rendered the paragraph at all although three of its stop conditions rest on one, and `GATE` named the performance target in ck/hip/hipblaslt, which collides now that the word has a defined meaning beside it. Both corrected.

## 3. Docs and CLI help

All five fixed, plus two more the sweep found: `operators/dense_gemm/gluon.md:83` ("The task's own `correctness_command` decides") and `operators/quant_fp4_mxfp/gluon.md:74`. The advice under the gluon cards is still sound — a scale-packing error produces plausible garbage a good average SNR will not catch — so it is retargeted at the driver's suite rather than deleted, and the lever card keeps "run the task's own commands yourself" as preparation for a consumer forge does not speak for.

`--snr-threshold`'s help and its reference row had inverted the truth rather than merely aged: the threshold is now what the driver's suite must clear.

## 4. The description

You were right and the cause is worth recording, because it would have hidden this from any local run: the environment that produced "315 passed, the failure set matches `origin/main` exactly" had no `pytest-asyncio`, so every async test failed identically on both sides and the four regressions sat inside the "pre-existing" count. Installing the plugin reproduced your four immediately.

The description is rewritten. Its test section now lists the ten tests that exist, names the three that were deleted and why, and its verification section says what was actually run and defers to CI for the full suite. The claim also appears in `6d015e7`'s message, which I have left rather than force-push a branch under review; say the word if you would rather it be rewritten.

Your "not blocking" list is unchanged by any of this — nothing in these three commits touches the budget plumbing, the apply-back wiring, the pathspec anchoring or `_ensure_git_committed`.

Still outstanding and worth your view: this keeps the gate keyed on the backend name. Your "or the gate is wrong" points at the alternative the description itself half-identifies — that the asymmetry is the bug, and a config the engine cannot parse should leave a candidate UNVERIFIED rather than failed, which would preserve the second verdict for consumers still shipping a v1-shaped config. I did not take it here because it keeps the engine reading the task's file, but I would rather hear your read before this merges than after.

xiaofei-zheng (approved):
Re-reviewed at `2d8654c56`. All four blocking items from the previous round are resolved. No blocking issues; LGTM.

**1. Broken loop tests — fixed.** `f55e71d5e` rewrites the four cases that the Step 7 gate invalidated rather than leaving them asserting the removed behavior. Verified the replacements actually exercise the paths they claim: `test_snr_pass_with_failing_assembly_acceptance_is_reverted` sets `kernel_backend = "assembly"`, writes a `numerical_validation` config via `_task`, then overwrites `driver.py` with a raising script — `correctness_command` exits nonzero, so `_run_canonical_suite` returns `output` carrying the tolerance text and `runner.py` maps the empty `outcome` to `canonical_correctness_failure`. `test_a_non_assembly_keep_rests_on_the_driver_not_on_the_task_config` parametrizes over shipping a rejecting config and asserts `"[canonical]"` never reaches stdout, which is the right assertion for a step that no longer runs.

**2. Stale gate prompt — fixed, and better than asked.** `918ebbb2d` replaces the single `CANONICAL_GATE_PROMPT` with `canonical_gate_prompt(backend)` behind the shared `runs_task_suite_acceptance` predicate, which `loop/runner.py:3792` now also reads, so the two cannot drift again. Checked completeness: all nine modules in `KERNEL_BACKEND_PROMPT_MODULES` take the new `canonical_gate` parameter, `base.py:72` is the only call site, and no import cycle is introduced (`loop/scoring.py` imports stdlib only; both `__init__.py` files are empty). `TestAcceptanceGateMatchesTheLoop` pins it in both directions. hipblaslt gaining a gate paragraph it never rendered is a fix, not scope creep — three of its stop conditions rested on one.

**3. Stale docs — fixed.** `2d8654c56` covers every location named: `architecture.md:29`, `optimization-loop.md:21-28`, `cli.md:93`, `cli.py:669-672`, the gluon KB (`forge_integration.md`, `dense_gemm/gluon.md`, `quant_fp4_mxfp/gluon.md`). The orphaned `_WARMSTART_CANONICAL_TIMEOUT_CAP_SEC` comment is gone. `git grep` for the old claims returns nothing under `src/` or `docs/`.

**4. Description/diff mismatch — fixed.** The test section now enumerates what the diff actually contains and states which three cases were deleted rather than moved, with the coverage that replaces them.

Also checked this round: the nine prompt hashes are re-snapshotted with the per-backend reason the snapshot file requires; `experience_integration.py` has no leftover canonical references; `asyncio` is still used after the warm-start removal. CI is green on all checks at `2d8654c56`.

jiaqiang-dot-liu:
## PR #1596 -- Let the rewrite pipeline's caller own its task shape, its files and its stages

What it does: the engine judged every KEEP by re-running one consumer's `config.yaml` evaluator, so a
consumer that declares evaluation differently could not earn a keep at all; Step 7 is now gated on
`runs_task_suite_acceptance()`, which only assembly satisfies, and the same predicate picks the gate
paragraph each of the nine backend prompts renders. Alongside it, the rewrite port commits the whole
attempt directory rather than one declared file, and `--applyback/--no-applyback` lets a caller take
the standalone kernel without paying for, or being judged on, the framework patch.

Blocking issues: 3

1. [X3] The gluon knowledge card keeps the paragraph `gluon/prompts.py` just corrected. [verified]
   Problem: `src/kernelforge/data/local_knowledge/languages/gluon/skills/optimize/gluon_levers/forge_integration.md:82`
   still reads "The task's `compile_command` often builds a **smaller shape** than the one the loop
   benchmarks, and a Gluon path with a shape or arch constraint that the benchmark satisfies and the
   compile check does not will fail acceptance after passing everything else." That is the sentence
   rewritten at `src/kernelforge/kernel_backends/gluon/prompts.py:80`, in a file this PR already
   edited at :32 and :190. Nothing covers it: `build_forge_knowledge`
   (`knowledge/local_index.py:249`) inlines INDEX maps only -- "full cards on disk, Read on demand" --
   so the nine re-snapshotted hashes do not reach it, and
   `test_only_a_task_suite_backend_claims_forge_runs_the_task_config` tests for
   "`correctness_command`", which this line does not contain.
   Impact: gluon no longer runs that step, and `gluon/prompts.py:42` and :87 both tell the agent to
   read this card. It shapes its fallback for a compile check that never runs, which is the drift the
   PR exists to remove -- one copy short.
   Action: retarget :81-84 at the downstream consumer's build, as :80 already does.

2. [C2] The port commit is fixed; the optimize loop that follows still drops a sibling module. [verified]
   Problem: `rewrite_by_flydsl/runner.py:432` now commits `attempt.relative_root`, on the stated
   premise that a producer may put part of its implementation in a module beside the entry point.
   Every iteration after that runs through `loop/runner.py`, which stages with `git add -u` (:1187)
   plus paths matching `commit_new_paths` (:1193) -- and `rewrite_by_flydsl/optimize.py` passes none.
   A new module the implementer creates in the attempt directory is not merely unstaged: `_new_paths`
   (:1056) enumerates via `_list_untracked` (:1030), which runs
   `git ls-files --others --exclude-standard`, so a path under the gitignored `.forge_rewrite/` is not
   in the refused list either.
   Impact: the same failure the PR describes for the port -- a commit holding an entry point whose
   import is not there -- for every KEEP in the campaign, with no `[git] N new file(s) outside
   commit_new_paths` line and no REVERT removing it. `best_commit` then names a tree no stage measured.
   Action: extend the fix to the optimize stage -- seed `commit_new_paths` with the attempt directory
   for the nested loop, or state in the body why the port is the only producer of this condition given
   `port_loop.py:155` declares a single-file target.

3. [X3] `--no-applyback` returns the reserve; three operator-facing strings still name it. [verified]
   Problem: with the stage declined, `without_applyback()` sets `applyback_reserve_sec` to 0 and
   `search_stop_unix == deadline_unix`. Unchanged text still attributes the stop to a reserve:
   `rewrite_by_flydsl/runner.py:249` "less than 20 minutes remain; no PORT session may start",
   :586 "20-minute finalization reserve reached after PORT; skipping forge-loop", and
   `port_loop.py:137,189,211` "PORT stopped at the 20-minute finalization reserve".
   Impact: :249 is not only a log line -- `_setup_failed` stores it as `failure_detail`, which
   `emit_result` writes into the result JSON a consumer parses to attribute the failure. On a
   `--no-applyback` run it names a reserve the run returned, when the cause was the deadline itself.
   Action: derive the wording from `rewrite_budget.applyback_reserve_sec`, or drop the number.

Checked: loop/scoring.py, loop/runner.py (Step 7, _list_untracked, _new_paths, the keep staging at
:1187), loop/canonical_correctness.py, loop/path_ownership.py, kernel_backends/base.py and all nine
prompts modules, knowledge/local_index.py, knowledge/experience_integration.py,
agent_backends/workspace_guard.py, rewrite_by_flydsl/{runner,budget,report,protocol,attempt,port_loop,applyback}.py,
the three changed knowledge cards, and the three changed docs pages.
Ran: `git add -f -- <dir> :(exclude)<dir>/__pycache__/** :(exclude)<dir>/**/__pycache__/**` in a
scratch repo with caches at two depths (rc 0, both excluded, siblings staged); a tree-wide sweep for
surviving `compile_command`/`correctness_command` claims; `build_result` call-site and
`build_system_prompt` caller enumeration; numstat against the body's +421/-281 over 30 files and its
"Ten added" tests (9 functions, 10 cases).
Dropped in refutation: the `_PYCACHE_EXCLUDES` list is narrower than `path_ownership.RUNTIME_DIRECTORY_NAMES`,
but `flydsl_cache` and `jit` live under `experiments_dir/aiter_cache` (`loop/aiter_cache.py:76-78`),
not under the attempt directory, so the wider list has nothing to exclude here.
No independent reader was available, so the three findings carry only my own refutation pass.
Base: 060a1e4b6 | Head: 2d8654c56 | CI at head: 11 of 12 shards success, py3.10 shard 1 and both
CodeQL analyze jobs still in progress, nothing failing, mergeable MERGEABLE.

